PR #28, opened 2021-07-09
Rename to axum

### LINKED ISSUE #6: Name...?

davidpdrsn:
Fixes https://github.com/davidpdrsn/awebframework/issues/6


shofetim:
Why is it called "Axum"? Is it after the [ancient African empire](https://en.wikipedia.org/wiki/Axum)?

davidpdrsn:
Its named after Aksum in Ethiopia which is a "tower like thing". We wanted the name to be a reference to tower, which axum is based on.

PR #53, opened 2021-07-31
Add unix domain socket example

davidpdrsn:
Not actually related to Axum, can be implemented directly with Hyper, but I figure its nice to have for demonstration and might help catch accidental breaking changes in the future.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -12,3 +12,4 @@
 - [`websocket`](../examples/websocket.rs) - How to build an app that handles WebSocket connections.
 - [`error_handling_and_dependency_injection`](../examples/error_handling_and_dependency_injection.rs) - How to handle errors and dependency injection using trait objects.
 - [`tokio_postgres`](../examples/tokio_postgres.rs) - How to use a tokio-postgres and bb8 to query a database.
+- [`unix_domain_socket`](../examples/unix_domain_socket.rs) - How to run an Axum server over unix domain sockets.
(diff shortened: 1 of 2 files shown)


danielecr:
Now it is possible to use an unix socket just replacing the line in
https://github.com/tokio-rs/axum/blob/main/examples/hello-world/src/main.rs#L15
with `tokio::net::UnixSocket::bind(path)` and it works with hyper 1.0
the current example is outdated but does something more (really I did not tried what is that peer staff).

It is a good example for connection infos, but really there is missing example for a working unix socket, that would be very short.
Can I do a PR for the smaller version?

davidpdrsn:
> Now it is possible to use an unix socket just replacing the line in
https://github.com/tokio-rs/axum/blob/main/examples/hello-world/src/main.rs#L15
with tokio::net::UnixSocket::bind(path) and it works with hyper 1.0

How? serve requires a TcpListener.

davidpdrsn:
I think perhaps you should just go straight to using hyper directly. I don’t wanna complicate `serve` and friends more than necessary.

The intention is that hyper-util will have higher level conveniences so it makes sense to iterate there, not in axum.

danielecr:
I agree, I am using hyper indeed.
With hyper I just changed 2 lines of code, and it listen on unix socket.

But still your example for using unix socket has been commented out in the newer axum.
And looking at the code in serve there is little diff with hyper example.
I can say the only added staff is IncomingStream.
Still, I didn't looked at tracing or logging staff, and extractors maybe.

Issue #43, opened 2021-07-30
Extractor for request remote address


Roguelazer:
### Motivation

I'm playing with axum and I noticed that there doesn't seem to be any way to extract the remote address (or even just IP) of the other end of the connection. As far as I can tell there's no way to do this anywhere in Tower; `tower`'s example logging middleware doesn't even log the remote address.

### Proposal

Any of the following perhaps:

   1. Extractor that gives you a `SocketAddr` of the peer
   2. An attribute on the `Request` with the peer information
   3. Perhaps some way to get at the underlying `hyper::server::AddrStream`?

### Alternatives

Uh, I guess put another HTTP server in front of `axum` that copies the remote address into a header?

davidpdrsn:
Makes sense to provide!

Without having through about this too much it might be possible to change [`RoutingDsl::into_make_service`](https://docs.rs/axum/0.1.1/axum/routing/trait.RoutingDsl.html#method.into_make_service) to return some custom service that stores the remote address in a request extension, which can be picked up by an extractor.

davidpdrsn:
Here is a POC:

```rust
use async_trait::async_trait;
use axum::{
    extract::{FromRequest, RequestParts},
    prelude::*,
    routing::BoxRoute,
    AddExtension,
};
use futures_util::future::Ready;
use hyper::server::conn::AddrStream;
use std::{
    convert::Infallible,
    net::SocketAddr,
    task::{Context, Poll},
};
use tower::{Service, ServiceBuilder};
use tower_http::add_extension::AddExtensionLayer;

#[tokio::main]
async fn main() {
    hyper::Server::bind(&SocketAddr::from(([127, 0, 0, 1], 3000)))
        .serve(MakeSvc)
        .await
        .unwrap();
}

fn app() -> BoxRoute<Body> {
    route(
        "/",
        get(|RemoteAddr(addr)| async move {
            println!("`{}` connected", addr);
        }),
    )
    .boxed()
}

struct MakeSvc;

impl Service<&AddrStream> for MakeSvc {
    type Response = AddExtension<BoxRoute<Body>, RemoteAddr>;
    type Error = Infallible;
    type Future = Ready<Result<Self::Response, Self::Error>>;

    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
        Poll::Ready(Ok(()))
    }

    fn call(&mut self, conn: &AddrStream) -> Self::Future {
[...]

PR #55, opened 2021-07-31
Add extractor for remote connection info

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/43

With this you can get the remote address like so:

```rust
use axum::{prelude::*, extract::ConnectInfo};
use std::net::SocketAddr;

let app = route("/", get(handler));

async fn handler(ConnectInfo(addr): ConnectInfo<SocketAddr>) -> String {
    format!("Hello {}", addr)
}

// Starting the app with `into_make_service_with_connect_info` is required
// for `ConnectInfo` to work.
let make_svc = app.into_make_service_with_connect_info::<SocketAddr, _>();

hyper::Server::bind(&"0.0.0.0:3000".parse().unwrap())
    .serve(make_svc)
    .await
    .expect("server failed");
```

This API is fully generic and supports whatever transport layer you're using with Hyper. I've updated the unix domain socket example to extract `peer_creds` and `peer_addr`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Implement `Stream` for `WebSocket`.
 - Implement `Sink` for `WebSocket`.
+- Add extractor for remote connection info.
 
 ## Breaking changes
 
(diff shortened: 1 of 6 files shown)


Roguelazer:
I've only had time to give this a cursory glance, but it looks great! Thanks for looking into this so fast!

PR #60, opened 2021-07-31
Add test helpers

### LINKED ISSUE #59: Add helper fn to test a JSON response
### Motivation
As of now, to test a REST endpoint that returns a JSON object, two steps are needed; the first one, to get the body of the response using hyper, and the second one to convert it to JSON with serde:
```rust
    #[tokio::test]
    async fn json() {
        let app = app();

        let response = app.oneshot(Request::builder()...).await.unwrap();

        let body = hyper::body::to_bytes(response.into_body()).await.unwrap(); 
        let body: Value = serde_json::from_slice(&body).unwrap();

        assert_eq!(body, json!({ "data": [1, 2, 3, 4] }));
    }
```

### Proposal
It would be great to have a single fn that does all the job, so one can write, for example:
```rust
    #[tokio::test]
    async fn json() {
        let app = app();

        let response = app.oneshot(Request::builder()...).await.unwrap();

        let body: Value = test::read_response_json(response).unwrap();

        assert_eq!(body, json!({ "data": [1, 2, 3, 4] }));
    }
```

though this would require, probably, the creation of a `test` module

ufoscout:
This could fix #59

### DIFF
--- a/examples/testing.rs
+++ b/examples/testing.rs
@@ -34,10 +34,10 @@ fn app() -> BoxRoute<Body> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use axum::test::*;  // for `app.oneshot()`, `read_response_json()`
     use http::StatusCode;
     use serde_json::{json, Value};
     use std::net::{SocketAddr, TcpListener};
-    use tower::ServiceExt; // for `app.oneshot()`
 
     #[tokio::test]
     async fn hello_world() {
@@ -76,8 +76,7 @@ mod tests {
 
         assert_eq!(response.status(), StatusCode::OK);
 
-        let body = hyper::body::to_bytes(response.into_body()).await.unwrap();
-        let body: Value = serde_json::from_slice(&body).unwrap();
+        let body: Value = read_response_json(response).await;
         assert_eq!(body, json!({ "data": [1, 2, 3, 4] }));
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -613,6 +613,7 @@ pub mod handler;
 pub mod response;
 pub mod routing;
 pub mod service;
+pub mod test;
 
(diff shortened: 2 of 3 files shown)


davidpdrsn:
Hm I'm not sure this is something that belongs in Axum 🤔 I don't feel doing

```rust
let body = hyper::body::to_bytes(response.into_body()).await.unwrap(); 
let body: Value = serde_json::from_slice(&body).unwrap();
```

Is that bad and if users want something more convenient they can add utilities to their own crates.

One can also spawn the app in a task and call it via reqwest, which gives even more high level methods. [That is how Axum itself is tested](https://github.com/tokio-rs/axum/blob/main/src/tests.rs#L36-L40).

I wouldn't mind adding a method like this to the test example so users have something to copy/paste.

ufoscout:
@davidpdrsn 
I agree that it is quite easy to write such a helper directly in your own project, but, for sure, I would prefer to have it provided by the framework itself to avoid being forced of declaring it in every project.

In addition, I created this PR because at the moment you are forced to add these two dev-dependencies only to perform a test:
```toml
[dev-dependencies]
hyper = "0.14"
tower = "0.4"
```
This PR would avoid it.

davidpdrsn:
You always need hyper to run the app and axum already depends on tower so those dependencies are included regardless.

I personally don't think copy/pasting a small function like this into your projects is a big deal 🤷 

If we were to add test helpers I would rather we add something thats been carefully designed and provides a great user experience. Something along the lines of what rocket or warp provides.

davidpdrsn:
I've asked in Discord if people have thoughts on this. Lets see if anyone replies.

davidpdrsn:
Sorry but I think I'm going to close this. As mentioned making testing easy is important but I think it requires more careful design.

PR #78, opened 2021-08-01
add chat example

programatik29:


examples/chat.rs
```diff
@@ -0,0 +1,125 @@
...
+
+    // Send joined message to all subscribers.
+    let msg = format!("{} joined.", username);
+    let _ = state.tx.send(msg);
+
+    // This task will receive broadcast messages and send text message to our client.
+    let mut send_task = tokio::spawn(async move {
+        while let Ok(msg) = rx.recv().await {
+            // In any websocket error, break loop.
+            if let Err(_) = sender.send(Message::text(msg)).await {
```
davidpdrsn, line 78:
```suggestion
            if sender.send(Message::text(msg)).await.is_err() {
```

examples/chat.rs
```diff
@@ -0,0 +1,125 @@
...
+        .unwrap();
+}
+
+async fn websocket(stream: WebSocket, state: extract::Extension<Arc<AppState>>) {
+    let state = state.0;
+
+    // By splitting we can send and receive at the same time.
+    let (mut sender, mut receiver) = stream.split();
+
+    // Empty string.
```
davidpdrsn, line 46:
```suggestion
    // Username gets set in the receive loop, if its valid
```

Probably a bit more descriptive.

examples/chat.rs
```diff
@@ -0,0 +1,125 @@
...
+
+use axum::handler::get;
+use axum::response::Html;
+use axum::routing::RoutingDsl;
+use axum::ws::{ws, Message, WebSocket};
+use axum::{extract, route, AddExtensionLayer};
+
+// Our shared state
+struct AppState {
+    user_map: Mutex<HashMap<String, ()>>,
```
davidpdrsn, line 17: What do you think about using `HashSet<String>` instead of `HashMap<String, ()>`? They're basically the same.

examples/chat.rs
```diff
@@ -0,0 +1,125 @@
...
+
+use futures::{sink::SinkExt, stream::StreamExt};
+
+use tokio::sync::broadcast;
+
+use axum::handler::get;
+use axum::response::Html;
+use axum::routing::RoutingDsl;
+use axum::ws::{ws, Message, WebSocket};
+use axum::{extract, route, AddExtensionLayer};
```
davidpdrsn, line 13: I would say `axum::prelude::*` is recommended but its up to you.

davidpdrsn (commented):
Looks awesome! Thanks for putting this together.

I had a couple of minor comments.

programatik29:
I updated as you suggested. Thanks for comments.

davidpdrsn (approved):

davidpdrsn:
Thanks!

PR #81, opened 2021-08-01
Changed http-body dependency version to 0.4.2

minghuaw:
## Motivation

`http_body::Full` (used here in [src/ws/mod.rs line 72](https://github.com/tokio-rs/axum/blob/55c1a294206bc7e9fce940a81582e875e98eefda/src/ws/mod.rs#L72)) is not available in versions before 0.4.2, and this may lead to error when version 0.4.0 or 0.4.1 is pulled from crates.io.

## Solution

Changing the dependency version of `http-body` from 0.4 to 0.4.2. This will ensure the availability of `http_body::Full`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ async-trait = "0.1"
 bytes = "1.0"
 futures-util = "0.3"
 http = "0.2"
-http-body = "0.4"
+http-body = "0.4.2"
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
Thanks!

Wanna add a line to the changelog as well?

minghuaw:
I feel like this change is pretty minor and adding it to the change log is not necessary.

davidpdrsn:
Hm sure

PR #83, opened 2021-08-02
Add support for WebSocket protocol negotiation.

sunli829:
I added support for WebSocket sub-protocol negotiation. Users can use `axum::ws::ws_with_protocols` function to specify a list of known sub-protocols by the server.

## Motivation

Most WebSocket-based specifications need to specify the sub-protocol name, such as GraphQL over WebSocket. If the server cannot return the negotiated sub-protocol name, the client will disconnect.

## Solution

I added the `axum::ws::ws_with_protocols` function, which allows to specify a list of known sub-protocols. If the sub-protocol requested by the user matches one of it, add `Sec-WebSocket-Protocol` in the upgrade response header and specify the matching sub-protocol name, otherwise the `Sec-WebSocket-Protocol` header will not be returned, and the client will disconnect.

### DIFF
--- a/src/ws/mod.rs
+++ b/src/ws/mod.rs
@@ -73,6 +73,7 @@ use http_body::Full;
 use hyper::upgrade::{OnUpgrade, Upgraded};
 use sha1::{Digest, Sha1};
 use std::pin::Pin;
+use std::sync::Arc;
 use std::{
     borrow::Cow, convert::Infallible, fmt, future::Future, marker::PhantomData, task::Context,
     task::Poll,
@@ -96,6 +97,7 @@ where
     WebSocketUpgrade {
         callback,
         config: WebSocketConfig::default(),
+        protocols: Vec::new().into(),
         _request_body: PhantomData,
     }
 }
@@ -169,6 +171,7 @@ impl_ws_handler!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T1
 pub struct WebSocketUpgrade<F, B, T> {
     callback: F,
     config: WebSocketConfig,
+    protocols: Arc<[Cow<'static, str>]>,
     _request_body: PhantomData<fn() -> (B, T)>,
 }
 
@@ -180,6 +183,7 @@ where
         Self {
             callback: self.callback.clone(),
             config: self.config,
+            protocols: self.protocols.clone(),
             _request_body: PhantomData,
         }
     }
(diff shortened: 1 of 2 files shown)


src/ws/mod.rs
```diff
@@ -96,6 +96,29 @@ where
...
+}
+
+/// Create a new [`WebSocketUpgrade`] service that will call the closure with
+/// each connection.
+///
+/// `protocols` is a list of known protocols. If the protocol name specified by `Sec-WebSocket-Protocol` header
+/// to match any of them, the upgrade response will include `Sec-WebSocket-Protocol` header and return the protocol name.
+///
+/// See the [module docs](crate::ws) for more details.
+pub fn ws_with_protocols<F, B, T>(
```
davidpdrsn, line 111: I think we should add this as a builder method on `WebSocketUpgrade` similar to [the other methods](https://docs.rs/axum/0.1.2/axum/ws/struct.WebSocketUpgrade.html).

davidpdrsn: I think we should also include a code example. Could just use the graphql protocol name.

src/ws/mod.rs
```diff
@@ -169,6 +192,7 @@ impl_ws_handler!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T1
 pub struct WebSocketUpgrade<F, B, T> {
     callback: F,
     config: WebSocketConfig,
+    protocols: &'static [&'static str],
```
davidpdrsn, line 195: I'm thinking if requiring `'static` is perhaps too restrictive here 🤔 

I'm wondering if we could make the struct field `Arc<[Cow<'static, str>]>`. That way its still cheap to clone and supports both string literals and `String`s. Using `Arc<[_]>` should be fine since we don't never have to mutate the list.

I think the builder method on `WebSocketUpgrade` can then be defined like this:

```rust
fn protocols(self, protocols: I) -> Self
where
    I: IntoIterator,
    I::Item: Into<Cow<'static, str>>,
{
    let protocols = protocols
        .into_iter()
        .map(Into::into)
        .collect::<Vec<_>>()
        .into();

    // ...
}
```

I feel that gives pretty good ergonomics while being light weight.

What do you think?

davidpdrsn (changes requested):
Thanks for doing this! Its definitely something that makes sense to have!

I think the main implementation in `fn call` looks good but I had a few suggestions regarding the interface.

Let me know what you think.

src/ws/mod.rs
sunli829: You are right, I didn't keep the API style uniform.

src/ws/mod.rs
sunli829: You are right, this is more flexible. 🙂

sunli829:
Please review again. 🙂

davidpdrsn (approved):
Great! Thanks for doing this 😊

PR #85, opened 2021-08-02
Add chat example documentation.

pepoviola:
Hi @davidpdrsn, first let me say thanks for this awesome work!! I was playing with the examples and I think this little changes can help people to easily run the `chat` example.

## Motivation

Allow people to easy run the `chat` example by check the docs in `chat.rs` or with the hint when run 
```
cargo run --example chat
```

## Solution

- Add example of how to run
- Add `required-features`
- Add `chat` to example list.

Thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,3 +80,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [package.metadata.playground]
 features = ["ws"]
+
+[[example]]
+name = "chat"
+required-features = ["ws"]
--- a/examples/chat.rs
+++ b/examples/chat.rs
@@ -3,7 +3,7 @@
 //! Run with
 //!
 //! ```
-//!     cargo run --features="ws" --example chat
+//! cargo run --features=ws --example chat
 //! ```
 
 use std::collections::HashSet;
--- a/examples/chat.rs
+++ b/examples/chat.rs
@@ -1,3 +1,11 @@
+//! Example chat application.
+//!
+//! Run with
+//!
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -80,3 +80,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [package.metadata.playground]
 features = ["ws"]
+
+[[example]]
+name = "chat"
+required-features = ["ws"]
```
davidpdrsn, line 86: Whoa! I didn't realize you can do this. Thats cool! We should totally add for the other examples as well. Doesn't have to be done in this PR though.

examples/chat.rs
```diff
@@ -1,3 +1,11 @@
+//! Example chat application.
+//!
+//! Run with
+//!
+//! ```
+//!     cargo run --features="ws" --example chat
```
davidpdrsn, line 6: nitpick:

```suggestion
//! cargo run --features=ws --example chat
```

davidpdrsn: This would also make sense to have for all other example. Doesn't have to be done in this PR though.

davidpdrsn (changes requested):
Looks good!

I had one small nitpick.

pepoviola:
Hi @davidpdrsn, thanks for the review! I already applied the suggestion. 

Is nice to have the `required-features` for examples, I can go over and add for the other examples as well in another pr.

Thanks!

davidpdrsn (approved):

PR #94, opened 2021-08-03
Make imports consistent.

sunli829:
I don't know if this is a coding style standard, but I know that many popular projects use this import order. 😁

### DIFF
--- a/examples/chat.rs
+++ b/examples/chat.rs
@@ -10,14 +10,12 @@ use std::collections::HashSet;
 use std::net::SocketAddr;
 use std::sync::{Arc, Mutex};
 
-use futures::{sink::SinkExt, stream::StreamExt};
-
-use tokio::sync::broadcast;
-
 use axum::prelude::*;
 use axum::response::Html;
 use axum::ws::{ws, Message, WebSocket};
 use axum::AddExtensionLayer;
+use futures::{sink::SinkExt, stream::StreamExt};
+use tokio::sync::broadcast;
 
 // Our shared state
 struct AppState {
--- a/examples/error_handling_and_dependency_injection.rs
+++ b/examples/error_handling_and_dependency_injection.rs
@@ -9,6 +9,8 @@
 
 #![allow(dead_code)]
 
+use std::{net::SocketAddr, sync::Arc};
+
 use axum::{
     async_trait,
     extract::{Extension, Json, UrlParams},
@@ -19,7 +21,6 @@ use axum::{
 use http::StatusCode;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
-use std::{net::SocketAddr, sync::Arc};
 use uuid::Uuid;
 
 #[tokio::main]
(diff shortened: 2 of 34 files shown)


davidpdrsn:
The order of imports isn't something I'm very concerned with tbh. The one used today is the rustfmt's default. Can you configure rustfmt to use this ordering?

davidpdrsn:
It seems rustfmt has [an `group_imports ` option](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#group_imports) which when set to `StdExternalCrate` should match this style. That option isn't stable yet however.

Until rustfmt stable is able to automatically format imports the way we want I'm reluctant to change the formatting as it creates churn and will not be maintained unless done automatically.

Do you know of a way to maintain this style automatically?

sunli829:
rustfmt can't do this automatically, I did it manually, but I saw other projects do it, which seems more clear. 😁

davidpdrsn:
> I did it manually

In that case I think we should wait with this. Even though I'm sorry for discarding all this tedious manual work 😕

The order used now is the default ordering used by rustfmt when all the imports are in one block (no newlines between) so at least that gives some consistency.

sunli829:
I respect your decision. 😁

jplatte:
What I've done in a bunch of projects before is have one Stable CI job that runs `cargo test` (and `cargo check` for different feature sets) and a one Nightly CI job that runs `cargo fmt` and `cargo clippy` (benefitting from new lints as they come out, with the downside that there's also false positives now and then).

davidpdrsn:
What do you do for running rustfmt locally then? Override it to use nightly somehow?

jplatte:
Well stable rustfmt still applies all non-experimental formatting options for me and with the codebase being consistently formatted I hardly ever accidentally introduce something that nightly rustfmt would format differently. Occasionally I might `cargo +nightly fmt`.

ZhangHanDong:
The `group_imports` option doesn't solve this problem either, and you end up having to manually group the imported modules in order to improve code readability. Of course these tasks can be done after the project is stable.

jplatte:
@ZhangHanDong What do you mean by that? I'm aware of [two](https://github.com/rust-lang/rustfmt/issues/4709) [bugs](https://github.com/rust-lang/rustfmt/issues/4723) affecting `group_imports`, but it doesn't sound like that's what you're referring to.

ZhangHanDong:
@jplatte 

What I mean is that relying on `group_imports` alone doesn't achieve the desired effect either.

The only option currently available for `group_imports` is `StdExternalCrate`, which can be divided into three groups.

1. std, core and alloc,
2. external crates.
3. self, super and crate imports.

But wouldn't external crates have to be manually grouped by the developer?

jplatte:
Not sure what additional grouping of external crates you're talking about. From a quick glance at this PR it seems like it implements exactly what rustfmt would do too with `group_imports = "StdExternalCrate"`.

ZhangHanDong:
@jplatte  You'll understand if you look at the changes in this pr.

Screen Shot 2021-08-10 at 22 34 32

group_imports = "StdExternalCrate " is useless for this kind of grouping and needs to be manually adjusted by the developer. 

Does this make it clear?

PR #95, opened 2021-08-03
Use `pin-project-lite` instead of `pin-project`

sunli829:
The compilation speed has not improved much, at least it looks like this on my Mac. But pin_project_lite is not a procedural macro, so the IDE can correctly analyze the return type of the `project` function. I believe this can make the code easier to maintain.

### DIFF
--- a/src/ws/mod.rs
+++ b/src/ws/mod.rs
@@ -277,110 +277,112 @@ where
         let this = self.clone();
         let protocols = self.protocols.clone();
 
-        ResponseFuture(Box::pin(async move {
-            if req.method() != http::Method::GET {
-                return response(StatusCode::NOT_FOUND, "Request method must be `GET`");
-            }
+        ResponseFuture {
+            future: Box::pin(async move {
+                if req.method() != http::Method::GET {
+                    return response(StatusCode::NOT_FOUND, "Request method must be `GET`");
+                }
 
-            if !header_contains(&req, header::CONNECTION, "upgrade") {
-                return response(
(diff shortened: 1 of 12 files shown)


davidpdrsn:
Thanks for doing this! Using `pin-project-lite` totally makes sense.

As discussed [here](https://github.com/tokio-rs/axum/pull/94) though I think we should keep the import ordering as it was. That will also make this change smaller and easier to review.

sunli829:
> As discussed [here](https://github.com/tokio-rs/axum/pull/94) though I think we should keep the import ordering as it was. 

Yes, I am working on this and it will be done soon. 😁

davidpdrsn:
Thank you 😃

sunli829:
I have done it. 😁

src/ws/mod.rs
```diff
@@ -212,6 +216,47 @@ impl<F, B, T> WebSocketUpgrade<F, B, T> {
...
+        I: IntoIterator,
+        I::Item: Into<Cow<'static, str>>,
+    {
+        self.protocols = protocols
+            .into_iter()
+            .map(Into::into)
+            .collect::<Vec<_>>()
+            .into();
+        self
+    }
```
davidpdrsn, line 259: Seems you need to rebase/merge main as this was merged in the other PR.

sunli829:
Sorry, it's done now. 😂

davidpdrsn (approved):
Thanks!

PR #97, opened 2021-08-03
Add 404 page example

mkieblesz:
## Motivation

Very common problem of how to handle route misses makes it a good candidate for an example.

## Solution

Add 404 to example list.

### DIFF
--- a/examples/404.rs
+++ b/examples/404.rs
@@ -1,0 +1,44 @@
+//! Run with
+//!
+//! ```not_rust
+//! cargo run --example 404
+//! ```
+
+use axum::{
+    body::{box_body, Body, BoxBody},
+    prelude::*,
+};
+use http::{Response, StatusCode};
+use std::net::SocketAddr;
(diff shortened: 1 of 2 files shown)


examples/404.rs
```diff
@@ -0,0 +1,44 @@
...
+use http::{Response, StatusCode};
+use std::net::SocketAddr;
+use tower::util::MapResponseLayer;
+
+#[tokio::main]
+async fn main() {
+    tracing_subscriber::fmt::init();
+
+    // build our application with a route
+    let app = route("/", get(handler)).layer(MapResponseLayer::new(map_404));
```
davidpdrsn, line 20:
```suggestion
    let app = route("/", get(handler))
        // make sure this is added as the very last thing
        .layer(MapResponseLayer::new(map_404));
```

davidpdrsn: Probably good to make this clear

davidpdrsn (approved):

PR #98, opened 2021-08-03
Implement SSE using responses

KaiJewson:
## Motivation

SSE is currently implemented as an entire service. However, this is an overly general abstraction, since SSE really only cares about the response format, so it makes more sense to implement it as a regular response. [Warp also implements SSE as a reply](https://docs.rs/warp/0.3/warp/filters/sse/fn.reply.html) and not a `Filter`. Implementing SSE as a response also simplifies the implementation by removing the need for `SseHandler` and its associated macros.

## Solution

Implement SSE as another response inside the `response` module.

## Unresolved questions

- `sse(stream)`, `Sse::new(stream)` or both? I currently just implemented the former.
- Should `sse` be re-exported at the crate root?


KaiJewson:
I agree, the current system works fine for now. It can wait for 0.2 :)
-------- Original message --------From: David Pedersen ***@***.***> Date: 2021-08-03  18:00  (GMT+00:00) To: tokio-rs/axum ***@***.***> Cc: Kai Jewson ***@***.***>, Author ***@***.***> Subject: Re: [tokio-rs/axum] Implement SSE using responses (#98) 
Of course yeah. This is much nicer!
Although in its current form it's a breaking change meaning it'll have to wait until 0.2. I dunno when that'll be but there is a milestone tracking things for 0.2.
Do you think we can live with the current implementation until 0.2 or should we find a non-breaking solution? Personally I think we should wait with this until 0.2 as the current setup isn't too bad.

davidpdrsn:
Wanna try to merge `main`? I've merged a bunch of PRs.

KaiJewson:
Sorry, I can't right now, I'm not at my PC :P. I'll be able to next week though.

davidpdrsn:
No worries. Enjoy your weekend 😊

KaiJewson:
I have merged master. Now that `IntoResponse` supports custom bodies I switched to a custom `HttpBody` implementation for the SSE responses, so we no longer have to box the stream, and we get a more specific error type rather than just `BoxError`. To avoid `Sync` bounds on the inner stream, I added a dependency on [`SyncWrapper`](https://docs.rs/sync_wrapper/0.1/sync_wrapper/struct.SyncWrapper.html). [Hyper also uses this type internally](https://github.com/hyperium/hyper/blob/684f2fa76d44fa2b1b063ad0443a1b0d16dfad0e/src/body/body.rs#L61) for its `Body` type which we were using before, but it uses a [copy-pasted local version](https://github.com/hyperium/hyper/blob/684f2fa76d44fa2b1b063ad0443a1b0d16dfad0e/src/common/sync_wrapper.rs) which we can't do because of `#![forbid(unsafe_code)]`.

src/response/sse.rs
```diff
@@ -0,0 +1,387 @@
...
+use tower::BoxError;
+
+/// Create a new [`Sse`] response that will respond with the given stream of
+/// [`Event`]s.
+///
+/// See the [module docs](self) for more details.
+pub fn sse<S>(stream: S) -> Sse<S>
+where
+    S: TryStream<Ok = Event> + Send + 'static,
+    S::Error: Into<BoxError>,
```
davidpdrsn, line 57: Is this bound actually needed?

src/response/sse.rs
```diff
@@ -0,0 +1,387 @@
...
+            .field("stream", &format_args!("{}", std::any::type_name::<S>()))
+            .field("keep_alive", &self.keep_alive)
+            .finish()
+    }
+}
+
+impl<S, E> IntoResponse for Sse<S>
+where
+    S: Stream<Item = Result<Event, E>> + Send + 'static,
+    E: Into<BoxError>,
```
davidpdrsn, line 94: Same here. Is this bound needed? We don't appear to return BoxError anywhere.

davidpdrsn (commented):
Great! Thanks for working this. I had two minor comments but looks good overall!

src/response/sse.rs
davidpdrsn: Never mind. I see now. It's a bound required by the trait. Carry on!

davidpdrsn:
Wanna update the changelog as well? Then it's good to merge 😊

PR #104, opened 2021-08-03
Remove unused imports from example

PatatasDelPapa:
Remove unused imports from the first crate documentation example.

## Motivation
Axium minimal hello world has unused imports that can confuse newbies to rust and the framework on why they are there.

## Solution
Remove the unused imports as they really are unused


davidpdrsn (approved):
Thanks!

Issue #101, opened 2021-08-03
Make it possible to merge routes


davidpdrsn:
Being able to merge routes/apps like this would be really cool

```rust
let app = route("/foo", get(|| async {}));
let other_route = route("/bar", get(|| async {}));

// this accepts `GET /foo` and `GET /bar`
let merged = app.or(other_route);
```

Things to consider:
- It should work with other things than `Route` like `BoxRoute`, `Nested`, basically anything that implements `RouteDsl`.

PR #108, opened 2021-08-03
Add `RouteDsl::or` to combine routes

davidpdrsn:
With this you'll be able to do:

```rust
let one = route("/foo", get(|| async { "foo" }))
    .route("/bar", get(|| async { "bar" }));

let two = route("/baz", get(|| async { "baz" }));

let app = one.or(two);
```

Fixes https://github.com/tokio-rs/axum/issues/101


davidpdrsn:
I think this is almost ready. However there is one small foot gun I would like to try and fix. This should not compile:

```rust
let svc_one = service_fn(|_: Request<Body>| async {
    Ok::<_, hyper::Error>(Response::new(Body::empty()))
})
.handle_error::<_, _, hyper::Error>(|_| Ok(StatusCode::INTERNAL_SERVER_ERROR));

let svc_two = svc_one.clone();

let app = svc_one.or(svc_two);
```

Since `svc_one` isn't created with `route` or `nest` it will swallow all requests. No actual routing will take place. It would be ideal if this didn't compile at all. Basically we need `svc_one` to _not_ implement `RoutingDsl`. I'll look into that in a separate PR.

src/routing.rs
```diff
@@ -507,7 +543,10 @@ impl<E> RoutingDsl for EmptyRouter<E> {}
 
 impl<E> crate::sealed::Sealed for EmptyRouter<E> {}
 
-impl<B, E> Service<Request<B>> for EmptyRouter<E> {
+impl<B, E> Service<Request<B>> for EmptyRouter<E>
+where
+    B: Send + Sync + 'static,
```
davidpdrsn, line 548: This additional trait bound makes this a breaking change. However I doubt anyone will actually hit this in practice.

davidpdrsn:
The issue mentioned [here](https://github.com/tokio-rs/axum/pull/108#issuecomment-892811637) will be fixed by https://github.com/tokio-rs/axum/pull/120.

PR #110, opened 2021-08-04
Re-export http crate.

sunli829:
I highly recommend re-export the `http` crate, as it is really commonly used. 😁


davidpdrsn:
I've been intentionally conservative with re-exports. I've seen it lead to issues with incompatible versions a few times. However you're right that `http` is used very commonly so it might be worth to make an exception for.

Hyper is also used a lot though, probably always. Should we re-export that as well? That doesn't feel right but I cannot explain why 🤔

sunli829:
That's exactly what I thought, I don't want to have to learn `hyper` usage while using `Axum`, although it's not hard, `Axum` should contain all the necessary functions, just like `Warp` does. 😁

sunli829:
Just like what `warp::Server` does, I think it's great.

davidpdrsn:
I see that warp exports both http and hyper, but with `#[doc(hidden)]`. That makes things harder to discover on docs.rs.

I don't think axum wrapping `hyper::Server` is right. The idiomatic way to use axum should be using hyper directly. In an earlier version of the API you were able to do `route(...).serve(addr)` but that was removed for this reason.

I'm not sure where to draw the line. Should axum also re-export tower, tower-http, and tokio? Users are likely to also require those. I don't necessarily think our goal should be that getting started with axum will only require `axum = "0.1"` in your `Cargo.toml`.

jplatte:
I think the real solution here is [public & private dependencies](https://github.com/rust-lang/rust/issues/44663) to be distinguished in Cargo, on which there seems to have been some progress recently, but it's looks to still be a decent bit away from stabilization.

Then people would still have to add dependencies to a bunch of things explicitly to use Axum¹, but Cargo would give a very clear error when the versions are mismatched.

> I've seen it lead to issues with incompatible versions a few times.

I think with all of these crates being part of the public interface anyways, there should be no compatibility concerns here. AFAIK it's just making things part of the public API that don't have to be which is problematic (since you then have to release breaking changes to upgrade when it could have been avoided).

¹ by the way, what's the right capitalization, `axum` or `Axum`? 😄

sunli829:
Yes, I didn't think about this, maybe `Axum` only needs to do what it should do, which is also very good. 😁

davidpdrsn:
> I think the real solution here is public & private dependencies to be distinguished in Cargo, on which there seems to have been some progress recently, but it's looks to still be a decent bit away from stabilization.

Ah I wasn't aware of that RFC. Gonna give it a read.

> I think with all of these crates being part of the public interface anyways, there should be no compatibility concerns here. 

Yeah thats true.

So which crates do you think we should re-export? `http`, `hyper`, and `tower`? You need tower for Service and ServiceExt which are useful for testing. Re-exporting tokio is perhaps a bit much 😜

src/lib.rs
```diff
@@ -745,6 +742,7 @@ pub mod ws;
 mod tests;
 
 pub use async_trait::async_trait;
+pub use http;
```
davidpdrsn, line 745:
```suggestion
#[doc(no_inline)]
pub use http;
```

davidpdrsn:
Cool! Lets remain cautious for now and only re-export `http` and `hyper::Server`. We can always add more in the future should someone request it.

src/lib.rs
sunli829: That's great! 🙂

davidpdrsn (approved):

davidpdrsn:
> ¹ by the way, what's the right capitalization, axum or Axum? 😄

`axum` :)

jplatte:
By the way AFAIK whole-crate re-export aren't inlined by default either, but it obviously doesn't hurt to add `#[doc(no_inline)]` either :)

PR #114, opened 2021-08-04
Tell clippy about MSRV

jplatte:
See https://github.com/rust-lang/rust-clippy#configuration


src/lib.rs
```diff
@@ -705,12 +705,7 @@
     missing_docs
 )]
 #![deny(unreachable_pub, broken_intra_doc_links, private_in_public)]
-#![allow(
-    elided_lifetimes_in_paths,
-    // TODO: Remove this once the MSRV bumps to 1.42.0 or above.
```
davidpdrsn, line 710: I had totally forgotten this was there. On CI we check against version 1.42 but that seems to be broken because `ContentLengthLimit` uses const generics which didn't land until 1.51.

Shouldn't adding this `.clippy.toml` also make clippy complain about `ContentLengthLimit`?

davidpdrsn (approved):
Lets get this in and then look at CI separately.

PR #116, opened 2021-08-04
Make ws::Message an enum for easier frame type matching

gbaranski:
## Motivation

This change makes frame type matching much easier.

Instead of:
```rust
let message = ...;
if message.is_text() {
    let text = message.to_str().unwrap();
     //e.g parse_json(text)
} else if message.is_binary() {
    let bytes = message.as_bytes();
    // e.g parse_cbor(bytes)
} else if message.is_pong() {
    let bytes = message.as_bytes();
    // something(bytes)
} else if message.is_ping() {
    let bytes = message.as_bytes();
    // something(bytes)
} else if message.is_close() {
    let (code, reason) = message.close_frame().unwrap(); 
    // close(code, reason)
}
```
we can do
```rust
let message = ...;
match message {
    Message::Text(text) => {
        // e.g parse_json(text);
    }
    Message::Binary(binary) => {
        // e.g parse_cbor(bytes);
    }
    Message::Ping(ping) => {
        // something(ping)
    }
    Message::Pong(pong) => {
        // something(pong)
    },
    Message::Close(close) => {
        // close(close.code, close.reason);
    }
}
```
I've also updated examples.

## Solution

Copy `tungstenite::Message` and `tungstenite::protocol::CloseCode` to `axum::ws` so that no tungstenite API is publicly exposed, so even if they make a breaking change, we don't need to.

I am not sure about legal reasons, don't we need to include their license for parts I've copied from `tungstenite`? They're also using MIT License.

Note that it is a breaking change in axum, theoritically we could keep old methods on `Message` to keep backwards compatibility, but that's up to you.


src/ws/coding.rs
```diff
@@ -0,0 +1,173 @@
...
+            1010 => Extension,
+            1011 => Error,
+            1012 => Restart,
+            1013 => Again,
+            1015 => Tls,
+            1..=999 => Bad(code),
+            1016..=2999 => Reserved(code),
+            3000..=3999 => Iana(code),
+            4000..=4999 => Library(code),
+            _ => Bad(code),
```
davidpdrsn, line 150: Odd that they're using `Bad` here instead of implementing `TryFrom` so you can return a `Result` 🤔

src/ws/coding.rs
```diff
@@ -0,0 +1,173 @@
...
+    #[doc(hidden)]
+    Tls,
+    #[doc(hidden)]
+    Reserved(u16),
+    #[doc(hidden)]
+    Iana(u16),
+    #[doc(hidden)]
+    Library(u16),
+    #[doc(hidden)]
+    Bad(u16),
```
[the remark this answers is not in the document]
gbaranski: https://datatracker.ietf.org/doc/html/rfc6455#section-7.4.1

they must not be used, but they exist for other reasons.
e.g for CloseCode::Tls
>    1015 is a reserved value and MUST NOT be set as a status code in a
      Close control frame by an endpoint.  It is designated for use in
      applications expecting a status code to indicate that the
      connection was closed due to a failure to perform a TLS handshake
      (e.g., the server certificate can't be verified).

src/ws/coding.rs
gbaranski: Maybe they didn't want to confuse things with CloseCode::Error, no idea.

src/ws/coding.rs
```diff
@@ -0,0 +1,173 @@
+//! Various codes defined in RFC 6455.
+//
+// This code comes from https://github.com/snapview/tungstenite-rs/blob/master/src/protocol/frame/coding.rs
```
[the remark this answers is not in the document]
gbaranski: Should we include MIT or Apache license? https://github.com/snapview/tungstenite-rs repo has both.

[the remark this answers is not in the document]
davidpdrsn: Let's go with MIT.

src/ws/coding.rs
davidpdrsn: Perhaps. Don't think it's a big deal.

davidpdrsn (approved):
Sweet! Thanks for working on this.

davidpdrsn:
Wanna try to merge `main`? I've merged a bunch of PRs.

src/ws/coding.rs
```diff
@@ -0,0 +1,202 @@
...
+// THE SOFTWARE.
+
+use std::{
+    convert::{From, Into},
+    fmt,
+};
+
+/// Status code used to indicate why an endpoint is closing the WebSocket connection.
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+pub enum CloseCode {
```
davidpdrsn, line 32: A thought I had regarding this.

What if instead of axum exposing the `CloseCode`, we instead exposed a `u16` and then users could convert that into a `CloseCode` using [`From<u16>` for `tungstenite::protocol::frame::coding::CloseCode`](https://docs.rs/tungstenite/0.14.0/tungstenite/protocol/frame/coding/enum.CloseCode.html#impl-From%3Cu16%3E). Then we wouldn't have to copy over all this stuff and maintain it going forward.

What do you think?

gbaranski: Hmm, that's good idea. I'll implement this in a while.

gbaranski:
> Wanna try to merge `main`? I've merged a bunch of PRs.

Done.

davidpdrsn (approved):
Awesome! Wanna add a line to the changelog as well?

Evil-li:
I hope Message enum support Json ，not merely Text

Issue #82, opened 2021-08-02
Query extractor given an bad request error when query string is empty


11nk:
```rust
#[derive(Deserialize)]
pub struct Pagination {
    pub size: Option<u64>,
    pub page: Option<u64>,
}
```
in such a case, the query string maybe empty.

programatik29:
You can make query extractor optional.

```rust
use std::net::SocketAddr;
use axum::prelude::*;
use serde::Deserialize;

#[derive(Deserialize)]
struct Pagination {
    size: Option<u64>,
    page: Option<u64>,
}

#[tokio::main]
async fn main() {
    let app = route("/", get(handler));

    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
    hyper::Server::bind(&addr)
        .serve(app.into_make_service())
        .await
        .unwrap();
}

async fn handler(p: Option<extract::Query<Pagination>>) -> String {
    let p = match p {
        Some(p) => p.0,
        None => return String::new(),
    };

    let mut s = String::new();

    if let Some(size) = p.size {
        s.push_str(&format!("size: {}\n", size));
    }

    if let Some(page) = p.page {
        s.push_str(&format!("page: {}\n", page));
    }

    s
}
```

davidpdrsn:
Fixed in https://github.com/tokio-rs/axum/pull/117

PR #117, opened 2021-08-04
Fix query extractor given an bad request error when query string is empty.

sunli829:
Related: #82

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for WebSocket protocol negotiation. ([#83](https://github.com/tokio-rs/axum/pull/83))
 - Use `pin-project-lite` instead of `pin-project`. ([#95](https://github.com/tokio-rs/axum/pull/95))
 - Re-export `http` crate and `hyper::Server`. ([#110](https://github.com/tokio-rs/axum/pull/110))
+- Fix `Query` and `Form` extractors giving bad request error when query string is empty. ([#117](https://github.com/tokio-rs/axum/pull/117))
 
 ## Breaking changes
 
--- a/src/extract/form.rs
+++ b/src/extract/form.rs
@@ -50,20 +50,20 @@ where
 
     #[allow(warnings)]
     async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
-        if !has_content_type(&req, "application/x-www-form-urlencoded")? {
-            Err(InvalidFormContentType)?;
-        }
-
         if req.method().ok_or(MethodAlreadyExtracted)? == Method::GET {
             let query = req
                 .uri()
                 .ok_or(UriAlreadyExtracted)?
                 .query()
-                .ok_or(QueryStringMissing)?;
+                .unwrap_or_default();
             let value = serde_urlencoded::from_str(query)
                 .map_err(FailedToDeserializeQueryString::new::<T, _>)?;
             Ok(Form(value))
         } else {
+            if !has_content_type(&req, "application/x-www-form-urlencoded")? {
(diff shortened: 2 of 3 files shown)


src/extract/query.rs
```diff
@@ -51,7 +51,7 @@ where
             .uri()
             .ok_or(UriAlreadyExtracted)?
             .query()
-            .ok_or(QueryStringMissing)?;
+            .unwrap_or_default();
```
sunli829, line 54: I don't think it is necessary to judge that the query string is empty, because `serde_urlencoded` will handle it. 🙂

davidpdrsn: Yeah that makes sense.

I guess `extract::Form` has the same bug then. Wanna look at that as well? The fix should be identical.

We might be able to deprecate `QueryStringMissing` as well then since its not used anywhere else.

sunli829: I will check it out now.🙂

src/extract/form.rs
```diff
@@ -50,20 +50,20 @@ where
...
                 .uri()
                 .ok_or(UriAlreadyExtracted)?
                 .query()
-                .ok_or(QueryStringMissing)?;
+                .unwrap_or_default();
             let value = serde_urlencoded::from_str(query)
                 .map_err(FailedToDeserializeQueryString::new::<T, _>)?;
             Ok(Form(value))
         } else {
+            if !has_content_type(&req, "application/x-www-form-urlencoded")? {
```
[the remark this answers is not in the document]
davidpdrsn: I think you're right 👍

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for WebSocket protocol negotiation. ([#83](https://github.com/tokio-rs/axum/pull/83))
 - Use `pin-project-lite` instead of `pin-project`. ([#95](https://github.com/tokio-rs/axum/pull/95))
 - Re-export `http` crate and `hyper::Server`. ([#110](https://github.com/tokio-rs/axum/pull/110))
+- Fix query extractor given an bad request error when query string is empty. ([#117](https://github.com/tokio-rs/axum/pull/117))
```
davidpdrsn, line 14:
```suggestion
- Fix `Query` and `Form` extractors giving bad request error when query string is empty. ([#117](https://github.com/tokio-rs/axum/pull/117))
```

davidpdrsn (approved):
Great! And thanks for adding tests 😊

One quick change to the wording of the changelog but otherwise it looks good!

sunli829:
I removed `QueryStringMissing` because it is no longer used.

src/extract/rejection.rs
```diff
@@ -39,13 +39,6 @@ define_rejection! {
     pub struct HeadersAlreadyExtracted;
 }
 
-define_rejection! {
-    #[status = BAD_REQUEST]
-    #[body = "Query string was invalid or missing"]
-    /// Rejection type for [`Query`](super::Query).
-    pub struct QueryStringMissing;
-}
```
davidpdrsn, line 47: We can't remove this as thats a breaking change. We can deprecate it

```rust
define_rejection! {
    #[status = BAD_REQUEST]
    #[body = "Query string was invalid or missing"]
    #[deprecated(since = "0.1.3", note = "No longer used. See https://github.com/tokio-rs/axum/pull/117")]
    /// Rejection type for [`Query`](super::Query).
    pub struct QueryStringMissing;
}
```

and then remove it in 0.2

sunli829:
`#[deprecated]` caused some nasty warnings, I don't know what to do. 😂

davidpdrsn:
Alright. Then lets just merge this without the deprecation. Then I'll look into it.

davidpdrsn (approved):

PR #118, opened 2021-08-04
Replace `hyper::Server` with `axum::Server` in docs

davidpdrsn:
Hopefully makes it easier to get started for some.

Intentionally didn't change README.md because I wanna make it clear that axum directly runs no hyper, which isn't clear unless you can see from the docs that `axum::Server` is a re-export.

### DIFF
--- a/examples/multipart_form.rs
+++ b/examples/multipart_form.rs
@@ -21,7 +21,7 @@ async fn main() {
     // run it with hyper
     let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
     tracing::debug!("listening on {}", addr);
-    hyper::Server::bind(&addr)
+    axum::Server::bind(&addr)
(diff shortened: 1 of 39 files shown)


examples/multipart_form.rs
```diff
@@ -21,7 +21,7 @@ async fn main() {
     // run it with hyper
```
jplatte, line 21: Maybe also standardize these comments across the examples? :)

jplatte (commented):
Maybe for the readme also use `axum::Server` but add a comment above that it's simply a re-export of the hyper type? I could imagine people copy-paste from readme to get started.

examples/multipart_form.rs
davidpdrsn: Sure!

Issue #109, opened 2021-08-04
Ability to do more complex route matching


mkieblesz:
### Motivation

- full path match
- multiple groups in a single path segment

### Proposal

There could be a lot of ways of doing this, but the idea is basically the same - allow to pass custom regex to each capture group in route string. Imo best would be to allow to specify name of the the passed group (lookup custom converter in flask). These groups would be predefined in some sort of CustomConverterMap<string, Regex> and somehow passed to the route function (wip).

```rust
# how it is now
route("/greet/:id", handler)

# how it could be
route("/greet/<int:id>", handler)
```

How others are doing it: 
- flask way: https://werkzeug.palletsprojects.com/en/2.0.x/routing/#custom-converters
- django way: https://docs.djangoproject.com/en/3.2/ref/urls/#re-path
- rails way: https://guides.rubyonrails.org/routing.html#segment-constraints

### Alternatives

Probably the only alternative is changing the spec for url patterns of the application.

davidpdrsn:
Note to self: Make sure we don't add something that might conflict with https://github.com/tokio-rs/axum/issues/50

mkatychev:
To follow up on https://github.com/tokio-rs/axum/pull/122#event-5117809154 here. Would it be reasonable to accommodate enums for the route path? 

I think regex is a very solid foundation but it would be great to allow matching on request method + uri. I realize looping through route aliases is probably not the most efficient way of doing grouping so it would be good to have some functional control of path + method similar to how `Request<Body>`:
```rust
match (req.method(), req.uri().path()) {
    (Method::GET, "/" | "/home") => Some("/" get(home_handler)),
    (Method::POST, "/" | "/home") => Some("/" post(home_handler)),
    _ => (None),
}
```

This can also help avoid precedent logic: https://github.com/tokio-rs/axum/issues/62#issuecomment-890380738.

I think the current pattern of `.route(path, method(handler))` too strictly bisects the routing logic. Something silimar to how the `Handler` trait allows closures to be passed in would go a long way into allowing simple yet flexible routing:
```rust
let app = route("/", get(|| async { Ok::<_, Infallible>(StatusCode::OK) }));
```

davidpdrsn:
Here are my thoughts:

axum's current routing consisting of fixed paths (`route("/foo/bar", ...)`) or named parameters (`route("/users/:id", ...)`) is honestly all I've ever needed, so thats why axum shipped with that. However I am open to adding more flexible routing and agree that it has great utility.

[...]

PR #122, opened 2021-08-04
Add support for multiple endpoints using a given handler 

mkatychev:
## Motivation
This PR implements the bullet point below https://github.com/tokio-rs/axum/issues/109:
> * multiple groups in a single path segment

## Solution

* Added `RoutePath` to be used en lieu of `&str` in `route`:

https://github.com/tokio-rs/axum/blob/af3dbeaea604a83eb184594773d0e84a5ab85175/src/routing.rs#L76-L83

```diff
- pub fn route<S, B>(description: &str, service: S) -> Route<S, EmptyRouter<S::Error>>
+ pub fn route<S, B, P>(description: P, service: S) -> Route<S, EmptyRouter<S::Error>>
```

* Moved `route::PathPattern` visibility from `pub(crate)` to `pub`


mkatychev:
From the guidelines:
> Bug fixes and new features should include tests.

I would include tests but latest `main` is failing quit a few tests for me. 
Happy to move this PR into draft.

src/lib.rs
```diff
@@ -790,9 +790,10 @@ pub mod prelude {
 /// # Panics
 ///
 /// Panics if `description` doesn't start with `/`.
-pub fn route<S, B>(description: &str, service: S) -> Route<S, EmptyRouter<S::Error>>
+pub fn route<S, B, P>(description: P, service: S) -> Route<S, EmptyRouter<S::Error>>
```
mkatychev, line 793: should `description` be changed to `path`? Description is usually inert metadata but `P` in this case is used for routing logic.

davidpdrsn:
Thanks for doing this!

However I'm not comfortable merging this. I would like to spend more time researching and comparing other frameworks and then designing an API that solves as much of https://github.com/tokio-rs/axum/issues/109 as possible - while enabling future optimizations to axum's routing we might want to make.

While I think your solution is good it feels a bit like a bandaid and I'm worried it might conflict with a more flexible routing API that we design for https://github.com/tokio-rs/axum/issues/109.

Hope you understand 😊

By the way, a general OSS tip is to first ask a maintainer if they would be open to a change before you start working on a PR. Then you don't run the risk of writing something that doesn't fit with where the authors want to take the project.

davidpdrsn:
> I would include tests but latest main is failing quit a few tests for me.

Oh and `cargo test --all-features` should do the trick (works on my machine).

mkatychev:
> By the way, a general OSS tip is to first ask a maintainer if they would be open to a change before you start working on a PR. Then you don't run the risk of writing something that doesn't fit with where the authors want to take the project.

No worries, I worked on this out of curiosity and to get familiar working with the codebase so no time lost.

PR #123, opened 2021-08-04
Set RUST_LOG environment var for all examples using tracing

spencergilbert:
Signed-off-by: Spencer Gilbert <spencer.gilbert@gmail.com>

## Motivation

As discussed in Discord, it would be nice to configure tracing for all of the examples
provided.

## Solution

I added environment variable based filtering (with fallbacks) at what seemed reasonable
for each example. I could add additional instructions on setting RUST_LOG in the readme
if desired.

### DIFF
--- a/examples/multipart_form.rs
+++ b/examples/multipart_form.rs
@@ -14,7 +14,7 @@ use std::net::SocketAddr;
 async fn main() {
     // Set the RUST_LOG, if it hasn't been explicitly defined
     if std::env::var("RUST_LOG").is_err() {
-        std::env::set_var("RUST_LOG", "multipart_form=debug")
+        std::env::set_var("RUST_LOG", "multipart_form=debug,tower_http=debug")
     }
     tracing_subscriber::fmt::fmt()
         .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
--- a/examples/multipart_form.rs
+++ b/examples/multipart_form.rs
@@ -12,7 +12,13 @@ use std::net::SocketAddr;
 
 #[tokio::main]
 async fn main() {
-    tracing_subscriber::fmt::init();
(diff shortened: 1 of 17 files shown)


examples/multipart_form.rs
```diff
@@ -12,7 +12,13 @@ use std::net::SocketAddr;
 
 #[tokio::main]
 async fn main() {
-    tracing_subscriber::fmt::init();
+    // Set the RUST_LOG, if it hasn't been explicitly defined
+    if std::env::var("RUST_LOG").is_err() {
+        std::env::set_var("RUST_LOG", "multipart_form=debug")
```
davidpdrsn, line 17:
```suggestion
        std::env::set_var("RUST_LOG", "multipart_form=debug,tower_http=debug")
```

Since the example is using `TraceLayer`

davidpdrsn (commented):
Thanks! I bet this is going to help at least a few 😊

lnicola:
I think `.with_env_filter(tracing_subscriber::EnvFilter::from_default_env())` is quite redundant here. `try_init` already does that.

davidpdrsn:
@lnicola wanna submit a follow up PR?

lnicola:
I will, but I can't right now.

examples/multipart_form.rs
spencergilbert: Oh I missed that this had any tower_http wiring, thanks!

spencergilbert:
Ah, TIL - if you don't get to it @lnicola I can pick it up tonight

lnicola:
Yeah, I asked because I thought it might be intentional.

davidpdrsn:
I cannot keep up with all the ways to initialize tracing subscribers honestly

Issue #42, opened 2021-07-30
Extractor similar to `actic_web::web::Path`


davidpdrsn:
Axum already has [`UrlParams`](https://docs.rs/axum/0.1.1/axum/extract/struct.UrlParams.html) and [`UrlParamsMap`](https://docs.rs/axum/0.1.1/axum/extract/struct.UrlParamsMap.html) for extracting data from request URIs. However I was recently made aware of [`actic_web::web::Path`](https://docs.rs/actix-web/3.3.2/actix_web/web/struct.Path.html) which seems more powerful since it relies on `serde::Deserialize`. I think that is something we should look into.

sunli829:
Maybe we can copy this file directly?

https://github.com/actix/actix-net/blob/master/actix-router/src/de.rs

davidpdrsn:
Yes I had the same thought 👍 

Would have to read through the code though and make sure it fits with the rest of axum. There are also some commented tests towards the bottom which we should look into.

And include their license at the top of file.

davidpdrsn:
I'm also curious if we even need [`UrlParams`](https://docs.rs/axum/0.1.1/axum/extract/struct.UrlParams.html) and [`UrlParamsMap`](https://docs.rs/axum/0.1.1/axum/extract/struct.UrlParamsMap.html) when this is merged. Would be nice if `Path` could do everything.

sunli829:
Yes, `Path` looks like it can replace `UrlParams` and `UrlParamsMap`, I will look into it tomorrow.

PR #124, opened 2021-08-05
Implement path extractor.

sunli829:
Fixes https://github.com/tokio-rs/axum/issues/42

### DIFF
--- a/src/extract/path/de.rs
+++ b/src/extract/path/de.rs
@@ -1,0 +1,671 @@
+use crate::routing::UrlParams;
+use crate::util::ByteStr;
+use serde::{
+    de::{self, DeserializeSeed, EnumAccess, Error, MapAccess, SeqAccess, VariantAccess, Visitor},
+    forward_to_deserialize_any, Deserializer,
+};
+use std::fmt::{self, Display};
+
+/// This type represents errors that can occur when deserializing.
+#[derive(Debug, Eq, PartialEq)]
+pub(crate) struct PathDeserializerError(pub(crate) String);
+
+impl de::Error for PathDeserializerError {
+    #[inline]
+    fn custom<T: Display>(msg: T) -> Self {
+        PathDeserializerError(msg.to_string())
+    }
+}
+
+impl std::error::Error for PathDeserializerError {
+    #[inline]
+    fn description(&self) -> &str {
+        "path deserializer error"
+    }
+}
+
+impl fmt::Display for PathDeserializerError {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            PathDeserializerError(msg) => write!(f, "{}", msg),
+        }
+    }
+}
+
+macro_rules! unsupported_type {
+    ($trait_fn:ident, $name:literal) => {
+        fn $trait_fn<V>(self, _: V) -> Result<V::Value, Self::Error>
+        where
+            V: Visitor<'de>,
+        {
+            Err(PathDeserializerError::custom(concat!(
+                "unsupported type: ",
+                $name
+            )))
+        }
+    };
+}
+
+macro_rules! parse_single_value {
+    ($trait_fn:ident, $visit_fn:ident, $tp:literal) => {
+        fn $trait_fn<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+        where
+            V: Visitor<'de>,
+        {
+            if self.url_params.0.len() != 1 {
+                return Err(PathDeserializerError::custom(
+                    format!(
+                        "wrong number of parameters: {} expected 1",
+                        self.url_params.0.len()
+                    )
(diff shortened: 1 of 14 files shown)


sunli829:
I want to replace `UrlParams` and `UrlParamsMap` with `Path` in all examples. What do you think? @davidpdrsn

davidpdrsn:
> I want to replace UrlParams and UrlParamsMap with Path in all examples. What do you think?

Yeah sounds great!

> And I did not mark UrlParams as deprecated, because this will also cause a lot of warnings, so I am going to leave this to you.

👍

sunli829:
I also modified all the tests for `UrlParams` and `UrlParamsMap` because they are no longer needed.

sunli829:
This is a description copied from `UrlParams`, but I found that there is an [`extracting_url_params_multiple_times`](https://github.com/tokio-rs/axum/blob/a8eb26b672eabfec9172b3d79cd9b9eb65fb7d7c/src/tests.rs#L283) test, so is it allowed to extract multiple times or only once?

sunli829:
I think I have done it, and did not directly copy the `de.rs` of `actix-web`. In fact, I understood the code and built it by hand, so I don't think it is necessary to add their license.

davidpdrsn:
Great work! I'll review this later today.

src/extract/path/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+/// use uuid::Uuid;
+///
+/// #[derive(Deserialize)]
+/// struct Request {
+///     user_id: Uuid,
+///     team_id: Uuid,
+/// }
+///
+/// async fn users_teams_show(
+///     Path(Request { user_id, team_id }): Path<(Uuid, Uuid)>,
```
davidpdrsn, line 60:
```suggestion
/// struct Params {
///     user_id: Uuid,
///     team_id: Uuid,
/// }
///
/// async fn users_teams_show(
///     Path(Params { user_id, team_id }): Path<Params>,
```

Request might be confusing since it conflicts with `http::Request`.

src/extract/path/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+        &mut self.0
+    }
+}
+
+impl<T> Path<T> {
+    /// Unwrap into inner `T` value.
+    #[inline]
+    pub fn into_inner(self) -> T {
+        self.0
+    }
```
davidpdrsn, line 94: I think since `Path`'s field is `pub` we should just leave this out.

src/extract/path/de.rs
```diff
@@ -0,0 +1,671 @@
...
+                        "wrong number of parameters: {} expected 1",
+                        self.url_params.0.len()
+                    )
+                    .as_str(),
+                ));
+            }
+
+            let value = self.url_params.0[0].1.parse().map_err(|_| {
+                PathDeserializerError::custom(format!(
+                    "can not parse {:?} to a {}",
```
davidpdrsn, line 68: nitpick:

```suggestion
                    "can not parse `{:?}` to a `{}`",
```

src/extract/path/de.rs
```diff
@@ -0,0 +1,671 @@
...
+                    self.url_params.0[0].1.as_str(),
+                    "bool"
```
davidpdrsn, line 70: Seems this shouldn't always be `"bool"` 🤔 

```suggestion
                    $tp
```

src/extract/path/de.rs
```diff
@@ -0,0 +1,671 @@
...
+
+macro_rules! parse_value {
+    ($trait_fn:ident, $visit_fn:ident, $ty:literal) => {
+        fn $trait_fn<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+        where
+            V: Visitor<'de>,
+        {
+            let v = self.value.parse().map_err(|_| {
+                PathDeserializerError::custom(format!(
+                    "can not parse {:?} to a {}",
```
davidpdrsn, line 328:
```suggestion
                    "can not parse `{:?}` to a `{}`",
```

davidpdrsn (commented):
I'm really excited for this! Using `Path` is so much nicer than `UrlParams` and `UrlParamsMap`!

I had a few minor comments but otherwise I think this looks good. Great job!

sunli829:
I have changed these. 🙂

davidpdrsn (approved):

davidpdrsn:
By the way small thing: I would prefer if you didn't force push. It makes it a little easier to review exactly whats changed in the commits 😊

sunli829:
Well, I will not force push again after your review next time. 🙂

johannescpk:
> It makes it a little easier to review exactly whats changed in the commits

Since GitHub doesn't make it obvious and some people seem to not know about: It's possible to click on the "force-push" text to see the diff of the force-pushed changed. Tho that doesn't replace follow-up changes with a commit message of course

PR #126, opened 2021-08-05
docs: add community showcase

gbaranski:
Something similar to what [actix](https://github.com/actix/examples) has.


davidpdrsn (approved):
Thanks!

PR #128, opened 2021-08-05
Simplify `tracing-subscriber` initialization

lnicola:
Addresses https://github.com/tokio-rs/axum/pull/123#issuecomment-893456697

## Motivation

This is the default, so let's simplify the code a bit. Addresses


davidpdrsn (approved):
Thanks!

Issue #84, opened 2021-08-02
handles `HEAD` requests automatically when there exists a `GET` route that would otherwise match.  `405 Method Not Allowed` returns for now.


songww:
### Proposal

handles HEAD requests automatically when there exists a GET route that would otherwise match. It does this by stripping the body from the response, if there is one. You can also specialize the handling of a HEAD request by declaring a route for it;  won't interfere with HEAD requests your application explicitly handles.

### Alternatives

Just like rocket did.

[head-requests](https://rocket.rs/v0.5-rc/guide/requests/#head-requests)

davidpdrsn:
I've looked into this and I don't think a clean solution is possible given the current design of the router.

When you compose two routes with `head(handler).get(other_handler)` neither `head` nor `get` knows if they're the "last" route. That is to say a `get` route has no way to know if there is another route that accepts `head` that is also defined and so it cannot change its behavior based on that.

One possible solution is to make `get` accept both `GET` and `HEAD` request but always remove the response body on `HEAD`. This is a breaking change however since `head(handler).get(other_handler)` would now call `other_handler` instead of `handler` on `HEAD` requests (routes are matched bottom to top).

Not sure if entirely removing explicit matching on `HEAD` requests would make things less confusing. That way `get` could always accept `GET` and `HEAD` without a `head` method getting in the way. I think that is quite drastic though, and removing the ability to match on `HEAD` feels odd. You can also get around it using `any` I suppose.

I checked with warp (the framework axum is most alike from a design perspective) and it handles `HEAD` requests the same way as axum does today. So `warp::get().map(|| "Hi there!");` does _not_ accept `HEAD` requests. It seems [an issue](https://github.com/seanmonstar/warp/issues/882) was opened about that recently so lets see what happens there.

Honestly I would be fine with leaving things as is. User's who want to support `HEAD` requests can explicitly add handlers for them. In my experience their usage is fairly niche. Don't think I've ever personally sent a `HEAD` request 🤔 At least not knowingly.

What do you think @songww @jplatte?

jplatte:
> Don't think I've ever personally sent a `HEAD` request thinking

REST services used from browsers commonly need `HEAD` implemented for all endpoints AFAIK (though there's also many that don't, and I really should look into how that works). But it's really wasteful to hit the DB to generate a response body when the response is then stripped and only the headers returned, so the right thing would probably be to implement head manually, or have a macro that generates both head and get impl from one function, i.e. sth. like

```rust
[...]

PR #129, opened 2021-08-05
Correctly handle HEAD requests

davidpdrsn:
HEAD requests will be routed to GET handlers, unless an explicit HEAD handler exists:

```rust
// this will receive `GET /` and `HEAD /` and send both to `handler`
route("/", get(handler));

// `GET /` goes to `handler`, `HEAD /` goes to `other_handler`
route("/", get(handler).head(other_handler));
```

Reponses to `HEAD` requests will always have the body stripped.

Fixes https://github.com/tokio-rs/axum/issues/84

### DIFF
--- a/src/service/future.rs
+++ b/src/service/future.rs
@@ -6,7 +6,8 @@ use crate::{
 };
 use bytes::Bytes;
 use futures_util::ready;
-use http::{Request, Response};
+use http::{Method, Request, Response};
+use http_body::Empty;
 use pin_project_lite::pin_project;
 use std::{
     future::Future,
@@ -85,6 +86,7 @@ pin_project! {
     {
         #[pin]
         pub(super) inner: crate::routing::future::RouteFuture<S, F, B>,
+        pub(super) req_method: Method,
     }
 }
 
@@ -96,6 +98,13 @@ where
     type Output = Result<Response<BoxBody>, S::Error>;
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        self.project().inner.poll(cx)
+        let this = self.project();
+        let response = futures_util::ready!(this.inner.poll(cx))?;
+        if this.req_method == &Method::HEAD {
+            let response = response.map(|_| box_body(Empty::new()));
+            Poll::Ready(Ok(response))
+        } else {
+            Poll::Ready(Ok(response))
+        }
     }
 }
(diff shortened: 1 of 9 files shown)


davidpdrsn:
@jplatte @songww I looked a bit more into https://github.com/tokio-rs/axum/issues/84 and think I found a fix. Are you able to test this and see if its working correctly for you?

It unfortunately required some breaking changes so will be released in 0.2.

src/service/future.rs
```diff
@@ -50,3 +55,40 @@ where
...
+    S: Service<Request<B>, Response = Response<BoxBody>>,
+    F: Service<Request<B>, Response = Response<BoxBody>, Error = S::Error>,
+{
+    type Output = Result<Response<BoxBody>, S::Error>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let mut res: Response<_> = ready!(this.f.poll(cx)?);
+
+        // HEAD must not contain a body according to
```
songww, line 84: > A response to a HEAD method should not have a body. If it has one anyway, that body must be ignored: any representation headers that might describe the erroneous body are instead assumed to describe the response which a similar GET request would have received.

see: [mdn](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD)

It is useful for request part of a document that the server should return. 
For file download with Range header, it tells how large the document is.

Sorry my english is poor.

davidpdrsn: Good catch! Fixed in 4c1a1db6d69c8

davidpdrsn:
@jplatte would you be able to give this a try?

jplatte:
Nope, sorry. I'm not actually using axum in any projects yet 😄

davidpdrsn:
Ah okay. Fair enough 😊

KaiJewson:
I don't like how magic this solution feels. For simplicity's sake, I think we would be better off making `get` accept both `GET` and `HEAD`, and `head(handler).get(other_handler)` not being supported. If users really wanted to list their routes in that order, we maybe could support `head(handler).or(get(other_handler))` or something.

davidpdrsn:
What about this solution is too magical?

> I think we would be better off making get accept both GET and HEAD

That's basically what this does except it also removes response bodies from HEAD requests, as per the spec.

KaiJewson:
> What about this solution is too magical?

It automatically disables the head-matching part of `get` when it detects a `head` filter present. This is kind of opaque to the user, and somewhat counterintuitive if they expected each `OnMethod` to be fully modular.

> That's basically what this does except it also removes response bodies from HEAD requests, as per the spec.

I definitely agree that response bodies should be removed from HEAD requests. My suggestion is to keep `get` fully consistent in its behaviour, so that it _always_ matches both GET and HEAD even if there are other HEAD handlers. Users can handle HEADs explicitly by checking for HEAD before the `get` filter: `get(handler).head(other_handler)`.

davidpdrsn:
> It automatically disables the head-matching part of get when it detects a head filter present. This is kind of opaque to the user, and somewhat counterintuitive if they expected each OnMethod to be fully modular.

I agree with that actually 🤔 I'm gonna change it to work like that when getting up to date with `main` I think.

PR #140, opened 2021-08-06
Common JSON wrapper type for response and request.

### LINKED ISSUE #102: Common JSON wrapper type for response and request
Currently there is `axum::extract::Json` and `axum::response::Json`, this introduces naming conflicts if handler accepts Json and returns Json.

Actix have `actix_web::web::Json` which can do at the same time.

sunli829:
Fixes #102


jplatte (commented):
This doesn't expose the `Json` type at the crate root as described in https://github.com/tokio-rs/axum/issues/102#issuecomment-894100691

sunli829:
> This doesn't expose the `Json` type at the crate root as described in [#102 (comment)](https://github.com/tokio-rs/axum/issues/102#issuecomment-894100691)

Yes, I think maybe exporting in `prelude` is better? 🙂

jplatte:
I don't think so personally, but not my decision anyways 🙂

davidpdrsn:
I think we should export it from the crate level. Doesn't belong in the prelude imo.

davidpdrsn:
Arguably we should slim down the prelude but thats a separate topic.

sunli829:
Okay, I changed it. 😂

jplatte (commented):
Tested this locally, you'll have to change the re-exports in `response` and `extract` to

```rust
#[doc(no_inline)]
pub use crate::Json;
```

to have them show up as re-exports of the type in the crate root / to make the crate root re-export the "canonical" path (which I understood was the intended way to display this in the docs).

src/lib.rs
```diff
@@ -739,6 +739,7 @@ pub use http;
 pub use hyper::Server;
 pub use tower_http::add_extension::{AddExtension, AddExtensionLayer};
 
+#[doc(no_inline)]
 pub use crate::json::Json;
```
jplatte, line 743: Sorry if it wasn't clear, but this should be the canonical definition without the attribute; the re-exports in `response.rs` and `extract.rs` should re-export this one (rather than the original definition) with the `doc(no_inline)` attribute.

sunli829: So all re-export types should be added with `#[doc(no_inline)]`?

sunli829:
It seems reasonable now.🙂

jplatte:
Yeah this is what I was thinking of.

src/json.rs
```diff
@@ -0,0 +1,136 @@
...
+use hyper::Response;
+use serde::{de::DeserializeOwned, Serialize};
+use std::ops::Deref;
+
+/// JSON Extractor/Response
+///
+/// When used as an Extractor, it can deserializes request bodies into some type that
+/// implements [`serde::Serialize`]. If the request bodies cannot be parsed, or it does not contain
+/// the `Content-Type: application/json`t header, will reject the request and return a
+/// `400 Bad Request` response.
```
davidpdrsn, line 20:
```suggestion
/// JSON Extractor/Response
///
/// When used as an extractor, it can deserialize request bodies into some type that
/// implements [`serde::Serialize`]. If the request body cannot be parsed, or it does not contain
/// the `Content-Type: application/json` header, it will reject the request and return a
/// `400 Bad Request` response.
```

src/json.rs
```diff
@@ -0,0 +1,136 @@
...
+/// use serde::Deserialize;
+///
+/// #[derive(Deserialize)]
+/// struct CreateUser {
+///     email: String,
+///     password: String,
+/// }
+///
+/// async fn create_user(payload: extract::Json<CreateUser>) {
+///     let payload: CreateUser = payload.0;
```
davidpdrsn, line 35:
```suggestion
/// async fn create_user(extract::Json(payload): extract::Json<CreateUser>) {
///     // payload is a `CreateUser`
```

src/json.rs
```diff
@@ -0,0 +1,136 @@
...
+///     // ...
+/// }
+///
+/// let app = route("/users", post(create_user));
+/// # async {
+/// # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
+/// # };
+/// ```
+///
+/// When used as a `Response`, it can serialize any type that implements [`serde::Serialize`] to `JSON`,
```
davidpdrsn, line 46:
```suggestion
/// When used as a response, it can serialize any type that implements [`serde::Serialize`] to `JSON`,
```

src/json.rs
```diff
@@ -0,0 +1,136 @@
...
+///     "data": 42,
+/// });
+///
+/// let response: Response<Body> = Json(json).into_response();
+///
+/// assert_eq!(
+///     response.headers().get(CONTENT_TYPE).unwrap(),
+///     "application/json",
+/// );
+/// ```
```
davidpdrsn, line 66: I think this example should show creating a handler function and returning `impl IntoResponse`. Like the other example above. I imagine thats how most people will use `Json` as a response.

I realize you just copied the old example but I still think it would be nice to change.

src/json.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+}
+
+impl<T> Deref for Json<T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
davidpdrsn, line 105: Should we add `DerefMut` while we're at it? Just for good measure 😊

src/json.rs
```diff
@@ -0,0 +1,136 @@
...
+use std::ops::Deref;
+
+/// JSON Extractor/Response
+///
+/// When used as an Extractor, it can deserializes request bodies into some type that
+/// implements [`serde::Serialize`]. If the request bodies cannot be parsed, or it does not contain
+/// the `Content-Type: application/json`t header, will reject the request and return a
+/// `400 Bad Request` response.
+///
+/// # Example
```
davidpdrsn, line 22:
```suggestion
/// # Extractor example
```

src/json.rs
```diff
@@ -0,0 +1,136 @@
...
+/// let app = route("/users", post(create_user));
+/// # async {
+/// # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
+/// # };
+/// ```
+///
+/// When used as a `Response`, it can serialize any type that implements [`serde::Serialize`] to `JSON`,
+/// and will automatically set `Content-Type: application/json` header.
+///
+/// # Example
```
davidpdrsn, line 49:
```suggestion
/// # Response example
```

davidpdrsn (commented):
Looks good! I had a few minor comments about the docs.

src/json.rs
sunli829: It should be added. 🙂

sunli829:
Thanks,  have done it. 🙂

davidpdrsn (approved):

PR #141, opened 2021-08-06
Cleanup

### LINKED ISSUE #115: CI doesn't seems to not actually run for MSRV
When setting up CI for axum I just copied the setup from Tower thinking that would be fine. However I didn't notice it included an MSRV of 1.40 [which is tested on CI](https://github.com/tokio-rs/axum/blob/main/.github/workflows/CI.yml#L15).

axum requires at least Rust 1.51 due the use of const generics in [`ContentLengthLimit`](https://github.com/tokio-rs/axum/blob/main/src/extract/content_length_limit.rs#L24). So it seems the 1.40 step on CI isn't actually running, since it should be failing 🤔 

We should fix CI so it actually does perform the checks and bump our MSRV to 1.51. If it turns out we use a feature not available on 1.51 I would be fine with requiring an even newer version, but it depends on the exact case.

jplatte:
Commit descriptions should explain all changes :)

Fixes #115.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,6 @@
 authors = ["David Pedersen <david.pdrsn@gmail.com>"]
 categories = ["asynchronous", "network-programming", "web-programming"]
 description = "Web framework that focuses on ergonomics and modularity"
-documentation = "https://docs.rs/axum/0.1.3"
 edition = "2018"
 homepage = "https://github.com/tokio-rs/axum"
 keywords = ["http", "web", "framework"]
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rust: [stable, 1.40.0]
+        rust: [stable, 1.51]
     steps:
     - uses: actions/checkout@master
     - uses: actions-rs/toolchain@v1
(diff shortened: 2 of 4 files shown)


jplatte:
This now removes one check from CI (second-to-last commit) but should be better than the status quo (due to #115).

davidpdrsn:
`create_url_params` is only called in tests. So changing the call sites to use a slice is fine.

Thanks for doing this by the way! Gonna review it over the next few days.

Cargo.toml
```diff
@@ -2,7 +2,6 @@
 authors = ["David Pedersen <david.pdrsn@gmail.com>"]
 categories = ["asynchronous", "network-programming", "web-programming"]
 description = "Web framework that focuses on ergonomics and modularity"
-documentation = "https://docs.rs/axum/0.1.3"
```
taiki-e, line 5: docs.rs selects the latest version automatically, but I don't think crate.io sets the docs URL automatically. ([example](https://crates.io/crates/ruma))

```diff
+ documentation = "https://docs.rs/axum"
```

jplatte: I think you're blocking the script from docs.rs that adds the link on crates.io :)

taiki-e: Oh, sorry! You are right.

davidpdrsn:
I'm gonna merge this and fix CI on main. Would like to have this in when I start merging some of the many open PRs.

PR #144, opened 2021-08-06
Add oauth2 example

fluunke:
## Motivation
In my projects I like to provide logins/sessions using oauth2, since it allows my friends to easily use the service.
This example is tailored to Discord, but can be easily modified to work with any oauth2 provider.

## Solution
I've added a new example which implements an oauth2 client, allowing the user to log in using Discord and save the session.
It shows how sessions can be optional or required in a handler.
Trying to access a restricted route when not logged in redirects the user to the login page. A route to destroy the session is also included :)


Cargo.toml
```diff
@@ -59,6 +59,10 @@ uuid = { version = "0.8", features = ["serde", "v4"] }
 async-session = "3.0.0"
 tokio-stream = "0.1.7"
 
+# for oauth example
+oauth2 = "4.1"
+http-types = "2.11.1"
```
davidpdrsn, line 64: Is it possible to use the things in `http` or `headers`?

`http-types` provides many of the same things as `http` so including both shouldn't be necessary.

examples/oauth.rs
```diff
@@ -0,0 +1,261 @@
+//! Example OAuth (Discord) implementation.
+//!
+//! Thanks to pbzweihander for their tide-google-oauth-example!
+//! It helped me understand oauth2.
+//!
+//! Run with
+//!
+//! ```not_rust
+//! CLIENT_ID=123123 CLIENT_SECRET=secret
+//! cargo run --example oauth --features=headers
```
davidpdrsn, line 10: This being one line makes it easier to copy.

```suggestion
//! CLIENT_ID=123123 CLIENT_SECRET=secret cargo run --example oauth --features=headers
```

examples/oauth.rs
```diff
@@ -0,0 +1,261 @@
+//! Example OAuth (Discord) implementation.
+//!
+//! Thanks to pbzweihander for their tide-google-oauth-example!
+//! It helped me understand oauth2.
```
davidpdrsn, line 4: I think we should remove this. Feels a bit out of place 😅

examples/oauth.rs
```diff
@@ -0,0 +1,261 @@
...
+}
+
+// Session is optional
+async fn index(user: Option<User>) -> impl IntoResponse {
+    match user {
+        Some(u) => format!(
+            "Hey {}! You're logged in!\nYou may now access `/protected`.\nLog out with `/logout`.",
+            u.username
+        ),
+        None => "You're not logged in.\nVisit `/auth/discord` to do so.".to_string(),
```
davidpdrsn, line 115: Should this redirect directly to `/auth/discord`?

examples/oauth.rs
```diff
@@ -0,0 +1,261 @@
...
+    }
+}
+
+pub(crate) async fn discord_auth(client: Extension<BasicClient>) -> impl IntoResponse {
```
davidpdrsn, line 119:
```suggestion
async fn discord_auth(client: Extension<BasicClient>) -> impl IntoResponse {
```

I don't all these functions need to be `pub(crate)`. Could confuse people.

examples/oauth.rs
```diff
@@ -0,0 +1,261 @@
...
+        .header("Location", "/")
+        .header(SET_COOKIE, cookie.to_string())
+        .status(302);
+
+    r.body(Body::empty()).unwrap()
+}
+
+// Utility to save some lines of code
+pub struct Redirect(String);
+impl<'a> IntoResponse for Redirect {
```
davidpdrsn, line 207:
```suggestion
impl IntoResponse for Redirect {
```

Doesn't seem to be used.

examples/oauth.rs
```diff
@@ -0,0 +1,261 @@
...
+    let r = http::Response::builder()
+        .header("Location", "/")
+        .header(SET_COOKIE, cookie.to_string())
+        .status(302);
+
+    r.body(Body::empty()).unwrap()
+}
+
+// Utility to save some lines of code
+pub struct Redirect(String);
```
davidpdrsn, line 206: I've actually thought about adding something like this directly to axum. Seems quite useful. Have to think about the API though. But its cool that you arrived at the same thing 😊

examples/oauth.rs
```diff
@@ -0,0 +1,261 @@
...
+            .unwrap()
+        {
+            Some(c) => c,
+            None => return Err(AuthRedirect),
+        };
+
+        let user = match session.get::<User>("user") {
+            Some(u) => u,
+            None => return Err(AuthRedirect),
+        };
```
davidpdrsn, line 257: These three `match`es could also be written using `.ok_or(AuthRedirect)?` which is a bit cleaner imo.

davidpdrsn (changes requested):
Cool! Thanks for adding this. I think this example shows nicely how to put the pieces together 😊

I had a few suggestions but nothing major.

Cargo.toml
fluunke: I couldn't find similar cookie-building features in `http` or `headers`, but I've replaced http-types with the more purpose-built `cookie` crate, which pulls in less deps.

examples/oauth.rs
fluunke: The index page demonstrates how to easily make a session optional :)

Cargo.toml
davidpdrsn: Nice 👍

davidpdrsn (approved):
Thanks!

davidpdrsn:
It seems some of the new dependencies have crates that are no longer maintained. I wonder if they have cargo features to turn those dependencies off 🤔

fluunke:
Dang, seems like `cookie` uses an old version of `time` which pulls that in (a PR for an update exists: https://github.com/SergioBenitez/cookie-rs/pull/179). I'd have to disable a feature of a dependency's dependency, which doesn't seem to be a thing :(

Should I just write out the cookie manually for the time being? :smile:

davidpdrsn:
> Should I just write out the cookie manually for the time being?

If it isn't too much work 😊

examples/oauth.rs
```diff
@@ -0,0 +1,245 @@
...
+struct AuthRequest {
+    code: String,
+    state: String,
+}
+
+async fn login_authorized(
+    query: Query<AuthRequest>,
+    store: Extension<MemoryStore>,
+    oauth_client: Extension<BasicClient>,
+) -> impl IntoResponse {
```
gbaranski, line 158: Wouldn't making structures responses be better?

So instead of returning `IntoResponse`, you can return a `Result<Json<T>, E>`, where T has Serialize/Deserialize implementation, and E has IntoResponse. That could also help removing those `.unwrap()`'s at the bottom, which aren't best way to handle errors.

examples/oauth.rs
```diff
@@ -0,0 +1,245 @@
...
+        .json::<User>()
+        .await
+        .unwrap();
+
+    // Create a new session filled with user data
+    let mut session = Session::new();
+    session.insert("user", &user_data).unwrap();
+
+    // Store session and get corresponding cookie
+    let cookie = store.0.store_session(session).await.unwrap().unwrap();
```
gbaranski, line 184:
```suggestion
    let cookie = store.store_session(session).await.unwrap().unwrap();
```
After applying my suggestion with argument destructing, this can be removed.

examples/oauth.rs
```diff
@@ -0,0 +1,245 @@
...
+
+#[derive(Debug, Deserialize)]
+struct AuthRequest {
+    code: String,
+    state: String,
+}
+
+async fn login_authorized(
+    query: Query<AuthRequest>,
+    store: Extension<MemoryStore>,
```
gbaranski, line 156:
```suggestion
    Extension(store): Extension<MemoryStore>,
```

You can destruct parameters inside of arguments, that can save you from additional `.0` on `store`

gbaranski (changes requested):
I have few suggestions about this PR.

btw, in my project [houseflow](https://github.com/gbaranski/houseflow/tree/master/server/src/oauth), I've already implemented OAuth2, however I'm still rewriting this part of codebase from actix-web to axum, but that could be used as a reference for now since the API of those both libraries are similar.

examples/oauth.rs
fluunke: Yes, unwraps should be avoided, but the example felt kind of long already and I didn't want to implement a whole error handler myself for simplicity's sake (at that point one should probably look at the proper error handling example).

examples/oauth.rs
```diff
@@ -0,0 +1,245 @@
...
+// Utility to save some lines of code
+struct Redirect(String);
+impl IntoResponse for Redirect {
+    fn into_response(self) -> http::Response<Body> {
+        let builder = http::Response::builder()
+            .header("Location", self.0)
+            .status(StatusCode::FOUND);
+        builder.body(Body::empty()).unwrap()
+    }
+}
```
davidpdrsn, line 207: We've some changes to `IntoResponse`.

I think this should work:

```suggestion
impl IntoResponse for Redirect {
    type Body = Body;
    type BodyError = type BodyError = <Self::Body as axum::body::HttpBody>::Error;
    
    fn into_response(self) -> http::Response<Body> {
        let builder = http::Response::builder()
            .header("Location", self.0)
            .status(StatusCode::FOUND);
        builder.body(Body::empty()).unwrap()
    }
}
```

Same for `AuthRedirection` below.

davidpdrsn (approved):

davidpdrsn:
Thanks for working on this!

PR #156, opened 2021-08-07
Fix `Uri` extractor not being the full URI if using `nest`

davidpdrsn:


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Make `FromRequest` default to being generic over `axum::body::Body` ([#146](https://github.com/tokio-rs/axum/pull/146))
 - Implement `std::error::Error` for all rejections ([#153](https://github.com/tokio-rs/axum/pull/153))
+- Fix `Uri` extractor not being the full URI if using `nest` ([#156](https://github.com/tokio-rs/axum/pull/156))
```
lnicola, line 12: This is a breaking change actually.

davidpdrsn: Right! Just realized it was also mentioned in the docs that it worked like this. I think adding an extractor that provides the old behavior might be nice.

PR #163, opened 2021-08-08
Support `multipart` conversion to `HashMap<String, Field>`

mrxiaozhuox:
## Motivation

When I use `axum` to try to get value from `form-data`, it need use `next_field + while` to get value.

But I think if the data can be make a `HashMap` will be more better.

for example: 

```rust
let map: HashMap<String, Field> = multipart.map();

let v = map.get("value").unwrap();

```

## Solution

I create a new function: `map`, it can get all `field` from `Multipart` and make it to a `HashMap`.

but if developer want to use `next_field`, it also ok.

if you have any suggest or problem, please tell me. thanks.

### DIFF
--- a/examples/multipart_form.rs
+++ b/examples/multipart_form.rs
@@ -60,10 +60,34 @@ async fn accept_form(
         },
     >,
 ) {
-    while let Some(field) = multipart.next_field().await.unwrap() {
-        let name = field.name().unwrap().to_string();
-        let data = field.bytes().await.unwrap();
 
-        println!("Length of `{}` is {} bytes", name, data.len());
+    let v = multipart.map().await.unwrap();
+
+    let text = v.0;
+    let file = v.1;
+
+    println!("Form-Data: Text\n");
+    for i in text {
+
+        println!("{}: {:?}", i.0, i.1);
+
     }
+
+    println!("\nForm-Data: File\n");
+    for i in file {
+
+        println!("{}: size({})", i.0, i.1.content.len());
+
+    }
+
+    println!();
+
+    // This is another way to use it
+    //
+    // while let Some(field) = multipart.next_field().await.unwrap() {
+    //     let name = field.name().unwrap().to_string();
+    //     let data = field.bytes().await.unwrap();
+    //
+    //     println!("Length of `{}` is {} bytes", name, data.len());
+    // }
 }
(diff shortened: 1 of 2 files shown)


src/extract/multipart.rs
```diff
@@ -80,6 +81,21 @@ impl Multipart {
...
+
+        let mut result = HashMap::new();
+
+        while let Some(mut field) = self.next_field().await? {
+            let name = field.name()?.to_string();
+            result.insert(name, field);
+        }
+
+        Ok(result)
+    }
```
davidpdrsn, line 97: Did you actually compile this? I don't think it compiles, or at least it _shouldn't_.

There can only be one `Field` live at any point, since `Field` contains `&mut Multipart`. Multiple `Field`s from the same `Multipart` would mean there are multiple `&mut Multipart` which is unsound. Here you're storing multiple `Field`s in the map at the same time. This is multers ["Field exclusivity"](https://docs.rs/multer/2.0.1/multer/struct.Multipart.html#field-exclusivity) enforced via lifetimes.

mrxiaozhuox: Sorry, I'll try to fix it. I just made a brief proposal, and I wonder if you have a problem with him.

examples/multipart_form.rs
```diff
@@ -60,10 +60,34 @@ async fn accept_form(
...
+    }
+
+    println!("\nForm-Data: File\n");
+    for i in file {
+
+        println!("{}: size({})", i.0, i.1.content.len());
+
     }
+
+    println!();
```
mrxiaozhuox, line 83: I have modify my code, now the function `map` will return a tuple: `(HashMap<String, String>, HashMap<String, UploadFile>)`

First element is all the `Text` field, and the second element is all the `File` field.

mrxiaozhuox (commented):
and i have try it in the `examples`, it's can be use. 😃😃😃

davidpdrsn:
> Sorry, I'll try to fix it. I just made a brief proposal, and I wonder if you have a problem with it.

Thats fine, but in the future please make sure your code compiles before you submit a PR. Its a little disrespectful of my time to submit a PR that you didn't test. You could also have opened an issue with the proposal or asked in `#axum` on the discord.

I've been thinking about this and I'm sorry but I don't think we're going to merge this. While I am open to adding utility methods to make multipart requests easier to use I don't think this API is right. Returning `(HashMap<String, String>, HashMap<String, UploadFile>)` is a very odd type that I would personally be surprised to find a crate. As the code isn't complicated I think its better for users to build whatever utilities is right for them, directly in their code.

Issue #50, opened 2021-07-31
Support generating OpenAPI/Swagger docs


hronro:
If you are wring a RESTful API service, the OpenAPI/Swagger documentation is very useful. It would be great if `axum` is able to generate them automatically.

Michael-J-Ward:
Just documenting some prior art in this area

- [dropshot](https://github.com/oxidecomputer/dropshot/blob/1c6f4d156c670edc6476ca5bcdffc33a7001cbc7/dropshot_endpoint/src/lib.rs#L3-L5) creaed by oxide computing
- [weave](https://github.com/jsdw/seamless) - which goes a step further and can generate typescript clients to consume the api

aesteve:
A (hopefully helpful?) source of inspiration in OpenAPI generation design: [tapir](https://tapir.softwaremill.com/en/latest/openapi.html) which is a Scala server/router comparable with a philosophy comparable to Axum (router / builder pattern and no macro / annotation stuff).

The parts I'm finding interesting in the design: it's all methods added to existing structures. You already have the router/server you built using the library, which knows about the routes, parameters, body types, etc. And there's a `toOpenAPI` method which generates the appropriate OpenAPI definition builder. This OpenAPI definition can then be exposed in YAML or JSON format, and mounted on the appropriate endpoint of your choice. But the OpenAPI definition being an object (let's assume it would be a struct in Rust) also allows for enhancing the docs with comments, descriptions, server information, etc. which is a really nice feature. Annotation (or macro, in rust) driven libraries tend to be quite tedious in this regard, where would you put the annotation for the global app description? The app doc version? You have to annotate something? But what? 

-----

A good first step could be to try to generate an OpenAPI definition from the existing structs in Axum.
For example: mapping endpoints to OpenAPI operations, using a set of Rust OpenAPI structs matching their concepts (Operation, Response, Schema, etc.) using existing crates like https://crates.io/crates/openapiv3 for instance .

From this point on, we could get a glimpse at what pieces of information are missing. 
For instance: examples for query parameters will be missing from the UrlParams extractor I suppose. Same goes for Request/Response bodies schemas, or operations (get/post/put/...) descriptions.

There is the hardest part imo, and could lead to creating another crate? 
Can extractors be re-worked, in some way, to include such missing information? Is it desirable? Will it make the crate too cluttered, or too complex? That's a very tough question at this point.

-----

A note on "re-usability". 
[...]

PR #170, opened 2021-08-08
very-much-work-in-progress: Support generating OpenAPI docs

davidpdrsn:
This will probably be a long running PR and will eventually contain some version of OpenAPI docs generation.

Nothing is set in stone, everything you see is work in progress, and I have no clue if/when we'll ship anything.

With that said here is an example:

```rust
use axum::{
    extract::{Extension, Query},
    open_api::{self, ToQueryParameter},
    prelude::*,
    response::IntoResponse,
    AddExtensionLayer, Json,
};
use openapiv3::OpenAPI;
use serde::Deserialize;
use std::sync::Arc;

#[tokio::main]
async fn main() {
    let app = route("/api/users", get(get_users).post(|| async {}));

    let open_api = open_api::to_open_api(&app);

    let app = app
        .route("/openapi.json", get(open_api_json))
        .layer(AddExtensionLayer::new(Arc::new(open_api)));

        .await
        .unwrap();
}

async fn open_api_json(Extension(open_api): Extension<Arc<OpenAPI>>) -> impl IntoResponse {
    Json(open_api)
}

async fn get_users(_: Query<Pagination>) -> &'static str {
    "users"
}

#[derive(Deserialize)]
struct Pagination {
    offset: Option<usize>,
    limit: Option<usize>,
}

// we're gonna need #[derive(ToQueryParameter)] for this
// way too much boilerplate here :/
impl ToQueryParameter for Pagination {
    fn to_query_parameter() -> open_api::Query {
        use openapiv3::*;

        let mut obj = ObjectType::default();

        let offset_schema_data = SchemaData {
            nullable: true,
            ..Default::default()
        };
        let offset_schema = Box::new(Schema {
            schema_data: offset_schema_data,
            schema_kind: SchemaKind::Type(Type::Number(NumberType::default())),
        });
        obj.properties
            .insert("offset".to_string(), ReferenceOr::Item(offset_schema));

        let limit_schema_data = SchemaData {
            nullable: true,
            ..Default::default()
        };
        let limit_schema = Box::new(Schema {
            schema_data: limit_schema_data,
            schema_kind: SchemaKind::Type(Type::Number(NumberType::default())),
        });
        obj.properties
            .insert("limit".to_string(), ReferenceOr::Item(limit_schema));

        let schema = Schema {
            schema_data: SchemaData::default(),
            schema_kind: SchemaKind::Type(Type::Object(obj)),
        };

        let parameter_data = ParameterData {
            name: "Pagination".to_string(),
            description: None,
            required: false,
            deprecated: Some(false),
            format: ParameterSchemaOrContent::Schema(ReferenceOr::Item(schema)),
            example: None,
            examples: Default::default(),
            explode: None,
            extensions: Default::default(),
        };

        open_api::Query {
            parameter_data,
            allow_reserved: false,
            style: QueryStyle::default(),
            allow_empty_value: Some(true),
        }
    }
}
```

This responds with

```json
{
  "openapi": "",
  "info": {
    "title": "",
    "version": ""
  },
  "paths": {
    "/api/users": {
      "get": {
        "parameters": [
          {
            "in": "query",
            "name": "Pagination",
            "deprecated": false,
            "schema": {
              "type": "object",
              "properties": {
                "offset": {
                  "nullable": true,
                  "type": "number"
                },
                "limit": {
                  "nullable": true,
                  "type": "number"
                }
              }
            },
            "style": "form",
            "allowEmptyValue": true
          }
        ],
        "responses": {
          "default": {
            "description": "Plain text",
            "content": {
              "text/plain": {
                "schema": {
                  "type": "string"
                },
                "encoding": {
                  "text/plain": {
                    "contentType": "text/plain"
                  }
                }
              }
            }
          }
        }
      },
      "post": {
        "responses": {
          "default": {
            "description": "Always empty"
          }
        }
      }
    }
  }
}
```

I haven't actually checked this against a client generating/linter but doesn't seem too horrible.

## Hard questions

- [ ] How do we support handlers with layers applied?
- [x] Is `BoxRoute` going to be problematic since it erases type info?
- [ ] How do we allow users to customize the schema?
- [ ] Path parameters
- [x] Operation ids?

Fixes https://github.com/tokio-rs/axum/issues/50

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ version = "0.1.3"
 default = []
 ws = ["tokio-tungstenite", "sha-1", "base64"]
 multipart = ["multer", "mime"]
+open_api = ["openapiv3", "indexmap"]
 
 [dependencies]
 async-trait = "0.1"
@@ -41,6 +42,8 @@ base64 = { optional = true, version = "0.13" }
 headers = { optional = true, version = "0.3" }
 multer = { optional = true, version = "2.0.0" }
 mime = { optional = true, version = "0.3" }
+openapiv3 = { optional = true, version = "0.5.0" }
+indexmap = { optional = true, version = "1.7" }
 
 [dev-dependencies]
 askama = "0.10.5"
@@ -111,3 +114,7 @@ required-features = ["headers"]
 [[example]]
 name = "oauth"
 required-features = ["headers"]
+
+[[example]]
+name = "open_api"
+required-features = ["open_api"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ tracing-subscriber = "0.2"
 uuid = { version = "0.8", features = ["serde", "v4"] }
 async-session = "3.0.0"
 tokio-stream = "0.1.7"
+serde_yaml = "0.8.17"
 
 # for oauth example
 oauth2 = "4.1"
--- a/examples/open_api.rs
+++ b/examples/open_api.rs
@@ -8,9 +8,10 @@
 
 use axum::{
     extract::{Extension, Query},
-    open_api::{self, ToQueryParameter},
+    open_api::{self, *},
     prelude::*,
     response::IntoResponse,
+    routing::BoxRoute,
     AddExtensionLayer, Json,
 };
 use openapiv3::OpenAPI;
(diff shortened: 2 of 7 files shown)


ufoscout:
@davidpdrsn 
I am not sure if you are aware of it or if it can be of any help, anyway, this crate permits to perform introspection of `Serde` objects: https://github.com/novifinancial/serde-reflection
It works at runtime by building a [container](https://novifinancial.github.io/serde-reflection/serde_reflection/enum.ContainerFormat.html) representation of the struct itself.
I guess it can be used to implement `ToQueryParameter` directly for the `Serialize/Deserialize` traits so for every struct that derives from them you could get automatic openapi description.

ufoscout:
FYI, I spent some time with [serde-reflection](https://github.com/novifinancial/serde-reflection) and I found some bad limitations with the support of some serde attributes (e.g. `#[serde(tag = "type")]`, `#[serde(flatten)]`) and types with generics.
At this point, I think a more interesting approach is the one used by [okapi](https://github.com/GREsau/okapi) which is based on the well established [schemars](https://github.com/GREsau/schemars) crate.

davidpdrsn:
I don't think I'll get time to work on this for 0.3. Routing is gonna be the focus for 0.3 so maybe openapi will be 0.4 (if if requires breaking changes). I don't like just having stale PRs open so I'll close this for now.

PR #171, opened 2021-08-09
Use tower http auth layer in kv store example

SuperFluffy:
With https://github.com/hyperium/http-body/pull/46 merged and released
in http-body 0.4.3, the key value store example can rely on tower_http's
`RequireAuthorizationLayer` instead of using the now removed custom
version (as per the TODO in the comment above it).

### DIFF
--- a/examples/key_value_store.rs
+++ b/examples/key_value_store.rs
@@ -7,8 +7,7 @@
 //! ```
 
 use axum::{
-    async_trait,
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):

davidpdrsn:
@SuperFluffy seems CI is failing.

SuperFluffy:
@davidpdrsn Strict rustfmt rules. Fixed.

PR #173, opened 2021-08-10
Reduce futures-util features

paolobarbolini:
## Motivation

`futures-util` enables the `async-await-macro` feature by default, which imports `futures-macro`, `proc-macro-hack` and `proc-macro-nested`, which are unnecessary and slow to compile.

## Solution

Disable unnecessary features.


davidpdrsn:
Thanks! Didn't realize!

PR #178, opened 2021-08-13
Simple example showing how to add CORS headers to responses

heliumbrain:
## Motivation

It would be nice to have a simple example to define CORS headers 

## Solution

CORS example 😄

### DIFF
--- a/examples/cors.rs
+++ b/examples/cors.rs
@@ -5,30 +5,29 @@
 //! ```
 
 use axum::prelude::*;
-use http::{HeaderValue, header};
-use tower::ServiceBuilder;
+use http::{header, HeaderValue};
 use std::net::SocketAddr;
+use tower::ServiceBuilder;
 use tower_http::set_header::SetResponseHeaderLayer;
 
 #[tokio::main]
 async fn main() {
     tracing_subscriber::fmt::init();
 
-	// define the CORS-headers you want to add as a Tower service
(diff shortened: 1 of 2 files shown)


jplatte:
This needs to be changed not to respond 'method not allowed' to OPTIONS requests, no?

examples/cors.rs
```diff
@@ -5,30 +5,29 @@
...
+        .layer(SetResponseHeaderLayer::<_, Body>::if_not_present(
+            header::ACCESS_CONTROL_ALLOW_ORIGIN,
+            HeaderValue::from_static("*"),
+        ))
+        .into_inner();
 
     // create application with a route and add the layer off middleware including a handler for OPTIONS requests.
-    let app = route("/", get(handler).options(|| async { "" }))
-	.layer(cors_middleware);
+    let app = route("/", get(handler).options(|| async { "" })).layer(cors_middleware);
```
jplatte, line 30: If done this way, the empty options handler has to be added for every single route. I think this should be done similarily to the 404 example, with a layer that checks `request.method() == Method::OPTIONS && response.status() == StatusCode::METHOD_NOT_ALLOWED` (not sure whether there's some util thing that will make implementing that easier).

davidpdrsn:
I think I'll close this for now. The [tower-http PR](https://github.com/tower-rs/tower-http/pull/112) should be merged soon-ish.

Issue #181, opened 2021-08-14
Add XHTML response


kellytk:
### Discussed in https://github.com/tokio-rs/axum/discussions/180

<div type='discussions-op-text'>

<sup>Originally posted by **kellytk** August 14, 2021</sup>
HTML response [hardcodes a `text/html` content type](https://github.com/tokio-rs/axum/blob/main/src/response/mod.rs#L414), however XHTML5 requires[0]:

> When a document is transmitted with an XML MIME type, such as application/xhtml+xml, then it is treated as an XML document by Web browsers, to be parsed by an XML processor.

Where please is it documented how that should be configured with `axum`?  If it isn't, modify the response's header in the handler, mostly duplicate [HTML response](https://github.com/tokio-rs/axum/blob/main/src/response/mod.rs#L398) to an XHTML response pull request, other?

[0] https://www.w3.org/TR/html51/introduction.html 1.6. HTML vs XHTML</div>

PR #182, opened 2021-08-14
Add XHTML response

kellytk:
Resolves https://github.com/tokio-rs/axum/issues/181.

### DIFF
--- a/src/response/mod.rs
+++ b/src/response/mod.rs
@@ -422,6 +422,33 @@ impl<T> From<T> for Html<T> {
     }
 }
 
+/// An XHTML response.
+///
+/// Will automatically get `Content-Type: application/xhtml+xml`.
+#[derive(Clone, Copy, Debug)]
+pub struct Xhtml<T>(pub T);
+
+impl<T> IntoResponse for Xhtml<T>
+where
+    T: Into<Full<Bytes>>,
+{
+    type Body = Full<Bytes>;
+    type BodyError = Infallible;
+
+    fn into_response(self) -> Response<Self::Body> {
+        let mut res = Response::new(self.0.into());
+        res.headers_mut()
+            .insert(header::CONTENT_TYPE, HeaderValue::from_static("application/xhtml+xml"));
+        res
+    }
+}
+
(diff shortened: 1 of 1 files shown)


davidpdrsn:
Rocket has something similar called Xml which gives a text/xml content type https://docs.rs/rocket/0.4.10/rocket/response/content/struct.Xml.html

What's the difference?

kellytk:
`application/xhtml+xml` is what's [required](https://github.com/tokio-rs/axum/discussions/180#discussion-3521428) for XHTML.

davidpdrsn:
Alright.

I'm actually not sure axum needs built-in support for Xhtml as it's used very rarely. See https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/XHTML

I think users putting this in their own apps, if they need it, is reasonable.

kellytk:
Would it be an option to support modifying a response's header in the handler?  I'd prefer to find a solution that doesn't involve lugging around diffs for a fairly fundamental task.

Re: "used very rarely", even if true, it's still a part of the latest HTML5 spec.

kellytk:
Another idea, would you be open to this PR if XHTML support was behind a feature flag?

davidpdrsn:
> Would it be an option to support modifying a response's header in the handler?

Possibly. If we can find the right api for it. You're welcome to make a proposal in an issue.

> I'd prefer to find a solution that doesn't involve lugging around diffs for a fairly fundamental task.

You can always put this in your own "axum-utils" crate and include that in all your projects.

> Re: "used very rarely", even if true, it's still a part of the latest HTML5 spec.

Some things can still be out of scope. I'd rather provide a good API that users can use to build what's right for their app, rather than building everything into axum. It's a question of balance and where to draw the line.

davidpdrsn:
> Another idea, would you be open to this PR if XHTML support was behind a feature flag?

I don't think so. I would like to reserve feature flags for features that require additional dependencies and/or would be difficult to build yourself. Such as websockets.

kellytk:
I'll make a separate crate, thanks.

PR #183, opened 2021-08-15
correct hyper link to add the .rs suffix.

ChillFish8:
This fixes the broken redirect for `tracing_aka_logging` which current redirects to what it thinks is a folder instead of a file, resulting in a 404.

This is not a code change.


davidpdrsn (approved):
Thanks!

PR #192, opened 2021-08-16
Add `Redirect` response

### LINKED ISSUE #188: Add `Redirect` response
### Motivation

The oauth example contains [a little `Redirect` utility](https://github.com/tokio-rs/axum/blob/main/examples/oauth.rs#L198-L210). I think that is quite useful and so we should move it to axum proper.

### Proposal

I think an API similar to tower-http's [`Redirect`](https://docs.rs/tower-http/0.1.1/tower_http/services/redirect/struct.Redirect.html) leaf service would be good. So something like

```rust
Redirect::temporary(uri)
Redirect::permanent(uri)
Redirect::with_status_code(uri, status)
```

davidpdrsn:
Fixes #188

### DIFF
--- a/examples/oauth.rs
+++ b/examples/oauth.rs
@@ -9,14 +9,13 @@
 use async_session::{MemoryStore, Session, SessionStore};
 use axum::{
     async_trait,
+    body::{Bytes, Empty},
     extract::{Extension, FromRequest, Query, RequestParts, TypedHeader},
     prelude::*,
-    response::IntoResponse,
+    response::{IntoResponse, Redirect},
     AddExtensionLayer,
 };
-use http::header::SET_COOKIE;
-use http::StatusCode;
-use hyper::Body;
+use http::{header::SET_COOKIE, HeaderMap};
 use oauth2::{
     basic::BasicClient, reqwest::async_http_client, AuthUrl, AuthorizationCode, ClientId,
     ClientSecret, CsrfToken, RedirectUrl, Scope, TokenResponse, TokenUrl,
@@ -118,7 +117,7 @@ async fn discord_auth(Extension(client): Extension<BasicClient>) -> impl IntoRes
         .url();
 
     // Redirect to Discord's oauth service
-    Redirect(auth_url.into())
+    Redirect::permanent(auth_url.to_string().parse().unwrap())
 }
 
 // Valid user session required. If there is none, redirect to the auth page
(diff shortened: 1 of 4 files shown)


examples/oauth.rs
```diff
@@ -187,35 +186,20 @@ async fn login_authorized(
...
 impl IntoResponse for AuthRedirect {
-    type Body = Body;
-    type BodyError = hyper::Error;
+    type Body = Empty<Bytes>;
+    type BodyError = <Self::Body as axum::body::HttpBody>::Error;
 
-    fn into_response(self) -> http::Response<Body> {
-        Redirect("/auth/discord".to_string()).into_response()
+    fn into_response(self) -> http::Response<Self::Body> {
+        Redirect::permanent("/auth/discord".parse().unwrap()).into_response()
```
heliumbrain, line 202: I'm not sure if ´308´ is the best status code for a redirect like this. To my knowledge that would tell search engines that http:://example.com/auth/discord no longer exists and it will only index the page it points to. Maybe that only matters in a frontend-setting though now that I think about it.

jplatte: The more important thing is browsers may cache this redirect. I don't think I've ever seen permanent redirects in auth code.

scott-ling (changes requested):
it's nice and clean, 

but suggest a minor documentation change in `redirect.rs` as it i susing prelude::* instead of full list of `use`'s.

I know it is a pain but prelude::* hides details and there is another Issue relating to removing preludes where possible.

examples/oauth.rs
scott-ling: they can happen but should not happen, however, the original tower version had a version where you can provide the status code - based on what you feel is best as long as this is still an option the base return is not as important. 

In most systems, I have used and built we mapped 308 -> 301 as they are essentially the same and most don't check for 308.  

side note: 301 you can change the method in the redirect -> 308 methods is persisted so in discord example 308 is 100% correct as it persists the method.

src/response/redirect.rs
```diff
@@ -0,0 +1,77 @@
...
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308
+    pub fn permanent(uri: Uri) -> Self {
+        Self::with_status_code(StatusCode::PERMANENT_REDIRECT, uri)
+    }
+
+    // This is intentionally not public since other kinds of redirects might not
+    // use the `Location` header, namely `304 Not Modified`.
+    //
+    // We're open to adding more constructors upon request, if they make sense :)
```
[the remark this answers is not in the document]
scott-ling: suggest an ability to set status code with re-direct will fix all issues, as sometimes you may redirect and not want to signal it downstream.

examples/oauth.rs
heliumbrain: I think a 302 Found is fine to use aswell

davidpdrsn: I've changed it to 302 Found. Thats also what the example used originally actually.

PR #195, opened 2021-08-16
Remove `axum::prelude`

### LINKED ISSUE #185: Remove `axum::prelude`
### Motivation

I added `axum::prelude` mainly to make documentation examples shorter and have less imports. But preludes have fallen out of style over the last few years for a few reasons:

- Glob imports make things harder to understand. Scanning a list of imports and seeing everything that gets pulled in is nice. Can't do that when using glob imports which preludes require.
- Auto import in rust-analyzer has gotten pretty reliable even for methods on traits.
- They're doing a dis-service to new users since it makes things harder to find in the docs.

### Proposal

Remove the prelude and update all the docs and examples. Shouldn't be hard but will most likely be tedious work.

### Alternatives

None that come to mind.

ZFLloyd:
## Motivation

Fixes #185

## Solution

Imported explicitely what was necessary after removing the prelude.


davidpdrsn (approved):
🥳

Issue #159, opened 2021-08-08
How should `Uri` extractor work when combined with `nest`?


davidpdrsn:
In https://github.com/tokio-rs/axum/pull/156 the behavior of `Uri` when used as an extractor for nested routes changed. But as mentioned by @lnicola [here](https://github.com/tokio-rs/axum/pull/156#discussion_r684741806) the old behavior might be desirable in some cases.

I would like to provide both options but am unsure about what the most reasonable setup is:

1. Should `Uri` always be the original URI, regardless of `nest`, and then add another extractor like `NestedUri` (name pending) that provides the URI with the prefix stripped.
2. Or the other way around: `Uri` extracts the stripped URI and then we provide an `OriginalUri` extractor.

I'm personally in favor of 1. because I think this behavior is less surprising. To me nesting is a router concern and not something that should leak into user code in this way.

@programatik29 @jplatte @lnicola what do you think?

jplatte:
Maybe there's even cases where you would want a certain `nest` to affect the URI being extracted, but others not? Could be useful when people host their blog on `site.tld/blog`, their wiki on `site.tld/wiki` and so on rather than using subdomains. It's not that common anymore but I've recently seen with Matrix that that's still a thing (for whatever reason).

Since we don't have default fn parameters in Rust, maybe we could have a new function akin to `nest` that does the same thing except it does affect the `Uri` extractor, while keeping the behavior for `nest` the same way it is now, after #156.

davidpdrsn:
Oh yeah thats an interesting idea! To make it a config thing of `nest` and not the extractor 🤔

lnicola:
I've needed both in the same handler: the stripped Uri to extract information from it (I actually have an extractor), but also the whole one to make a redirect URL for authentication.

Getting the other one through an extractor would work for me.

davidpdrsn:
@lnicola Which do you think is most reasonable of the two options I listed?

lnicola:
No preference. 1 might be slightly nicer, but 2 seems cleaner to implement.

davidpdrsn:
I went with 1 and added `NestedUri` https://github.com/tokio-rs/axum/pull/161

lnicola:
Ah, I meant the other way around. But it's fine :sweat_smile:.

davidpdrsn:
I realized always showing the original URI breaks serving static files and so I've had to revert it. See https://github.com/tokio-rs/axum/pull/197 for more details.

PR #197, opened 2021-08-17
Change nested routes to see the URI with prefix stripped

davidpdrsn:
In https://github.com/tokio-rs/axum/issues/159 we decided that nested services should always see the full original URI when using `Uri` as an extractor. However I've since realized that that doesn't work as it breaks serving static files like so:

```rust
let app = nest(
    "/static",
    axum::service::get(ServeDir::new(".")).handle_error(...)
);
```

The request URI that `ServeDir` sees will still contain the `/static` prefix, which isn't correct.

I've therefore reverted the behavior to the way it was in 0.1 where nested services will see the URI with the matched prefix stripped.

Also added an `OriginalUri` extractor that always gives the original URI, regardless of nesting.


lnicola:
Can we instead have a service wrapper that replaces the `Uri` with the nested one?

davidpdrsn:
One could write a middleware like that outside of axum. I'm also not sure which approach is best but since most people are kinda "meh both are fine I guess" and one option breaks a use case I think I'll move forward with this until someone complains :)

lnicola:
Aaand now I have to change my code again :grimacing:.

PR #213, opened 2021-08-19
Update README to remove use of prelude

hamza1311:
## Motivation

Prelude was removed as part of https://github.com/tokio-rs/axum/pull/195 but this usage was not removed. This PR fixes that.

## Solution

Not applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,8 +30,7 @@ use axum::{
     extract,
     handler::{get, post},
     http::StatusCode,
-    response,
-    response::IntoResponse,
+    response::{self, IntoResponse},
     route,
     routing::RoutingDsl,
 };
--- a/README.md
+++ b/README.md
@@ -26,7 +26,15 @@ applications written using [`hyper`] or [`tonic`].
 ## Usage example
 
 ```rust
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,7 +26,15 @@ applications written using [`hyper`] or [`tonic`].
 ## Usage example
 
 ```rust
-use axum::{prelude::*, response::IntoResponse, http::StatusCode};
+use axum::{
+    extract,
+    handler::{get, post},
+    http::StatusCode,
+    response,
+    response::IntoResponse,
```
davidpdrsn, line 34:
```suggestion
    response::{self, IntoResponse},
```

davidpdrsn (commented):
Thanks!

I had one little nitpick

davidpdrsn:
Actually. Now that I think of it. I don't think we should merge this yet. Users coming to the repo will expect the readme to reflect whats currently on crates.io. So until 0.2 is released I think its better we keep the old readme.

davidpdrsn:
I fixed the readme manually in 77f7b51.

But thanks regardless 😊

PR #216, opened 2021-08-20
Add `print-request-response` example

davidpdrsn:
Someone asked about this on Discord. I think its worth adding as an
example.

### DIFF
--- a/examples/print-request-response/src/main.rs
+++ b/examples/print-request-response/src/main.rs
@@ -1,0 +1,64 @@
+//! Run with
+//!
+//! ```not_rust
+//! cargo run -p example-print-request-response
+//! ```
+
+use axum::{
+    body::{Body, BoxBody, Bytes},
+    handler::post,
+    http::{Request, Response},
+    Router,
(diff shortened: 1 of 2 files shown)


examples/print-request-response/src/main.rs
```diff
@@ -0,0 +1,64 @@
...
+    let res = Response::from_parts(parts, Body::from(bytes));
+    Ok(res)
+}
+
+async fn buffer_and_print<B>(direction: &str, body: B) -> Result<Bytes, BoxError>
+where
+    B: axum::body::HttpBody<Data = Bytes>,
+    B::Error: Into<BoxError>,
+{
+    let bytes = hyper::body::to_bytes(body).await.map_err(Into::into)?;
```
paulzhang5511, line 59: ```hyper``` should not be relied on

davidpdrsn: Why not? Axum already uses hyper.

If you want you can write it yourself using `http_body::Body::data`.

PR #218, opened 2021-08-20
feat: expose tower log feature

johannescpk:
## Motivation

`tower` has as [only `default` feature `log`](https://github.com/tower-rs/tower/blob/ee131aaf467d2f1dcf9e60086823cf540ee94a0d/tower/Cargo.toml#L27), which toggles `tracing/log`. That's problematic in code relying on `Send+Sync`. This has been done in warp and hyper as well:

- https://github.com/seanmonstar/warp/pull/861
- https://github.com/hyperium/hyper/issues/2326

## Solution

We don't enable `tower`'s default features and expose the `log` feature instead. I guess this should be considered a breaking change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,9 +15,10 @@ version = "0.1.3"
 members = ["examples/*"]
 
 [features]
-default = []
+default = ["tower-log"]
 ws = ["tokio-tungstenite", "sha-1", "base64"]
 multipart = ["multer", "mime"]
+tower-log = ["tower/log"]
 
 [dependencies]
 async-trait = "0.1"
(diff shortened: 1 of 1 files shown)


davidpdrsn:
Thanks! Maybe it should be enabled by default though. Its default in warp and tower 🤔

johannescpk:
Good point, that also doesn't make it a breaking change. Updated. :+1:

jplatte:
This is (technically) a breaking change because it changes the behavior for crates that depend on axum with `default-features = false` (although there was previously no reason to ever do that).

davidpdrsn:
It wont be released until 0.2 anyway. @jplatte do you think its worth mentioning in the changelog?

jplatte:
Yeah, I'm sure some people will want to get rid of any unnecessary features so telling them there is now sth. to disable would be nice.

davidpdrsn:
Cool! @johannescpk wanna submit a quick follow up PR? 🙏

PR #223, opened 2021-08-20
Revert "Simplify handler trait (#221)"

davidpdrsn:
This reverts commit 44c58bdf5f43020f92633b7cf79174b1c69cb049.

### DIFF
--- a/src/handler/future.rs
+++ b/src/handler/future.rs
@@ -2,7 +2,10 @@
 
 use crate::body::{box_body, BoxBody};
 use crate::util::{Either, EitherProj};
-use futures_util::{future::BoxFuture, ready};
+use futures_util::{
+    future::{BoxFuture, Map},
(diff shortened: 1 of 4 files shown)


ZizhengTai:
What's the reason for reverting this?

davidpdrsn:
It broke @ChillFish8's app.

ZizhengTai:
Do we lose anything (in addition to simpler code) without these simplifications? Like compilation time improvement?

davidpdrsn:
Nope. I was hoping this would improve compile times but it didn't during my testing. Thought it was still worth merging but not if it breaks stuff.

ZizhengTai:
Maybe consider merging it in 0.2?

davidpdrsn:
As it wasnt an important change I don't think we will merge it.

PR #224, opened 2021-08-20
Fallback to calling next route if no methods match

davidpdrsn:
This removes a small foot gun from the routing.

This means matching different HTTP methods for the same route that
aren't defined together now works.

So `Router::new().route("/", get(...)).route("/", post(...))` now
accepts both `GET` and `POST`. Previously only `POST` would be accepted.

### DIFF
--- a/src/routing/mod.rs
+++ b/src/routing/mod.rs
@@ -706,10 +706,6 @@ where
 
         let mut res = Response::new(crate::body::empty());
 
-        // if request.extensions().get::<OrDepth>().is_some() {
-        //     res.extensions_mut().insert(FromEmptyRouter { request });
-        // }
-
         // XXX: we can't actually ship this because of
         // https://github.com/hyperium/hyper/issues/2621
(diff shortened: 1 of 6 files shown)


davidpdrsn:
cc @dbofmmbt 😊

src/routing/mod.rs
```diff
@@ -713,39 +728,6 @@ struct FromEmptyRouter<B> {
...
-///
-/// This is to work around https://github.com/hyperium/hyper/issues/2621.
-///
-/// Since ours can be nested we have to track the depth to know when we're
-/// leaving the top most `Or`.
-///
-/// Hopefully when https://github.com/hyperium/hyper/issues/2621 is resolved we
-/// can remove this nasty hack.
-#[derive(Debug)]
-struct OrDepth(usize);
```
davidpdrsn, line 727: https://github.com/hyperium/hyper/issues/2621 was fixed in h2 0.3.4 which was just published.

Issue #225, opened 2021-08-21
404  Don't work.


paulzhang5511:
### Version
cargo 1.54.0 (5ae8d74b3 2021-06-22)

axum 0d8f8b7

### Platform

windows linux subsystem(Ubuntu 20.04)

### Description
404 handle not working when visiting a non-existent page.

Example:
https://github.com/paulzhang5511/axum-compile-slow/tree/issue-404

Get http://127.0.0.1:8000/hello 
Result: Cannot display "not found" correctly

```rust
async fn map_404() -> &'static str {
    "not found"
}
```

PR #226, opened 2021-08-21
Fix 404 example

davidpdrsn:
Forgot to actually set the correct status code 🤦

Fixes https://github.com/tokio-rs/axum/issues/225

### DIFF
--- a/examples/global-404-handler/src/main.rs
+++ b/examples/global-404-handler/src/main.rs
@@ -6,7 +6,8 @@
 
 use axum::{
     handler::{get, Handler},
-    response::Html,
+    http::StatusCode,
+    response::{Html, IntoResponse},
     Router,
 };
 use std::net::SocketAddr;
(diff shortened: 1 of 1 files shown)


examples/global-404-handler/src/main.rs
```diff
@@ -6,7 +6,8 @@
 
 use axum::{
     handler::{get, Handler},
-    response::Html,
+    http::StatusCode,
+    response::{Html, IntoResponse},
```
paulzhang5511, line 10:
```rust
 response::{Html, IntoResponse},
```
To
```rust
response::IntoResponse,
```

paulzhang5511:
It still doesn't work.

image

```rust
 async fn map_404() -> impl IntoResponse {
    (http::StatusCode::NOT_FOUND, "not found")
}
```
Cannot display "not found" correctly

davidpdrsn:
It does work but @paulzhang5511 had `nest("/", ...)` which will capture all requests, so the 404 handler is never called. If you wanna do that you should instead do [this](https://github.com/tokio-rs/axum/blob/ca4d9a2bb99746c291ca8ad28dd0c5574fa46893/examples/global-404-handler/src/main.rs).

PR #228, opened 2021-08-21
Mention using `Path<HashMap>` to capture all params

davidpdrsn:
Might not be entirely obvious that you can do this so makes sense to mention in the docs.


jplatte:
Would `Vec<(String, String)>` also work? I feel like a type like std's `HashMap` that's tuned to perform well with huge numbers of entries is kind of overkill for this kind of use case.

davidpdrsn:
@jplatte Yep that works too. I personally prefer maps but I've updated the example to show `Vec` as well.

Issue #186, opened 2021-08-16
Remove tower from axum's public API


davidpdrsn:
### Motivation

There are currently a few places where things from tower appears in axum's public API. Since tower's versioning is a bit more loose, than say tower-service, I think it makes sense to remove.

### Proposal

Remove tower from axum's public API most likely by vendoring the parts.

I think have an exhaustive list but these come to mind:

- `RoutingDsl::into_make_service` returns `tower::make::Shared`
- `tower::BoxError` is still used in some bounds. Can just be vendored since its just a type alias.

There are probably other places I'm forgetting.

ZizhengTai:
Sorry for probably a noob question, but what does "vendoring" mean in this context?

davidpdrsn:
Copying code from a dependency into axum so we can control the API, or in this case to avoid breaking changes propagating into axum. For example `tower::make::Shared` is very simple so copying it into axum wouldn't be too bad.

pepoviola:
Hi @davidpdrsn, this sounds like a fun issue to work and contribute is nobody start to work on it yet.
Thx!

davidpdrsn:
@pepoviola go for it!

For `tower::BoxError` that should just become `pub type BoxError = Box<dyn std::error::Error + Send + Sync>` in `lib.rs`. For `tower::make::Shared` I think we should just vendor it, rename it to `MakeRouteService` and make the constructor private.

I'm not sure if there are other tower things in axum's public API. Would be neat if there was a tool that could figure that out 🤔

pepoviola:
Great! thanks @davidpdrsn! I will also look if there is some tool we can use for scan the public api.

davidpdrsn:
@pepoviola Not sure if you started working on this yet but I'm gonna look into it today. Would like to get 0.2 out next week.

pepoviola:
Hi @davidpdrsn, I didn't make time to make any progress. I saw your pr and looks awesome 🚀🙌
Thanks!

PR #229, opened 2021-08-21
Remove tower from axum's public API

davidpdrsn:
Instead rely on `tower-service` and `tower-layer`. `tower` itself is
only used internally.

Fixes https://github.com/tokio-rs/axum/issues/186

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,8 @@ members = ["examples/*"]
 default = ["tower-log"]
 ws = ["tokio-tungstenite", "sha-1", "base64"]
 multipart = ["multer", "mime"]
-tower-log = ["tower/log"]
+# tower-log = ["tower/log"]
+tower-log = []
 
 [dependencies]
 async-trait = "0.1"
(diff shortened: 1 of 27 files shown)


Cargo.toml
```diff
@@ -18,7 +18,8 @@ members = ["examples/*"]
 default = ["tower-log"]
 ws = ["tokio-tungstenite", "sha-1", "base64"]
 multipart = ["multer", "mime"]
-tower-log = ["tower/log"]
+# tower-log = ["tower/log"]
+tower-log = []
```
jplatte, line 22: Is this change intentional?

src/routing/future.rs
```diff
@@ -179,3 +181,9 @@ where
         self.project().inner.poll(cx)
     }
 }
+
+opaque_future! {
+    /// Response future from [`MakeRouteService`] services.
+    pub type MakeRouteServiceFuture<S> =
+        futures_util::future::Ready<Result<S, Infallible>>;
```
jplatte, line 188: With axum's current MSRV you could use the `std` equivalent (introduced in 1.48.0). Or does that not work due to dependencies using this type?

Cargo.toml
davidpdrsn: Nope 😊 going through everything now making sure its good.

src/routing/future.rs
davidpdrsn: Whoa I did not realize `std::future::Ready` was stable 🎊 Gonna try using that instead!

Cargo.toml
davidpdrsn: Done

src/routing/future.rs
davidpdrsn: I'll address this in a follow up PR.

PR #254, opened 2021-08-24
Simplify tls-rustls example

programatik29:
## Motivation

Current `tls-rustls` example might be inconvenient for some people.

## Solution

Rename current example to `low-level-rustls` and add a high level example in its place.


davidpdrsn (approved):
👌

PR #255, opened 2021-08-24
Add Redirect::to constructor

### LINKED ISSUE #248: Add a 303 constructor to Redirect
### Motivation

This should be the go-to redirect response code for GET-only redirects. Rocket even has this as [`Redirect::to`](https://docs.rs/rocket/0.4.10/rocket/response/struct.Redirect.html#method.to) (the other constructors have the same names as the ones in axum).

### Proposal

Add a constructor to `Redirect` for creating a 303 response. I don't care much about the name, `Redirect::to` like in Rocket or `Redirect::see_other` seem like viable options.

### Alternatives

Provide a constructor with a status code parameter. I don't think this is actually a good idea, and it was intentionally left out so far too.

jplatte:
The motivation for this is established in the issue it fixes.

Resolves #248

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- 👀
+- **added:** Add `Redirect::to` constructor ([#255](https://github.com/tokio-rs/axum/pull/255))
 
 # 0.2.0 (23. August, 2021)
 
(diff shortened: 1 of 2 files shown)


jplatte:
Seems like I can't get GitHub to register this PR as closing #248 :confused:

davidpdrsn:
Looks alright for me

image

davidpdrsn (commented):
Wanna add a note to the changelog as well?

jplatte:
Oh I thought if it just said "Linked" it wouldn't auto-close. For some reason it doesn't show that small text for me.

Issue #259, opened 2021-08-24
Reverse route ordering (for 0.3.0)


jplatte:
## Feature Request

Please reverse the route ordering such that routes registered first (earlier in a `.route()` chain) have higher precedence.

### Motivation

It's quite unusual to have one's index route last, instead of first. Proof:

* [actix-web docs](https://actix.rs/docs/url-dispatch/)
* [a Rocket example with routing](https://github.com/SergioBenitez/Rocket/blob/42a0fb8afe0ad94024cbb2a131390b53b99937c1/examples/pastebin/src/main.rs#L57)
* [a Gotham example with routing](https://github.com/gotham-rs/gotham/blob/5c50f807abbe60d94b80d19907c2967dc15c53d8/examples/routing/http_verbs/src/main.rs#L25-L46)

all have the index route first.

### Proposal

Reverse the order in which routes are attempted to use.

### Alternatives

Don't have `"/:param"` match the index route (i.e. require path params to be non-empty). This would at least solve the issue as presented here, though there would still be the issue of having to declare a `"/:param"` route before `"/static_page"`, not afterwards as I'd personally expect.

davidpdrsn:
> Don't have `"/:param"` match the index route (i.e. require path params to be non-empty). This would at least solve the issue as presented here, though there would still be the issue of having to declare a `"/:param"` route before `"/static_page"`, not afterwards as I'd personally expect.

I think it makes sense to do both: reverse the order of routes, and don't match captures match empty params. I would personally find that behavior a bit surprising

davidpdrsn:
I thought I would just provide some context as for why axum matches routes bottom to top:

axum's router is structure very much like `Iterator`. When you write `some_vec.into_iter().map(...).filter(...)` the type you get back is `Filter<Map<vec::IntoIter>>`. Thus the `Filter` is called first, even though it was added last in the chain.

If you define an axum router like so

```rust
Router::new()
    .route("/", root)
    .route("/foo", foo);
```

That is roughly the same as doing

```rust
Route::new("/foo", foo, Route::new("/", root, Router::new())
```

So for each `.route()` you add, you're adding another `Route` struct that wraps the whole previous thing. The final type is something along the lines of

```
Route {
    svc: foo,
    fallback: Route {
        svc: root,
        fallback: EmptyRoute, // created by Router::new()
    },
}
```

[...]

PR #264, opened 2021-08-24
Fix URI captures matching empty segments

davidpdrsn:
It was never the intention that `/:key` should match `/`. This fixes that.

Part of https://github.com/tokio-rs/axum/issues/259

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+Just me testing auto-merge
+
 - **fixed:** Fix URI captures matching empty segments. This means requests with
   URI `/` will no longer be matched by `/:key` ([#264](https://github.com/tokio-rs/axum/pull/264))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- **fixed:** Fix URI captures matching empty segments. This mean requests with
-  URI `/` will no longer be matched by `/:key`
+- **fixed:** Fix URI captures matching empty segments. This means requests with
+  URI `/` will no longer be matched by `/:key` ([#264](https://github.com/tokio-rs/axum/pull/264))
 
 # 0.2.1 (24. August, 2021)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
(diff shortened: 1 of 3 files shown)


davidpdrsn:
@jplatte Would you consider this a breaking change? I'm thinking no, since it was a bug and not intended behavior but I can see arguments for or againts.

jplatte:
I agree that this should be considered a bugfix, especially since axum doesn't have many users yet and whoever is already using it in production can only blame themselves 😄 

You even [state that it follows the official semver spec](https://github.com/tokio-rs/axum/blob/9a082a74b0a1641b44e31cfaac9cbc69a2e0e238/CHANGELOG.md#L6), which ([unlike Cargo's semver flavor](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html?highlight=considers#caret-requirements)) means pre-1.0 any version bump can contain breaking changes 😜

CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- 👀
+- **fixed:** Fix URI captures matching empty segments. This mean requests with
+  URI `/` will no longer be matched by `/:key`
```
davidpdrsn, line 11:
```suggestion
- **fixed:** Fix URI captures matching empty segments. This means requests with
  URI `/` will no longer be matched by `/:key` ([#264](https://github.com/tokio-rs/axum/pull/264))
```

CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- 👀
+- **fixed:** Fix URI captures matching empty segments. This means requests with
+  URI `/` will no longer be matched by `/:key` ([#264](https://github.com/tokio-rs/axum/pull/264))
```
davidpdrsn, line 11:
```suggestion
Just me testing auto-merge

- **fixed:** Fix URI captures matching empty segments. This means requests with
  URI `/` will no longer be matched by `/:key` ([#264](https://github.com/tokio-rs/axum/pull/264))
```

CHANGELOG.md
```diff
@@ -7,7 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- 👀
+Just me testing auto-merge
+
+- **fixed:** Fix URI captures matching empty segments. This means requests with
+  URI `/` will no longer be matched by `/:key` ([#264](https://github.com/tokio-rs/axum/pull/264))
```
davidpdrsn, line 13:
```suggestion
- **fixed:** Fix URI captures matching empty segments. This means requests with
  URI `/` will no longer be matched by `/:key` ([#264](https://github.com/tokio-rs/axum/pull/264))
```

PR #266, opened 2021-08-25
feat: add some inline attributes to tower::Service to improve performance

PureWhiteWu:
Add some inline attributes to enable cross-crate inline to improve performance.


davidpdrsn:
Do you have a benchmark that actually shows a difference in performance?

Issue #261, opened 2021-08-24
Path extraction uses raw, percent-encoded path segments


jplatte:
### Version

0.2.1

### Platform

`Linux jp-notebook 5.13.12-arch1-1 #1 SMP PREEMPT Wed, 18 Aug 2021 20:49:03 +0000 x86_64 GNU/Linux`

### Description

<short summary of the bug>

I tried this code:

```rust
// TurboFish deserialization expects a string of the form `::<{guts}>` or `<{guts}>::`.
pub async fn turbofish(Path(turbofish): Path<TurboFish>) -> impl IntoResponse {
    // ...
}

// in async fn main
let app = Router::new()
    .route("/:turbofish", get(routes::turbofish))
    // ...
```

I expected to see this happen: Route extraction succeeds, for paths such as `/%3A%3A<_>`

Instead, this happened: Route extraction failed (until I added percent decoding to the `Deserialize` impl where it really doesn't belong IMHO)

PR #272, opened 2021-08-26
Percent decode automatically in `extract::Path`

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/261

@jplatte what you think?


lnicola:
Just throwing out some random thoughts:

 - not every framework does percent decoding, it might be less surprising to have a special extractor for it
 - if not, this should be clearly mentioned in the `Path` documentation
 - a `PathBuf` extractor would also useful, and should probably percent-decode automatically (and check for path traversal too)
 - in my app I also want non-UTF-8 paths to work (using https://crates.io/crates/os-str-bytes), but that's more of a niche use case

davidpdrsn:
> not every framework does percent decoding, it might be less surprising to have a special extractor for it

I think it should be the other way around. Have `Path` decode by default, and then provide a more niche extractor (possibly in another crate) that doesn't. I imagine working with utf8 is more common in rust.

> if not, this should be clearly mentioned in the Path documentation

Definitely!

> a PathBuf extractor would also useful, and should probably percent-decode automatically (and check for path traversal too)

That's orthogonal to this but is something I've thought about as well. Contributions are welcome 😊

lnicola:
Yeah, the raw path should still be available in the request, so that's fine.

src/util/mod.rs
```diff
@@ -34,6 +34,39 @@ impl ByteStr {
...
+    pub(crate) fn new<S>(s: S) -> Self
+    where
+        S: AsRef<str>,
+    {
+        let byte_str = if let Ok(decoded) =
+            percent_encoding::percent_decode(s.as_ref().as_bytes()).decode_utf8()
+        {
+            ByteStr::new(decoded)
+        } else {
+            ByteStr::new(s)
```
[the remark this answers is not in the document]
davidpdrsn: I think I managed to find a fix. The extractor will now fail if one of the params contained invalid utf-8 after decoding.

CHANGELOG.md
```diff
@@ -7,13 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- **fixed:** Fix `extract::Path` not percent decoding params ([#272](https://github.com/tokio-rs/axum/pull/272))
```
jplatte, line 10: This should be `**changed**` and the PR labeled 'breaking change'. Doing more or less percent encoding clearly changes observable behavior.

src/extract/path/de.rs
```diff
@@ -561,18 +560,16 @@ mod tests {
...
     {
-        UrlParams(
-            values
-                .into_iter()
-                .map(|(k, v)| (ByteStr::new(k), ByteStr::new(v)))
-                .collect(),
-        )
+        values
+            .into_iter()
+            .map(|(k, v)| (ByteStr::new(k), PercentDecodedByteStr::new(v).unwrap()))
```
jplatte, line 571: I think this `.unwrap()` should be an `.expect("message")`. Without having read the rest of this PR I can't tell why this wouldn't sometimes cause a panic.

src/util/mod.rs
```diff
@@ -34,6 +34,39 @@ impl ByteStr {
...
+    {
+        let byte_str = if let Ok(decoded) =
+            percent_encoding::percent_decode(s.as_ref().as_bytes()).decode_utf8()
+        {
+            ByteStr::new(decoded)
+        } else {
+            return None;
+        };
+
+        Some(Self(byte_str))
```
jplatte, line 53: This seems unnecessarily verbose with the "early" `return`, but I guess that's a just matter of taste.

src/routing/mod.rs
```diff
@@ -626,24 +626,49 @@ where
     }
 }
 
-#[derive(Debug)]
-pub(crate) struct UrlParams(pub(crate) Vec<(ByteStr, ByteStr)>);
+// we store the potential error here such that users can handle invalid path
+// params using `Result<Path<T>, _>`. That wouldn't be possible if we
+// returned an error immediately when decoding the param
+pub(crate) struct UrlParams(
+    pub(crate) Result<Vec<(ByteStr, PercentDecodedByteStr)>, InvalidUtf8InPathParam>,
```
jplatte, line 633: I wonder whether this should be `Vec<Result<(ByteStr, PercentDecodedByteStr), InvalidUtf8InPathParam>>`. Would make sense even without really "supporting" non-UTF8 segments if the path deserializer supported [`IgnoredAny`](https://docs.rs/serde/1.0.130/serde/de/struct.IgnoredAny.html).

CHANGELOG.md
davidpdrsn: Bah you're right. One more thing that has to wait for 0.3 😞

src/extract/path/de.rs
davidpdrsn: This is test code.

src/routing/mod.rs
davidpdrsn: I think I'll just leave it as it. Can always improve things if someone has a use case.

src/util/mod.rs
davidpdrsn: Sure

PR #286, opened 2021-08-28
Make JSON and HTTP1 support optional

jplatte:
This is a breaking change because it changes the API surface for users who use `default-features = false`.


jplatte:
Semi-blocked on https://github.com/hyperium/hyper/pull/2639 (the failing CI could be fixed in different ways but that seemed like the best solution to me).

PR #294, opened 2021-08-31
Remove repeated constraint

sidkshatriya:
The constraint for `B` is repeated. Remove it.

---
P.S. Thank you for this most elegant looking library. I'm seriously considering using axum for a project of mine. Thank you for providing a lot of examples -- they really increase confidence towards the project. BTW are there some more examples I can help contribute? Please let me know and I'll give it a try!

I have a couple of questions:
- Are there some third party libraries I can use for form handling in Axum? The current form handling example seems to be quite basic and I'm looking for a lot more form handling power. For instance, Rocket seems to have powerful form handling capabilities.
- I'm looking for more options around cookies and authentication. Again, Rocket seems to have some more options here. Are there some good third party libraries that I can use for authentication and cookies? Things like CSRF protection etc. will be very useful.

In general I am finding axum extremely well thought out, consistent and "clean". Please keep up the great work :-) !

### DIFF
--- a/src/handler/mod.rs
+++ b/src/handler/mod.rs
@@ -316,7 +316,6 @@ macro_rules! impl_handler {
             Fut: Future<Output = Res> + Send,
             B: Send + 'static,
             Res: IntoResponse,
-            B: Send + 'static,
             $head: FromRequest<B> + Send,
             $( $tail: FromRequest<B> + Send,)*
         {


davidpdrsn:
Thanks!

Would you mind submitting those two questions as separate [discussions](https://github.com/tokio-rs/axum/discussions). Then I'll answer there. I think that'll make the answers more searchable.

sidkshatriya:
I've added the two questions as separate discussion points. Also do let me know if you have some more examples in mind that I could help with...

PR #297, opened 2021-08-31
Add very basic validation to the `form` example

sidkshatriya:
All HTML forms inevitably have some sort of validation. 

This PR adds very basic form validation for the `form` example using the [validator](https://docs.rs/validator/0.14.0/validator/) crate (validator crate originally suggested by @jplatte at https://github.com/tokio-rs/axum/discussions/295#discussioncomment-1258438 )

Please let me know if this is a useful addition. If things are not done in an idiomatic fashion, please suggest how I can make them better...

### DIFF
--- a/examples/form/src/main.rs
+++ b/examples/form/src/main.rs
@@ -36,7 +36,7 @@ async fn show_form(errors: Option<String>) -> Html<String> {
         <html>
             <head></head>
             <body>
-                <!-- In a more realistic application we would need to make sure things are properly escaped -->
+                <!-- In a more realistic application we would use a templating language (see templates example) -->
                 <pre>{}</pre>
                 <form action="/" method="post">
                     <label for="name">
@@ -71,7 +71,7 @@ async fn accept_form(Form(input): Form<Input>) -> Html<String> {
     match input.validate() {
         Ok(_) => String::from("Form submitted successfully!").into(),
         // Here the validation errors are presented as-is. In a more realistic
-        // application we would need to customize/process the validation errors to be able to
+        // application we would customize / process the validation errors to be able to
         // show to the user in a way that they could understand it
         Err(e) => show_form(Some(format!("Form error: {}", e))).await,
     }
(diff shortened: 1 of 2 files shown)


examples/form/src/main.rs
```diff
@@ -28,13 +29,15 @@ async fn main() {
...
-async fn show_form() -> Html<&'static str> {
-    Html(
+async fn show_form(errors: Option<String>) -> Html<String> {
+    Html(format!(
         r#"
         <!doctype html>
         <html>
             <head></head>
             <body>
+                <!-- In a more realistic application we would need to make sure things are properly escaped -->
```
jplatte, line 39: How about

```suggestion
                <!-- In a more realistic application we would use a templating language (see templates example) -->
```

examples/form/src/main.rs
```diff
@@ -51,15 +54,25 @@ async fn show_form() -> Html<&'static str> {
...
     email: String,
 }
 
-async fn accept_form(Form(input): Form<Input>) {
+async fn accept_form(Form(input): Form<Input>) -> Html<String> {
     dbg!(&input);
+    match input.validate() {
+        Ok(_) => String::from("Form submitted successfully!").into(),
+        // Here the validation errors are presented as-is. In a more realistic
+        // application we would need to customize/process the validation errors to be able to
```
jplatte, line 74: 'need' is a strong word 😄 

```suggestion
        // application we would customize / process the validation errors to be able to
```

davidpdrsn:
Thanks for doing this! I think this kind of example is worth having but I think it fits better as a separate example. That way we keep the current example focused on showing how `Form` works with `serde`, and the example with validation can be expanded a bit to also show basic error handling. What do you think?

sidkshatriya:
Yes, I think this can be made into a separate example!

davidpdrsn:
@sidkshatriya just following up. Are you still up for splitting it into a separate example?

sidkshatriya:
@davidpdrsn Thanks for following up -- sorry I got into other things. I'm not sure when I will come back to this, but I can try.

davidpdrsn:
An example using validator was added in https://github.com/tokio-rs/axum/pull/352

Issue #240, opened 2021-08-22
Implement routing in `Router`


KaiJewson:
### Motivation

- Axum does not warn when routes are ambiguous.
- Routing is O(n).
- 404 handling is currently a bit hacky.

### Proposal

Move all route matching into the `Router` type itself. I [have implemented a sketch of this proposal on the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=151f1b36a50c0b6a5cf3157a79c3700c), based on an imaginary `Node` type that does all the actual routing, and can be optimized to O(log n) because it knows all the routes up front. It is similar to [`matchit::Node`](https://docs.rs/matchit/0.4/matchit/struct.Node.html) but with a larger API; we might be able to upstream the required features there and use it. The sketch uses un-newtyped future types, which would obviously be changed in the actual library. We also might want to make the implementation details (`Destinations`, `RouteDestination`, etc) inaccessible (public in a private module) to avoid polluting the public API.

Now that routing is all done in one place, 404 handling can be implemented simply as another generic field of the `Router` type, which would be an `Empty404` service by default, but can be changed to any service.

### Drawbacks

The current design makes `route("/", get(a)).route("/", post(b))` not work anymore; it will panic at server startup. Similarly `route("/", get(a)).or(route("/", post(b)))` panics as well. However I don't think this pattern is particularly useful. Axum only supported it before because it wasn't able to create an error there.

The implementation uses an hlist to store the list of services the routes point to, meaning routing is technically O(n). It should be really easy for the optimizer to turn it into O(1), but I haven't checked so we're kind of at the mercy of LLVM there. That said, even if it is O(n) it's just a sequence of integer comparisons which should be pretty cheap.

### Alternatives

- Have a single non-generic `Router` type that boxes all internal services, but otherwise operates in a similar fashion to above. This would make it a _lot_ easier to pass around and return routers, even when `.boxed()` exists. Not sure how it would impact performance.

davidpdrsn:
I've read through your implementation and my initial impression is that it feels like a lot of changes, and new API, for not much gain.

> Axum does not warn when routes are ambiguous.

By "ambiguous" do you something like `route("/foo", _).route("/:key", _)`? If so, is that actually a problem in practice?

> Routing is O(n)

I don't think this is an issue. The actual work performed by hyper or your app is likely taking up way more time than routing.

> 404 handling is currently a bit hacky

[...]

PR #301, opened 2021-09-01
reverse routing

dbofmmbt:
Closes issue #259

## Solution

Instead of nesting routing methods using the `fallback`, let's wrap them with `Or`. [Discussion thread on Discord](https://discord.com/channels/500028886025895936/870760546109116496/882726358697017344).

## Pending Tasks

TODO

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -165,7 +165,7 @@
 //! // since its defined last.
 //! let (status, body) = call_service(&mut app, Method::GET, "/foo").await;
 //! assert_eq!(status, StatusCode::OK);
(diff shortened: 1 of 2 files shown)


davidpdrsn:
I have observed `or` having compile time issues when composing lots of them. Is this also the case here?

I also think if we're going to make breaking to the routing (this changes a return type) then we should consider a different design along the lines of https://github.com/tokio-rs/axum/issues/240

davidpdrsn:
I'll close this for now. Feel free to re-open if you're still working on it.

PR #305, opened 2021-09-05
fix examples link in README.md

ImSingee:


davidpdrsn (approved):
Thanks! Can't believe we didn't catch this until now 😅

Issue #306, opened 2021-09-07
[Example request]: simple jwt example


silvioprog:
It would be nice to provide a simple JWT example, something like [this](https://github.com/zupzup/rust-jwt-example).

What do you think?

Z4RX:
Agree, it would be nice to enrich axum examples with jwt one. But it should be listed as community showcase because it will use lot of third-party crates.

Btw I can do it in 1-2 days.

davidpdrsn:
Sounds like a good idea! Something using `jsonwebtoken` would be nice.

silvioprog:
I'm going try to implement an example and send as PR. (btw I just started studying Axum, so it will be a naive example 😅)

silvioprog:
Done: https://github.com/tokio-rs/axum/pull/308. 😅

Z4RX:
Also made [example](https://github.com/Z4RX/axum_jwt_exampl), but a bit more opinioned and complex :D

silvioprog:
@Z4RX Woooow, amazing example! 🤩

What do you think about adding the two examples? The first one shows the basic concepts of JWT, and this second shows a full integration of JWT + Axum + PG.

Z4RX:
@silvioprog as I mentioned above, I think that it should be added to something like `community examples` section because it covers db connection, error handling, project layout and those themes  are very opinioned and shouldn't be shown as official example.

silvioprog:
> @silvioprog as I mentioned above, I think that it should be added to something like `community examples` ...

The first, second, or both? (the first one dependes only of `jsonwebtoken`)

Z4RX:
Your example should be added into repository examples

Z4RX:
I think this issue should be closed now @silvioprog

silvioprog:
Oops, I forgot to close it. Thanks a lot guys!

PR #308, opened 2021-09-07
Add JWT example (#306)

silvioprog:


examples/jwt/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    // Extract the authorization header value (i.e the token)
+    let token = authorization.0.token();
+
+    // Authenticate the user and returns his data
+    let validation = Validation {
+        sub: Some("b@b.com".to_string()),
+        ..Validation::default()
+    };
+    let token_data = decode::<Claims>(token, &DecodingKey::from_secret(JWT_SECRET), &validation)
+        .map_err(|_| AuthError::InvalidToken)?;
```
davidpdrsn, line 84: I imagine this is something users are gonna do a lot. What do you think about wrapping it up into its own extractor, so its. easier to reuse for other routes? Probably good to show even though its only used in one route here.

examples/jwt/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+    // Here you can check the user credentials from a database
+    if payload.client_id != "foo" || payload.client_secret != "bar" {
+        return Err(AuthError::WrongCredentials);
+    }
+
+    // Create the authorization token
+    let claims = Claims {
+        sub: "b@b.com".to_owned(),
+        company: "ACME".to_owned(),
```
davidpdrsn, line 106: My JWT knowledge is a bit rusty. Are these fields data about the token issuer, or the user?

examples/jwt/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+//
+// - try to visit the protected area using an invalid token
+//
+// curl -s \
+//     -w '\n' \
+//     -H 'Content-Type: application/json' \
+//     -H 'Authorization: Bearer blahblahblah' \
+//     http://localhost:3000/protected
+
+static JWT_SECRET: &[u8] = b"secret";
```
davidpdrsn, line 49: Should probably have a warning about not doing this in production and a hint telling users what to actually do in prod.

davidpdrsn (commented):
Very nice! Thanks for adding this. I had a few comments/questions.

examples/jwt/src/main.rs
jplatte: Or just do it better right away. I think this example should depend on `once_cell` either way to only construct the decoding key once, then it can also easily grab the secret from an env var in the initialization of the `DecodingKey`.

examples/jwt/src/main.rs
Z4RX: Usually `sub` field contains some identification information like user id or uuid. So both fields are about user

examples/jwt/src/main.rs
silvioprog: I totally agree. Tonight I'll read the extractors documentation to check how to implement them. Any example showing it will be appreciated. :smiley:

examples/jwt/src/main.rs
silvioprog: @Z4RX is right. The values `b@b.com`/`ACME` were extracted from the [`jsonwebtoken`'s example](https://github.com/Keats/jsonwebtoken/blob/master/examples/validation.rs#L15).

examples/jwt/src/main.rs
silvioprog: A very simple way was used in the [`oauth` example](https://github.com/tokio-rs/axum/blob/main/examples/oauth/src/main.rs#L6): using environment variables. What do ou thing about to use this same approach in the JWT example?

examples/jwt/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+
+#[tokio::main]
+async fn main() {
+    // Set the RUST_LOG, if it hasn't been explicitly defined
+    if std::env::var("RUST_LOG").is_err() {
```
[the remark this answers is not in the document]
davidpdrsn: I think its fine for now. Could do another PR where we change that in all the examples.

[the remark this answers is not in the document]
silvioprog: Hmm, all Axum examples used `std::env::var("RUST_LOG").is_err()`, I got it from [`todos`](https://github.com/tokio-rs/axum/blob/main/examples/todos/src/main.rs#L37).

examples/jwt/src/main.rs
jplatte: Yes, that's what I was saying :)

I guess it would work without putting the `DecodingKey` into a `once_cell::sync::Lazy` even, by putting it into a request extension (wrapped in `Arc<_>`) instead. Either way `DecodingKey` (and `EncodingKey` too) should only be constructed once, not once per request since constructing it is somewhat expensive AFAIK.

The jsonwebtoken docs for both types include:

> This key can be re-used so make sure you only initialize it once if you can for better performance

silvioprog: @jplatte Tonight I'll try it, thanks a lot! :smiley:

silvioprog: @jplatte, I'm trying to get the secret from an environment variable and initialize the `DecodingKey`/`EncodingKey` once, so suppose this struct:

```rust
#[derive(Debug)]
struct Keys<'a> {
    encoded: EncodingKey,
    decoded: DecodingKey<'a>,
}

impl<'a> Keys<'a> {
    fn new(secret: &'a [u8]) -> Self {
        Self {
            encoded: EncodingKey::from_secret(secret),
            decoded: DecodingKey::from_secret(secret),
        }
    }
}
```

what to you thing about this code using `lazy_static!`?:

```rust
lazy_static! {
    static ref JWT_SECRET: String = std::env::var("JWT_SECRET").expect("JWT_SECRET must be set");
}

lazy_static! {
    static ref KEYS: Keys<'static> = Keys::new(JWT_SECRET.as_bytes());
}
```

also, it can use `once_cell` instead:

```rust
static JWT_SECRET: Lazy<String> =
    Lazy::new(|| std::env::var("JWT_SECRET").expect("JWT_SECRET must be set"));

static KEYS: Lazy<Mutex<Keys>> = Lazy::new(|| Mutex::new(Keys::new(JWT_SECRET.as_bytes())));
```

Please let me know the right path to follow ...

jplatte: I think you don't need to propagate the lifetime of `DecodingKey` to the `Keys` struct, that lifetime parameter is going away with the next `jsonwebtoken` release anyways. You can just use `.into_static()` to get a `DecodingKey<'static>`, and I don't think the keys should be mutable (`Mutex` should not be necessary):

```rust
#[derive(Debug)]
struct Keys {
    encoded: EncodingKey,
    decoded: DecodingKey<'static>,
}

impl Keys {
    fn new(secret: &[u8]) -> Self {
        Self {
            encoded: EncodingKey::from_secret(secret),
            decoded: DecodingKey::from_secret(secret).into_static(),
        }
    }
}

static KEYS: Lazy<Keys> = Lazy::new(|| {
    let secret = std::env::var("JWT_SECRET").expect("JWT_SECRET must be set");
    Keys::new(secret.as_bytes())
});

File: examples/jwt/src/main.rs [hunk 5]
```diff
@@ -150,6 +150,21 @@ impl IntoResponse for AuthError {
     }
 }
 
+#[derive(Debug)]
+struct Keys {
+    encoded: EncodingKey,
+    decoded: DecodingKey<'static>,
```
jplatte, line 156: One more nit, I think `encoding` and `decoding` makes more sense for these, since the key itself isn't what's being de- / encoded.

examples/jwt/src/main.rs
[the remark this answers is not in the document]
silvioprog: Issued at https://github.com/tokio-rs/axum/issues/310.

examples/jwt/src/main.rs
silvioprog: Done. Thanks a lot! Now I'm going to implement https://github.com/tokio-rs/axum/pull/308#discussion_r704077062.

examples/jwt/src/main.rs
silvioprog: Done in 6016c65.

silvioprog:
Hey guys, do you have more reviews to apply to this PR? (Just waiting the workflow approval... 🙂)

examples/jwt/src/main.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
+        // Extract the token from the authorization header
+        let token = TypedHeader::<Authorization<Bearer>>::from_request(req)
+            .await
+            .map(|TypedHeader(Authorization(bearer))| bearer.token().to_owned())
+            .map_err(|_| AuthError::InvalidToken)?;
+        // Decode the user data
+        let token_data = decode::<Claims>(&token, &KEYS.decoding, &Validation::default())
+            .map_err(|_| AuthError::InvalidToken)?;
```
jplatte, line 137: If the `map` closure content is moved out there should be no need to use `.to_owned()`:

```suggestion
        let TypedHeader(Authorization(bearer)) =
            TypedHeader::<Authorization<Bearer>>::from_request(req)
                .await
                .map_err(|_| AuthError::InvalidToken)?;
        // Decode the user data
        let token_data = decode::<Claims>(bearer.token(), &KEYS.decoding, &Validation::default())
            .map_err(|_| AuthError::InvalidToken)?;
```

davidpdrsn (approved):
LGTM!

PR #309, opened 2021-09-08
Add JWT auth example as community showcase

Z4RX:
## Motivation
#306


Z4RX:
I'm not sure that this example can belong to `community showcase` probably we need to add a new section

davidpdrsn (approved):
I think its fine.

silvioprog:
> I'm not sure that this example can belong to `community showcase` probably we need to add a new section

I'm very new to Axum, so sorry for the question, but where is the community showcase page? 😅

jplatte:
https://github.com/tokio-rs/axum/tree/main/examples#community-showcase

silvioprog:
> https://github.com/tokio-rs/axum/tree/main/examples#community-showcase

@jplatte Thanks a lot!

PR #311, opened 2021-09-09
Add a way to get 'static Field from Multipart

HK416-is-all-you-need:
Introduce a way to get owned `Field` in order to allow user to satisfy `'static` bounds.

This is necessary to allow user safely access underlying `'static` Field.
It obviously has the same caveat as underlying library, but it is necessary as otherwise you make it impossible to wrap such stream into anything that has `'static + Stream` bounds like `https://docs.rs/rusoto_core/0.47.0/rusoto_core/struct.ByteStream.html`

### DIFF
--- a/src/extract/multipart.rs
+++ b/src/extract/multipart.rs
@@ -84,6 +84,77 @@ impl Multipart {
             Ok(None)
         }
     }
+
+    /// Yields the next [`OwnedField`] if available.
+    ///
+    /// Only one instance of `OwnedField` is allowed at the time.
+    ///
+    /// Unless previous instance is consumed, this function shall return error.
+    pub async fn next_owned_field(&mut self) -> Result<Option<OwnedField>, MultipartError> {
+        let field = self.inner
+                        .next_field()
+                        .await
+                        .map_err(MultipartError::from_multer)?;
+
+        if let Some(field) = field {
+            Ok(Some(OwnedField(field)))
+        } else {
+            Ok(None)
+        }
+    }
+}
+
+/// A single field in a multipart stream.
+#[derive(Debug)]
+#[repr(transparent)]
+pub struct OwnedField(multer::Field<'static>);
+
+impl OwnedField {
+    /// The field name found in the
+    /// [`Content-Disposition`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition)
+    /// header.
+    pub fn name(&self) -> Option<&str> {
+        self.0.name()
+    }
+
+    /// The file name found in the
+    /// [`Content-Disposition`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition)
+    /// header.
+    pub fn file_name(&self) -> Option<&str> {
+        self.0.file_name()
+    }
+
+    /// Get the content type of the field.
+    pub fn content_type(&self) -> Option<&Mime> {
+        self.0.content_type()
+    }
+
+    /// Get a map of headers as [`HeaderMap`].
+    pub fn headers(&self) -> &HeaderMap {
+        self.0.headers()
+    }
+
+    /// Get the full data of the field as [`Bytes`].
+    pub async fn bytes(self) -> Result<Bytes, MultipartError> {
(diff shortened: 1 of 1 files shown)


davidpdrsn:
I think this particular design choice in multer is quite unfortunate honestly. I think using ownership to check these things statically is way nicer so if possible I would like to keep axum's API the way it is.

If what you're looking for is a fully owned stream of bytes you can do this:

```rust
use axum::extract::multipart::{Multipart, MultipartError};
use std::io;

async fn accept_form(mut multipart: Multipart) {
    let stream = async_stream::stream! {
        while let Some(field) = multipart.next_field().await.map_err(multipart_error_to_io)? {
            yield field.bytes().await.map_err(multipart_error_to_io)
        }
    };

    let bytes_stream = rusoto_core::ByteStream::new(stream);

    // ...
}

fn multipart_error_to_io(err: MultipartError) -> io::Error {
    io::Error::new(io::ErrorKind::Other, err)
}
```

Would this work for you use case? You don't have to use `async_stream` I just find it quite nice in this particular case.

HK416-is-all-you-need:
Using `Field::bytes` is ok if you do not care about efficiency.
I'd prefer allocate by chunks via `Stream`, for simple playground it might be ok, but if we want to have proper solution we need to provide way to get `Stream + 'static` without actually sacrificing benefits of using `Stream` interface.

Right now I have to `transmute` to correct lifetime which is UB because it changes lifetime of reference to `Multipart` object.

If you do not like my idea to expanding API, then how about this:
```rust
pub async fn next_raw_field(&mut self) -> Result<Option<impl Stream + 'static + Send + Sync>, MultipartError> 
```

We can simplify and just return underlying type.
Obviously limitation applies and you have to drop it before using anotehr `next*` call, but I think it is much better than telling user to allocate the whole bunch of memory to read file at once.

Kestrer:
> I think this particular design choice in multer is quite unfortunate honestly.

I originally was going to implement the fully statically checked API, but then [SergioBenitez pointed out how many lifetimes it would infect code with](https://github.com/rousan/multer-rs/pull/33#issuecomment-841697288) so we went with the dynamic one. It probably won't change now, but at least you know why.

davidpdrsn:
> Using Field::bytes is ok if you do not care about efficiency, for simple playground it might be ok.

Can you elaborate on this point? How is using `Field::bytes` inefficient? axum doesn't provide another way of getting at the data so have to call it regardless what API you put on top.

> I originally was going to implement the fully statically checked API, but then SergioBenitez pointed out how many lifetimes it would infect code with so we went with the dynamic one. It probably won't change now, but at least you know why.

Thanks for providing the context 😊

HK416-is-all-you-need:
> axum doesn't provide another way of getting at the data so have to call it regardless what API you put on top.

Ehm... Stream is implemented for `Field`
https://github.com/tokio-rs/axum/blob/main/src/extract/multipart.rs#L98-L107

davidpdrsn:
That forwards for multers stream implementation which is also what the `bytes` method calls. So if you wanted a stream of the individual chunks couldn't that be done with

```rust
let stream = async_stream::stream! {
    while let Some(mut field) = multipart.next_field().await.map_err(multipart_error_to_io)? {
        while let Some(chunk) = field.try_next().await.map_err(multipart_error_to_io)? {
            yield Ok::<_, io::Error>(chunk)
        }
    }
};
```

HK416-is-all-you-need:
This is rather inconvenient because we need to do several checks prior to actually starting pulling data and carrying `Multipart` around for that purpose is rather complicated.
Overall it is possible, but to be blunt it would be easier for me to just implement my own extractor that provides interface that I want rather than going all lengths to work around inconvenient lack of `'static` which could be provided so easily.

davidpdrsn:
Cool. If you're fine with implementing your own extractor then I think thats the better approach. I'll close this for now.

martijnarts:
@HK416-is-all-you-need did you ever end up implementing this extractor?

HK416-is-all-you-need:
@martijnarts you can just copy-paste code from axum and fix lifetimes yourself
There is not much to implement

PR #314, opened 2021-09-10
Show how to configure `TraceLayer`

davidpdrsn:
This isn't very obvious so makes sense to include in the example.


shanesveller (commented):
This definitely would've helped me this morning! ❤️

PR #315, opened 2021-09-11
remove MpscBuffer from boxed

programatik29:
## Motivation

`MpscBuffer` adds a lot of overhead when boxing.

## Solution

Implement a cloneable `BoxService` using `SyncWrapper` to make `async_trait` futures `Sync`.


src/routing/mod.rs
```diff
@@ -256,7 +253,7 @@ impl<S> Router<S> {
     /// routes.
     pub fn boxed<ReqBody, ResBody>(self) -> Router<BoxRoute<ReqBody, S::Error>>
     where
-        S: Service<Request<ReqBody>, Response = Response<ResBody>> + Send + 'static,
+        S: Service<Request<ReqBody>, Response = Response<ResBody>> + Clone + Send + Sync + 'static,
```
davidpdrsn, line 256: Adding these bounds is a breaking change. It was also required to make https://github.com/tokio-rs/axum/pull/270 work.

davidpdrsn:
I played around with this a bit a found a slightly shorter version https://github.com/tokio-rs/axum/pull/339. I don't think we to use `SyncWrapper` to make the future `Sync` because it already is `!Sync`.

Thanks for finding this solution that didn't require [dyn-clone](https://github.com/dtolnay/dyn-clone). I'll close this for now. https://github.com/tokio-rs/axum/pull/339 will be included in 0.3

PR #317, opened 2021-09-11
Add accessors to `TypedHeaderRejection` fields

### LINKED ISSUE #316: Change fields visibility in TypedHeaderRejection from private to public
### Motivation
_from discord discussion_
I m trying to implement centralised error handling
so i have enum which represents all errors
```rust
#[derive(Debug, thiserror::Error)]
pub enum Error { ... }
// From<Error>
pub type ApiError = (StatusCode, Json<Value>);
// used as return type from handlers
pub type ApiResult<T> = std::result::Result<T, ApiError>;
```
Also I have centralised type for error returning from handlers  and custom extractors (ApiError and ApiResult)

And impl from Error for ApiError. (all internal functions returns `Result<T, Error>`)
Simplified `From<Error> for ApiError`
```rust
impl From<Error> for ApiError {
    fn from(err: Error) -> Self {
        let status = match err {
            ...
            _ => StatusCode::INTERNAL_SERVER_ERROR,
        };
        let payload = json!({"message": err.to_string()});
        (status, Json(payload))
    }
}
```
I'd like understand which header failed to be extracted in simple way like this:
```rust
Error::AxumTypedHeaderError(e) => match e.name {
  AUTHORIZATION => StatucCode::UNAUTHORIZED,
  _ => StatucCode::BAD_REQUEST
}
```

### Proposal

make fields `name` and `reason` public in struct `TypedHeaderRejection`

### Alternatives

`extract::typed_header` module is private (but `Reason` enum should be exported)

AlisCode:
## Motivation

Fixes #316.

Allows to implement a custom extractor that can behave depending on the header rejection's details (name and reason). 
More details in the original issue.

## Solution

* Changes the field visibility from private to public. 
* Adds relevant documentation. 
* Exposes the inner `Reason` type publicly, because this doesn't compile otherwise.

### DIFF
--- a/src/extract/typed_header.rs
+++ b/src/extract/typed_header.rs
@@ -48,7 +48,7 @@ where
         } else {
             return Err(TypedHeaderRejection {
                 name: T::name(),
-                reason: Reason::Missing,
+                reason: TypedHeaderRejectionReason::Missing,
             });
         };
 
@@ -56,11 +56,11 @@ where
             Ok(Some(value)) => Ok(Self(value)),
             Ok(None) => Err(TypedHeaderRejection {
                 name: T::name(),
-                reason: Reason::Missing,
+                reason: TypedHeaderRejectionReason::Missing,
             }),
             Err(err) => Err(TypedHeaderRejection {
                 name: T::name(),
-                reason: Reason::Error(err),
+                reason: TypedHeaderRejectionReason::Error(err),
             }),
         }
     }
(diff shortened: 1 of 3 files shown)


Z4RX (commented):
`Reason` type should be exportable for having possibility to handle different reasons manually.
So those changes are not enough because `typed_header` module is private and all rejections is public from `axum::extract::rejection`.
IMHO simply reimport `Reason` type in `rejection` mode will be at least strange. So probably better way is making public only header name.

src/extract/typed_header.rs
```diff
@@ -79,12 +79,14 @@ impl<T> Deref for TypedHeader<T> {
 #[cfg_attr(docsrs, doc(cfg(feature = "headers")))]
 #[derive(Debug)]
 pub struct TypedHeaderRejection {
```
davidpdrsn, line 81: I think we should make this `#[non_exhaustive]` so users cannot construct this from the outside. Feels the safest to me and we can always change it if someone has a use-case.

src/extract/typed_header.rs
```diff
@@ -79,12 +79,14 @@ impl<T> Deref for TypedHeader<T> {
...
-    reason: Reason,
+    /// Name of the header that caused the rejection
+    pub name: &'static http::header::HeaderName,
+    /// Reason why the header extraction has failed
+    pub reason: Reason,
 }
 
 #[derive(Debug)]
-enum Reason {
+pub enum Reason {
```
davidpdrsn, line 89: The `typed_header` module isn't public so I don't think this type will actually be accessible. I think we should rename it to `TypedHeaderRejectionReason` it expose it in `axum::extract::rejection`, like the other rejection types.

davidpdrsn (changes requested):

davidpdrsn (approved):
Great! Thanks 😊 Wanna put a line in the changelog as well?

AlisCode:
Added the changelog entry as well

PR #339, opened 2021-09-19
Improve performance of `BoxRoute`

davidpdrsn:
This is based on #315 but slightly shorter.

It removes the need for a `tower::buffer::Buffer` in `BoxRoute` which
improves performance.


src/routing/mod.rs
```diff
@@ -256,7 +253,7 @@ impl<S> Router<S> {
     /// routes.
     pub fn boxed<ReqBody, ResBody>(self) -> Router<BoxRoute<ReqBody, S::Error>>
     where
-        S: Service<Request<ReqBody>, Response = Response<ResBody>> + Send + 'static,
+        S: Service<Request<ReqBody>, Response = Response<ResBody>> + Clone + Send + Sync + 'static,
```
davidpdrsn, line 256: These new bounds makes this a breaking change.

PR #352, opened 2021-09-28
Add validator example

Oodachi:
example of [validator](https://crates.io/crates/validator)


examples/validator/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+publish = false
+version = "0.1.0"
+
+[dependencies]
+anyhow = "1.0.44"
+axum = {path = "../.."}
+serde = {version = "1.0", features = ["derive"]}
+thiserror = "1.0.29"
+tokio = {version = "1.0", features = ["full"]}
+validator = {version = "0.14.0", features = ["derive"]}
```
jplatte, line 13: For consistency, spaces after `{` and before `}` should be added.

examples/validator/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+                (StatusCode::BAD_REQUEST, message)
+            }
+            _ => (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                StatusCode::INTERNAL_SERVER_ERROR.to_string(),
+            ),
+        }
+        .into_response()
+    }
+}
```
jplatte, line 78: Missing final newline.

examples/validator/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[derive(Debug, Deserialize, Validate)]
+pub struct NameInput {
+    #[validate(length(min = 1, message = "Can not be empty"))]
+    pub name: String,
+}
+
+async fn handler(Form(input): Form<NameInput>) -> Result<Html<String>, ServerError> {
+    input.validate()?;
```
davidpdrsn, line 47: I think it would be more idiomatic to create a `ValidatedForm` extractor that wraps `Form` but also calls `.validate()?`. Otherwise users have to remember to check the validation in each handler. Do you think something like that is possible?

examples/validator/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn into_response(self) -> Response<Self::Body> {
+        match self {
+            ServerError::ValidationError(_) => {
+                let message = format!("Input validation error: [{}]", self).replace("\n", ", ");
+                (StatusCode::BAD_REQUEST, message)
+            }
+            _ => (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                StatusCode::INTERNAL_SERVER_ERROR.to_string(),
+            ),
```
davidpdrsn, line 74:
```suggestion
            error => (
                StatusCode::INTERNAL_SERVER_ERROR,
                error.to_string(),
            ),
```

examples/validator/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+    response::{Html, IntoResponse},
+    Router,
+};
+use serde::Deserialize;
+use std::{convert::Infallible, net::SocketAddr};
+use thiserror::Error;
+use validator::Validate;
+
+#[tokio::main]
+async fn main() {
```
davidpdrsn, line 27: Should include the standard logging setup all the examples have.

examples/validator/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+    Ok(Html(format!("<h1>Hello, {}!</h1>", input.name)))
+}
+
+#[derive(Debug, Error)]
+pub enum ServerError {
+    #[error(transparent)]
+    ValidationError(#[from] validator::ValidationErrors),
+
+    #[error(transparent)]
+    InternalServerError(#[from] anyhow::Error),
```
davidpdrsn, line 58: Where does anyhow come from? Does `validator` use that?

davidpdrsn (changes requested):

examples/validator/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+use std::{convert::Infallible, net::SocketAddr};
+use thiserror::Error;
+use validator::Validate;
+
+#[tokio::main]
+async fn main() {
+    if std::env::var_os("RUST_LOG").is_none() {
+        std::env::set_var("RUST_LOG", "example_validator=debug")
+    }
+    tracing_subscriber::fmt::init();
```
davidpdrsn, line 32:
```suggestion
    tracing_subscriber::fmt::init();
    
```

examples/validator/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+    #[validate(length(min = 1, message = "Can not be empty"))]
+    pub name: String,
+}
+
+async fn handler(
+    ValidatedForm(input): ValidatedForm<NameInput>,
+) -> Result<Html<String>, ServerError> {
+    Ok(Html(format!("<h1>Hello, {}!</h1>", input.name)))
+    // Err(ServerError::InternalServerError("custom message".to_string()))
+}
```
davidpdrsn, line 57:
```suggestion
) -> Html<String> {
    Html(format!("<h1>Hello, {}!</h1>", input.name))
}
```

examples/validator/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+#[derive(Debug, Error)]
+pub enum ServerError {
+    #[error(transparent)]
+    ValidationError(#[from] validator::ValidationErrors),
+
+    #[error(transparent)]
+    AxumFormRejection(#[from] axum::extract::rejection::FormRejection),
+
+    #[error("Internal server error: {0}")]
+    InternalServerError(String),
```
davidpdrsn, line 88: This doesn't seem to actually be used.

```suggestion
```

davidpdrsn (changes requested):
Nice! Much better 😊

examples/validator/Cargo.toml
```diff
@@ -5,9 +5,12 @@ publish = false
...
+axum = { path = "../.." }
+serde = { version = "1.0", features = ["derive"] }
 thiserror = "1.0.29"
-tokio = {version = "1.0", features = ["full"]}
-validator = {version = "0.14.0", features = ["derive"]}
+tokio = { version = "1.0", features = ["full"] }
+validator = { version = "0.14.0", features = ["derive"] }
+tracing = "0.1"
+tracing-subscriber = "0.2"
+http-body = "0.4.3"
```
jplatte: Now there's a final newline missing here 😄

Also I think it would be nice if the dependencies were sorted alphabetically.

davidpdrsn (approved):
Thanks!

PR #363, opened 2021-10-03
"matchit" based router

davidpdrsn:
This fixes https://github.com/tokio-rs/axum/issues/240 by using [matchit](https://crates.io/crates/matchit) to do all the routing.

I'll expand the PR description when its ready.


davidpdrsn:
cc @ibraheemdev 😊

ibraheemdev:
:eyes:

I've been meaning to fix the fact that routes like `/a/static` and `/a/:dynamic` are conflicting, I'll try to get a fix out soon because that will probably be a cause of complaint.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+- Big internal refactoring of routing leading to several improvements ([#363])
+  - Wild card routes like `.route("/api/users/*", service)` is now supported.
```
jplatte, line 11:
```suggestion
  - Wildcard routes like `.route("/api/users/*", service)` are now supported.
```

davidpdrsn:
> I've been meaning to fix the fact that routes like /a/static and /a/:dynamic are conflicting, I'll try to get a fix out soon because that will probably be a cause of complaint.

Probably yeah but I don't think its a big issue tbh.

src/routing/mod.rs
```diff
@@ -36,10 +35,22 @@ mod or;
...
+        use std::sync::atomic::{AtomicU64, Ordering};
+        static ID: AtomicU64 = AtomicU64::new(0);
+        Self(ID.fetch_add(1, Ordering::SeqCst))
+    }
+}
+
 /// The router type for composing handlers and services.
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub struct Router<S> {
```
jplatte, line 51: I am surprised this is still a generic type. So even though route matching is no longer done through the "type chain" of routes, that type chain still exists?

jplatte (commented):
Here's my thoughts from a cursory review :)

One thing I've noticed that doesn't relate to a specific code location: There doesn't seem to be any `#[should_panic]` tests for overlapping routes.

davidpdrsn:
> One thing I've noticed that doesn't relate to a specific code location: There doesn't seem to be any #[should_panic] tests for overlapping routes.

I'm adding those as docs tests right this moment.

src/routing/mod.rs
davidpdrsn: The type chain still exists since we need to store the services/handlers somewhere. They are stored together with a unique id and that same id is stored in matchit's `Node` type. When a request comes in we match it with the `Node` which returns an id if a route matches. We then call nested chain of services which compares the id with the service and if it matches it'll call the service otherwise they'll call the next service in the chain which repeats the same process.

CHANGELOG.md
davidpdrsn: Applied manually.

src/routing/mod.rs
jplatte: Is OpenAPI schema generation still the main reason you wouldn't want to store type-erased routes in the nodes instead? Or am I making that sound easier than it actually is?

davidpdrsn: Thats the still the main reason. Would also have a small perf impact but I'm not too worried about that.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+- Big internal refactoring of routing leading to several improvements ([#363])
+  - Wildcard routes like `.route("/api/users/*rest", service)` is now supported.
```
jplatte, line 11: You missed part of the suggestion:

```suggestion
  - Wildcard routes like `.route("/api/users/*rest", service)` are now supported.
```

davidpdrsn: god damn english 😅

src/lib.rs
```diff
@@ -340,6 +273,41 @@
...
+//!
+//! ```rust,no_run
+//! use axum::{
+//!     handler::get,
+//!     Router,
+//! };
+//!
+//! let app = Router::new()
+//!     // this matches any request that starts with `/api`
+//!     .route("/api/*rest", get(|| async { /* ... */ }));
```
EdorianDark, line 288: Since also "/api/bar" matches this route, I think it would be easier to understand to leave "rest" out of "/api/*rest"

davidpdrsn: What do you mean? There is no `bar` in this route? And wildcard captures must be named so `/api/*` is not valid.

src/routing/mod.rs
```diff
@@ -213,13 +254,54 @@ impl<S> Router<S> {
...
-            svc,
-            fallback,
-        })
+    pub fn nest<T>(mut self, path: &str, svc: T) -> Router<Nested<T, S>> {
+        let id = RouteId::next();
+
+        let path = if path == "/" {
+            format!("/*{}", NEST_TAIL_PARAM)
+        } else {
+            format!("{}/*{}", path, NEST_TAIL_PARAM)
```
EdorianDark, line 290: Would nesting with "/{}" instead of "/*{}" also work?
Then nesting would behave more like other routes and users still could Wildcard routes themself.

src/lib.rs
EdorianDark: If wildcard captures must be named, then my comment wrong.
Thanks for answering!

src/routing/mod.rs
davidpdrsn: No. Nested routes must match all routes start with the prefix given by the user. That is what this does.

EdorianDark: Ok thanks

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Adding a conflicting route will now cause a panic instead of silently making
     a route unreachable.
   - Route matching is faster as number of routes increase.
+  - The routes `/foo` and `/:key` are considered to overlap and will cause a
+    panic when constructing the router. This might be fixed in the future.
+    Please open an issue if this is important to you.
```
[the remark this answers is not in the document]
davidpdrsn: By the way I'm not a hurry to ship 0.3 so if we'd have to wait a bit for it thats totally fine.

Issue #375, opened 2021-10-08
Json's FromRequest content type is not general enough


nkconnor:
https://github.com/tokio-rs/axum/blob/2bedcc285bf06765bad545ea5200a3cee09616ff/src/json.rs#L106

`impl FromRequest for Json` looks for an exact match on `application/json`, but, other formats should probably match too. There is a convention of using mime type suffix in JSON schema, e.g. with CloudEvents the content type is `application/cloudevents+json`

PR #377, opened 2021-10-08
Expand json mimes

nkconnor:
Solves #375 with minimal changes.

An alternative is to use the `mime` crate which adds it as a default dependency.


davidpdrsn:
I think adding `mime` as a default dependency is fine since it has no other dependencies. I'm working on a version that uses that.

PR #385, opened 2021-10-14
Fix compile time issues with `Router::boxed`

davidpdrsn:
This adds `BoxRouterLayer` (created with `BoxRouter::layer`) which is used to box routes and removes `Router::boxed`. Rust is apparently bad a checking trait bounds on functions (such as `Router::boxed`) but deals just fine with trait bounds in `impl`s 🤷 

Curiously doing `.layer(BoxRouter::layer())` is slow, but `.layer(BoxRoute::<Body>::layer())` is fast even though `BoxRouter`'s first generic parameter defaults to `Body` 🤔 So the docs specifically call out using `.layer(BoxRoute::<Body>::layer())`. I've tried making it required at compile time to always use a turbofish but haven't found a way.

I still wanna do a bit more testing hence the draft. If things turn out to work well I think we should consider backporting this to 0.2 but deprecating `Router::boxed` instead of removing it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   breaking change if you depend on axum with `default_features = false`. ([#286])
 - **breaking:** Remove `routing::Layered` as it didn't actually do anything and
   thus wasn't necessary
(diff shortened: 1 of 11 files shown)


davidpdrsn:
Superseded by https://github.com/tokio-rs/axum/pull/401

PR #401, opened 2021-10-24
Fix compile time regression by boxing routes internally

davidpdrsn:
After much testing this appears to fix the compile time regressions introduced by Rust 1.56. Kinda wish we didn't have to do this but I don't think there is another way. We can always change things if/when the bug is fixed in Rust.

I think we're gonna have to ship this as 0.3 and push the existing things on the [0.3 milestone](https://github.com/tokio-rs/axum/milestone/2) to 0.4. I don't like blocking users from updating to the 2021 edition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- Improve performance of `BoxRoute` ([#339])
-- Expand accepted content types for JSON requests ([#378])
+- **fixed:** Fix regressions in compile times introduced by Rust 1.56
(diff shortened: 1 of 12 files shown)


paolobarbolini:
At least speaking for me unless it requires them multiple versions to fix this issue I'd wait to see what happens in 1.57.

https://github.com/rust-lang/rust/pull/89831 seems to fix it

```bash
# if you haven't installed it yet
cargo install rustup-toolchain-install-master
# let's try this version https://github.com/rust-lang/rust/pull/89831#issuecomment-946147976
rustup-toolchain-install-master c80bb240f9760b0b0ed2e6045d60508129d31f26
cargo +c80bb240f9760b0b0ed2e6045d60508129d31f26 check # this fixes it for me
```

PR #440, opened 2021-11-01
Remove `Sync` requirement from response bodies

davidpdrsn:
As we learned [in Tonic] bodies don't need to be `Sync` because they can
only be polled from one thread at a time.

This changes axum's bodies to no longer require `Sync` and makes
`BoxBody` an alias for `UnsyncBoxBody<Bytes, axum::Error>`.

Also cleans up some body error trait bounds and a few misc other bounds
I noticed while cleaning this up.

[in Tonic]: https://github.com/hyperium/tonic/issues/117

### DIFF
--- a/src/error_handling/mod.rs
+++ b/src/error_handling/mod.rs
@@ -114,8 +114,8 @@ where
     S: Service<Request<ReqBody>, Response = Response<ResBody>> + Clone,
     F: FnOnce(S::Error) -> Res + Clone,
     Res: IntoResponse,
-    ResBody: http_body::Body<Data = Bytes> + Send + Sync + 'static,
-    ResBody::Error: Into<BoxError> + Send + Sync + 'static,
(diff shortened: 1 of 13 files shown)


src/error_handling/mod.rs
```diff
@@ -114,8 +114,8 @@ where
     S: Service<Request<ReqBody>, Response = Response<ResBody>> + Clone,
     F: FnOnce(S::Error) -> Res + Clone,
     Res: IntoResponse,
-    ResBody: http_body::Body<Data = Bytes> + Send + Sync + 'static,
-    ResBody::Error: Into<BoxError> + Send + Sync + 'static,
+    ResBody: http_body::Body<Data = Bytes> + Send + 'static,
+    ResBody::Error: Into<BoxError>,
```
LucioFranco, line 118: Nice!

LucioFranco (approved):

PR #443, opened 2021-11-02
Add example showing how to handle empty vs missing query params

davidpdrsn:
As discussed in discord. I think it makes sense to have an example for this so we have something to link to when people ask.

### DIFF
--- a/examples/query-params-with-empty-strings/Cargo.toml
+++ b/examples/query-params-with-empty-strings/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "example-query-params-with-empty-strings"
+version = "0.1.0"
+edition = "2018"
+publish = false
+
+[dependencies]
+axum = { path = "../.." }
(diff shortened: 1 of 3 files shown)


jplatte:
I'm a little worried about offering this as the "one" solution. `String::into_deserializer` + `T::deserialize` might not be suitable for everyone's use cases. `serde_with` offers [a different solution](https://docs.rs/serde_with/1.11.0/serde_with/rust/string_empty_as_none/index.html) that uses `FromStr`, maybe mention that too?

davidpdrsn:
Examples aren't meant to be the one way true way to do something but I see why someone might get that impression. I'll mention `serde_with` as well.

jplatte:
I think this example is also a bit special because the docs link to it.

davidpdrsn:
Yeah thats true. I've updated the example to mention serde_with.

jplatte (approved):

PR #444, opened 2021-11-02
Add `Handler::{into_make_service, into_make_service_with_connect_info}`

davidpdrsn:
These are useful if you want to run a handler without a `Router`, for
example to make a proxy.


jplatte:
> for example to make a proxy.

So would it make sense to update the `http-proxy` example to use this?

davidpdrsn:
@jplatte uhh yeah good idea! I'll do that now.

davidpdrsn:
Hm actually most of the http-proxy example is about handling `CONNECT`. So not sure this makes sense to use for that exact example.

PR #450, opened 2021-11-02
Add readme example project

jgould22:
## Motivation

Users first encountering Axum on github might want to run the example in the `README.md` however it requires a few dependencies to run.

## Solution

This pull request provides a `readme-example` project in the examples directory with a `cargo.toml` that includes the dependencies required to run it

It makes a small code change to the example to log at the info level and also makes a small change to the `README.md` to provide a link directly to the readme-example.


README.md
```diff
@@ -50,7 +50,7 @@ async fn main() {
     // run our app with hyper
     // `axum::Server` is a re-export of `hyper::Server`
     let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
-    tracing::debug!("listening on {}", addr);
+    tracing::info!("listening on {}", addr);
```
davidpdrsn, line 53: I don't see much reason to change this and its inconsistent with the other example. I think we should just keep it like it was.

examples/readme-example/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "example-readme"
```
davidpdrsn, line 2: I think this should follow the convention set by the other examples. So the directory is called `readme` while the crate is called `example-readme`.

examples/readme-example/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition = "2018"
+publish = false
+
+[dependencies]
+axum = { path = "../.." }
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0.68"
+tokio = { version = "1.0", features = ["full"] }
+tracing = "0.1"
+tracing-subscriber = "0.3"
```
davidpdrsn, line 13: The other examples still uses tracing-subscriber 0.2 so I think CI will complain about duplicate dependencies. Anyway I think we should go with 0.2 and update all the examples in one go.

```suggestion
tracing-subscriber = "0.2"
```

examples/readme-example/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+    let app = Router::new()
+        // `GET /` goes to `root`
+        .route("/", get(root))
+        // `POST /users` goes to `create_user`
+        .route("/users", post(create_user));
+
+    // run our app with hyper
+    // `axum::Server` is a re-export of `hyper::Server`
+    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
+    tracing::info!("listening on {}", addr);
```
davidpdrsn, line 31:
```suggestion
    tracing::debug!("listening on {}", addr);
```

README.md
```diff
@@ -92,7 +92,9 @@ struct User {
 }
 ```
 
-See the [crate documentation][docs] for way more examples.
+You can find this [example][readme-example] as well as other example projects in the [example directory][examples].
+
+There are also a number of usage examples in the [crate documentation][docs].
```
davidpdrsn, line 97: I think the previous wording is fine.

```suggestion
See the [crate documentation][docs] for way more examples.
```

davidpdrsn (changes requested):
Thanks for doing this! I had a few comments. Mostly minor things.

README.md
jgould22: Sure, I thought it was just a little more new user friendly since it doesn't print anything by default with the suggested cargo run

Either way is fine with me

davidpdrsn (approved):
Thanks!

PR #456, opened 2021-11-03
Reorganize tests

davidpdrsn:
This breaks up the large `crate::tests` module by moving the tests into places that are more focused. For example tests of JSON serialization are moved to the `crate::json` module. The remaining routing tests have been moved to `crate::routing::tests`.

I generally prefer having tests close to the code they're testing. Makes it easier to see how/if something is tested.


.github/workflows/CI.yml
```diff
@@ -64,7 +64,6 @@ jobs:
       run: cargo hack check --each-feature --no-dev-deps --all
 
   test-versions:
```
jplatte, line 66: I think you can remove `stable` from this job's `matrix` because the `test-examples` job will run all of the regular tests as well.

davidpdrsn: Moved things around a bit in 2decbe357b41f

PR #458, opened 2021-11-03
Move axum crate into workspace subfolder

davidpdrsn:
Over time I imagine we're gonna have other crates in this repo that
provide utilities or integrations for axum. This prepares for that by
moving the main axum crate into its own folder.

~~The README situation is a bit annoying because we want `./README.md`
for viewing the repo on github but `axum/README.md` for crates.io. For
now I've just copy/pasted it and added CI step to make sure they're
identical.~~


jplatte:
Re. `README.md`, have you tried a symlink in either direction?

davidpdrsn:
I haven't. I'm a bit worried that it'll break how things look on crates.io. [Tokio](https://github.com/tokio-rs/tokio) doesn't do it either which is where I looked first.

jplatte:
What if the root one is the symlink (if GitHub still renders it then)? That can't break crates.io since the workspace root isn't included in crate tarballs.

davidpdrsn:
Good idea! That seems to work https://github.com/tokio-rs/axum/tree/workspace

davidpdrsn:
omg ci is green 🟢 🕺

jplatte (approved):

jplatte:
GitHub gets really confused when you replace a file with a symlink 😄 

Screenshot_2021-11-03_123258

PR #459, opened 2021-11-03
Add `axum-openapi` crate

davidpdrsn:
This is a proof of concept implementation of OpenAPI generation for axum.

I've been thinking that it would be nice if we could iterate on openapi stuff in a separate crate, so we wouldn't have to ship breaking versions of axum itself all the time. When we've found an API we're happy with and its been tested by users, we can move it into axum itself.

So this introduces a new crate called `axum-openapi` (nothing on crates.io yet). The basic idea is creating types that wrap those in axum, provide an identical API, but also support generating openapi schemas for your routes. I think this is a decent way to experiment with it without changing a lot in axum.

Example:

```rust
let app = OpenApiRouter::new()
    .route("/", get(|| async { Json(json!({ "foo": "bar" })) }))
    .route("/foo", get(|| async {}).post(|| async {}));

let (router, openapi) = app.into_parts();

// we serialize our schema and print it
println!("{}", serde_yaml::to_string(&openapi).unwrap());

// or generate some routes to serve it to clients
let router_with_openapi_schema = router.merge(schema_routes(openapi).into_router());
```

Calling `GET /openapi.yaml` returns

```yaml
---
openapi: ""
info:
  title: ""
  version: ""
paths:
  /:
    get:
      responses:
        default:
          description: ""
          content:
            application/json: {}
  /foo:
    get:
      responses: {}
    post:
      responses: {}
```

There is still lots to do but hopefully this illustrates one possible path forward.

---

Notes to self:
- https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md
- https://docs.rs/okapi/0.7.0-rc.1/okapi/openapi3/struct.OpenApi.html
- https://docs.rs/schemars/0.8.6/schemars/gen/struct.SchemaSettings.html#method.openapi3

### DIFF
--- a/axum-openapi/src/lib.rs
+++ b/axum-openapi/src/lib.rs
@@ -1,0 +1,291 @@
+#![allow(
+    clippy::field_reassign_with_default,
+    clippy::new_without_default,
+    clippy::type_complexity,
+    dead_code,
+    unreachable_code,
+    unused_imports,
+    unused_mut,
+    unused_variables
+)]
+
+use axum::{
+    body::{Body, BoxBody, Bytes, Full, HttpBody},
+    http::{header, HeaderValue, Request, Response, StatusCode},
+    response::IntoResponse,
+    routing::Route,
+    BoxError, Json, Router,
+};
+use openapiv3::{HeaderStyle, MediaType, OpenAPI, Operation, PathItem, ReferenceOr, Responses};
+use std::{borrow::Cow, convert::Infallible, future::Future};
+use tower_layer::Layer;
+use tower_service::Service;
+
+pub mod handler_method_routing;
+
+pub struct OpenApiRouter<B = Body> {
+    router: Router<B>,
+    schema: OpenAPI,
+}
+
+impl<B> OpenApiRouter<B>
+where
+    B: Send + 'static,
+{
+    pub fn new() -> Self {
+        Self::with_openapi(Default::default())
+    }
+
+    pub fn with_openapi(openapi: OpenAPI) -> Self {
+        Self {
+            router: Default::default(),
+            schema: openapi,
+        }
+    }
+
+    pub fn route<T>(mut self, path: &str, service: T) -> Self
+    where
+        T: Service<Request<B>, Response = Response<BoxBody>, Error = Infallible>
+            + ToPathItem
+            + Clone
+            + Send
+            + 'static,
+        T::Future: Send + 'static,
+    {
+        // TODO(david): correct path templating with `{key}` instead of `:key`
+        // TODO(david): does openapi have wildcards?
+
+        self.schema
+            .paths
+            .insert(path.to_string(), ReferenceOr::Item(service.to_path_item()));
+
+        self.router = self.router.route(path, service);
+        self
+    }
+
+    pub fn nest<T>(mut self, path: &str, service: T) -> Self
+    where
+        T: Service<Request<B>, Response = Response<BoxBody>, Error = Infallible>
+            + Clone
+            + Send
+            + 'static,
+        T::Future: Send + 'static,
+    {
+        self.router = self.router.nest(path, service);
+        self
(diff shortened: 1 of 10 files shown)


davidpdrsn:
@jplatte what do you think about this approach?

axum-openapi/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+    where
+        T: Service<Request<B>, Response = Response<BoxBody>, Error = Infallible>
+            + ToPathItem
+            + Clone
+            + Send
+            + 'static,
+        T::Future: Send + 'static,
+    {
+        // TODO(david): correct path templating with `{key}` instead of `:key`
+        // TODO(david): does openapi have wildcards?
```
jplatte, line 56: I found this weird in the changelog too.. By "wildcards", you mean multi-segment variables / placeholders, i.e. `*foo` as opposed to `:foo`, right?

jplatte:
I like this approach! Seems like method routing should get an overhaul based on this & the earlier discussion on Discord, doesn't it? ^^

davidpdrsn:
> Seems like method routing should get an overhaul based on this & the earlier discussion on Discord, doesn't it? ^^

Very much so! Its a shame we just shipped 0.3 😜

Flowneee:
Nice :+1:! I personally struggle with OpenAPI in every single project, which involve HTTP API. 

Seems like `okapi` actually generate JSON schema for types, not OpenAPI, and they is not [exactly compatible](https://swagger.io/docs/specification/data-models/keywords/). I still don't know, is this lib generate correct OpenAPI schema (maybe they transform JSON schema to OpenAPI somewhere inside library). May I suggest to take a look at [this lib](https://github.com/Rexagon/opg) for generating OpenAPI spec for types (maybe not only for this)? We use it in some of our projects, works well.

davidpdrsn:
As far as I understand it okapi is for openapi. On crates.io it says "Structs for OpenAPI (AKA Swagger) documents".

schemars is for json schemas for rust types, which are part of openapi schema. This is mentioned on theirs docs and by the fact that the two crates work together.

So I think it'll work fine.

kellytk:
> Its a shame we just shipped 0.3

Brief note to encourage, as a user tracking changes, changes whenever they're unavoidable in improving axum.  Semver'ish, changelog, and migration guide entries are appreciated and 0.3 is too early to confine evolution IMHO.

davidpdrsn:
Haha sure we will have 0.4 eventually but not less than a week after 0.3

davidpdrsn:
An update on this: I haven't had much motivation to work on this lately so if someone wants to continue the work that'd be much appreciated 😊

This PR should contain the overall structure and hopefully it's clear how to otherwise just ping me if you questions. The goal is also to develop things in an external crate so the work doesn't necessarily need to happen in this repo.

FSMaxB:
I will take a closer look at this, try to use it and try to fix what's missing for my use-case (starting in the next couple of weeks). Based on if that works with my use-case and the then gained experience with this approach of adding OpenAPI support, I'll decide if I'll be able to pick this up in general.

tasn:
@FSMaxB, how is it looking? Any luck with getting this to work? (🙏🏻)

tasn:
Thanks @FSMaxB for the update!

I took a brief look over it (because frankly, I don't fully understand the Axum trait magic), and saw you made it into an external crate. I wonder whether it makes sense, since it's an external crate, to cut corners and have something less than perfect/more manual as a first step and then improve it as we go. I think breaking its API would be fine by most people who need it just so they can get started.

On a related note: I don't think we have the trait wizardy on our team to help with figuring out tricks to do it, but we can definitely help with the heavy lifting once there's a plan (e.g. show us how it's done for POST, we can replicate it for PUT, or whatever is relevant). Let me know if there's anything we can do to help.

PR #460, opened 2021-11-03
Fix link to license in `README.md`

david-perez:


davidpdrsn (commented):
I still get a 404 if I click the link in the readme from your branch. You sure this is correct?

jplatte:
It will probably work for the root `README.md` (symlink) on GitHub, but not the actual readme. Maybe link to an online version of the license instead (e.g. https://opensource.org/licenses/MIT)?

davidpdrsn:
This was done in https://github.com/tokio-rs/axum/pull/464.

PR #463, opened 2021-11-04
Derive `Debug` for `Router` instead of implementing manually

david-perez:
I believe `derive` produces the same implementation.


jplatte (approved):

davidpdrsn (approved):
Yeah that makes sense. Was just a holdover from how things used to work.

PR #465, opened 2021-11-05
Update low-level-rustls example

programatik29:
## Motivation

Up to date examples are good and low-level-rustls example can cover more use cases.

## Solution

Update low-level-rustls example.


davidpdrsn (approved):

davidpdrsn:
Thanks!

PR #468, opened 2021-11-05
Add `axum-handle-error-extract` crate

davidpdrsn:
For a project I'm working on with axum I have a "application renderer" extractor thing that takes HTML templates and renders those with a standard application layout around it. It needs to be an extractor because it grabs data from the session to show in the layout like whether the user is logged in or not.

However I couldn't easily use this same setup for handling errors from middleware since `HandleErrorLayer` doesn't support running extractors. I think that is something that makes sense to support. If you don't need to run any extractors you just don't specify any and things work like they did before.

I also think it makes sense to support async functions for handling errors in case users wanna do some async logging or whatnot while handling the error.

So this PR introduces a new crate called `axum-handle-error-extract` which has its own `HandleErrorLayer` that supports extractors and async functions.

For axum 0.4 I'm thinking we should replace the current `HandleErrorLayer` with this new more powerful version. Thats of course a breaking change but figured we could at least release it as a standalone crate so users can try it. Pretty neat that axum is flexible enough to support something like this 😊

## Example usage

```rust
use axum::{
    Router,
    BoxError,
    response::IntoResponse,
    http::{StatusCode, Method, Uri},
    routing::get,
};
use tower::{ServiceBuilder, timeout::error::Elapsed};
use std::time::Duration;
use axum_handle_error_extract::HandleErrorLayer;

let app = Router::new()
    .route("/", get(|| async {}))
    .layer(
        ServiceBuilder::new()
            // timeouts produces errors, so we handle those with `handle_error`
            .layer(HandleErrorLayer::new(handle_error))
            .timeout(Duration::from_secs(10))
    );

// our handler take can 0 to 16 extractors and the final argument must
// always be the error produced by the middleware
async fn handle_error(
    method: Method,
    uri: Uri,
    error: BoxError,
) -> impl IntoResponse {
    if error.is::<Elapsed>() {
        (
            StatusCode::REQUEST_TIMEOUT,
            format!("{} {} took too long", method, uri),
        )
    } else {
        (
            format!("{} {} failed: {}", method, uri, error),
        )
    }
}
```

### DIFF
--- a/axum-handle-error-extract/Cargo.toml
+++ b/axum-handle-error-extract/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "axum-handle-error-extract"
+version = "0.1.0"
+authors = ["David Pedersen <david.pdrsn@gmail.com>"]
+categories = ["asynchronous", "network-programming", "web-programming"]
+description = "Error handling layer for axum that supports extractors and async functions"
+edition = "2018"
+homepage = "https://github.com/tokio-rs/axum"
+keywords = ["http", "web", "framework"]
+license = "MIT"
+readme = "README.md"
+repository = "https://github.com/tokio-rs/axum"
+
+[dependencies]
+axum = { path = "../axum" }
+tower-service = "0.3"
+tower-layer = "0.3"
+tower = { version = "0.4", features = ["util"] }
+pin-project-lite = "0.2"
+
+[dev-dependencies]
+tower = { version = "0.4", features = ["util", "timeout"] }
--- a/axum-handle-error-extract/src/lib.rs
+++ b/axum-handle-error-extract/src/lib.rs
@@ -387,3 +387,16 @@ impl Future for ResponseFuture {
         self.project().future.poll(cx)
     }
 }
+
+/// Extension trait to [`Service`] for handling errors by mapping them to
+/// responses.
+///
+/// See [module docs](self) for more details on axum's error handling model.
+pub trait HandleErrorExt<B>: Service<Request<B>> + Sized {
+    /// Apply a [`HandleError`] middleware.
+    fn handle_error<F>(self, f: F) -> HandleError<Self, F, B> {
+        HandleError::new(self, f)
+    }
+}
+
+impl<B, S> HandleErrorExt<B> for S where S: Service<Request<B>> {}
--- a/axum-handle-error-extract/src/lib.rs
+++ b/axum-handle-error-extract/src/lib.rs
@@ -248,7 +248,7 @@ where
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("HandleError")
             .field("inner", &self.inner)
-            .field("f", &std::any::type_name::<F>())
+            .field("f", &format_args!("{}", std::any::type_name::<F>()))
             .finish()
     }
 }
--- a/axum-handle-error-extract/src/lib.rs
+++ b/axum-handle-error-extract/src/lib.rs
@@ -248,7 +248,7 @@ where
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("HandleError")
             .field("inner", &self.inner)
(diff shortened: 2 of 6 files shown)


axum-handle-error-extract/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+license = "MIT"
+readme = "README.md"
+repository = "https://github.com/tokio-rs/axum"
+
+[dependencies]
+axum = { path = "../axum" }
+tower-service = "0.3"
+tower-layer = "0.3"
+tower = { version = "0.4", features = ["util"] }
+pin-project-lite = "0.2"
```
jplatte, line 19: If the dependencies are meant to be sorted alphabetically, this one should move up to after axum 😉

If you want to automate sorting / check it in CI here and in other `Cargo.toml`s, there is [cargo-sort](https://github.com/DevinR528/cargo-sort).

axum-handle-error-extract/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+    clippy::unnested_or_patterns,
+    rust_2018_idioms,
+    future_incompatible,
+    nonstandard_style,
+    missing_debug_implementations,
+    missing_docs
+)]
+#![deny(unreachable_pub, private_in_public)]
+#![allow(elided_lifetimes_in_paths, clippy::type_complexity)]
+#![forbid(unsafe_code)]
```
jplatte, line 136: FWIW: Assuming that these lint configurations are (mostly) the same as in axum, you can enable them for the entire workspace with [a small hack](Assuming that these lint configurations are (mostly) the same as in axum, you can enable them for the entire workspace with [a small hack](https://github.com/EmbarkStudios/rust-ecosystem/issues/22#issuecomment-947011395). The downside is that Rust <1.55 will choke on the necessary config, so an extra workaround (likely `rm -r .cargo` at the start of the MSRV CI job) would be required.

axum-handle-error-extract/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+}
+
+impl<S, F, E> fmt::Debug for HandleError<S, F, E>
+where
+    S: fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("HandleError")
+            .field("inner", &self.inner)
+            .field("f", &format_args!("{}", std::any::type_name::<F>()))
```
jplatte, line 251: Wouldn't `&type_name()`, w/o the `format_args!`, work too?

axum-handle-error-extract/Cargo.toml
davidpdrsn: Hehe I don't care too much about keeping those sorted.

axum-handle-error-extract/src/lib.rs
davidpdrsn: That seems to also enable the lints for all the examples which is a little annoying 🤔

axum-handle-error-extract/src/lib.rs
davidpdrsn: Yeah I think thats right! No idea why we wrote it like this in tower. Seems to do the same.

jplatte: I think I know now actually 😅

W/o `format_args` it will add quotes, right?

axum-handle-error-extract/src/lib.rs
jplatte: Oh, right. Some might be useful for examples but most probably not.

I guees *if* you move all the library crates into a subdir at some point this will make more sense (since `.cargo` is directory structure based and doesn't have to be next to a `Cargo.toml`).

axum-handle-error-extract/src/lib.rs
davidpdrsn: Oh yeah thats true. I'll revert to keep the `format_args`.

davidpdrsn:
I think I'll go ahead and merge this now. We can always improve things later as we add more crates and use this in practice.

PR #474, opened 2021-11-06
Add `Router::route_layer`

davidpdrsn:
This addresses something thats been bothering me for some time: Most middleware need to run regardless if the request matches a route or not. For example you don't wanna skip logging for unmatched requests.

However middleware such as authorization only make sense to run for matching requests. This previously wasn't possible to express and you'd have to manually apply the middleware to each handler. Consider this:

```rust
Router::new()
    .layer(RequireAuthorizationLayer::bearer("password"));
```

Calling `GET /foo` with an invalid token would receive `401 Unauthorized` as expected however calling some unknown route like `GET /not-found` would also return `401 Unauthorized`. I think this is unexpected and have seen a few users ask questions about it.

It happened because the 404 you'd otherwise see is generated by a fallback service stored on `Router`. When adding a layer to the router the layer would also be applied to the fallback, which in the case of auth means the fallback would never be called for unauthorized requests.

I think what axum does today is the right default however I still think we should support this somehow. Especially since [`extractor_middleware`](https://docs.rs/axum/0.3.1/axum/extract/fn.extractor_middleware.html) is mainly useful for auth but it doesn't work great today due to this gotcha.

This PR proposes adding `Router::layer_on_matching_route` which only applies layers to routes, not the fallback, which fixes the issue. I'm not a big fan of the name `layer_on_matching_route`, would like something shorter, but I think it communicates the purpose decently.

The generics are a bit different since the request body used on the routes and the fallback must match, so layers that changes the request body type are not compatible with `layer_on_matching_route`. Such middleware are very rare so that should be fine.

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- **added:** Add `Router::layer_on_matching_route` for applying middleware that
+- **added:** Add `Router::route_layer` for applying middleware that
   will only run on requests that match a route ([#474])
 
 [#474]: https://github.com/tokio-rs/axum/pull/474
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -7,7 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- **added:** Add `Router::layer_on_matching_route` for applying middleware that
+  will only run on requests that match a route ([#474])
+
+[#474]: https://github.com/tokio-rs/axum/pull/474
 
 # 0.3.1 (06. November, 2021)
 
--- a/axum/src/docs/routing/layer_on_matching_route.md
+++ b/axum/src/docs/routing/layer_on_matching_route.md
@@ -1,0 +1,28 @@
+Apply a [`tower::Layer`] to the router that will only run if the request matches
+a route.
+
+This works similarly to [`Router::layer`] except the middleware will only run if
+the request matches a route. This is useful for middleware that return early
+(such as authorization) which might otherwise convert a `404 Not Found` into a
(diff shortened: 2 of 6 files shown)


davidpdrsn:
I also played around with `tide` a bit and it seems to have the same issue:

```rust
use async_trait::async_trait;
use tide::Request;
use tide::{Middleware, Next};

#[async_std::main]
async fn main() -> tide::Result<()> {
    let mut app = tide::new();
    app.at("/").get(handler);

    app.with(Auth);

    app.listen("127.0.0.1:3000").await?;
    Ok(())
}

async fn handler(_req: Request<()>) -> tide::Result {
    Ok("Hi there!".into())
}

struct Auth;

#[async_trait]
impl<T> Middleware<T> for Auth
where
    T: Clone + Send + Sync + 'static,
{
    async fn handle(&self, req: Request<T>, next: Next<'_, T>) -> tide::Result {
        let values = if let Some(values) = req.header("authorization") {
            values
        } else {
            return Err(tide::Error::from_str(
                tide::StatusCode::Unauthorized,
                "unauthorized",
            ));
        };

        if values.iter().any(|value| value == "Bearer foo") {
            Ok(next.run(req).await)
        } else {
            Err(tide::Error::from_str(
                tide::StatusCode::Unauthorized,
                "unauthorized",
            ))
        }
    }
}
```

Couldn't find a fix.

davidpdrsn:
Actually I just realized that changing the fallback after adding the middleware also fixes it

```rust
Router::new()
    .layer(RequireAuthorizationLayer::bearer("password"))
    .fallback((|| async { StatusCode::NOT_FOUND }).into_service());
```

I wouldn't expect users to think of this (I didn't until just now) and I still think this proposal makes sense.

kellytk:
>I wouldn't expect users to think of this

I wouldn't have.

Regarding naming, as it only applies to a matching route, `.route_layer`?

davidpdrsn:
> Regarding naming, as it only applies to a matching route, .route_layer?

Oh yeah thats better 👍

Issue #477, opened 2021-11-07
If already have a cookie such as `key=value`, exmples/sessions will cause a panic.


kokihonda:
If the browser has a cookie such as  `csrf=xxxxx`, `store.load_session(cookie)`(line 102) will cause a panic.

And I think cookie should be in the form of `key=value`, but currently it is only value.

kokihonda:
# Cause
I ran this program after I ran a different program that listens for localhost.

It seems that this program panicked because the previous program set the value `csrf=xxx` to cookie.

So I realized that there is no problem in normal use. I apologize for the bother.

PR #479, opened 2021-11-07
Update examples/sessions(#477)

kokihonda:

### DIFF
--- a/examples/sessions/src/main.rs
+++ b/examples/sessions/src/main.rs
@@ -17,9 +17,16 @@ use axum::{
     routing::get,
     AddExtensionLayer, Router,
 };
+use cookie::Cookie;
+use lazy_static::lazy_static;
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use std::net::SocketAddr;
 use uuid::Uuid;
+
+lazy_static! {
+    static ref SESSION_ID: String = Uuid::new_v4().to_string();
+}
 
 #[tokio::main]
 async fn main() {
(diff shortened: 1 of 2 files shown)


examples/sessions/src/main.rs
```diff
@@ -124,3 +135,18 @@ impl UserId {
...
+        let kv: Vec<&str> = cookie.split("=").collect();
+        if kv.len() < 2 {
+            return HashMap::new();
+        }
+        let key = kv[0].trim_start();
+        let value = kv[1].trim_end();
+        map.insert(key.to_string(), value.to_string());
+    }
+    map
+}
```
davidpdrsn, line 152: Do you know if `cookie` has a built in solution to this? Perhaps using [`CookieJar`](https://docs.rs/cookie/0.15.1/cookie/struct.CookieJar.html). Having to parse cookies yourself doesn't feel something you should be doing.

examples/sessions/src/main.rs
```diff
@@ -17,10 +17,17 @@ use axum::{
...
 };
+use cookie::Cookie;
+use lazy_static::lazy_static;
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use std::net::SocketAddr;
 use uuid::Uuid;
 
+lazy_static! {
+    static ref SESSION_ID: String = Uuid::new_v4().to_string();
```
davidpdrsn, line 28: What is this for?

davidpdrsn (changes requested):
Can you provide some more context for these changes? What was the bug? How did you address it? What are the tradeoffs? Is this something we'd expect every use to have to deal with?

examples/sessions/src/main.rs
kokihonda: Sorry for the lack of explanation. For example, if we run the program with a cookie of `csrf=xxx', it will panic as follows.
````
2021-11-07T13:37:06.702658Z DEBUG example_sessions: listening on 127.0.0.1:3000
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Invalid byte 61, offset 4.', examples_sessions\src\main. rs:102:79
````

So I tried to deal with the problem by parsing the cookie, creating a `HashMap` with the key and value of the cookie, and passing the value of the cookie given by this program to `load_session`.

> Do you know if cookie has a built in solution to this? Perhaps using CookieJar. Having to parse cookies yourself doesn't feel something you should be doing.

I only knew about the `cookie` crate😢 Yes, it's not something I would do.

examples/sessions/src/main.rs
kokihonda: If the cookie's key is always the same, the program will assume that he have already set the cookie when the program runs multiple times.

To prevent this problem, I used uuid.

kokihonda:
As I mentioned in the issue, I noticed that under normal circumstances it is not a problem, so I close the pull request.  I apologize for the bother.

Issue #451, opened 2021-11-03
Overlapping routes with differing methods cause a panic in 0.3.0


dwalker109:
### Version

axum v0.3.0

### Platform

Darwin Kernel Version 20.3.0 (macOS Big Sur)

### Description

I upgraded to 0.3.0 tonight and now see a panic when running my app: 
`Invalid route: insertion failed due to conflict with previously registered route: /`

This is caused by having both a GET and POST handler handing the same path. For example:

```rust
pub fn routes() -> Router {
    Router::new()
        .route("/", post(create_post))
        .route("/", get(index_get))
}

Router::new()
    .nest("/themes", routes())
```

Both routes were previously reachable in 0.2.*, though they were (naturally) boxed before.

kellytk:
Shouldn't those be a single route?

`Router::new().route("/", post(handler).get(handler))` is what I'm using successfully with 0.3.

davidpdrsn:
Yep @kellytk is right. You have to add the route once with handlers for both methods.

This is necessary with the changes made to routing in 0.3.

jplatte:
From Discord it sounded like it wouldn't be _that_ hard to support this (especially if the pending method router refactoring goes through), so could we re-open this issue for tracking that?

davidpdrsn:
I don't remember talking about it but I don't think it'll be easy. I'm thinking the easiest solution would be if matchit's `Node` provided a way to look up if a route already exists and get its corresponding key. Then once #484 is merged we could probably do some stuff to attempt to merge the method routers if we find one already exists.

cc @ibraheemdev

ibraheemdev:
> I'm thinking the easiest solution would be if matchit's Node provided a way to look up if a route already exists and get its corresponding key.

You can just call `node.at(route).ok().is_some()` (ie: the url path `/foo/:bar` will match the route `/foo/:bar`).

The way I would do this, as I mentioned earlier, is by storing a `HashMap<Method, Node<Service>>` instead of a single `Node<MethodRouter>`:
```rust
app.route("/", get(foo))           => map.get(GET).insert("/", foo)
app.route("/", get(foo).post(bar)) => map.get(GET).insert("/", foo), map.get(POST).insert("/", bar)
```

There wouldn't be any difference between `.route("/", post(handler).get(handler))` and if you registered them both individually, because both just stick the routes in the map.

[...]

PR #484, opened 2021-11-08
Add new `MethodRouter`

davidpdrsn:
This introduces a new crate called `axum-method-router`. It contains a new `MethodRoute` that works very similarly how to `Router` works where each route is boxed internally. This should lead to similar improvements like better compile times.

The API is also more consistent work `Router` as the new `MethodRouter` has gained `merge`, `layer`, `route_layer`, and `fallback` methods. There is also now also only one type for routing to services and handlers.

With this can do stuff like this:

```rust
Router::new().route(
    "/",
    post(_)
        .layer(concurrency_limit)
        .get(_)
        .merge(put(_).patch(_).route_layer(auth))
        .fallback(_)
        .layer(logging)
)
```

Similarly to https://github.com/tokio-rs/axum/pull/468 the idea is to release this as a standalone crate and make it the default in 0.4

The PR looks bigger than it really is. There is a lot of repetition for defining everything for all HTTP methods.

## TODO

- [ ] Write docs


jplatte:
This doesn't (on its own) solve the

```rust
.route("/foo", get(_))
.route("/foo", post(_))
```

issue, right?

davidpdrsn:
I've decided to just integrate this directly into axum without a separate crate. Thinking we might wanna ship 0.4 sooner rather than later. Its been pretty quiet lately with feature requests and bug reports. I'll submit that as a new PR. It will also include a fix for https://github.com/tokio-rs/axum/issues/451.

PR #485, opened 2021-11-09
Split first sentence into its own paragraph in TypedHeader docs

jplatte:
It's pretty uncommon to have a big first paragraph. I think it's also the breaking point to tell rustdoc what to show in a module overview.


davidpdrsn (approved):
👍

PR #486, opened 2021-11-09
Type aware `empty_string_as_none` implementation

imbolc:
The original version of `empty_string_as_none` worked only with string parameters, now it accepts other flat types

### DIFF
--- a/examples/query-params-with-empty-strings/src/main.rs
+++ b/examples/query-params-with-empty-strings/src/main.rs
@@ -5,7 +5,8 @@
 //! ```
 
 use axum::{extract::Query, routing::get, Router};
-use serde::{de::IntoDeserializer, Deserialize, Deserializer};
+use serde::{de, Deserialize, Deserializer};
+use std::{fmt, str::FromStr};
 
 #[tokio::main]
 async fn main() {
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
Looks good to me.

Wanna fix the "field is never read" lints as well? Can just add `#[allow(...)]` on the fields.

@jplatte what do you think?

imbolc:
For some reason I don't see any warnings:
```sh
axum/examples/query-params-with-empty-strings$ cargo check
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s

axum/examples/query-params-with-empty-strings$ rustc --version --verbose
rustc 1.56.1 (59eed8a2a 2021-11-01)
binary: rustc
commit-hash: 59eed8a2aac0230a8b53e89d4e99d55912ba6b35
commit-date: 2021-11-01
host: x86_64-unknown-linux-gnu
release: 1.56.1
LLVM version: 13.0.0
```

davidpdrsn:
Yeah me neither. Not sure if it's a GitHub actions thing. It's quite annoying 😞

jplatte:
Yeah, so GitHub actions doesn't actually test the HEAD of the PR by default, it checks what the repo would look like after merge. You can change that: https://github.com/actions/checkout#Checkout-pull-request-HEAD-commit-instead-of-merge-commit

> @jplatte what do you think?

I suggested this change (https://github.com/ruma/ruma/pull/751#discussion_r741300463) so I approve 😄

> Wanna fix the "field is never read" lints as well? Can just add `#[allow(...)]` on the fields.

These warnings were just fixed in #487 😉

jplatte (approved):

PR #487, opened 2021-11-09
Use nightly clippy & rustfmt for more style checks

jplatte:
## Motivation

Improving the readability of the codebase.

## Solution

Use nightly clippy & rustfmt and configure them to catch more things. We've been doing this in [Ruma](https://github.com/ruma/ruma) for a while now and haven't suffered from nightly instability much (once every few months a clippy lint produces some false-positives, but that's it).

## Notes

I am not a fan of the `mod.rs` module style, but it clearly seemed prevalent in this repo. To enforce the opposite, one would activate the `clippy::mod_module_files` lint.


davidpdrsn (approved):
Nice!

I prefer `mod.rs` as you probably guessed since it put all the files in the same folder.

jplatte:
Opened a rustfmt issue for one of the issues I saw when enabling `wrap_comments = true` + `comment_width = 100`: https://github.com/rust-lang/rustfmt/issues/5073

There were more, but those seemed like they would require more investigation.

PR #489, opened 2021-11-09
Implement `FromRequest` for `http::request::Parts`

davidpdrsn:
Its a convenient way to extract everything from the request except the
body. Also makes sense for axum to provide this since other crates
can't.

### DIFF
--- a/axum/src/extract/request_parts.rs
+++ b/axum/src/extract/request_parts.rs
@@ -323,10 +323,49 @@ where
     }
 }
 
+#[async_trait]
+impl<B> FromRequest<B> for http::request::Parts
+where
+    B: Send,
+{
+    type Rejection = RequestPartsAlreadyExtracted;
+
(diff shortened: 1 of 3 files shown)


axum/src/extract/request_parts.rs
```diff
@@ -323,10 +323,49 @@ where
...
+        let extensions = Extensions::from_request(req).await?;
+
+        let mut temp_request = Request::new(());
+        *temp_request.method_mut() = method;
+        *temp_request.uri_mut() = uri;
+        *temp_request.version_mut() = version;
+        *temp_request.headers_mut() = headers;
+        *temp_request.extensions_mut() = extensions;
+
+        let (parts, _) = temp_request.into_parts();
```
jplatte, line 347: It's unfortunate that there is no way to exhaustively create `request::Parts`.

jplatte (approved):

axum/src/extract/request_parts.rs
davidpdrsn: Yeah 😞

PR #490, opened 2021-11-09
Implement `IntoResponse` for `http::response::Parts`

davidpdrsn:
Not sure there are many use cases for this but still thinks it makes to
provide since other crates can't.

### DIFF
--- a/axum/src/response/mod.rs
+++ b/axum/src/response/mod.rs
@@ -250,20 +250,7 @@ impl IntoResponse for http::response::Parts {
     type BodyError = Infallible;
 
     fn into_response(self) -> Response<Self::Body> {
-        let Self {
-            status,
-            version,
-            headers,
-            extensions,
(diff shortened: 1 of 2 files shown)


axum/src/response/mod.rs
```diff
@@ -245,6 +245,28 @@ impl_into_response_for_body!(hyper::Body);
...
+            extensions,
+            ..
+        } = self;
+
+        let mut res = Response::new(Empty::new());
+        *res.status_mut() = status;
+        *res.version_mut() = version;
+        *res.headers_mut() = headers;
+        *res.extensions_mut() = extensions;
+        res
```
jplatte, line 266: Why not use `Response::from_parts`?

davidpdrsn: lol good idea :P

jplatte (approved):

PR #496, opened 2021-11-11
Make `.nest("/", service)` work again (with a warning)

jplatte:
## Motivation

Lots of people don't read the changelog or miss what it says about `.nest("/", svc)`.

## Solution

Make it work again by forwarding to `.fallback`, but raising a warning because `.fallback` is more explicit and panic messages if you do both are not going to be great.

### DIFF
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -40,6 +40,7 @@ tower = { version = "0.4.10", default-features = false, features = ["util", "buf
 tower-http = { version = "0.1", features = ["add-extension", "map-response-body"] }
 tower-layer = "0.3"
 tower-service = "0.3"
+tracing = "0.1"
 
 # optional dependencies
 base64 = { optional = true, version = "0.13" }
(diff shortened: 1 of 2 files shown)


davidpdrsn:
Hm I'm wondering if this is actually doing the users a disservice by making it seem like `nest("/", _)` works.

What do you think about instead storing in a `bool` in `Router` for whether `.nest("/", _)` has been called, and if so when adding another causes a panic, check if that has been called and then change the panic message to mention using `fallback` instead?

jplatte:
I'm a fan of deprecation before removal and since this is possible, it seems better to delay making it not work. Then again most people who will move to axum 0.3 probably have done so already. Your suggestions sounds like an improvement over what we have now, so if you're more comfortable with that than with this PR, don't let me stop you 🙂

davidpdrsn:
Right, I'm a bit worried that removing this in the future would break things again for some users. So I think adding a hint in the panic message is safer. I'll cook up a PR for that.

Will close this for now but thanks for bringing it up!

PR #497, opened 2021-11-11
Move `axum-debug` crate to workspace

programatik29:


davidpdrsn (approved):
Thank you @programatik29! I think I'll just merge it as is and make some smaller adjustments here and there myself.

PR #498, opened 2021-11-11
Clean up axum-debug crate

davidpdrsn:
Mostly just brings the crate more in line with the rest of the crates in
the workspace.

I've also removed the axum-debug-macros crate since axum-debug only
contained one re-export from axum-debug-macros. So we didn't need two
crates. Can always bring the "backend" crate back if we need things in
axum-debug that aren't proc-macros.


axum-handle-error-extract/src/lib.rs
```diff
@@ -125,7 +125,6 @@
     clippy::option_option,
     clippy::verbose_file_reads,
     clippy::unnested_or_patterns,
-    clippy::self_named_module_files,
```
jplatte, line 128: Why did you remove this?

axum-debug/CHANGELOG.md
```diff
@@ -9,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **breaking:** Removed `debug_router` macro.
 - **breaking:** Removed `check_service` function.
 - **breaking:** Removed `debug_service` function.
+- **breaking:** No longer re-exports `axum`.
```
programatik29, line 13: Version `0.1.0` does not re-export `axum`. I did re-export change while moving the crate in.

axum-handle-error-extract/src/lib.rs
davidpdrsn: I got

```
warning: unknown lint: `clippy::self_named_module_files`
   --> axum-handle-error-extract/src/lib.rs:128:5
    |
128 |     clippy::self_named_module_files,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: did you mean: `clippy::self_named_constructors`
    |
    = note: `#[warn(unknown_lints)]` on by default
```

So I guess its a nightly lint and not sure we should have nightly lints disabled for contributors working on stable.

jplatte: Wdym. by having lints disabled for contributors? People could "just" ignore the warning when working on stable.

davidpdrsn: I still get the warning when I'm hacking on the code and I pull the "show me all the warnings window". I assume getting rid of the warning would require running nightly.

davidpdrsn: I don't think telling people to "just ignore the warnings" is great 😕

axum-debug/src/lib.rs
```diff
@@ -84,21 +79,331 @@
...
+
+        quote_spanned! {span=>
+            {
+                debug_handler(#ident);
+
+                fn debug_handler<F, Fut, Res, #(#generics),*>(_f: F)
+                where
+                    F: ::std::ops::FnOnce(#(#generics),*) -> Fut,
+                    Fut: ::std::future::Future<Output = Res>,
+                    Res: ::axum_debug::axum::response::IntoResponse,
```
programatik29, line 373: This needs to be `::axum::response::IntoResponse`.

davidpdrsn: Fixed in 50b841b0f8a16

davidpdrsn: Gonna add some tests in the next PR

axum-handle-error-extract/src/lib.rs
jplatte: > I assume getting rid of the warning would require running nightly.

You could also `#[allow(unknown_lints)]` but that obviously has its downsides as well.

Anyways, I can understand not wanting to enable lints that stable clippy doesn't recognize.

PR #506, opened 2021-11-13
axum-debug: Move axum dependency to dev-dependencies

jplatte:
There is no reason for axum-debug to depend on axum, it's more of a sibling dependency (the code *generated by* axum-debug needs `axum` to be in scope, not axum-debug itself).

For the doc tests, it needs to be available though.

### DIFF
--- a/axum-debug/Cargo.toml
+++ b/axum-debug/Cargo.toml
@@ -18,7 +18,7 @@ proc-macro = true
 proc-macro2 = "1.0"
 quote = "1.0"
 syn = { version = "1.0", features = ["full"] }
-axum = { path = "../axum", version = "0.3" }
 
 [dev-dependencies]
+axum = { path = "../axum", version = "0.3" }
 trybuild = "1.0"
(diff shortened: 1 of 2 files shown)


jplatte:
Hrm, rustdoc won't resolve the `axum` links w/o the dependency. I think the right solution here would be to not use intra-rustdoc links here, because the versions of axum and axum-debug aren't inherently related.

davidpdrsn:
Makes a lot of sense. I actually just noticed that the version of axum-debug on crates.io depends on axum 0.2. So we should probably do a release once this is merged.

> Hrm, rustdoc won't resolve the `axum` links w/o the dependency. I think the right solution here would be to not use intra-rustdoc links here, because the versions of axum and axum-debug aren't inherently related.

Yeah I think thats fine. Can just replace the links with links to `https://docs.rs/...`

jplatte:
> we should probably do a release once this is merged.

Let me get one more PR in :)

PR #508, opened 2021-11-13
Only run axum-debug tests on stable

davidpdrsn:
Quick workaround for CI failing


davidpdrsn:
Uhm okay CI doesn't actually fail 🤔

PR #520, opened 2021-11-15
`axum-debug`: fix UI tests' failures when using rustc 1.54

david-perez:
The MSRV is 1.54. However, `rustc`'s messages in the tests use 1.56.
This commit is the result of running `TRY_BUILD=overwrite cargo test` on
the `axum-debug` crate with 1.54.

### DIFF
--- a/axum-debug/tests/fail/argument_not_extractor.stderr
+++ b/axum-debug/tests/fail/argument_not_extractor.stderr
@@ -2,10 +2,7 @@ error[E0277]: the trait bound `bool: FromRequest` is not satisfied
  --> tests/fail/argument_not_extractor.rs:4:23
   |
 4 | async fn handler(foo: bool) {}
-  |                       ^^^^ the trait `FromRequest` is not implemented for `bool`
-  |
-note: required by a bound in `handler::{closure#0}::debug_handler`
- --> tests/fail/argument_not_extractor.rs:4:23
-  |
-4 | async fn handler(foo: bool) {}
-  |                       ^^^^ required by this bound in `handler::{closure#0}::debug_handler`
+  |                       ^^^^
+  |                       |
+  |                       the trait `FromRequest` is not implemented for `bool`
+  |                       required by this bound in `handler::{closure#0}::debug_handler`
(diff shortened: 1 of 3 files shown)


davidpdrsn:
This is being fixed in https://github.com/tokio-rs/axum/pull/518. That changes it to only run axum-debug's ui tests on stable. So I guess we don't need this? I can file it as a separate PR if necessary.

david-perez:
What's the reasoning for #518 running the tests with stable, when `axum`'s MSRV is 1.54?

davidpdrsn:
axum != axum-debug. Its only axum-debug's tests we don't run, but we still axum-debug on 1.54. I'm confident that if it builds on 1.54 then it also works.

david-perez:
`axum-debug` is a crate that [currently advertises MSRV of 1.54](https://github.com/tokio-rs/axum/blob/e2e9f1ff68dc54f5a4a90c1d20b93d619bccee0e/axum-debug/README.md?plain=1#L18); is that correct? If it is, I think everyone expects a crate to `cargo test` without errors on its MSRV.

---

I think `axum-debug`'s MSRV should be kept at the same MSRV than `axum`, because users will depend on both `axum` and `axum-debug` when developing their projects. By ensuring both crates build and test successfully in CI at the same MSRV, we avoid the possibility of breaking any users...

> I'm confident that if it builds on 1.54 then it also works.

however slim that may be. (As the code stands, I'm confident too, but what benefit does introducing the disparity offer?)

davidpdrsn:
How do you propose we keep the UI tests passing both stable and 1.54 given the output is different? Have two versions of the expected output in the repo? I'm not saying that's a bad thing necessarily but it's at least something we have to maintain.

david-perez:
> Have two versions of the expected output in the repo? I'm not saying that's a bad thing necessarily but it's at least something we have to maintain.

And have the test detect with which `rustc` version it is being invoked? Yes, I can't think of another way.

---

I think it would be the ideal scenario, to have _both_ ends of the range [`MSRV`, `stable`] tested in CI.

However, this PR would only make it so that the left end of the range is tested, and after thinking about it some more, if we're only going to test one end of the range, I don't think there is a compelling reason to favor one over the other, for a crate of this nature.

So if you don't feel inclined to test both ends of the range, feel free to close this PR. Otherwise, let me know and I can try rework it and add both outputs to the repo, and do some `rustc` version detection in the tests with [`version_check`](https://docs.rs/version_check/0.9.3/version_check/).

davidpdrsn:
I think since we build the crate for all versions and run doctests we will be fine. I'll close this for now.

PR #522, opened 2021-11-16
Drop `Sync` requirement on the future returned by a handler

david-perez:
Since it doesn't need to be polled from different threads at the same
time.

---

I wonder why the requirement is there, since it appears the future is never polled from different threads at the same time. Perhaps this is a holdover from #440?


axum-debug/src/lib.rs
```diff
@@ -346,7 +346,7 @@ mod debug {
 
                 fn debug_handler<F, Fut, #(#generics),*>(_f: F)
                 where
-                    F: ::std::ops::FnOnce(#(#generics),*) -> Fut + Clone + Send + Sync + 'static,
+                    F: ::std::ops::FnOnce(#(#generics),*) -> Fut + Clone + Send + 'static,
                     Fut: ::std::future::Future + Send,
```
davidpdrsn, line 350: The future is not required to be `Sync`, the closure that returns it is. The `Fut` type is the future. `F` is the closure.

Probably still not necessary since it requires `FnOnce` so you must have an owned closure to call.

I'm curious if you've run into some issue caused by this.

david-perez: Ah, true. I've reworded the PR title and commit message.

> I'm curious if you've run into some issue caused by this.

No, just learning :) I thought the signature was odd.

davidpdrsn: > No, just learning :) I thought the signature was odd.

Thats fine 😊

I'm a bit unsure if this could have unforeseen consequences later down the road 🤔 @jplatte What do you think about this?

jplatte: Since `Sync` only has meaning when you take a reference to the thing and a `&impl FnOnce` is quite useless (you can't call it unless you have ownership), I think this change should be fine. In general I'm always in favor of keeping generics as simple as possible.

davidpdrsn: Thanks for confirming my understanding 😊 This is good to merge then.

PR #524, opened 2021-11-16
Bump version to 0.4.0

davidpdrsn:
Not releasing 0.4.0 yet but work has begun in `main`.

### DIFF
--- a/axum-debug/Cargo.toml
+++ b/axum-debug/Cargo.toml
@@ -20,5 +20,5 @@ quote = "1.0"
 syn = { version = "1.0", features = ["full"] }
 
 [dev-dependencies]
-axum = { path = "../axum", version = "0.3" }
+axum = { path = "../axum", version = "0.4" }
 trybuild = "1.0"
--- a/axum-handle-error-extract/Cargo.toml
+++ b/axum-handle-error-extract/Cargo.toml
@@ -12,7 +12,7 @@ readme = "README.md"
 repository = "https://github.com/tokio-rs/axum"
 
 [dependencies]
-axum = { version = "0.3.2", path = "../axum" }
+axum = { version = "0.4", path = "../axum" }
 tower-service = "0.3"
 tower-layer = "0.3"
 tower = { version = "0.4", features = ["util"] }
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "axum"
-version = "0.3.3"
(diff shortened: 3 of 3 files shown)


jplatte (changes requested):
I strongly recommend against bumping the version before doing the release, since this makes it impossible to use the `[patch]` section to replace the released version with what's in the git repo.

davidpdrsn:
Not impossible. This works

```toml
[dependencies]
axum = "0.4"

[patch.crates-io]
axum = { git = "https://github.com/tokio-rs/axum.git", branch = "0.4.0" }
```

[tracing also does it](https://github.com/tokio-rs/tracing/blob/master/tracing/Cargo.toml#L11) and makes it more clear that things have changed to people browsing the code on github.

jplatte:
Yes, it works if your dependency is on "0.4", but since that version is not released nobody is going to have a dependency on 0.4. When trying to use it on a project that (indirectly) depends on 0.3, you get the following warning:

```
warning: Patch `axum v0.4.0 (https://github.com/tokio-rs/axum.git?branch=0.4.0#d173a1a7)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
```

and the `patch` becomes ineffective.

davidpdrsn:
How could that ever work? If we keep the version at 0.3.x there will still be breaking changes which other dependencies cannot incorporate, unless you fork those as well and add a patch.

jplatte:
Other crates depending on axum are likely to only rely on parts of its interface such as `IntoResponse` and `FromResponse`. In such cases, they would continue to work against a `[patch]`ed-in git version.

As an example, let's say I have two crates in separate repos, `axum-extra-extractors` and `my-site`, plus a local workdir of `axum` where I want to experiment with a change to the router that I would later like to upstream if it works out.

If the version in the local workdir is the same as the latest released one, I can simply add `[patch.crates-io]` -> `axum = { path = "../axum" }` to the `my-site` repo and use it that way.
If the version in the local workdir is not the same as the latest released one, I have to either

* change it back, which will likely lead to that change accidentally being committed
* change axum to a path dependency in both my `axum-extra-extractors` workdir (assuming there is one) and `my-site` workdir, and change the `axum-extra-extractors` dependency as well

davidpdrsn:
Alright. I don't think this a big deal regardless.

PR #529, opened 2021-11-16
Panic instead of silently discarding fallbacks

### LINKED ISSUE #488: Panic if merging two routers that each have a fallback
Currently when you `Router::merge` two routes that each have a fallback it'll pick the fallback of the router on the right hand side, and if only one has a fallback it'll pick that one. This might lead users to think that multiple fallbacks are supported (see https://github.com/tokio-rs/axum/discussions/480#discussioncomment-1601938), which it isn't.

I'm wondering if, in 0.4, we should change it such that merging two routers that each have a fallback causes a panic telling you that multiple fallbacks are not supported. This would also be consistent with the general routing which panics on overlapping routes.

The reason multiple fallbacks are not allowed is that `Router` tries hard to merge all your routes in one `matchit::Node` rather than nesting them somehow. This leads to significantly nicer internals and better performance.

@jplatte what do you think?

davidpdrsn:
This introduces two new possible panics when constructing routers:

- If merging two routers that each have a fallback. Previously that left
  side fallback would be silently discarded.
- If nesting a router that has a fallback. Previously it would be
  silently discarded.

Overall this should make things more explicit and users shouldn't have
to worry "why isn't my fallback" working.

Fixes #488


jplatte:
> If nesting a router that has a fallback.

Why is this? It should be possible to have a fallback for `/foo/unrecognized` that's different from `/unrecognized`, no?

davidpdrsn:
Its unfortunately no so straight forward with how `nest` is currently implemented:

If the service you're trying to nest [is actually a `Router`](https://github.com/tokio-rs/axum/blob/main/axum/src/routing/mod.rs#L184) then we simply clone all the routes from the nested router to the outer router. That means we end up one "flat" router in the end which removes pretty much all the [complexity nesting otherwise leads to](https://github.com/tokio-rs/axum/blob/64a3e98d6cb1e694580a7d63478569312b2a15f3/src/routing/mod.rs#L774-L778).

Adding a wildcard route like `.route("/nest-path/*axum_fallback", nest_fallback)` doesn't work since that conflicts with the nested routes.

jplatte:
> Adding a wildcard route like `.route("/nest-path/*axum_fallback", nest_fallback)` doesn't work since that conflicts with the nested routes.

Oh, so while `/foo` and `/:var` don't conflict, `/foo` and `/*wildcard` do?

davidpdrsn:
Yep. That was the source of the whole `nest("/", _)` thing.

jplatte (approved):

PR #536, opened 2021-11-17
Add `axum-extra` crate

davidpdrsn:
Empty for now but now we have a place to put stuff.

I'll make a PR for moving over https://github.com/davidpdrsn/axum-resource.

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -1,5 +1,4 @@
 [package]
-authors = ["David Pedersen <david.pdrsn@gmail.com>"]
 categories = ["asynchronous", "network-programming", "web-programming"]
 description = "Extra utilities for axum"
 edition = "2018"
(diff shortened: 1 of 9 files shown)


axum-extra/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+authors = ["David Pedersen <david.pdrsn@gmail.com>"]
```
jplatte, line 2: By the way, this field hasn't been required for a while now (and `cargo init` no longer sets it), because it can be very annoying to have info that can change such as an email (and name to lesser extent) in a place where it can't be updated like crates.io.

jplatte (approved):

axum-extra/Cargo.toml
davidpdrsn: Ah yes thats right! I'll remove it here and for the other crates as well.

PR #544, opened 2021-11-19
Change how `Resource`s are added to `Router`s

davidpdrsn:
I think being able to create `Resource`s without going through closures,
as was previously required, is nicer:

```rust
let users = Resource::named("users")
    .index(|| async { "users#index" })
    .create(|| async { "users#create" });

let app = Router::new().with(users);
```

Among other things this allows you to build and compose `Resource`s
independently and return them from functions.

---

I also think a general trait for things that can provide routes is nice:

```rust
pub trait HasRoutes<B = Body> {
    fn routes(self) -> Router<B>;
}
```

This works with `RouterExt::with`.

I don't feel strongly about the name `HasRoutes` so feel free to
bikeshed. Could also be called `IntoRouter` 🤷 

For example it allows attribute macros like

```rust
async fn foo() {}
```

To expand to

```rust
struct foo;

impl HasRoutes for foo {
    fn routes(self) -> Router<B> {
        // the original function
        async fn foo() {}
        Router::new().route("/", get(foo))
    }
}
```

and then be added like so

```rust
Router::new().with(foo);
```

### DIFF
--- a/axum-extra/src/routing/resource.rs
+++ b/axum-extra/src/routing/resource.rs
@@ -188,8 +188,8 @@ impl<B: Send + 'static> Resource<B> {
 }
 
 impl<B> HasRoutes<B> for Resource<B> {
-    fn routes(&self) -> Router<B> {
-        self.router.clone()
+    fn routes(self) -> Router<B> {
+        self.router
     }
 }
 
--- a/axum-extra/src/routing/resource.rs
+++ b/axum-extra/src/routing/resource.rs
@@ -1,3 +1,4 @@
+use super::HasRoutes;
 use axum::{
     body::{Body, BoxBody},
     handler::Handler,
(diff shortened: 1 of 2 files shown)


axum-extra/src/routing/resource.rs
```diff
@@ -52,27 +54,17 @@ pub struct Resource<B = Body> {
...
-        T::Future: Send + 'static,
-    {
-        self.router = self.router.route(path, svc);
-        self
+    /// Create a `Resource` with the given name.
+    ///
+    /// All routes will be nested at `/{resource_name}`.
+    pub fn named(resource_name: &str) -> Self {
+        Self {
+            name: resource_name.to_string(),
```
jplatte, line 62: Nitpick: I hate `str_ref.to_string()` because it uses the name of an operation that should be done very rarely (stringification of something that is not a string) in a very boring way and so results in one paying less attention to `.to_string()` calls overall.

I would recommend using `.into()` or `.to_owned()` instead.

There's also a clippy lint for this; if you agree with my reasoning you could turn it on for all the axum crates: https://rust-lang.github.io/rust-clippy/master/index.html#str_to_string

jplatte (approved):
For some reason I hadn't thought of this possibility when I did the rewrite. I agree it looks nicer! 👍🏼

axum-extra/src/routing/resource.rs
davidpdrsn: It's mostly a habit of mine. I don't feel strongly about it. Feel free to add the lint and change existing callsites.

PR #550, opened 2021-11-21
Allow extractor resources to be used external to axum

Dispersia:
Hello, the current extractors only allow use within the crate itself, even though custom extractors would want to be able to reuse this logic.

## Motivation

in making a custom extractor, wanted to reuse the existing MissingExtension/ExtensionAlreadyExtracted

## Solution

just remove the (crate) part of from err

### DIFF
--- a/axum/src/macros.rs
+++ b/axum/src/macros.rs
@@ -86,10 +86,10 @@ macro_rules! define_rejection {
     ) => {
         $(#[$m])*
         #[derive(Debug)]
-        pub struct $name(pub(crate) crate::Error);
+        pub struct $name(pub crate::Error);
 
         impl $name {
-            pub(crate) fn from_err<E>(err: E) -> Self
+            pub fn from_err<E>(err: E) -> Self
             where
                 E: Into<crate::BoxError>,
             {


Dispersia:
for some reason I am not able to create a draft on this repo, however I still have an issue that I can't figure out. I am not able to expose "ExtensionsAlreadyExtracted".

Under: https://github.com/tokio-rs/axum/blob/69fee5864d9463d3646e89cf055019bc702c297b/axum/src/macros.rs#L58
it should match the rule of no fields. However, when actually trying to use it I get the error "Found struct `ExtensionsAlreadyExtracted` constructor is not visible here due to private fields" even though I see no private fields, and it doesn't then match the second macro expansion, so I don't understand how I'm getting this error. Any help, or an explanation if these aren't meant to be exposed would be appreciated, thanks!

davidpdrsn:
It is intentional that the rejections cannot be constructed outside axum as it gives us the ability to change them without making breaking changes.

I would instead recommend that you create your own rejection types as needed.

davidpdrsn:
> Also, can you explain the "ExtensionsAlreadyExtracted" rule? I don't understand quite what the comment means, as there can be multiple extensions that all extract in a single handler, no?

`RequestParts` allows users to [remove all extensions from the request](https://docs.rs/axum/0.3.4/axum/extract/struct.RequestParts.html#method.take_extensions). This is necessary for implementing some extractors such as `Request`. So this rejection will be returned if some other extractor has taken the extensions. The `Extension` extractor does not call this method but instead clones the single type it needs.

Dispersia:
Makes sense, thanks! I'll just create unique rejections then.

Issue #551, opened 2021-11-22
axum_debug reports error when a handler has multiple extractors.


kenmasumitsu:
### Version

axum-debug = "0.2.1"
[69fee5864d9463d3646e89cf055019bc702c297b ](https://github.com/tokio-rs/axum/commit/69fee5864d9463d3646e89cf055019bc702c297b)

### Platform

Darwin 20.6.0 Darwin Kernel Version 20.6.0: Tue Oct 12 18:33:42 PDT 2021; root:xnu-7195.141.8~1/RELEASE_X86_64 x86_64

### Description

<short summary of the bug>

I tried this code:

```

#[debug_handler]
async fn get_user_things(
    Path(user_id): Path<Uuid>,
    pagination: Option<Query<Pagination>>,
) -> impl IntoResponse {
    let Query(pagination) = pagination.unwrap_or_default();

    println!("{:?}, {:?}", user_id, pagination);
    Html("<h1>Hello, World!</h1>")
}

```

I expected to see this happen: No error reports.

Instead, this happened: shows the report below

```
the name `__axum_debug_check_get_user_things_from_request` is defined multiple times
`__axum_debug_check_get_user_things_from_request` must be defined only once in the value namespace of this modulerustcE0428
main.rs(48, 20): previous definition of the value `__axum_debug_check_get_user_things_from_request` here
```
-->

davidpdrsn:
I know exactly that the issue is. I'll get it fixed 😊

davidpdrsn:
Are you able to test https://github.com/tokio-rs/axum/pull/552?

kenmasumitsu:
I confirmed the issue was resolved with #552.

davidpdrsn:
Fix is shipped in axum-debug v0.2.2

kenmasumitsu:
I appreciate your quick fix!

PR #552, opened 2021-11-22
Fix `#[debug_handler]` for multiple extractors

davidpdrsn:
It generated a function for each extractor to check the type but those
functions didn't have unique names.

Fixed by including all idents in the `arg: Extractor` token tree in the
name of the function generated. Not sure there is a simpler way to fix
it.

Fixes https://github.com/tokio-rs/axum/issues/551

### DIFF
--- a/axum-debug/CHANGELOG.md
+++ b/axum-debug/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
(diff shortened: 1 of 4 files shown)


jplatte:
> Fixed by including all idents in the `arg: Extractor` token tree in the name of the function generated. Not sure there is a simpler way to fix it.

How about a counter? 😄

davidpdrsn:
lol omg yes. Welcome to monday morning I guess. My brain was stuck in the "global counter" thing like my earlier PR.

jplatte (approved):

PR #554, opened 2021-11-22
feat: default to charset=utf-8 for text content type

PureWhiteWu:
## Motivation

When using axum, I found that if I returned content in non-English(e.g. Chinese), then all browsers (include Chrome, Safari, Firefox) will display unreadable chars.

After investigating, I found that this is because axum only returns `Content-Type: text/plain` instead of `Content-Type: text/plain; charset=utf-8`.

I think adding `charset=utf-8` as the default charset for `text` content type makes sense.

## Solution

I added `charset=utf-8` as the default charset for `text` content type and used the `mime` crate to avoid repeat and hard code of header values.

### DIFF
--- a/axum/src/extract/form.rs
+++ b/axum/src/extract/form.rs
@@ -60,7 +60,7 @@ where
                 .map_err(FailedToDeserializeQueryString::new::<T, _>)?;
             Ok(Form(value))
         } else {
-            if !has_content_type(req, "application/x-www-form-urlencoded")? {
+            if !has_content_type(req, mime::WWW_FORM_URLENCODED.as_ref())? {
                 return Err(InvalidFormContentType.into());
             }
 
@@ -115,7 +115,7 @@ mod tests {
                 .method(Method::POST)
                 .header(
                     http::header::CONTENT_TYPE,
-                    "application/x-www-form-urlencoded",
+                    mime::WWW_FORM_URLENCODED.as_ref(),
                 )
                 .body(http_body::Full::<bytes::Bytes>::new(
                     serde_urlencoded::to_string(&value).unwrap().into(),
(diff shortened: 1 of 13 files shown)


davidpdrsn (approved):
Thanks! This looks good to me.

@jplatte Do you wanna take a quick look as well?

jplatte (approved):
Looks good to me as well. I was a little worried about the extra dependency at first but I've had a quick skim of its docs & source and it really seems minimal (nothing like mime guessing based on magic bytes or file extensions in there) and is maintained by seanmonstar so doesn't even increase the potential for supply-chain attacks either.

axum/src/extract/form.rs
```diff
@@ -60,7 +60,7 @@ where
                 .map_err(FailedToDeserializeQueryString::new::<T, _>)?;
             Ok(Form(value))
         } else {
-            if !has_content_type(req, "application/x-www-form-urlencoded")? {
+            if !has_content_type(req, mime::WWW_FORM_URLENCODED.as_ref())? {
```
jplatte, line 63: One small thing though, we can probably update this helper function to take `&Mime`, right?

PureWhiteWu: I thought about it for a while, and felt that change this to take `&Mime` is not necessary now, because we still need to call `mime.as_ref()` in `has_content_type`:
```
pub(crate) fn has_content_type<B>(
    req: &RequestParts<B>,
    expected_content_type: &mime::Mime,
) -> Result<bool, HeadersAlreadyExtracted> {
    ...
    Ok(content_type.starts_with(expected_content_type.as_ref()))
}
```

This helper function can take `&str` now, which is much more flexible than can only take `&mime::Mime` which also has no gain.

What do you think?

davidpdrsn:
One more thing I noticed. We should probably mention this in the changelog.

axum/src/extract/form.rs
jplatte: I think it will actually be useful to restrict the function like that because it will enforce (more) consistent usage of the constants from the mime crate.

PureWhiteWu: That's right, I will do that.

PureWhiteWu:
I've fixed the comment and added the changelog

davidpdrsn (approved):

PureWhiteWu:
@davidpdrsn Hi, thanks for your review!
You may need to approve the workflow for this pr~

PureWhiteWu:
@davidpdrsn 
Sorry for the mistake. Should be fine now.
I've tested locally.

Issue #561, opened 2021-11-24
Move `FromRequest` and `IntoResponse` into their own crate?


davidpdrsn:
Most crates that provide integrations with axum only need the `FromRequest` and `IntoResponse` traits. Examples include [maud](https://crates.io/crates/maud), [tower-request-id](https://crates.io/crates/tower-request-id), [tower-cookies](https://crates.io/crates/tower-cookies), [axum-flash](https://crates.io/crates/axum-flash), [and others](https://crates.io/crates/axum/reverse_dependencies).

These traits are somewhat more stable than the rest of axum. They haven't changed since 0.2 back in August, and also wont change in 0.4. So it is inconvenient for libraries since they have to bump their axum version to pull in changes that are essentially irrelevant, such as routing.

Because of this I think we should consider moving `FromRequest` and `IntoResponse` to their own crate alongside implementations for non-local types such as the strings from `http` and `bytes`. The traits would be re-exported from axum in the location they exist today.

I think we should be quite conservative about what goes into this new crate. For example I wouldn't move `Html` and `Json` even though those might be useful to some libraries. Ideally only what _has_ to be moved should be moved.

Basically the same pattern as `futures-core` and `tower-service` and many other crates I'm sure.

I suggest naming this crate `axum-core`.

@jplatte thoughts?

kellytk:
I've two packages that extend axum.

* The first uses `async_trait`, `FromRequest`, and `RequestParts`.
* The second uses `IntoResponse`.

For any project that may be extended, I move user-facing types into a separate `project-types` package.

davidpdrsn:
> I wonder whether there is some clever way to avoid that while still being compatible with hyper without (much) additional boilerplate when depending on the "full" axum crate.

Uh yeah thats worth exploring when we do this. Not depending hyper would mean loosing [`impl FromRequest<Body> for Body`](https://docs.rs/axum/0.3.4/src/axum/extract/request_parts.rs.html#295-301). I'm not sure how much its used in the wild but feels a bit unfortunate to loose.

jplatte:
I think it wouldn't hurt too much to have `axum::extractors::Body` as `struct Body(pub hyper::body::Body);` since full `Body` extraction is uncommon enough. I'm more worried about the type parameter defaults.

davidpdrsn:
I agree with that 👍

davidpdrsn:
I gave it a shot in https://github.com/tokio-rs/axum/pull/564

[...]

PR #564, opened 2021-11-24
Move `FromRequest` and `IntoResponse` into new `axum-core` crate

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/561

What changed:

- Default for `B` type param in `FromRequest` and `RequestParts` is gone 😞 Users will now have to do `impl FromRequest<axum::body::Body> ...` to get the previous behavior. This is annoying I think. Not quite sure if getting rid of the dependency on hyper is worth it. Haven't made up my mind quite yet.
- `BodyAlreadyExtracted`, `HeadersAlreadyExtracted`, and `ExtensionsAlreadyExtracted` now have public constructors (they're unit structs) since they're used in both axum-core and axum.
- `axum::Error` has moved to `axum_core::Error` (still re-exported from the old location) and its `new` constructor is now public since its used in both crates. I'm not very happy about this since users outside of axum shouldn't create this type. It only exists as a wrapper that makes `BoxError` implement `std::error::Error`.
- `JsonRejection` has changed a bit. It now contains `BytesRejection` instead of `BodyAlreadyExtracted` since it calls `Bytes::from_request`.
- Had to copy `hyper::body::to_bytes` into axum-core. Luckily it had no dependencies on hyper itself.
- `hyper::Body` no longer implements `FromRequest`. We previously had [`RawBody`](https://docs.rs/axum/0.3.4/axum/extract/struct.RawBody.html) as the wrapper for it. I think thats a fine.
- [`Headers`](https://docs.rs/axum/0.3.4/axum/response/struct.Headers.html) needs to live in core because of `impl IntoResponse for (StatusCode, Headers)` and friends. Not very happy with this but loosing those impls would hurt ergonomics so think the tradeoff is fine.
- Had to duplicate some macros between the two crates. No biggie imo.

These things are obviously breaking so will have to be released in 0.4

## TODO

- [x] Changelog


jplatte:
I have an idea on how to resolve the type parameter default issue! We can have a `hyper0_14` crate feature that adds the dependency and the default. I'll elaborate when I'm home, currently on the move.

jplatte:
Okay, so to elaborate: The crate feature will be a `default` feature such that people can just do the obvious thing. When a new version of hyper is released, we would still put releasing a new axum-core version on the todo list, **but** people could upgrade hyper independently of axum and kick out the old hyper version from their dependencies entirely by making sure that all of their indirect axum-core dependencies use `default-features = false` and then they get rid of any reliance on the default and add the same setting (I guess this will be threaded through `axum`s features then).

With this scheme, we could even keep the `FromRequest<hyper::body::Body>` impl *and additionally* add more `FromRequest<hyper::body::Body>` impls for different hyper versions under separate feature flags. I.e. if this goes through as axum-core 0.1 and we start off with a `hyper0_14` feature, once 0.15 releases we

* add `hyper0_15` as an off-by-default feature that only adds a trait impl in a non-breaking release
* prepare for axum-core 0.2 which switches both the default Cargo feature and the conditional associated type default to hyper 0.15

tesaguri:
You can do something like `pub type RequestParts<B = crate::body::Body> = axum_core::extract::RequestParts<B>;` in `axum` crate. This cannot be applied to `FromRequest` trait (not without [RFC 1733], at least), though.

[RFC 1733]: <https://rust-lang.github.io/rfcs/1733-trait-alias.html>

jplatte:
> You can do something like `pub type RequestParts<B = crate::body::Body> = axum_core::extract::RequestParts<B>;` in `axum` crate.

Ooohh, this is also an interesting possibility. It doesn't allow to keep the `Body` extraction as-is, but otherwise sounds better than my (somewhat crazy) idea 😄

> This cannot be applied to `FromRequest` trait (not without [RFC 1733](https://rust-lang.github.io/rfcs/1733-trait-alias.html), at least), though.

Actually, it should not be a problem to just have a separate `FromRequest` trait in axum like this:

```rust
trait FromRequest<B = crate::body::Body>: axum_core::FromRequest<B> {}
impl<B, T> FromRequest<B> for T where T: axum_core::FromRequest<B> {}
```

Which effectively works as a trait alias.

davidpdrsn:
> ```rust
> trait FromRequest<B = crate::body::Body>: axum_core::FromRequest<B> {}
> impl<B, T> FromRequest<B> for T where T: axum_core::FromRequest<B> {}
> ```

This would mainly benefit users using `FromRequest` in trait bounds, right? Because to implement the trait you'd still have to implement the version from core, which doesn't have the default. And the only times you're using `FromRequest` in bounds is probably when implementing `FromRequest`, which makes it less useful 🤔

davidpdrsn:
It would be so nice if `hyper::Body` was in its own crate 😞

davidpdrsn:
Also worth noting that there is discussion in hyper about splitting the `Body` type into separate types that each are more focused https://github.com/hyperium/hyper/issues/2345. What that means for axum is hard to say but I guess that would break whatever default type we have.

jplatte:
> This would mainly benefit users using `FromRequest` in trait bounds, right? Because to implement the trait you'd still have to implement the version from core, which doesn't have the default. And the only times you're using `FromRequest` in bounds is probably when implementing `FromRequest`, which makes the alias less useful 🤔

Right, yeah. The alternative would be to copy the core `FromRequest` definition and only have a `Sized` supertrait for both, then adding a blanket impl that delegates. With that setup, you could implement the version that has the default.

FWIW, I think trait aliases also won't allow `impl`s on them.

davidpdrsn:
I'll try an implement the suggestions here and see what it feels like to use them.

davidpdrsn:
I’m leaning towards dropping the default types. It’s less ergonomic but also easier to understand and easy to document. When hyper moves towards 1.0 we can reevaluate.

davidpdrsn:
@jplatte How do you feel about the state of this? As I wrote above I'm fine with it, even though its slightly less ergonomic. Would rather keep things conceptually simple and explore different solutions in the future.

I would also like to get 0.4 out sometime soon-ish together with axum-extra since I need those for something else I'm building 🤓

jplatte:
The ergonomic hit only applies to writing custom extractors now with `IntoResponse` no longer being generic, right? I think that'd be fine. It should already be the best practice to implement it generically anyways.

davidpdrsn:
> The ergonomic hit only applies to writing custom extractors now with `IntoResponse` no longer being generic, right?

Yep that’s correct.

david-perez:
@davidpdrsn Shouldn't the macros in the `axum-core` crate now be depended upon from the `axum` crate, instead of maintaining two copies?

davidpdrsn:
That would require the macros to be public API, which I don’t want.

david-perez:
@davidpdrsn Why the `#[allow(deprecated)]`s ([example](https://github.com/tokio-rs/axum/blob/main/axum-core/src/macros.rs#L13)) in the `macros.rs` file?

davidpdrsn:
For when we have deprecated enum variants in some of the rejections.

Please hop on discord for future questions like this. Otherwise everyone watching the repo keep getting notifications.

david-perez:
> For when we have deprecated enum variants in some of the rejections.

I was confused because `#[allow(deprecated)]` is not in the other `impl` blocks (`Display`, `Error`, `Default`), so I don't understand how it could work if you add a deprecated enum variant.

> Please hop on discord for future questions like this. Otherwise everyone watching the repo keep getting notifications.

Sorry, I don't have Discord. I thought that asking a question about a particular piece of code in the PR where it was last touched was appropiate. If you don't want watchers to get notified about questions, you should probably update the readme, since [it states that opening issues to ask questions is fine](https://github.com/tokio-rs/axum/blob/3841ef44d563021a2bf335173ff72c574914d358/axum/README.md?plain=1#L124), and that also creates notifications for watchers.

davidpdrsn:
> I was confused because #[allow(deprecated)] is not in the other impl blocks (Display, Error, Default), so I don't understand how it could work if you add a deprecated enum variant.

Hm good question 🤔 I don't specifically remember.

> Sorry, I don't have Discord. I thought that asking a question about a particular piece of code in the PR where it was last touched was appropiate. If you don't want watchers to get notified about questions, you should probably update the readme, since it states that opening issues to ask questions is fine, and that also creates notifications for watchers.

If you don't have discord then its no big deal, the questions just feel slightly unrelated to this PR but its alright.

PR #565, opened 2021-11-25
Add `Cached` extractor

davidpdrsn:
`Cached` wraps another extractor and caches its result in request extensions.

### DIFF
--- a/axum-extra/src/extract/cached.rs
+++ b/axum-extra/src/extract/cached.rs
@@ -1,0 +1,232 @@
+use axum::{
+    async_trait,
+    body::{boxed, BoxBody},
+    extract::{
+        rejection::{ExtensionRejection, ExtensionsAlreadyExtracted},
+        Extension, FromRequest, RequestParts,
+    },
+    http::Response,
+    response::IntoResponse,
+};
+use std::{
+    fmt,
+    ops::{Deref, DerefMut},
+};
+
+/// Cache results of other extractors.
+///
+/// `Cached` wraps another extractor and caches its result in [request extensions].
+///
+/// This is useful if you have a tree of extractors that share common sub-extractors that
+/// you only want to run once, perhaps because they're expensive.
+///
+/// The cache purely type based so you can only cache one value of each type. The cache is also
+/// local to the current request and not reused across requests.
+///
+/// # Example
+///
+/// ```rust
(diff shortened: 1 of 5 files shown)


axum-extra/src/extract/cached.rs
```diff
@@ -0,0 +1,232 @@
...
+            Err(_) => {
+                let value = T::from_request(req).await.map_err(CachedRejection::Inner)?;
+
+                req.extensions_mut()
+                    .ok_or_else(|| {
+                        CachedRejection::ExtensionsAlreadyExtracted(
+                            ExtensionsAlreadyExtracted::default(),
+                        )
+                    })?
+                    .insert(value.clone());
```
jplatte, line 119: How about using `CacheEntry<T>` (where `CacheEntry` is just a private newtype) instead of `T` here so you can have both a `Cached<T>` and a "regular" request extension `T` on the same request?

This would also stop accidentally relying on the cache having been propagated earlier in request processing by using `Extension<T>` over `Cached<T>` somewhere.

davidpdrsn: Good idea!

axum-extra/src/extract/cached.rs
```diff
@@ -0,0 +1,232 @@
...
+/// Rejection used for [`Cached`].
+///
+/// Contains one variant for each way the [`Cached`] extractor can fail.
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum CachedRejection<T> {
+    #[allow(missing_docs)]
+    ExtensionsAlreadyExtracted(ExtensionsAlreadyExtracted),
+    #[allow(missing_docs)]
+    Inner(T),
```
jplatte, line 150: I was confused there for a bit thinking the `T` on `CachedRejection` was the same thing as the `T` on `Cached`. Maybe the type definition would be clearer if it was, and this variant was `Inner(T::Rejection)`?

davidpdrsn: Using the type of the extractor complicates the trait bounds for `IntoResponse for CachedRejection` and the others. I've changed the type param from `T` to `R` which is maybe a bit more clear.

jplatte: Yeah, I think this helps a lot actually 🙂

jplatte (approved):

PR #569, opened 2021-11-26
Avoid double-boxing bodies where possible

KaiJewson:
## Motivation

It can be inefficient if bodies are accidentally boxed twice - for example a handler might return a `Response<BoxBody>`, which `MethodRouter` would then box again.

## Solution

Ensure that bodies don't get boxed twice, by checking if they are already boxed and if so not doing anything.

### DIFF
--- a/axum/src/body/mod.rs
+++ b/axum/src/body/mod.rs
@@ -1,7 +1,6 @@
 //! HTTP body utilities.
 
-use crate::{BoxError, Error};
-use std::any::Any;
+use crate::{util::try_downcast, BoxError, Error};
 
 mod stream_body;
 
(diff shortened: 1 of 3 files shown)


KaiJewson:
Hmm, with this change might it be possible to modify `IntoResponse` to always return `BoxBody`? Since it will probably be boxed anyway somewhere down the line there would only be a performance difference in a tiny handful of cases. Doing so would drastically simplify lots of generic code I'm currently writing, so I think it would be quite nice. The original motivation for making `IntoResponse` have associated types was back when it returned the less general `hyper::Body` type, but now `BoxBody` exists that motivation isn't present anymore.

axum/src/body/mod.rs
```diff
@@ -22,12 +23,31 @@ pub use bytes::Bytes;
...
+}
+
+fn downcast_box_body<B: 'static>(body: B) -> Result<BoxBody, B> {
+    let mut body = Some(body);
+    if let Some(body) = <dyn Any>::downcast_mut::<Option<BoxBody>>(&mut body) {
+        Ok(body.take().unwrap())
+    } else {
+        Err(body.unwrap())
+    }
+}
```
davidpdrsn, line 43: We already have [this](https://github.com/tokio-rs/axum/blob/main/axum/src/routing/mod.rs#L660-L672) which does the same thing. Maybe we could move that to the root module and re-use it?

axum/src/body/mod.rs
```diff
@@ -22,12 +23,31 @@ pub use bytes::Bytes;
 pub type BoxBody = http_body::combinators::UnsyncBoxBody<Bytes, Error>;
 
 /// Convert a [`http_body::Body`] into a [`BoxBody`].
+///
+/// If the given body is already a [`BoxBody`], this function will not box the body again.
```
davidpdrsn, line 27: Hm this feels like an implementation detail thats probably not worth documenting 🤔

davidpdrsn (changes requested):

davidpdrsn:
Nice!

> Since it will probably be boxed anyway somewhere down the line there would only be a performance difference in a tiny handful of cases. Doing so would drastically simplify lots of generic code I'm currently writing, so I think it would be quite nice.

Wanna elaborate on this? I personally think being able to return any body type is quite ergonomic.

KaiJewson:
> Wanna elaborate on this? I personally think being able to return any body type is quite ergonomic. Its also one less than we'd have to explain to users.

My idea was to change the definition of `IntoResponse` to this:

```rs
pub trait IntoResponse: Sized {
    fn into_response(self) -> Response<BoxBody>;
}
```

The types `IntoResponse` would be implemented for wouldn't change at all, and we also wouldn't lose any flexibility because the trait bounds required to implement it would be the exact same. Its main benefit is allowing this code to just compile:

```rs
async fn index() -> impl IntoResponse {
    if condition() {
        "hello world".into_response()
    } else {
        ().into_response()
    }
}
```

It currently doesn't because the body types are different and I find that annoying - I can always box myself, but if Axum did that automatically it would be easier, and with this PR not any less fast.

davidpdrsn:
Ah yes I see. Not quite sure how I feel about it but definitely worth exploring for 0.4. Do wanna file an issue about that?

davidpdrsn (approved):

Issue #559, opened 2021-11-24
Expose PathDeserializerError in InvalidPathParam with additional data points


pnehrer:
### Motivation

As discussed in https://github.com/tokio-rs/axum/discussions/556, it would be great to be able to determine which path parameter (by name) failed to parse during extraction from request.

This would make it possible to customize the error response in an application-specific way. E.g., if path parameter `userId` failed to parse:

```
400 Bad Request
Content-Type: application/json; charset=utf-8
Connection: close

{
    "error": {
        "message": "Invalid parameter value.",
        "locationType": "path",
        "location": "userId"
    }
}
```

### Proposal

Perhaps instead of passing `impl Into<String>` when creating InvalidPathParam we could pass in the actual error type, PathDeserializerError (which would have to be public and exported).

Furthermore, we could add the name of the failed parameter to PathDeserializerError and expose it via a method, much like `serde_json::Error` exposes line(), column(), etc.

### Alternatives

Alternatively, we could just add the failing parameter name directly to PathParamsRejection, in addition to the error description when the deserializer returns an error. This would still be a breaking change since we'd have to add another tuple value or change PathParamsRejetction type to a struct.

PR #574, opened 2021-11-28
Provide more data in `Path` deserialization error

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/559

This changes the error returned by `Path` if deserialization fails, to provide more data about the exact error. This allows users to convert into their own errors. It also makes the default error messages more precise.

Also added a new example showing how to customize the error https://github.com/tokio-rs/axum/blob/path-error/examples/customize-path-rejection/src/main.rs. Its the exact same approach as [this](https://github.com/tokio-rs/axum/blob/main/examples/customize-extractor-error/src/main.rs) but can't hurt having an example for it.

# TODO

- [x] Write more tests
- [x] Add missing changelog item


davidpdrsn:
@pnehrer what do you think about this solution?

axum/CHANGELOG.md
```diff
@@ -37,13 +37,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **breaking:** The `Body` and `BodyError` associated types on the
   `IntoResponse` trait have been removed - instead, `.into_response()` will now
   always return `Response<BoxBody>` ([#571])
+- **breaking:** The variants of `Path`'s rejection type has been renamed ([#574])
```
jplatte, line 40:
```suggestion
- **breaking:** The variants of `Path`'s rejection type have been renamed ([#574])
```

jplatte, line 40:
```suggestion
- **breaking:** The variants of `Path`'s rejection type have been renamed ([#574])
```

axum/src/extract/path/mod.rs
```diff
@@ -121,8 +127,14 @@ use std::{
 /// # };
 /// ```
 ///
+/// # Providing detailed rejection output
+///
+/// If the URI cannot be deserialized into the target type the request will be rejected and an
+/// error response will be returned. See [`customize-path-rejection`] for an exapmle of how to customize that error.
```
jplatte, line 133:
```suggestion
/// error response will be returned. See [`customize-path-rejection`] for an example of how to customize that error.
```

axum/src/extract/path/mod.rs
```diff
@@ -157,20 +169,236 @@ where
...
+
+impl fmt::Display for PathDeserializationError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.kind.fmt(f)
+    }
+}
+
+impl std::error::Error for PathDeserializationError {}
+
+/// The kinds of errors that can happen we deserializing into a [`Path`].
```
jplatte, line 255:
```suggestion
/// The kinds of errors that can happen when deserializing into a [`Path`].
```

pnehrer:
This looks great! Thank you so much for doing this 👍

jplatte:
This is a little too big for me to fully review. Is there such a thing as a rustdoc diff? 😄

From my glancing over the changes before I found the internal `struct WrongNumberOfParameters<G>` API a little bit weird (would probably have used a struct literal instead of this kind of type-checked builder pattern myself) but since it's internal I don't think it's sth. to worry about.

The general idea of this seems good, the only other thing I can say is I think you missed documenting the rejection type rename in the changelog.

davidpdrsn:
I've been thinking that we should have more tests for this since generating the right errors is an important part of the API. So I'm gonna write some tests for the stuff in this PR. That might also make things slightly easier to review because you can more easily see the behavior of things without having to internalize how all the code works.

> From my glancing over the changes before I found the internal `struct WrongNumberOfParameters<G>` API a little bit weird (would probably have used a struct literal instead of this kind of type-checked builder pattern myself) but since it's internal I don't think it's sth. to worry about.

Hehe yeah for sure its over engineered but also kinda nice I think. But yeah as you say its a small internal API that we can always change later if we want.

> The general idea of this seems good, the only other thing I can say is I think you missed documenting the rejection type rename in the changelog.

Good catch! I'll fix that later.

jplatte (approved):

PR #578, opened 2021-11-30
Use `Self` in `RouterFuture`'s `impl` block

david-perez:
For consistency with the other method and the rest of the codebase.


davidpdrsn (approved):
Not sure this really matters but why not 🤷

PR #579, opened 2021-11-30
Reduce `NotFound`'s visibility

david-perez:
It is only used in the crate's `routing` module.


davidpdrsn (approved):
Not sure this really matters but why not 🤷

PR #588, opened 2021-12-04
Fix `debug_handler` for `Result<impl IntoResponse, Error>` and friends

### LINKED ISSUE #587: `#[debug_handler]` doesn't work with `Result<impl IntoResponse, SomeError>`
This produces code that doesn't compile

```rust
#[debug_handler]
async fn handler() -> Result<impl IntoResponse, Infallible> {
    Ok(())
}
```

Error is:

```
error[E0720]: cannot resolve opaque type
  --> examples/hello-world/src/main.rs:26:30
   |
26 | async fn handler() -> Result<impl IntoResponse, Infallible> {
   |                              ^^^^^^^^^^^^^^^^^ cannot resolve opaque type
```

It expands into

```rust
async fn handler() -> Result<impl IntoResponse, Infallible> {
    Ok(())
}

#[allow(warnings)]
async fn __axum_debug_check_handler_into_response() {
    #[allow(warnings)]
    async fn __axum_debug_check_handler_into_response_make_value(
    ) -> Result<impl IntoResponse, Infallible> {
        //      ^^^^^^^^^^^^^^^^^ cannot infer type of this from `panic!()`
        panic!()
    }
    let value = __axum_debug_check_handler_into_response_make_value().await;
    fn check<T>(_: T)
    where
        T: ::axum::response::IntoResponse,
    {
    }
    check(value);
}

#[allow(warnings)]
fn __axum_debug_check_handler_future() {
    async fn handler() -> Result<impl IntoResponse, Infallible> {
        Ok(())
    }
    let future = handler();
    fn check<T>(_: T)
    where
        T: ::std::future::Future + Send,
    {
    }
    check(future);
}
```

Should look into supporting that together with `Result<SomeResponse, impl IntoResponse>` and `Result<impl IntoResponse, impl IntoResponse>`.

davidpdrsn:
Fixes #587

### DIFF
--- a/axum-debug/src/lib.rs
+++ b/axum-debug/src/lib.rs
@@ -251,6 +251,24 @@ mod debug_handler {
         };
         let span = ty.span();
 
+        let declare_inputs = item_fn
+            .sig
+            .inputs
+            .iter()
+            .filter_map(|arg| match arg {
+                FnArg::Receiver(_) => None,
+                FnArg::Typed(pat_ty) => {
+                    let pat = &pat_ty.pat;
+                    let ty = &pat_ty.ty;
+                    Some(quote! {
+                        let #pat: #ty = panic!();
+                    })
+                }
+            })
+            .collect::<TokenStream>();
+
+        let block = &item_fn.block;
+
         let make_value_name = format_ident!(
             "__axum_debug_check_{}_into_response_make_value",
             item_fn.sig.ident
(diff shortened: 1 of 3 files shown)


davidpdrsn:
cc @programatik29

axum-debug/src/lib.rs
```diff
@@ -251,6 +251,24 @@ mod debug_handler {
...
+            .sig
+            .inputs
+            .iter()
+            .filter_map(|arg| match arg {
+                FnArg::Receiver(_) => None,
+                FnArg::Typed(pat_ty) => {
+                    let pat = &pat_ty.pat;
+                    let ty = &pat_ty.ty;
+                    Some(quote! {
+                        let #pat: #ty = panic!();
```
jplatte, line 264: I briefly thought that this could cause the same issue in a way but I don't think it can, since handler functions can't practically be generic (right?)

jplatte (approved):

axum-debug/src/lib.rs
davidpdrsn: Handlers can be generic but `debug_handler` doesn’t support that and panics. There might be ways to make it work but it’s a huge can of worms.

jplatte: Oh I guess it works with an explicit generic parameter and registering it under a route as `handler::<Foo, Bar>`. I was thinking of `impl Trait` in argument position.

PR #589, opened 2021-12-04
Add `OrDefault` extractor to `axum-extra`

davidpdrsn:
I've found this to be useful, especially for query params such as
pagination. So I think it makes sense to include in `axum-extra`.

### DIFF
--- a/axum-extra/src/extract/or_default.rs
+++ b/axum-extra/src/extract/or_default.rs
@@ -1,0 +1,175 @@
+use axum::{
+    async_trait,
+    extract::{FromRequest, RequestParts},
+};
+use std::{
+    convert::Infallible,
+    ops::{Deref, DerefMut},
+};
+
+/// Use an extractor's default value if it fails.
+///
+/// `OrDefault` wraps another extractors, that implements [`Default`], and uses its default value
+/// if the inner extractor fails.
+///
+/// # Example
+///
+/// Serializing query params for pagination and using default values for missing params:
+///
+/// ```rust
+/// use axum_extra::extract::OrDefault;
+/// use axum::extract::Query;
+/// use serde::Deserialize;
+///
+/// #[derive(Default, Deserialize)]
+/// struct Pagination {
+///     #[serde(default)]
+///     page: Page,
+///     #[serde(default)]
+///     per_page: PerPage,
+/// }
+///
+/// #[derive(Deserialize, Copy, Clone)]
+/// #[serde(transparent)]
+/// struct Page(usize);
+///
+/// impl Default for Page {
+///     fn default() -> Self {
+///         Self(1)
+///     }
+/// }
+///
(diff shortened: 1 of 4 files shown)


axum-extra/src/extract/or_default.rs
```diff
@@ -0,0 +1,175 @@
...
+    extract::{FromRequest, RequestParts},
+};
+use std::{
+    convert::Infallible,
+    ops::{Deref, DerefMut},
+};
+
+/// Use an extractor's default value if it fails.
+///
+/// `OrDefault` wraps another extractors, that implements [`Default`], and uses its default value
```
jplatte, line 12:
```suggestion
/// `OrDefault` wraps another extractors that implements [`Default`], and uses its default value
```

axum-extra/src/extract/or_default.rs
```diff
@@ -0,0 +1,175 @@
...
+/// struct PerPage(usize);
+///
+/// impl Default for PerPage {
+///     fn default() -> Self {
+///         Self(30)
+///     }
+/// }
+///
+/// async fn handler(
+///     pagination: OrDefault<Query<Pagination>>,
```
jplatte, line 53: Why is `OrDefault` even needed in this case? If there is not `?query_params`, does `Query` fail rather than passing an empty string to `serde_urlencoded`? I think the latter is what actix-web does and should make this example work without `OrDefault`.

jplatte (approved):
I don't really get the query example (see second review thread) but I think this could be useful for other reasons 👍🏼

axum-extra/src/extract/or_default.rs
davidpdrsn: Ah yeah thats a good point. If there is no query string it gives an empty string to `serde_urlencoded` so that would work without `OrDefault`. If some params are present but invalid `OrDefault` would catch the error and return the default instead. I'll update the example to mention that.

davidpdrsn: Hm maybe pagination isn't the best example 🤔

axum-extra/src/extract/or_default.rs
```diff
@@ -14,7 +14,7 @@ use std::{
 ///
 /// # Example
 ///
-/// Serializing query params for pagination and using default values for missing params:
+/// Serializing query params for pagination and using default values for missing or invalid params:
```
jplatte, line 17: This sounds like individual invalid params would have their default applied if deserialization fails, but they will all be defaulted if a single one if invalid. Which I think it another reason to choose a different example entirely.

davidpdrsn: Yeah I agree, this example isn't good. Do you have an idea? Otherwise I'll try and think of something. I agree `OrDefault` still feels useful :P

PR #591, opened 2021-12-06
Release axum, axum-core, and axum-debug

davidpdrsn:


jplatte:
The `axum` release didn't actually do anything, did it?

davidpdrsn:
It did. The response type is also re-exported from axum https://docs.rs/axum/0.4.1/axum/response/type.Response.html

jplatte:
Right. Then you should have bumped the `axum-core` dependency before releasing, because right now it's possible to upgrade axum to 0.4.1 without upgrading axum-core to 0.1.1 and have axum be broken by that 😉

davidpdrsn:
Ah right. I’ll fix that later today with another release of axum that depends on core 0.1.1. So many crates 😅

PR #594, opened 2021-12-06
Refactor `sse::Event`

KaiJewson:
SSE `Events` are currently somewhat costly to use: each field is a separate allocation and at the end there's two more allocations as it's converted to a `String` then `Bytes`. By instead having the `sse::Event` build a `BytesMut` as it goes along, there will be far fewer allocations overall.

- I added dependencies on `memchr` and `itoa` for the implementation; I judged it to be fine because they're already in the required dependency tree, even several times.
- The SSE event's fields will now be in the order of the methods called on `Event`, instead of a fixed one. This is allowed by the SSE spec AFAICT and it makes more sense.
- To avoid nonsense things like an SSE event with multiple `id` fields, it also keeps track of which fields have been filled in and shouldn't be filled in again via a bitflags.
- I noticed that Axum didn't check that values passed in to the SSE event methods were valid, so I added validation for them (e.g. `data` cannot contain carriage returns, `id` cannot contain null bytes).
- There was a bug that setting `data`/`event`/`id` to a value that began with a space would lead to that space being dropped, because Axum would generate `data: text` from the input `" text"`, which the browser interprets as just `text`. I fixed that and added a test for it.
- I removed the `Display` impl on `Event` since it's no longer used. But I could easily add it back.

Unfortunately these changes are breaking, although they probably won't break any actual code. And it doesn't touch `axum-core` so libraries won't break.

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -7,7 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- **breaking:** `sse::Event` is more strict about what field values it supports, disallowing any
+  SSE events that break the specification (such as field values containing carriage returns).
+- **breaking:** `sse::Event` now accepts types implementing `AsRef<str>` instead of `Into<String>`
+  as field values.
+- **fixed:** `sse::Event` will no longer drop the leading space of data values that have it.
 
 # 0.4.2 (06. December, 2021)
 
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -27,9 +27,11 @@ bytes = "1.0"
 futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
 http = "0.2.5"
 http-body = "0.4.4"
-mime = "0.3.16"
 hyper = { version = "0.14.14", features = ["server", "tcp", "stream"] }
+itoa = "0.4.8"
 matchit = "0.4.4"
+memchr = "2.4.1"
+mime = "0.3.16"
 percent-encoding = "2.1"
 pin-project-lite = "0.2.7"
 serde = "1.0"
--- a/axum/src/response/sse.rs
+++ b/axum/src/response/sse.rs
@@ -32,7 +32,7 @@ use crate::{
     response::{IntoResponse, Response},
     BoxError,
 };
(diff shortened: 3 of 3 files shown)


davidpdrsn:
Thanks for working on this! Probably wont have time to review it until next week.

From reading the description my initial thoughts are if there is a way to get some of these benefits without breaking changes and then perhaps pushing the breaking changes to the next major release. I would like to wait a bit with shipping another major since 0.4 shipped recently.

KaiJewson:
It depends what you consider to be breaking. There are five aspects of potential breakage, in order of least to most breaking:

1. Leading spaces won't be dropped. This is the most obviously a fix rather than a change, but it still technically changes behaviour.
1. Fields are now placed in order of calls on `Event` rather than a fixed order. There was no explicit guarantee of stability made either way beforehand.
1. Extra validation is performed on inputs. This can be considered breaking in that it can cause previously OK code to panic, however such code was already breaking the SSE spec (so likely if it was tested it would've errored anyway) and it could be classified as a fix rather than a breaking change.
1. Code will panic if methods like `data`/`id` are called multiple times where previously it would silently override the old value. This is unavoidable with the new design, as there's no easy way to go back and edit the contents of the `BytesMut`.
1. `impl AsRef<str>` bounds replace `impl Into<String>` bounds. I can easily go back on this, at the cost of allocations, which 0.5 can later fix.

So yeah, this change could be made with zero compile time breakage easily, the only question is whether you're OK with the potential runtime breakage.

davidpdrsn:
- 1 and 2: I wouldn't consider this a breaking change since its so minor.
- 3: This is a bug fix to me and not a breaking change.
- 4 and 5: I agree are breaking changes that should wait until 0.5

Would you be up for changing this (or making a new PR) that only does 1, 2, and 3 and then making another PR with 4 and 5 that we can include in 0.5?

What do you think @jplatte?

jplatte:
I hardly know anything about SSE so not sure whether 1 would qualify as a bugfix or not, but either way I agree that 1-3 should be fine for a non-breaking release while 4+5 should wait.

KaiJewson:
Okay, now everything else has been merged I will close this. Later I'll re-make the PR with the `BytesMut` part alone, once another breaking release is on the table.

davidpdrsn:
Thanks for all your work @KaiJewson 😃

PR #595, opened 2021-12-07
Support customizing the request body type in `#[debug_handler]`

davidpdrsn:
I ran into

```rust
#[debug_handler]
async fn handle(req: Request<BoxBody>) {}
```

claiming that `Request<BoxBody>` didn't implement `FromRequest`, [which it does.](https://github.com/tokio-rs/axum/blob/main/axum-core/src/extract/request_parts.rs#L9) Turns out thats because `#[debug_handler]` specifically requires `FromRequest<Body>` which doesn't work for `Request<BoxBody>`.

So this adds an option to configure the request body type:

```rust
#[debug_handler(body = BoxBody)]
async fn handle(req: Request<BoxBody>) {}
```


axum-debug/CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- Support checking `FromRequest` bounds for extractors who's request body is something else than
```
jplatte, line 10: Not 100% sure, but I think this should be

```suggestion
- Support checking `FromRequest` bounds for extractors whose request body is something else than
```

axum-debug/src/lib.rs
```diff
@@ -170,11 +189,31 @@ mod debug_handler {
...
+            while !input.is_empty() {
+                let ident = input.parse::<syn::Ident>()?;
+                if ident == "body" {
+                    input.parse::<Token![=]>()?;
+                    body_ty = Some(input.parse()?);
+                } else {
+                    return Err(syn::Error::new_spanned(ident, "unknown argument"));
+                }
+
+                let _ = input.parse::<Token![,]>();
```
jplatte, line 209: I think the idiomatic way to do this is [Punctuated::parse_terminated_with](https://docs.rs/syn/latest/syn/punctuated/struct.Punctuated.html#method.parse_terminated_with)

davidpdrsn: Hm how would that work 🤔? I'm not parsing multiple of the same thing. Specifically just parsing `body = Type` with optional trailing comma.

jplatte: The way I usually do this kind of thing is

```rust
enum MyAttribute {
    Body(syn::Type),
    // other variants may be added in the future
}

impl Parse for MyAttribute { /* ... */ }

let attributes = input.call(Punctuated::<MyAttribute, Token![,]>::parse_terminated)?;

let mut body = None;
for attr in attributes {
    match attr {
        // Or with a function that raises a duplicate-attribute
        // `syn::Error` if the `Option` is already `Some`
        MyAttribute::Body(b) => body = Some(b),
    }
}
```

davidpdrsn: How do you handle having multiple variants in `MyAttribute` then, given that you end up with a vec of enums and would have to search through it to find a `MyAttribute::Body` to get the value of that attribute. Maybe I'm misunderstanding you but feels a struct with fields is easier to access.

jplatte: That's what the for loop is for, same as the while loop in your code. It's just that it's mostly decoupled from the parsing. The `body` var in my example above was intended to mirror `body_ty` in the code.

It's probably more code than what you wrote actually, but more declarative, less imperative.

davidpdrsn: Ah okay yeah I get it now. I think I'll just leave it as it is. If it gets annoying if/when we add more attributes we can always reconsider.

axum-debug/CHANGELOG.md
jplatte: Looks like you accidentally changed this back.

jplatte (approved):

PR #602, opened 2021-12-10
Expose TypedHeaderRejection fields

CGreenburg:
## Motivation

Similar to the `customize-extractor-error` and `customize-path-rejection`, I'm trying to customize the response based on the rejection. I'm unable to do this without these fields being public. 

## Solution

The solution is to make these fields public so that I can match on `rejection.reason`


davidpdrsn:
[There are already accessors for the fields](https://docs.rs/axum/0.4.2/axum/extract/rejection/struct.TypedHeaderRejection.html#method.name). Do those work for your use case?

CGreenburg:
Oh boy. I totally missed that. 😅 Thanks.

Issue #596, opened 2021-12-08
Library not compiling for armv5te because of missing AtomicU64


DrSloth:
### Version
This error happened on version 0.3.3.

### Platform
The build was done on Debian 10 and Debian 11 and the error is expected and known (Rocket had the same error).
The build was for the `armv5te-unknown-linux-musleabi` which is not a high tier target but it is what the embedded 
Linux device (Kernel 2.5) i am using has. 

### Crates
As this is a compiler error every dependent of axum is affected

### Description
Trying to compile the crate results in immediate error because the RouteId uses AtomicU64 which some platforms do not support.
I solved this temporarily by creating a fork which uses AtomicU32 which seems like a reasonable solution. Using a Mutex
would also be a possible solution. I suggest adding a cfg query to use AtomicU32 on 32 bit platforms, i could adapt my changes for the newest version and create a pull request. This would be nice as Axum is perfect for older embedded platforms as it is very fast and its memory footprint isn't too large. 

Thank you and keep up the nice work.

PR #616, opened 2021-12-12
Use `AtomicU32` in `RouteId`

KaiJewson:
I don't think anyone is going to be creating >`u32::MAX` routes anyway so it should be fine, but I added a panic branch just in case, so silent bugs are never introduced. Also, there's no need to use `SeqCst`, it's not like there are any data dependencies associated with that atomic so `Relaxed` will do.

Fixes: #596.


davidpdrsn (approved):

davidpdrsn:
Thanks!

PR #621, opened 2021-12-12
Example for defining routes and handlers close together

davidpdrsn:
Some users have asked if its possible to have routes and handlers close together in the code, so one can easily see what the path is for a given handler. Often this is followed by asking for route macros like `#[get("/")]` which addresses this.

I think a decent, macro free, solution is making functions like this

```rust
fn root() -> Router {
    async fn handler() -> &'static str {
        "Hello, World!"
    }

    route("/", get(handler))
}
```

and then using `Router::merge` to combine things.

I think it makes sense to have an example for this so we have something to link to.

### DIFF
--- a/examples/routes-and-handlers-close-together/Cargo.toml
+++ b/examples/routes-and-handlers-close-together/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "example-routes-and-handlers-close-together"
+version = "0.1.0"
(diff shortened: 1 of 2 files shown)


jplatte (approved):
I'm not a fan of this style but even if it's not what people will write, maybe it's useful as an example of what a routing proc-macro could generate.

PR #633, opened 2021-12-16
pub use mime

lz1998:
## Motivation

`pub use mime`, then developers does not need to import `mime` in Cargo.toml.

They can use `use axum::mime`, it's more convenient.

## Solution

`pub use mime`

### DIFF
--- a/axum/src/lib.rs
+++ b/axum/src/lib.rs
@@ -409,6 +409,8 @@ pub use headers;
 pub use http;
 #[doc(no_inline)]
 pub use hyper::Server;
+#[doc(no_inline)]
+pub use mime;
 
 #[doc(inline)]
 #[cfg(feature = "json")]


davidpdrsn:
What do you need mime for that’s directly related to axum?

Re-exporting all dependencies is a slippery slope that I want to be careful with, unless there is good reason to.

davidpdrsn:
Technically using two versions of mine is fine. So it’s not the same as https://github.com/tokio-rs/axum/pull/630

lz1998:
> What do you need mime for that’s directly related to axum?
> 
> Re-exporting all dependencies is a slippery slope that I want to be careful with, unless there is good reason to.

I am writing my own `pub struct Json<T>(pub T);` which can automatically convert i64 to string, and parse i64 from string, because number in JavaScript only supports int53.

I copied [json.rs](https://github.com/tokio-rs/axum/blob/main/axum/src/json.rs) which depends on crate `mime`, and I found crate `mime` is missing.

SabrinaJewson:
FWIW Axum does already have `mime` in its public API - [`Field::content_type`](https://docs.rs/axum/latest/axum/extract/multipart/struct.Field.html#method.content_type) returns an `Option<&Mime>`. So technically this does not expose any additional surface not already exposed by Axum.

> I am writing my own `pub struct Json<T>(pub T);` which can automatically convert i64 to string, and parse i64 from string, because number in JavaScript only supports int53.

This sounds like a use for a custom `Serialize`/`Deserialize` implementation rather than a custom `Json` type.

jplatte:
I have a crate that might help you with your use case without having to create your own Json type: [js_int](https://lib.rs/crates/js_int)

lz1998:
> I have a crate that might help you with your use case without having to create your own Json type: [js_int](https://lib.rs/crates/js_int) 

I use an 'IdGenerator' based on timestamp, which will generate a i64 id, and I hope I can convert it automatically to string.

This issue helps me.
https://github.com/serde-rs/json/issues/833

Now I am going to write a struct similar to 'Json' which implements 'FromRequest' and 'IntoResponse'.

jplatte:
Oh, yeah I guess if it's larger than i53 and you're set on representing it as an i64 on the Rust side, that might be a reasonable solution. Still seems very unusual to me.

Back to this PR: I think mime should not be a public dependency, would it hurt much to instead make it an internal one by returning a string from that one method?

lz1998:
> Oh, yeah I guess if it's larger than i53 and you're set on representing it as an i64 on the Rust side, that might be a reasonable solution. Still seems very unusual to me.
> 
> Back to this PR: I think mime should not be a public dependency, would it hurt much to instead make it an internal one by returning a string from that one method? 

It can be an internal one, and I can import it in Cargo.toml by myself.

But I think json.rs is an official example, and it shows how to parse data from request. People can copy json.rs and run without other dependencies.

lz1998:
I found that mime is not maintained.

The last commit is 3 years ago, and there are many issues and PRs.

Perhaps it should be cloned to axum, then it will be easy to modify.

I don't know what's the best solution. Just an idea.

SabrinaJewson:
> Now I am going to write a struct similar to 'Json' which implements 'FromRequest' and 'IntoResponse'.

I think that should ideally be avoided. A better option would be to either use [`Config::field_attribute`](https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.field_attribute) to give every `i64` field an attribute `#[serde(with = "i64_as_string")]` (and define the `i64_as_string` module [like so](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4dae8be5f345bddcac6c9f9acace84d2)), or to map every JSON type you send and receive to the equivalent type, but with the `#[serde(with = "i64_as_string")]` attribute.

davidpdrsn:
> But I think json.rs is an official example, and it shows how to parse data from request. People can copy json.rs and run without other dependencies.

Just to be clear: json.rs is not an “example”. Examples are in the examples folder :) You’re of course free to copy/paste the code, but I don’t think we should specifically optimize the for that use case.

> Back to this PR: I think mime should not be a public dependency, would it hurt much to instead make it an internal one by returning a string from that one method?

Might be worth considering. I decided not to for the initial implementation since mine hadn’t been updated in a long time. Figured it was quite low risk. I agree it would be nice to change regardless.

PR #634, opened 2021-12-17
Using tokio RwLock instead of std in todo example

damienpontifex:
- Within async context, preferred to use tokio RwLock to not block the
  runtime if waiting for the lock

## Motivation

Maintain async best practices in examples by using types that understand async
context within the tokio runtime. It is my understanding that using a std RwLock
within tokio async context can lead to poor performance in higher throughput
scenarios. While I understand this is an example, I thought it nice to maintain that
best practice to support the code here

## Solution

Swap `std::sync::RwLock` for `tokio::sync::RwLock` and it's usage in todo example

### DIFF
--- a/examples/todos/src/main.rs
+++ b/examples/todos/src/main.rs
@@ -25,9 +25,10 @@ use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
     net::SocketAddr,
-    sync::{Arc, RwLock},
+    sync::Arc,
     time::Duration,
 };
+use tokio::sync::RwLock;
 use tower::{BoxError, ServiceBuilder};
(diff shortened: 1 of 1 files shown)


davidpdrsn:
I would argue that using a sync lock is actually correct in this case since the lock is not held across `await`s, the critical section is short, and the value held by the lock is "just data" and not IO resources.

More details [here](https://docs.rs/tokio/1.15.0/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use) and [here](https://ryhl.io/blog/async-what-is-blocking/).

programatik29:
Usage of `std::sync::RwLock` seems correct.

davidpdrsn:
Think I’ll close this but feel free to ask if you have more questions 😊

PR #638, opened 2021-12-18
refactor: refine session cookie example

ttys3:
## Motivation

Current session example use the cookie method to send the session ID.

but the cookie it send is not valid, the whole cookie looks like this:

```
wa4TKv3HOc9muZR1/HHgOOq/LGnKzSKYvhzlF4iI6cNXkqp5Luhx1S9BOUQkv1nupwDTDNTwUJny61SjPGIRXA==
```

most client will parse this into `key=value` format like this:

```
key: wa4TKv3HOc9muZR1/HHgOOq/LGnKzSKYvhzlF4iI6cNXkqp5Luhx1S9BOUQkv1nupwDTDNTwUJny61SjPGIRXA=

value: =
```

this is not the expected behavior

according to RFC examples: https://datatracker.ietf.org/doc/html/rfc6265#section-3.1
 
a simple set cookie header to send the cookie with name `SID` and value with `31d4d96e407aad42` like this:
```
   == Server -> User Agent ==

   Set-Cookie: SID=31d4d96e407aad42

   == User Agent -> Server ==

   Cookie: SID=31d4d96e407aad42
```

and rfc6265 5.2.  The Set-Cookie Header 
 https://datatracker.ietf.org/doc/html/rfc6265#section-5.2

>  4.1.  Set-Cookie
>   The Set-Cookie HTTP response header is used to send cookies from the
>   server to the user agent.

4.1.1.  Syntax  https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1

>   Informally, the Set-Cookie response header contains the header name
>   "Set-Cookie" followed by a ":" and a cookie.  Each cookie begins with
>   a name-value-pair, followed by zero or more attribute-value pairs.

So there are several problems with the current implementation:

1. the `Set-Cookie` send the cookie header in an  invalid format
2. the cookie extract from client assume that client only send one cookie, this is not the situation in most case.

the actual cookie send from client may like this:

```
CSRF-Token-YNVMA=mnmXXXXXXXXXXXXXX; phpMyAdmin=e406a99bfefcc27xxxxxxxxxxxxxxx; SID=UX7bR/cxxxxxxxxxxx; io=93-xxxxxxx; mw_installer_session=6b29879d9xxxxxxxxxxxxxxxx; cockpit=dj0yO2s9Nxxxxxxxxxxxxx==; _flash={"_":{"kind":"success","message":"Post succcessfully added"}}; axum_session=wa4TKv3HOc9muZR1/HHgOOq/LGnKzSKYvhzlF4iI6cNXkqp5Luhx1S9BOUQkv1nupwDTDNTwUJny61SjPGIRXA==
```

we need to parse the cookie header and extract the cookie name we needed, in here it is `axum_session`

## Solution

1. **cookie send from server to client**: change the set cookie header to follow `key=value`  **name-value-pair** format

2. **cookie receive from client**: parse cookie from client, use a fixed cookie name (like `axum_session` or `axum_sid`) to extract the session cookie back to ourself (as a lib it should allow user to change it, but here it is just working demo, so I think hardcoded name is OK)

(exists example: https://github.com/actix/actix-extras/blob/6676a509443306cd589e10e16d1617154c40ee23/actix-session/src/cookie.rs#L60  actix use  `actix-session` as the `default` cookie name, but it also allow caller to set it to a different one)

### DIFF
--- a/examples/sessions/src/main.rs
+++ b/examples/sessions/src/main.rs
@@ -91,10 +91,13 @@ where
             .await
             .expect("`MemoryStore` extension missing");
 
-        let session_cookie = Option::<TypedHeader<Cookie>>::from_request(req)
+        let cookie = Option::<TypedHeader<Cookie>>::from_request(req)
             .await
-            .unwrap()
-            .and_then(|cookie| Some(cookie.get(AXUM_SESSION_COOKIE_NAME)?.to_owned()));
+            .unwrap();
+
+        let session_cookie = cookie
+            .as_ref()
+            .and_then(|cookie| cookie.get(AXUM_SESSION_COOKIE_NAME));
 
         // return the new created session cookie for client
         if session_cookie.is_none() {
@@ -114,11 +117,11 @@ where
         tracing::debug!(
             "UserIdFromSession: got exists session cookie, {}={}",
             AXUM_SESSION_COOKIE_NAME,
-            session_cookie.as_ref().unwrap()
+            session_cookie.unwrap()
         );
         // continue to decode the session cookie
         let user_id = if let Some(session) = store
-            .load_session(session_cookie.as_ref().unwrap().to_owned())
+            .load_session(session_cookie.unwrap().to_owned())
             .await
             .unwrap()
         {
@@ -138,7 +141,7 @@ where
             tracing::debug!(
                 "UserIdFromSession: err session not exists in store, {}={}",
                 AXUM_SESSION_COOKIE_NAME,
-                session_cookie.as_ref().unwrap()
+                session_cookie.unwrap()
             );
             return Err((StatusCode::BAD_REQUEST, "No session found for cookie"));
         };
--- a/examples/sessions/src/main.rs
+++ b/examples/sessions/src/main.rs
@@ -91,12 +91,10 @@ where
             .await
             .expect("`MemoryStore` extension missing");
 
-        let session_cookie =
-            if let Ok(TypedHeader(cookie)) = TypedHeader::<Cookie>::from_request(req).await {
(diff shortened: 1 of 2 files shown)


examples/sessions/src/main.rs
```diff
@@ -81,25 +86,40 @@ where
 
         let headers = req.headers().expect("other extractor taken headers");
 
-        let cookie = if let Some(cookie) = headers
+        let session_cookie = if let Some(cookie) = headers
```
jplatte, line 89: I think the whole parsing logic here should just be replaced with `TypedHeader<Cookie>::from_request`.

examples/sessions/src/main.rs
```diff
@@ -79,28 +91,44 @@ where
             .await
             .expect("`MemoryStore` extension missing");
 
-        let headers = req.headers().expect("other extractor taken headers");
+        let session_cookie =
+            if let Ok(TypedHeader(cookie)) = TypedHeader::<Cookie>::from_request(req).await {
+                cookie.get(AXUM_SESSION_COOKIE_NAME).map(|x| x.to_owned())
+            } else {
+                None
+            };
```
jplatte, line 99:
```suggestion
            Option::<TypedHeader<Cookie>>::from_request(req).await.and_then(
                |cookie| Some(cookie.get(AXUM_SESSION_COOKIE_NAME)?.to_owned())
            );
```

jplatte: Could also keep the `Option<TypedHeader>` in a local variable so `to_owned()` is not needed.

jplatte (approved):
Looks good to me, but I'll wait for David to also have a look and do the merge.

davidpdrsn (approved):
LGTM! Thanks 😃

PR #639, opened 2021-12-18
Change error type for `RequestParts::try_into_request`

### LINKED ISSUE #632: Change error type for `RequestParts::try_into_request`
Probably make sense to use [`RequestAlreadyExtracted`](https://docs.rs/axum/0.4.2/axum/extract/rejection/enum.RequestAlreadyExtracted.html) since those are the only ways it can fail.

See https://github.com/tokio-rs/axum/discussions/631#discussioncomment-1814984 for more context.

nylonicious:
## Motivation

Fixes #632.

I'm not sure if `Error::downcast` should be removed or not as it was only used in `try_into_request`, so for now I marked it with `#[allow(dead_code)]`

### DIFF
--- a/axum-core/src/error.rs
+++ b/axum-core/src/error.rs
@@ -15,6 +15,7 @@ impl Error {
         }
     }
 
+    #[allow(dead_code)]
     pub(crate) fn downcast<T>(self) -> Result<T, Self>
     where
         T: StdError + 'static,
--- a/axum-core/src/error.rs
+++ b/axum-core/src/error.rs
@@ -14,17 +14,6 @@ impl Error {
             inner: error.into(),
         }
     }
-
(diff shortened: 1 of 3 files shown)


jplatte (commented):
I think it's fine to remove the crate-internal downcast function.

Since this is a breaking change, I don't think it can be merged into `main` now. @davidpdrsn do you want to create a `next` / `0.5-pre` or similar branch that breaking-change PRs can go to?

davidpdrsn:
Yeah I think we can just remove the method. It's in git if we need it back.

> Since this is a breaking change, I don't think it can be merged into main now

I'll make a branch that we can merge breaking changes into. I'll do that next week and then can we merge this branch into that.

davidpdrsn (approved):

PR #642, opened 2021-12-21
Replace public use of `mime` crate with `&str`

### LINKED ISSUE #637: Change return type of `Field::content_type`
It should return some kind of string, or a type axum owns, to avoid the public dependency on mime.

`Mime` implements `AsRef<str>` meaning we can probably return `&str` but should verify that it returns the whole mime type and not just a part.

nashley:
This PR introduces a breaking change.

## Motivation

Resolves #637.
Removes exposure of the `mime` crate in the public library.
This will allow users to handle the underlying `&str` instead and optionally parse `Content-Type` headers with their library of choice (if desired).
This also allows `axum` to remove the dependency on `mime` in the future without requiring another breaking change.

## Solution

`&str` has replaced `&mime::Mime` as [`axum::extract::multipart::Field::content_type`](https://docs.rs/axum/latest/axum/extract/multipart/struct.Field.html#method.content_type)'s return type

Searching docs for `mime` shows no results:
```sh
> cd axum/src/docs
> rg mime
> # (no results)
```

## Questions
- [ ] Does this PR need to be against a different branch other than `main` for the `0.5` release?
- [ ] Should there be more tests? I couldn't find a good way to create parameterized tests without adding another dependency or creating a potentially-esoteric macro.
- [ ] Should there be a doc test for `content_type()`? Seems like it'd either be excessive or incomplete.

### DIFF
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -53,7 +53,7 @@ tokio-tungstenite = { optional = true, version = "0.16" }
 
 [dev-dependencies]
 futures = "0.3"
-reqwest = { version = "0.11", default-features = false, features = ["json", "stream"] }
+reqwest = { version = "0.11", default-features = false, features = ["json", "stream", "multipart"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 tokio = { version = "1.6.1", features = ["macros", "rt", "rt-multi-thread", "net"] }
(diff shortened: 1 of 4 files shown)


davidpdrsn:
> Does this PR need to be against a different branch other than main for the 0.5 release?

I'm gonna make a branch for 0.5 soon and then we can merge this into that.

> Should there be more tests? I couldn't find a good way to create parameterized tests without adding another dependency or creating a potentially-esoteric macro.

Thank you for adding a test! I dont think we need to add more tests here 😅

> Should there be a doc test for content_type()? Seems like it'd either be excessive or incomplete.

Nah I think it's fine as is.

davidpdrsn (approved):

PR #645, opened 2021-12-21
axum-core 0.2: release branch

davidpdrsn:
Same idea as https://github.com/tokio-rs/axum/pull/644 but for axum-core, instead of axum.

---

This branch contains the changes we're planning to make in the next breaking release of axum-core. Note this is a draft since we don't want to merge it yet. Want to wait a bit longer before making another breaking release.

PRs merged:
- None.


davidpdrsn:
Never mind this PR. We decided to ship axum 0.5 and axum-core 0.2 at the same time. So we only need one release PR, which is https://github.com/tokio-rs/axum/pull/644.

PR #647, opened 2021-12-21
Make `sse::Event` build event as a `BytesMut`

SabrinaJewson:
Revival of #594

## Motivation

Using `sse::Event` currently incurs a lot of allocations as each field is stored in a separate `String`.

## Solution

Build each SSE event with a `BytesMut` as methods are called on it.


davidpdrsn:
I've rebased `axum-next` on `main` so it now includes the SSE tests. Wanna rebase and see if the tests still pass? If so this is good to merge!

SabrinaJewson:
Tests seem to pass :D

PR #648, opened 2021-12-21
Add trait IntoResponseHeaders

jplatte:
## Motivation

TODO

## Solution

See the code? 😅

## Notes

I think this is all of the hard stuff dealt with. It should now be straight-forward to move `Headers` into axum, implement `IntoResponseHeaders` for `TypedHeader` and move it into its own module (outside `response`) / re-export it from the `axum` crate root like `Json` and add `IntoResponse` impls for tuples with more `IntoResponseHeaders` items.

### DIFF
--- a/axum-core/src/response/headers.rs
+++ b/axum-core/src/response/headers.rs
@@ -1,11 +1,5 @@
-use super::{IntoResponse, Response};
-use crate::body::boxed;
-use bytes::Bytes;
-use http::{
-    header::{HeaderMap, HeaderName, HeaderValue},
-    StatusCode,
-};
-use http_body::{Empty, Full};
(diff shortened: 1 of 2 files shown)


jplatte:
The last commit message is wrong, but the branch is protected? Can't fix it..

jplatte:
Can't even push normally. @davidpdrsn please unbreak PR branches! 😅

I'll create a new PR now to show that this actually can pass tests with minimal changes, will then wait until pushing works again before continuing work on this.

davidpdrsn:
Whoops! Should be fixed now. GitHub’s branch protection rules and how it interacts with the ability to auto merge is so bad…

PR #649, opened 2021-12-21
Add trait IntoResponseHeaders

jplatte:
## Motivation

It would be nice to be able to return `TypedHeader` in the same positions as `Headers<_>`, but that doesn't work without moving it into `axum-core` due to coherence and that would be very unfortunate. Also, `Headers` only lives in `axum-core` due to the same coherence issues.

## Solution

Introduce a new trait `IntoResponseHeaders` and implement it for `HeaderMap` and `Headers`. Move `Headers` into axum.

## Notes

I think this is all of the hard stuff dealt with. It should now be straight-forward to ~~move `Headers` into axum~~ (done), implement `IntoResponseHeaders` for `TypedHeader` and move it into its own module (outside `response`) / re-export it from the `axum` crate root like `Json` and add `IntoResponse` impls for tuples with more `IntoResponseHeaders` items.

### DIFF
--- a/axum-core/src/response/headers.rs
+++ b/axum-core/src/response/headers.rs
@@ -1,11 +1,8 @@
-use super::{IntoResponse, Response};
-use crate::body::boxed;
-use bytes::Bytes;
+use super::{IntoResponse, IntoResponseHeaders, Response};
 use http::{
-    header::{HeaderMap, HeaderName, HeaderValue},
+    header::{HeaderName, HeaderValue},
     StatusCode,
 };
-use http_body::{Empty, Full};
 use std::{convert::TryInto, fmt};
 
 /// A response with headers.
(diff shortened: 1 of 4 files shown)


jplatte:
Due to the new blanket impl of `IntoResponse`, the error message for it not being implemented now mentions `IntoResponseHeaders`. Don't have the Rust issue link at hand, but it's a somewhat well-known issue.

davidpdrsn (commented):
I think this is looking really good! Getting these more or less duplicate impls combined into is a big win imo!

jplatte:
I think it makes sense to deal with the proposed `TypedHeader` changes and adding more impls of the form `(StatusCode, $repeat(impl IntoResponseHeaders, N), impl IntoResponse)` in separate PRs.

This one is useful as-is.

davidpdrsn:
Sorry for taking so long to review this. Hoping to get to it this weekend.

jplatte:
Rebased.

davidpdrsn (approved):

PR #652, opened 2021-12-24
Add SSE tests

davidpdrsn:
While reviewing https://github.com/tokio-rs/axum/pull/647 I thought some tests for SSE would help. So this adds at least some basic test coverage.

cc @SabrinaJewson

### DIFF
--- a/axum/src/response/sse.rs
+++ b/axum/src/response/sse.rs
@@ -455,8 +455,9 @@ impl KeepAliveStream {
 mod tests {
     use super::*;
     use crate::{routing::get, test_helpers::*, Router};
-    use futures::{prelude::*, stream};
+    use futures::stream;
     use std::{collections::HashMap, convert::Infallible};
+    use tokio_stream::StreamExt as _;
 
     #[test]
     fn leading_space_is_not_stripped() {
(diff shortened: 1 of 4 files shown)


davidpdrsn:
Actually it should be possible to write the keep alive test in a way that is less flaky and dependent on the timings of things. I'll play around.

jplatte:
I know next to nothing about SSE, this should be reviewed by somebody else.

davidpdrsn:
@SabrinaJewson are you up for reviewing this? You've done a bunch of stuff around SSE in axum.

axum/src/response/sse.rs
```diff
@@ -451,11 +451,120 @@ impl KeepAliveStream {
...
+        let event_fields = parse_event(&res.chunk_text().await.unwrap());
+        assert_eq!(event_fields.get("event").unwrap(), "three");
+        assert_eq!(event_fields.get("retry").unwrap(), "30000");
+        assert_eq!(event_fields.get("id").unwrap(), "unique-id");
+        assert!(event_fields.get("comment").is_none());
+
+        assert!(res.chunk_text().await.is_none());
+    }
+
+    #[tokio::test]
```
SabrinaJewson, line 513: If you change this to `#[tokio::test(start_paused = true)]` then we can use [`time::advance`](https://docs.rs/tokio/latest/tokio/time/fn.advance.html) to have the test run more deterministically and without any delay.

axum/src/response/sse.rs
```diff
@@ -451,11 +451,120 @@ impl KeepAliveStream {
...
+        let event_fields = parse_event(&res.chunk_text().await.unwrap());
+        assert_eq!(event_fields.get("data").unwrap(), "one");
+
+        let event_fields = parse_event(&res.chunk_text().await.unwrap());
+        assert_eq!(event_fields.get("comment").unwrap(), "keep-alive-text");
+
+        let event_fields = parse_event(&res.chunk_text().await.unwrap());
+        assert_eq!(event_fields.get("data").unwrap(), "two");
+
+        assert!(res.chunk_text().await.is_none());
```
SabrinaJewson, line 547: After this we can test that keep alives continue to be sent:

```rs
time::advance(KEEP_ALIVE_INTERVAL);
let event_fields = parse_event(&res.chunk_text().await.unwrap());
assert_eq!(event_fields.get("comment").unwrap(), "keep-alive-text");

assert!(res.chunk_text().await.is_none());
```

SabrinaJewson (commented):
Looks good overall.

axum/src/response/sse.rs
davidpdrsn: Good idea! I've changed the keep alive tests!

davidpdrsn:
@SabrinaJewson ready for another look!

SabrinaJewson (approved):
Seems fine 👍

PR #656, opened 2021-12-27
Add `middleware::from_fn` for creating middleware from async fns

davidpdrsn:
I feel axum is in a pretty good place in terms of ease of use. However one area that is still a big hurdle for some people is `tower::Service` which quickly comes up up when writing custom middleware. Suddenly you cannot use high level `async`/`await` syntax anymore and have to worry about pinning, `poll_ready`, and most likely lots of generics.

This is something I've been thinking about for awhile but hadn't found a good solution for. But today I discovered poem's [`middleware_fn`](https://github.com/poem-web/poem/blob/master/examples/poem/middleware_fn/src/main.rs) which I think is perfect for this!

So this PR introduces `axum_extra::middleware::from_fn` that works like this:

```rust
use axum::{
    Router,
    http::{Request, StatusCode},
    routing::get,
    response::IntoResponse,
};
use axum_extra::middleware::{self, Next};

async fn auth<B>(req: Request<B>, next: Next<B>) -> impl IntoResponse {
    let auth_header = req.headers().get(http::header::AUTHORIZATION);

    match auth_header {
        Some(auth_header) if auth_header == "secret" => {
            Ok(next.run(req).await)
        }
        _ => Err(StatusCode::UNAUTHORIZED),
    }
}

let app = Router::new()
    .route("/", get(|| async { /* ... */ }))
    .route_layer(middleware::from_fn(auth));
```

You have pretty much the same level of control as with `tower::Service` but with a higher level API.

### DIFF
--- a/axum-extra/src/middleware/middleware_fn.rs
+++ b/axum-extra/src/middleware/middleware_fn.rs
@@ -89,6 +89,7 @@ where
 impl<F> fmt::Debug for MiddlewareFnLayer<F> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("MiddlewareFnLayer")
+            // Write out the type name, without quoting it as `&type_name::<F>()` would
             .field("f", &format_args!("{}", type_name::<F>()))
             .finish()
     }
@@ -152,7 +153,7 @@ where
     }
 }
 
-/// The remainder of a middleware stack, including the endpoint.
+/// The remainder of a middleware stack, including the handler.
 pub struct Next<ReqBody> {
     inner: BoxCloneService<Request<ReqBody>, Response, Infallible>,
 }
--- a/axum-extra/src/middleware/middleware_fn.rs
+++ b/axum-extra/src/middleware/middleware_fn.rs
@@ -55,7 +55,7 @@ use tower_service::Service;
 ///
 /// let app = Router::new()
 ///     .route("/", get(|| async { /* ... */ }))
-///     .layer(middleware::from_fn(auth));
+///     .route_layer(middleware::from_fn(auth));
 /// # let app: Router = app;
 /// ```
 pub fn from_fn<F>(f: F) -> MiddlewareFnLayer<F> {
--- a/axum-extra/src/middleware/middleware_fn.rs
+++ b/axum-extra/src/middleware/middleware_fn.rs
@@ -1,0 +1,239 @@
+//! Create middleware from async functions.
+//!
+//! See [`from_fn`] for more details.
+
+use axum::{
+    body::{self, Bytes, HttpBody},
+    response::{IntoResponse, Response},
+    BoxError,
+};
+use http::Request;
+use pin_project_lite::pin_project;
+use std::{
+    any::type_name,
+    convert::Infallible,
+    fmt,
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tower::{util::BoxCloneService, ServiceBuilder};
+use tower_http::ServiceBuilderExt;
+use tower_layer::Layer;
+use tower_service::Service;
+
+/// Create a middleware from an async function.
+///
+/// `from_fn` requires the function given to
+///
+/// 1. Be an `async fn`.
+/// 2. Take [`Request`](http::Request) as the first argument.
+/// 3. Take [`Next<B>`](Next) as the second argument.
+/// 4. Return something that implements [`IntoResponse`].
+///
+/// # Example
+///
+/// ```rust
+/// use axum::{
+///     Router,
+///     http::{Request, StatusCode},
+///     routing::get,
+///     response::IntoResponse,
+/// };
+/// use axum_extra::middleware::{from_fn, Next};
+///
+/// async fn auth<B>(req: Request<B>, next: Next<B>) -> impl IntoResponse {
+///     let auth_header = req.headers().get(http::header::AUTHORIZATION);
+///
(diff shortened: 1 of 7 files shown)


axum-extra/src/middleware/middleware_fn.rs
```diff
@@ -0,0 +1,261 @@
...
+            .field("f", &format_args!("{}", type_name::<F>()))
+            .field("inner", &self.inner)
+            .finish()
+    }
+}
+
+/// The remainder of a middleware stack, including the endpoint.
+pub struct Next<ReqBody> {
+    inner: BoxCloneService<Request<ReqBody>, Response, Infallible>,
+}
```
davidpdrsn, line 158: I also considered calling this `Rest` but went with `Next` since its more common across other frameworks [such as tide](https://docs.rs/tide/latest/tide/struct.Next.html).

davidpdrsn:
Also worth noting that for simple middleware tower's `map_request`, `map_response`, etc, utilities gets you must of the way there but its not easy to write something like the auth middleware shown in the PR description with a custom middleware (or [`extractor_middleware`](https://docs.rs/axum/latest/axum/extract/fn.extractor_middleware.html), which have other limitations that `middleware::from_fn` doesn't have).

axum-extra/src/middleware/middleware_fn.rs
```diff
@@ -0,0 +1,261 @@
...
+            f: self.f.clone(),
+            inner,
+        }
+    }
+}
+
+impl<F> fmt::Debug for MiddlewareFnLayer<F> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("MiddlewareFnLayer")
+            .field("f", &format_args!("{}", type_name::<F>()))
```
jplatte, line 92: I didn't know that `fmt::Arguments` implements `Debug` in addition to `Display`. Maybe add a comment though? I still find it looks pretty weird.

```suggestion
            // Write out the type name, without quoting it as `&type_name::<F>()` would
            .field("f", &format_args!("{}", type_name::<F>()))
```

axum-extra/src/middleware/middleware_fn.rs
```diff
@@ -0,0 +1,261 @@
...
+    F: FnMut(Request<ReqBody>, Next<ReqBody>) -> Fut,
+    Fut: Future<Output = Out>,
+    Out: IntoResponse,
+    S: Service<Request<ReqBody>, Response = Response<ResBody>, Error = Infallible>
+        + Clone
+        + Send
+        + 'static,
+    S::Future: Send + 'static,
+    ResBody: HttpBody<Data = Bytes> + Send + 'static,
+    ResBody::Error: Into<BoxError>,
```
jplatte, line 117: That's quite a long list of constraints ^^

I wonder what cursed middleware one could write that actually makes use of the extra capabilities `FnMut` gives in addition to "plain" `Fn` 😅

axum-extra/src/middleware/middleware_fn.rs
```diff
@@ -0,0 +1,261 @@
...
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("MiddlewareFnLayer")
+            .field("f", &format_args!("{}", type_name::<F>()))
+            .field("inner", &self.inner)
+            .finish()
+    }
+}
+
+/// The remainder of a middleware stack, including the endpoint.
```
jplatte, line 155:
```suggestion
/// The remainder of a middleware stack, including the handler.
```

axum-extra/src/middleware/mod.rs
```diff
@@ -0,0 +1,6 @@
+//! Additional types for creating middleware.
+
+pub mod middleware_fn;
+
+#[doc(inline)]
```
jplatte, line 5: Why generate duplicate doc pages for `from_fn` and `Next` ?

axum-extra/src/middleware/middleware_fn.rs
davidpdrsn: Hehe yeah 😜 I'm not sure `FnMut` is that useful in practice actually but since we a `&mut self` I figure why not. [Tower also uses `FnMut`.](https://docs.rs/tower/latest/tower/filter/trait.Predicate.html#impl-Predicate%3CT%3E)

jplatte (approved):

johannescpk:
Probably a naive question, but how would one pass some context variables into the middleware function? I've tried wrapping the middleware itself into a "factory function", but failed to do so because of the `B` generics requirement. Basically something like

```rust
pub fn auth(some_context_variable: Context) -> impl Fn(Request<_>, Next<_>) -> impl IntoResponse {
    |req: Request<_>, next: Next: <_>| async move {
        // do based on `some_context_variable`
        Ok(next.run(req).await)
    }
}

**Comment by davidpdrsn** [comment]
@johannescpk You can do this:

```rust
use axum::{http::Request, response::IntoResponse, Router};
use axum_extra::middleware::{self, Next};
use std::net::SocketAddr;

#[derive(Clone)]
struct Context {}

#[tokio::main]
async fn main() {
    let ctx = Context {};

    let app = Router::new().layer(middleware::from_fn(move |req, next| {
        foo(ctx.clone(), req, next)
    }));

    axum::Server::bind(&addr)
        .await
        .unwrap();
}

async fn foo<B>(ctx: Context, req: Request<B>, next: Next<B>) -> impl IntoResponse {
    next.run(req).await
}
```

PR #661, opened 2021-12-27
Move header name validity checks to compile time

jplatte:
… using HeaderValue::from_static.

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -215,13 +215,15 @@ impl WebSocketUpgrade {
             callback(socket).await;
         });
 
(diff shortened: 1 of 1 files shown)


SabrinaJewson:
I don't think moving them into `const`s has any advantage, since LLVM's const folding will make them constants anyway if it can.

jplatte:
This does have an advantage in that it removes two runtime panic paths that detract from the main logic when reading. Also `from_str` isn't trivial, constant-folding it away is certainly possible but I wouldn't blindly assume that it happens.

davidpdrsn (approved):

PR #663, opened 2021-12-27
Fix axum-extra's optional feature documentation

SabrinaJewson:
## Motivation

* docs.rs doesn't display `axum_extra::response::ErasedJson`.
* If it was displayed, it doesn't have the purple feature box.

## Solution

Make sure to enable `--all-features` and `--cfg docsrs` when building `axum_extra` on docs.rs, and add the appropriate `doc(cfg)` to `ErasedJson`.

### DIFF
--- a/axum-extra/src/response/erased_json.rs
+++ b/axum-extra/src/response/erased_json.rs
@@ -28,6 +28,7 @@ use serde::Serialize;
 ///     }
 /// }
 /// ```
+#[cfg_attr(docsrs, doc(cfg(feature = "erased-json")))]
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


axum-extra/src/response/erased_json.rs
```diff
@@ -28,6 +28,7 @@ use serde::Serialize;
 ///     }
 /// }
 /// ```
+#[cfg_attr(docsrs, doc(cfg(feature = "erased-json")))]
```
jplatte, line 31: For future reference, this is needed because of https://github.com/rust-lang/rust/issues/83428.

jplatte (approved):

PR #669, opened 2021-12-27
Document how to pass state to `middleware::from_fn`

davidpdrsn:
I think this makes sense to have examples for.

### DIFF
--- a/axum-extra/src/middleware/middleware_fn.rs
+++ b/axum-extra/src/middleware/middleware_fn.rs
@@ -78,6 +78,7 @@ use tower_service::Service;
 /// async fn my_middleware<B>(
 ///     req: Request<B>,
 ///     next: Next<B>,
+///     state: State,
 /// ) -> impl IntoResponse {
 ///     // ...
 ///     # ()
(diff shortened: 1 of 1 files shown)


axum-extra/src/middleware/middleware_fn.rs
```diff
@@ -58,6 +58,86 @@ use tower_service::Service;
...
+///     state: State,
+/// ) -> impl IntoResponse {
+///     if state.in_maintenance {
+///         Err(StatusCode::SERVICE_UNAVAILABLE)
+///     } else {
+///         Ok(next.run(req).await)
+///     }
+/// }
+///
+/// let state = State {
```
jplatte, line 92: Wouldn't you usually want `Arc<Mutex<State>>` or something like that for this kind of state?

davidpdrsn: Yeah you probably would. I guess removing the fields from this example well makes it less cluttered with these kinds of things. It probably depends on the exact use case exactly how you wanna share some piece of state.

jplatte (approved):

PR #671, opened 2021-12-27
Add metrics example

davidpdrsn:
We often get questions from people asking about how to track metrics but
I've been reluctant to add an example for it since it required writing a
`tower::Service` middleware. However now that we have
`middleware::from_fn` I think it makes sense to have!

The longterm goal is for tower-http to provide something but I haven't
found the right design for it yet.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -166,4 +166,11 @@ jobs:
         CC: clang
       with:
         command: check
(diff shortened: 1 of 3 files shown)


jplatte:
Looks reasonable at a glance, but I have no experience with metrics. Should sbd. else review maybe?

davidpdrsn:
We've been using pretty much this exact setup at Embark for a while so should be good to go 😊

jplatte (approved):

PR #672, opened 2021-12-28
Store Bytes in ErasedJson

SabrinaJewson:
## Motivation

Storing a `Vec<u8>` can result in more data copies than necessary:

- `From<Vec<u8>>` for `Bytes` calls `.into_boxed_slice()` which can reallocate.
- Attempts to clone the `Bytes` will then copy the data again.

## Solution

Store a `Bytes` to begin with by writing to a `BytesMut`, making everything zero-copy.


jplatte (approved):
This makes a lot of sense; thanks!

PR #680, opened 2021-12-31
Fix the link to "Routing to services and backpressure"

patrick-gu:
In `Router::route`, the link to the "Routing to services and backpressure" section on the crate documentation page directed to <https://docs.rs/#routing-to-services-and-backpressure>. This change makes it direct to the main crate documentation page.

## Motivation

On the documentation for [`Router::route`](https://docs.rs/axum/latest/axum/routing/struct.Router.html#method.route), the link to [Routing to services and backpressure](https://docs.rs/axum/latest/axum/index.html#routing-to-services-and-backpressure) instead links to <https://docs.rs/#routing-to-services-and-backpressure>. This pull request fixes that, and makes the link correct.

## Solution

This updates the link to point to the correct page. This is similar to the link found here:
https://github.com/tokio-rs/axum/blob/8411b7873680d4b5a2d390bdd2f520186bb94c9c/axum/src/add_extension.rs#L10

### DIFF
--- a/axum/src/docs/routing/route.md
+++ b/axum/src/docs/routing/route.md
@@ -164,7 +164,7 @@ Routing to arbitrary services in this way has complications for backpressure
 ([`Service::poll_ready`]). See the [Routing to services and backpressure] module
 for more details.
 
-[Routing to services and backpressure]: /#routing-to-services-and-backpressure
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
Thanks!

PR #683, opened 2022-01-03
Document passing state to handlers via closure captures

davidpdrsn:
## Motivation

Passing state to handlers with `Extension` will give you runtime errors if you forget to add the `AddExtension` middleware or you try and extract the wrong type. However state can also be passed using closure captures, which doesn't have this issue. Its a little more verbose though.

## Solution

I think its still worth documenting since it might not be obvious to beginners that you can do this.

### DIFF
--- a/axum/src/lib.rs
+++ b/axum/src/lib.rs
@@ -227,9 +227,17 @@
 //! # Sharing state with handlers
 //!
 //! It is common to share some state between handlers for example to share a
-//! pool of database connections or clients to other services. That can be done
-//! using the [`AddExtension`] middleware (applied with [`AddExtensionLayer`])
-//! and the [`Extension`](crate::extract::Extension) extractor:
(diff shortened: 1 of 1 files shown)


axum/src/lib.rs
```diff
@@ -260,6 +268,68 @@
 //! # };
 //! ```
 //!
+//! The downside to this approach is that you'll get runtime errors
+//! (specifically a `500 Internal Server Error` response) if you try and extract
+//! an extension that doesn't exist, perhaps because you forgot add the
+//! middleware or because you're extracting the wrong type.
```
jplatte, line 274: Closure captures should also have lower runtime cost as the clones only happen when that specific route is invoked, right?

jplatte: Not sure whether it's worth to point that out in docs though.

jplatte (approved):

PR #686, opened 2022-01-03
Add some debug assertions

davidpdrsn:
As discussed in Discord its probably worth having debug assertions for these two cases, should we accidentally break things in the future.

### DIFF
--- a/axum/src/extract/request_parts.rs
+++ b/axum/src/extract/request_parts.rs
@@ -56,7 +56,13 @@ where
     async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
         let uri = Extension::<Self>::from_request(req)
             .await
-            .unwrap_or_else(|_| Extension(OriginalUri(req.uri().clone())))
+            .unwrap_or_else(|_| {
+                #[cfg(debug_assertions)]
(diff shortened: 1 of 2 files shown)


axum/src/extract/request_parts.rs
```diff
@@ -56,7 +56,13 @@ where
     async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
         let uri = Extension::<Self>::from_request(req)
             .await
-            .unwrap_or_else(|_| Extension(OriginalUri(req.uri().clone())))
+            .unwrap_or_else(|_| {
+                #[cfg(debug_assertions)]
+                panic!("`OriginalUri` extension should always be present, but it wasn't");
```
jplatte, line 61: The other panic makes sense to me but I'm unsure about this one. There is the possibility of `ExtensionsAlreadyExtracted`, which would be a programmer error rather than a bug in axum. Should we maybe change the `Rejection` type in 0.5?

Also it's easy to call this `FromRequest` impl on a `RequestParts` that wasn't created through the axum `Router`. As a somewhat crazy example, somebody could have their own router that uses `FromRequest`.

davidpdrsn: Thats a good point! I'll remove this one

jplatte (approved):

PR #687, opened 2022-01-03
Don't silently swallow errors in `OriginalUri` extractor

davidpdrsn:
Failing to extract `OriginalUri` could be the result of a programmer error so we shouldn't swallow those as that might result in wrong data being given to the user.

### DIFF
--- a/axum/src/extract/rejection.rs
+++ b/axum/src/extract/rejection.rs
@@ -61,8 +61,9 @@ define_rejection! {
 define_rejection! {
     #[status = INTERNAL_SERVER_ERROR]
     #[body = "No original uri was found for the matched route. This is likely a bug in axum. Please open an issue"]
-    /// Rejection type used if axum's internal representation of original URIs is missing. This
-    /// should never happen and is a bug in axum if it does.
+    /// Rejection type used if axum's internal representation of original URIs is missing.
+    ///
+    /// This should never happen and is likely a bug in axum if it does.
     pub struct MissingOriginalUri;
 }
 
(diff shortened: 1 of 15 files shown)


axum/src/extract/rejection.rs
```diff
@@ -58,6 +58,14 @@ define_rejection! {
     pub struct MissingPathParams;
 }
 
+define_rejection! {
+    #[status = INTERNAL_SERVER_ERROR]
+    #[body = "No original uri was found for the matched route. This is a bug in axum. Please open an issue"]
```
davidpdrsn, line 63: As noted previously, this might also be a programmer error if using the extractor outside of a `Router` but thats quite an edge case so I think this wording is fine.

jplatte: Maybe weaken the message just a small bit?

```suggestion
    #[body = "No original uri was found for the matched route. This is likely a bug in axum. Please open an issue"]
```

davidpdrsn: Sure@

axum/src/extract/rejection.rs
```diff
@@ -58,6 +58,14 @@ define_rejection! {
     pub struct MissingPathParams;
 }
 
+define_rejection! {
+    #[status = INTERNAL_SERVER_ERROR]
+    #[body = "No original uri was found for the matched route. This is likely a bug in axum. Please open an issue"]
+    /// Rejection type used if axum's internal representation of original URIs is missing. This
+    /// should never happen and is a bug in axum if it does.
```
jplatte, line 65:
```suggestion
    /// Rejection type used if axum's internal representation of original URIs is missing.
    ///
    /// This should never happen and is likely a bug in axum if it does.
```

jplatte:
What happened with this PR?

davidpdrsn:
wat. No clue. I'll reopen tomorrow.

Issue #684, opened 2022-01-03
Document extracting `OriginalUri` directly via request extensions


davidpdrsn:
It might not be entirely obvious but `OriginalUri` can be extracted directly via request extensions for use in a `tower::Service` middleware (which cannot easily run `FromRequest` extractors). That might be nice to use as the fallback for middleware that use `MatchedPath` which wont be present for fallback routes.

I think it makes sense to document that in similar style to the [`MatchedPath` path example](https://docs.rs/axum/latest/axum/extract/struct.MatchedPath.html).

`OriginalUri` will always be present if the request hits the router, regardless if there is a matching route or not. That should be documented as well.

PR #688, opened 2022-01-03
Document extracting `OriginalUri` directly via request extensions

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/684

### DIFF
--- a/axum/src/extract/request_parts.rs
+++ b/axum/src/extract/request_parts.rs
@@ -43,6 +43,43 @@ use sync_wrapper::SyncWrapper;
 /// # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
 /// # };
 /// ```
+///
+/// # Extracting via request extensions
+///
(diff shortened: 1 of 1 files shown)


davidpdrsn:
@jplatte I have no idea what happened to this PR and why it was closed a few days. But its back! 🚀

axum/src/extract/request_parts.rs
```diff
@@ -43,6 +43,43 @@ use sync_wrapper::SyncWrapper;
 /// # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
 /// # };
 /// ```
+///
+/// # Extracting via request extensions
+///
+/// `OriginalUri` can also be accessed from middleware via request extensions.
+/// This is useful for example with [`Trace`](tower_http::trace::Trace) to
+/// create a span that contains the original URI, if your service might be nested:
```
jplatte, line 51:
```suggestion
/// create a span that contains the full path, if your service might be nested:
```

jplatte (approved):

PR #698, opened 2022-01-11
Change `HeaderMap` extractor to clone the headers

davidpdrsn:
We often hear from users who are hit by `Headers taken by other extractor` errors because they used `HeaderMap` as an extractor combined with `Form<T>`, or some other extractor that happens to use the headers.

The reason [`RequestParts::take_headers`](https://docs.rs/axum/latest/axum/extract/struct.RequestParts.html#method.take_headers) exists in the first place is for performance reasons. But I think I've realized that it was probably a mistake because it causes issues like this. So I think we should change it!

This PR removes `RequestParts::take_headers` and changes the `HeaderMap` extractor to clone the headers. We could also consider other workarounds such as recording which extractor removed and which requires the headers, but I think this solution is the least surprising, and worth the small perf hit.

If someone is really concerned about cloning the headers you can still write an extractor that does `request_parts.headers_mut().take()` which would leave an empty `HeaderMap` in place. I don't think axum should provide such an extractor though --- then the problem is just gonna stay.

This PR doesn't touch "extensions already taken by another extractor" which is basically the same problem. I'm not very concerned about that since it feels more niche but probably worth looking into as well.

## TODO

- [x] Changelog
- [x] Document the cloning

### DIFF
--- a/axum-core/CHANGELOG.md
+++ b/axum-core/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `RequestParts::headers` returns `&HeaderMap`.
     - `RequestParts::headers_mut` returns `&mut HeaderMap`.
     - `HeadersAlreadyExtracted` has been removed.
-    - The `HeadersAlreadyExtracted` removed variant has been removed from these rejections:
+    - The `HeadersAlreadyExtracted` variant has been removed from these rejections:
         - `RequestAlreadyExtracted`
         - `RequestPartsAlreadyExtracted`
     - `<HeaderMap as FromRequest<_>>::Error` has been changed to `std::convert::Infallible`.
--- a/axum-core/CHANGELOG.md
+++ b/axum-core/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **breaking:** Using `HeaderMap` as an extractor will no longer remove the headers and thus
   they'll still be accessible to other extractors, such as `axum::extract::Json`. Instead
   `HeaderMap` will clone the headers. You should prefer to use `TypedHeader` to extract only the
-  header you need ([#698])
+  headers you need ([#698])
 
   This includes these breaking changes:
     - `RequestParts::take_headers` has been removed.
--- a/axum-core/CHANGELOG.md
+++ b/axum-core/CHANGELOG.md
@@ -7,7 +7,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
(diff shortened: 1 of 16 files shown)


axum/src/docs/extract.md
```diff
@@ -377,7 +373,7 @@ where
     type Rejection = (StatusCode, &'static str);
 
     async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
-        let user_agent = req.headers().and_then(|headers| headers.get(USER_AGENT));
+        let user_agent = req.headers().get(USER_AGENT);
```
jplatte, line 376: Might as well inline?

jplatte (approved):
I like the various code simplifications! I think we should be consistent and do the same thing for extensions, as well as adding some warnings about the mostly-unnecessary cloning (of headers you're not interested in) to the affected `FromRequest` impls.

axum-core/CHANGELOG.md
```diff
@@ -7,7 +7,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  they'll still be accessible to other extractors, such as `axum::extract::Json`. Instead
+  `HeaderMap` will clone the request. You should prefer to use `TypedHeader` to extract only the
+  header you need ([#698])
+
+  This includes these breaking changes:
+    - `RequestParts::take_headers` has been removed.
+    - `RequestParts::headers` returns `&HeaderMap`.
+    - `RequestParts::headers_mut` returns `&mut HeaderMap`.
+    - `HeadersAlreadyExtracted` has been removed.
+    - The `HeadersAlreadyExtracted` removed variant has been removed from these rejections:
```
jplatte, line 20:
```suggestion
    - The `HeadersAlreadyExtracted` variant has been removed from these rejections:
```

SabrinaJewson:
Would there be the possibility of having `RequestParts` use an `Arc<HeaderMap>`? Then the cloning would become cheap. I don't know how useful it is to be able to mutate headers during extraction.

davidpdrsn:
Hm not sure if it's worth it since it's only the `HeaderMap` extractor that got more expensive. If you don't use that things should work the same.

rich-murphey:
I believe most session implementations would depend on cookies or JWT obtained from the HeaderMap, and rarely if ever mutate the headers.  As it stands, would this make every API call that uses sessions more expensive?

davidpdrsn:
> As it stands, would this make every API call that uses sessions more expensive?

Like I just wrote, no it won't. 

_Only_ using `HeaderMap` as an extractor is more expensive.

If you need a single header for a handler you should use `TypedHeader`, or write your own extractor if the `headers` crate doesn't have the header you need.

If you need a header in your own extractor you just call `request_parts.headers()` which returns `&HeaderMap` and is free.

rich-murphey:
Thanks for the clarification!

PR #699, opened 2022-01-11
Remove `RequestParts::take_extensions`

davidpdrsn:
Same story as https://github.com/tokio-rs/axum/pull/698 but this time for extensions.

Removes `RequestParts::take_extensions` and `impl FromRequest for Extensions` meaning `Extensions` is no longer an extractor. Been thinking before that it probably didn't have much utility in the first place since all you can do is [`get`](https://docs.rs/http/latest/http/struct.Extensions.html#method.get) types, which `extract::Extension` can also do. Users can still use `RequestParts::extensions_mut` however to `std::mem::take` them if they really need to.

## TODO

- [x] Changelog

### DIFF
--- a/axum-core/src/extract/mod.rs
+++ b/axum-core/src/extract/mod.rs
@@ -115,7 +115,7 @@ impl<B> RequestParts<B> {
 
     /// Convert this `RequestParts` back into a [`Request`].
     ///
-    /// Fails if The request body has been extracted, that is [`take_body`] have
+    /// Fails if The request body has been extracted, that is [`take_body`] has
     /// been called.
     ///
     /// [`take_body`]: RequestParts::take_body
--- a/axum-core/src/extract/mod.rs
+++ b/axum-core/src/extract/mod.rs
@@ -115,14 +115,9 @@ impl<B> RequestParts<B> {
 
     /// Convert this `RequestParts` back into a [`Request`].
     ///
-    /// Fails if
+    /// Fails if The request body has been extracted, that is [`take_body`] have
+    /// been called.
     ///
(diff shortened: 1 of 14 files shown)


axum-core/src/extract/mod.rs
```diff
@@ -108,52 +108,38 @@ impl<B> RequestParts<B> {
...
-            extensions: Some(extensions),
+            extensions,
             body: Some(body),
         }
     }
 
     /// Convert this `RequestParts` back into a [`Request`].
     ///
-    /// Fails if
+    /// Fails if The request body has been extracted, that is [`take_body`] have
```
jplatte, line 118:
```suggestion
    /// Fails if The request body has been extracted, that is [`take_body`] has
```

axum-core/src/macros.rs
```diff
@@ -26,12 +26,6 @@ macro_rules! define_rejection {
         }
 
         impl std::error::Error for $name {}
-
-        impl Default for $name {
-            fn default() -> Self {
-                Self
-            }
-        }
```
jplatte, line 34: This change seems unrelated.

davidpdrsn: Since `BodyAlreadyExtracted` is the only rejection left in core that didn't wrap another error I thought it was nicer just add `#[derive(Default)]` directly at the definition site. Technically we could remove this whole half of the macro since its only used once.

jplatte: Okay, sure. 👍🏼 for getting rid of macro code that's only used once.

davidpdrsn: Done!

jplatte (approved):

PR #707, opened 2022-01-14
feat/mpack-extractor

Lainera:
TL;DR: Adds [`MessagePack`](https://msgpack.org/index.html) extractor via [rmp-serde](https://docs.rs/rmp-serde/latest/rmp_serde)

## Motivation

Using `axum` in a couple of projects and liking it a lot. Since I use it with `MessagePack`, decided to add extractor for the format. Given that `MessagePack` is not as ubiquitous as `JSON`, gated by feature flag. 

## Solution

Following implementation of `Json` extractor added similar functionality, docs & tests for `MessagePack`.

### DIFF
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -15,6 +15,7 @@ default = ["http1", "json", "tower-log"]
 http1 = ["hyper/http1"]
 http2 = ["hyper/http2"]
 json = ["serde_json"]
+messagepack = ["rmp-serde"]
 multipart = ["multer"]
 tower-log = ["tower/log"]
 ws = ["tokio-tungstenite", "sha-1", "base64"]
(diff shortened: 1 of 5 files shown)


davidpdrsn:
So this https://crates.io/crates/axum-msgpack?

I don't use of MessagePack is common enough for warrant an extractor directly in axum itself. It should be an external crate imo (and already is).

davidpdrsn:
I suppose we could put it in axum-extra 🤔

davidpdrsn:
Going to make a call here and say that we don't do this, for now at least. I encourage you to look at https://crates.io/crates/axum-msgpack instead.

You're very welcome to make a PR that adds that to [ECOSYSTEM.md](https://github.com/tokio-rs/axum/blob/main/ECOSYSTEM.md) to make it easier for people to discover 😊

PR #708, opened 2022-01-14
Update to changed compiler error in 1.58

davidpdrsn:
The trybuild tests needed to be updated for 1.58. We only run the tests
on stable (but build on other toolchains).


jplatte:
`axum-next` branch needs a rebase to fix the same CI issue there.

PR #711, opened 2022-01-18
Fix path route in extractor docs

AaronErhardt:
## Motivation

The [extract](https://docs.rs/axum/latest/axum/extract/index.html#common-extractors) docs suggest that you don't need `path/:variable` in path extractor routes but actually you do.

## Solution

Add `:user_id` to the path of the route.


davidpdrsn (approved):
Thanks!

PR #717, opened 2022-01-23
Move `Headers` into axum

davidpdrsn:
- Moves `Headers` from axum-core and into axum.
- Removes the blanket `impl<H> IntoResponse for H where H: IntoResponseHeaders`. This really hurt the error message from `#[debug_handler]` as it would mention the `IntoResponseHeaders` trait instead of `IntoResponse`.
- Implement `IntoResponse` for `HeaderMap` and `Headers`. Having to do this is a bit unfortunate but I'd rather have the better error message for `#[debug_handler]`.
- This also means `IntoResponseHeaders` can be implemented outside of axum-core. I guess thats fine 🤷

### DIFF
--- a/axum-core/src/response/mod.rs
+++ b/axum-core/src/response/mod.rs
@@ -144,38 +144,17 @@ pub trait IntoResponse {
 }
 
 /// Trait for generating response headers.
-///
-
-/// **Note: If you see this trait not being implemented in an error message, you are almost
-/// certainly being mislead by the compiler¹. Look for the following snippet in the output and
-/// check [`IntoResponse`]'s documentation if you find it:**
-///
-/// ```text
-/// note: required because of the requirements on the impl of `IntoResponse` for `<type>`
-/// ```
-///
-/// Any type that implements this trait automatically implements `IntoResponse` as well, but can
-/// also be used in a tuple like `(StatusCode, Self)`, `(Self, impl IntoResponseHeaders)`,
-/// `(StatusCode, Self, impl IntoResponseHeaders, impl IntoResponse)` and so on.
-///
-/// This trait can't currently be implemented outside of axum.
-///
-/// ¹ See also [this rustc issue](https://github.com/rust-lang/rust/issues/22590)
 pub trait IntoResponseHeaders {
-    /// The return type of [`.into_headers()`].
+    /// The return type of `into_headers`.
     ///
-    /// The iterator item is a `Result` to allow the implementation to return a server error
+    /// The iterator item is a [`Result`] to allow the implementation to return a server error
     /// instead.
     ///
-    /// The header name is optional because `HeaderMap`s iterator doesn't yield it multiple times
+    /// The header name is optional because [`HeaderMap`]s iterator doesn't yield it multiple times
     /// for headers that have multiple values, to avoid unnecessary copies.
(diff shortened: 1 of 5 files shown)


axum-core/src/response/mod.rs
```diff
@@ -149,38 +144,17 @@ pub trait IntoResponse {
...
-///
-/// ```text
-/// note: required because of the requirements on the impl of `IntoResponse` for `<type>`
-/// ```
-///
-/// Any type that implements this trait automatically implements `IntoResponse` as well, but can
-/// also be used in a tuple like `(StatusCode, Self)`, `(Self, impl IntoResponseHeaders)`,
-/// `(StatusCode, Self, impl IntoResponseHeaders, impl IntoResponse)` and so on.
-///
-/// This trait can't currently be implemented outside of axum.
```
jplatte, line 166: The current trait members are rather weird, do you really want them to be part of the public API nominally?

axum-core/src/response/mod.rs
```diff
@@ -356,17 +330,10 @@ impl IntoResponse for StatusCode {
     }
 }
 
-impl<H> IntoResponse for H
-where
-    H: IntoResponseHeaders,
-{
+impl IntoResponse for HeaderMap {
```
jplatte, line 333: I guess this makes sense. It's really annoying the previous solution worsens rustc's error reporting.

axum-core/src/response/mod.rs
davidpdrsn: Yeah they're quite odd at first glance. I suppose we could make them `#[doc(hidden)]` 🤔

axum-core/src/response/mod.rs
davidpdrsn: Yeah it is 😞 I still think its worth it overall

davidpdrsn: (including the introduction of `IntoResponseHeaders`)

jplatte: Yeah, agreed.

axum-core/src/response/mod.rs
jplatte: I think we should.

jplatte (approved):
LGTM with `#[doc(hidden)]` re-added.

axum-core/src/response/mod.rs
```diff
@@ -149,37 +144,18 @@ pub trait IntoResponse {
...
     ///
-    /// The header name is optional because `HeaderMap`s iterator doesn't yield it multiple times
+    /// The header name is optional because [`HeaderMap`]s iterator doesn't yield it multiple times
     /// for headers that have multiple values, to avoid unnecessary copies.
     #[doc(hidden)]
     type IntoIter: IntoIterator<Item = Result<(Option<HeaderName>, HeaderValue), Response>>;
 
     /// Attempt to turn `self` into a list of headers.
-    ///
-    /// In practice, only the implementation for `axum::response::Headers` ever returns `Err(_)`.
```
jplatte, line 182: Why was this deleted, by the way?

davidpdrsn: Ups not sure. Added it back in 208720be531df.

PR #718, opened 2022-01-23
Add `axum-macros` crate with `#[derive(FromRequest)]`

davidpdrsn:
See the docs for example usage.

Things I intentionally don't handle:
- `#[derive(FromRequest)]` on generic structs


examples/hello-world/Cargo.toml
```diff
@@ -5,5 +5,7 @@ edition = "2018"
 publish = false
 
 [dependencies]
-axum = { path = "../../axum" }
+axum = { path = "../../axum", features = ["headers"] }
+axum-macros = { path = "../../axum-macros" }
```
jplatte, line 9: Wouldn't it be better to have a separate macro / derive example and leave the hello-world example as-is?

axum-macros/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+use syn::parse::Parse;
+
+mod from_request;
+
+#[proc_macro_derive(FromRequest, attributes(from_request))]
+pub fn derive_from_request(item: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    expand_with(item, from_request::expand)
+}
+
+fn expand_with<F, T, K>(input: proc_macro::TokenStream, f: F) -> proc_macro::TokenStream
```
jplatte, line 11: What's the point of this abstraction?

axum-macros/src/from_request.rs
```diff
@@ -0,0 +1,186 @@
...
+                    format!("Unknown attribute: `{}`", ident),
+                ))
+            }
+        })
+        .collect::<syn::Result<Vec<_>>>()?;
+
+    let mut out = FromRequestAttrs::default();
+    for attr in attrs {
+        match attr {
+            Attr::FromRequest(from_requst_attrs) => {
```
jplatte, line 166: Typo: "requst"

axum-macros/src/from_request.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    let attrs = attrs
+        .iter()
+        .filter_map(|attr| attr.path.get_ident().map(|ident| (ident, attr)))
+        .map(|(ident, attr)| {
+            if ident == "from_request" {
+                attr.parse_args_with(Punctuated::parse_terminated)
+                    .map(Attr::FromRequest)
+            } else {
```
jplatte, line 154: If there are attributes other than `from_request`, they should simply be ignored rather than raising an error. They could be for a different macro (or builtin, though cfg attrs are expanded before derive macro expansion).

examples/hello-world/Cargo.toml
davidpdrsn: Yea I'm just using this for testing while developing.

axum-macros/src/lib.rs
davidpdrsn: Just removes some of the boilerplate. I expect this crate to get more macros overtime so I think it's nice. axum-debug could be moved in here.

axum-macros/src/from_request.rs
```diff
@@ -0,0 +1,381 @@
...
+        (syn::parse_quote!(::std::convert::Infallible), quote! {})
+    } else {
+        let rejection_ident = rejection_ident(&ident);
+        let rejection = extract_each_field_rejection(&ident, &fields, &vis)?;
+        (rejection_ident, rejection)
+    };
+
+    Ok(quote! {
+        #[::axum::async_trait]
+        impl<B> ::axum::extract::FromRequest<B> for #ident
```
jplatte, line 57: These impls should be marked [automatically_generated](https://doc.rust-lang.org/reference/attributes/derive.html#the-automatically_derived-attribute).

```suggestion
        #[::axum::async_trait]
        #[automatically_generated]
        impl<B> ::axum::extract::FromRequest<B> for #ident
```

davidpdrsn: Uhh not heard about `#[automatically_generated]` before! Where is that documented?

jplatte: At the place I linked to 😅

It can also improve compile times, this is the first reference to that I could find: https://github.com/rust-lang/rust/commit/24d410abd6924c175ee776ed84be851b615a555b

axum-macros/src/from_request.rs
```diff
@@ -0,0 +1,381 @@
...
+            let ident = attr.path.get_ident()?;
+            if ident == "from_request" {
+                Some(
+                    attr.parse_args_with(Punctuated::parse_terminated)
+                        .map(Attr::FromRequest),
+                )
+            } else {
+                None
+            }
+        })
```
jplatte, line 340:
```suggestion
        .filter(|attr| attr.path.is_ident("from_request"))
        .map(|attr| {
            attr.parse_args_with(Punctuated::parse_terminated).map(Attr::FromRequest)
        })
```

axum-macros/src/from_request.rs
davidpdrsn: oh lol sorry 😜 cool!

axum-macros/src/from_request.rs
davidpdrsn: Nice!

axum-macros/src/from_request.rs
davidpdrsn: Seems to be `#[automatically_derived]`

axum-macros/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+/// Note this requires the via extractor to be a generic tuple struct that implements `FromRequest`
+/// and has exactly one public field:
+///
+/// ```
+/// pub struct ViaExtractor<T>(pub T);
+///
+/// // impl<T, B> FromRequest<B> for ViaExtractor<T> { ... }
+/// ```
+///
+/// More complex via extractors are not supported and requires writing a manual implementation.
```
jplatte, line 128:
```suggestion
/// More complex via extractors are not supported and require writing a manual implementation.
```

axum-macros/src/from_request.rs
jplatte: Oh, right ^^

axum-macros/src/from_request.rs
```diff
@@ -0,0 +1,499 @@
...
+            .map(|field| {
+                let variant_name = rejection_variant_name(field)?;
+                Ok(quote! {
+                    Self::#variant_name(inner) => inner.fmt(f),
+                })
+            })
+            .collect::<syn::Result<Vec<_>>>()?;
+
+        quote! {
+            impl ::std::fmt::Display for #rejection_ident {
```
jplatte, line 283: There's this + two more `impl ::`s that don't have `#[automatically_derived]` yet.

davidpdrsn:
It now supports all the use cases I had in mind :)

axum-macros/src/from_request.rs
```diff
@@ -0,0 +1,533 @@
...
+
+            let field_ty = &field.ty;
+            let ty_span = field_ty.span();
+
+            let variant_name = rejection_variant_name(field)?;
+
+            let extractor_ty = if let Some((_, path)) = via {
+                if let Some(inner) = peel_option(field_ty) {
+                    quote_spanned! {ty_span=>
+                        Option<#path<#inner>>
```
jplatte, line 230: For consistency, if `peel_option` and `peel_result_ok` work with weird ways of writing `Option`, we shouldn't just assume it is in scope here right?

```suggestion
                        ::std::option::Option<#path<#inner>>
```

axum-macros/src/from_request.rs
```diff
@@ -0,0 +1,533 @@
...
+                    }
+                } else if let Some(inner) = peel_result_ok(field_ty) {
+                    quote_spanned! {ty_span=>
+                        Result<#path<#inner>, TypedHeaderRejection>
```
jplatte, line 234: Same here.

axum-macros/src/from_request.rs
```diff
@@ -0,0 +1,533 @@
...
+                })
+            }
+        }
+
+        #rejection
+    })
+}
+
+fn rejection_ident(ident: &syn::Ident) -> syn::Type {
+    let ident = quote::format_ident!("{}Rejection", ident);
```
jplatte, line 81: It's used a few times, why not `use quote::format_ident;`?

axum-macros/src/from_request.rs
```diff
@@ -0,0 +1,533 @@
...
+                together with `#[from_request(...)]` on the container",
+            ));
+        }
+    }
+
+    let path_span = path.span();
+
+    Ok(quote_spanned! {path_span=>
+        #[::axum::async_trait]
+        impl<B> ::axum::extract::FromRequest<B> for #ident
```
jplatte, line 423: Missing `#[automatically_derived]`.

jplatte (commented):
A few more comments about the main macro code, haven't read docs or looked closely at tests (not sure whether I need to).

davidpdrsn:
@jplatte I would appreciate that you take a look at the docs 😊 It outlines the supported use cases and perhaps you notice something that could be improved. I don't think you need to look at the tests.

axum-macros/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+/// #[derive(Clone)]
+/// struct State {
+///     // ...
+/// }
+/// ```
+///
+/// The rejection's `std::error::Error::source` implementation returns the inner rejection. This
+/// can be used to access source errors for example to customize rejection responses. Note this
+/// means the inner rejection types must themselves implement `std::error::Error`. All extractors
+/// in axum does this.
```
davidpdrsn, line 200: I'm thinking if this will turn out to be annoying in practice. It means extractors that use `(StatusCode, String)` or `Response` as rejection will not work, since they don't implement `Error`. Perhaps the `Error`, `Display`, and `Debug` impls should be opt-in and disabled by default 🤔

axum-macros/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+///
+/// #[derive(Clone)]
+/// struct State {
+///     // ...
+/// }
+///
+/// async fn handler(extractor: MyExtractor) {}
+/// ```
+///
+/// This requires that each field is an extractor (ie implements [`FromRequest`]).
```
jplatte, line 84:
```suggestion
/// This requires that each field is an extractor (i.e. implements [`FromRequest`]).
```

axum-macros/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+/// #[derive(Clone)]
+/// struct State {
+///     // ...
+/// }
+///
+/// async fn handler(extractor: MyExtractor) {}
+/// ```
+///
+/// Note this requires the via extractor to be a generic tuple struct that implements `FromRequest`
+/// and has exactly one public field:
```
jplatte, line 120: Maybe use the term 'newtype struct'?

```suggestion
/// Note this requires the via extractor to be a generic newtype struct (a tuple struct with
/// exactly one public field) that implements `FromRequest`:
```

axum-macros/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+/// pub struct ViaExtractor<T>(pub T);
+///
+/// // impl<T, B> FromRequest<B> for ViaExtractor<T> { ... }
+/// ```
+///
+/// More complex via extractors are not supported and require writing a manual implementation.
+///
+/// ## Optional fields
+///
+/// `#[from_request(via(...))]` supports `Option<_>` and `Result<_, _>` to make fields optional:
```
jplatte, line 132: Oohhh, now I get what the peel stuff is for. This is neat!

axum-macros/src/lib.rs
jplatte: I think it should be on by default. You could have an opt-out of `#[from_request(rejection_derive(!Display, !Error))]` for opting out (and allow deriving extra traits for the rejection type too 😉 – [prior art](https://github.com/ruma/ruma/blob/37095f88553b311e7a70adaaabe39976fb8ff71c/crates/ruma-common/src/directory.rs#L135)).

axum-macros/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+///
+/// async fn handler(state: State) {}
+/// ```
+///
+/// The rejection will be the "via extractors"'s rejection. For the previous example that would be
+/// [`axum::extract::rejection::ExtensionRejection`].
+///
+/// # Known limitations
+///
+/// Generics are not supported:
```
jplatte, line 226: \+ "currently"? 🙂

jplatte (approved):
Current code LGTM. The opt-out stuff could be a followup if it turns out to actually be needed. But I'm also okay with doing another review if it's going to be in this PR ^^

axum-macros/src/lib.rs
davidpdrsn: I like that!

davidpdrsn:
Yeah I'll do it in a follow up

PR #719, opened 2022-01-24
Move `middleware::from_fn` into axum

davidpdrsn:
I've been very happy with `axum_extra::middleware::from_fn` so for 0.5 I think we should move it into the main axum crate.

Also renamed the type from `MiddlewareFn` to `FromFn` to match std's `std::iter::FromFn`.

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -7,8 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- **added:** `axum_extra::middleware::from_fn` has been moved into the main axum crate. It is used
-  to easily crate a middleware from an async function ([#719])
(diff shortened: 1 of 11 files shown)


axum/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+- **added:** `axum_extra::middleware::from_fn` has been moved into the main axum crate. It is used
```
jplatte, line 10: For people who haven't seen it, "has been moved" might be confusing. I would suggest adding "from axum-extra" or just saying that it has been added.

davidpdrsn: Good point. I've rephrased it a bit.

jplatte (approved):
Lgtm! I like that the type name now matches the name of the function you use to construct it.

PR #722, opened 2022-01-25
Add sqlx example with migrations

FlakM:
Simple use case for sqlx based on tokio postgres example.
Sqlite database is created on execution in ./target directory
and migrations are then run against it.

Also I would like to thank you for creating such a wonderful library! Many kudos for awesome work!

## Motivation

I've got the chance to work with axum (<3!) and put this simple example for showcasing that using sqlx library is very straightforward. Example shows how to setup migrations and automatically create sqlite3 database if none exist at given path.
This example can be trivially used for example to add local backend for `async_cookies` used in sessions example.

## Solution

Add simple example of database creation, migrations execution and connection obtaining via extensions model.

### DIFF
--- a/examples/sqlx-sqlite/src/main.rs
+++ b/examples/sqlx-sqlite/src/main.rs
@@ -1,0 +1,119 @@
+//! Run with
+//!
+//! ```not_rust
+//! cargo run -p example-sqlx-sqlite
+//! ```
+//!
+//! Test with curl:
+//!
+//! ```not_rust
+//! curl 127.0.0.1:3000
+//! ```
+
+use axum::{
+    async_trait,
+    extract::{Extension, FromRequest, RequestParts},
+    http::StatusCode,
+    routing::get,
+    AddExtensionLayer, Router,
+};
+use sqlx::any::Any;
+use sqlx::migrate::MigrateDatabase;
+use sqlx::sqlite::SqlitePool;
+
+use std::net::SocketAddr;
+
+#[tokio::main]
+async fn main() {
+    // Set the RUST_LOG, if it hasn't been explicitly defined
+    if std::env::var_os("RUST_LOG").is_none() {
+        std::env::set_var("RUST_LOG", "example_tokio_postgres=debug")
+    }
+    tracing_subscriber::fmt::init();
+
+    let uri = std::env::var("DATABASE_URL")
+        .unwrap_or_else(|_| "sqlite://./target/db.sqlite3".to_string());
+
+    if !Any::database_exists(&uri).await.unwrap() {
+        Any::create_database(&uri).await.unwrap();
+        tracing::info!("created sqlite database at {}", uri);
+    }
+
+    let pool = SqlitePool::connect(&uri).await.unwrap();
+
+    // embeds migrations inside the binary
+    sqlx::migrate!().run(&pool).await.unwrap();
+
+    // build our application with some routes
+    let app = Router::new()
+        .route(
+            "/",
+            get(using_connection_pool_extractor).post(using_connection_extractor),
+        )
+        .layer(AddExtensionLayer::new(pool));
+
+    // run it with hyper
+    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
+    tracing::debug!("listening on {}", addr);
+    axum::Server::bind(&addr)
+        .serve(app.into_make_service())
+        .await
+        .unwrap();
+}
+
+// we can exact the connection pool with `Extension`
+async fn using_connection_pool_extractor(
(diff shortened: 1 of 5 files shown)


davidpdrsn:
Thanks!

However we already have an example using [tokio-postgres](https://github.com/tokio-rs/axum/blob/main/examples/tokio-postgres/src/main.rs) which is almost identical to this. So not sure this example adds much that isn't covered by other examples.

jplatte:
> So not sure this example adds much that isn't covered by other examples.

SQLx itself has horrible CI and the maintainers are still kind of slow to merge PRs, yet it is IMO the best SQL library for Rust. I think that for the overall ecosystem, having an example for axum + SQLx in an easy to discover location will be of great value 😄

examples/sqlx-sqlite/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+        .serve(app.into_make_service())
+        .await
+        .unwrap();
+}
+
+// we can exact the connection pool with `Extension`
+async fn using_connection_pool_extractor(
+    Extension(pool): Extension<SqlitePool>,
+) -> Result<String, (StatusCode, String)> {
+    let (result,) = sqlx::query_as("select username from users where id = $1")
```
jplatte, line 68:
```suggestion
    let result = sqlx::query_scalar("select username from users where id = $1")
```

examples/sqlx-sqlite/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        Ok(Self(conn))
+    }
+}
+
+async fn using_connection_extractor(
+    DatabaseConnection(conn): DatabaseConnection,
+) -> Result<String, (StatusCode, String)> {
+    let mut conn = conn;
+    let row: (String,) = sqlx::query_as("select username from users where id = $1")
```
jplatte, line 103:
```suggestion
    let row: String = sqlx::query_scalar("select username from users where id = $1")
```

jplatte (changes requested):

davidpdrsn:
Yeah that's true.

In that case I'm thinking if it makes sense to cut the migration stuff, as that's not axum specific. Doing a `select 1+1` should get the point across.

I would also prefer if the example used Postgres but that's less important I guess.

FlakM:
@davidpdrsn I thought that having runnable example without runtime dependencies would be more valuable, especially since there already is tokio-postgres example. I could include some docker snippet for setting pg up though and I will remove migrations if you prefer it this way :+1:

examples/sqlx-sqlite/src/main.rs
FlakM: oh thats nicer, thanks!

davidpdrsn:
I think connecting to the default postgres database similarly to how its done [here](https://github.com/tokio-rs/axum/blob/main/examples/tokio-postgres/src/main.rs#L29) would be best. Then there is one less thing for us to maintain/update.

I don't want to get docker involved in the examples. Thats way too much complexity.

jplatte:
What's the problem with the migrations? I've seen a bunch of people try to create their DB via `query` recently, clearly there is a need to show how it's actually done in practice.

FlakM:
You guys work fast :rocket: Give me a moment i will suggest something ;)

davidpdrsn:
> What's the problem with the migrations? I've seen a bunch of people try to create their DB via `query` recently¹, clearly there is a need to show how it's actually done in practice.

I don't think its our job to document how to handle migrations. I would rather users research the available options and pick the one that best suits their app.

jplatte:
But it's an example that shows how to use SQLx.. I know it's possible to use different crates to handle migrations together with SQLx, but using its builtin mirgations is the obvious way to go about it when using it for querying too.

davidpdrsn:
Its an example showing how to integrate sqlx with axum. I would rather keep the examples free to things like this because it forces the user to think about it, instead of just assume that our examples show best practices, which they don't.

I think a more complete example with migrations, docker, etc is more appropriate to link from [ECOSYSTEM.md](https://github.com/tokio-rs/axum/blob/main/ECOSYSTEM.md).

FlakM:
I've removed unwanted bits. If you change you mind I could leave some one off scripts to setup docker or add some functionality

examples/sqlx-sqlite/src/main.rs
FlakM: thanks thats nicer!

davidpdrsn (approved):
Thanks!

FlakM:
Commit massages and comments always lie!

PR #723, opened 2022-01-25
Match `ws-example` message type

mrxiaozhuox:
## Motivation

`websocket example` have not match the Message type.

## Solution

print different info for each type.


davidpdrsn:
Sorry for the bluntness, but why? How does this improve the example?

mrxiaozhuox:
the old example will print the `"Client says: MessageType::xxx()`, this is not good for explain the Message inside struct.

davidpdrsn (approved):
but wouldn't users have to look at the docs for `Message` no matter what and discover that its an enum?

I guess its fine. I just don't think it adds a whole lot to the example.

Issue #720, opened 2022-01-24
Merge macro crates


davidpdrsn:
Once https://github.com/tokio-rs/axum/pull/718 is merged I think we should move `axum_debug::debug_handler` in there and re-export it from axum-debug so its not a breaking change. Having just one macro crate is a bit nicer imo.

PR #724, opened 2022-01-25
Move axum-debug into axum-macros

davidpdrsn:
Turns out you can't actually re-export a proc-macro from another proc-macro crate. So the code still exists in axum-debug, but is marked as deprecated. My plan is to publish a new version of axum-debug after we've published axum-macros and then remove axum-debug from the repo.

This PR moved all the trybuild tests around a bit, hence the large number of changed files.

Fixes https://github.com/tokio-rs/axum/issues/720


axum-debug/README.md
```diff
@@ -7,6 +7,8 @@
 This is a debugging crate that provides better error messages for [`axum`]
 framework.
 
+Note this crate is deprecated. Use [axum-macros] instead.
```
jplatte, line 10: Maybe highlight this a bit?

```suggestion
**Note:** this crate is deprecated. Use [axum-macros] instead.
```

axum-debug/src/lib.rs
```diff
@@ -1,148 +1,19 @@
 //! This is a debugging crate that provides better error messages for [`axum`] framework.
 //!
-//! While using [`axum`], you can get long error messages for simple mistakes. For example:
+//! Note this crate is deprecated. Use [axum-macros] instead.
```
jplatte, line 3:
```suggestion
//! **Note:** this crate is deprecated. Use [axum-macros] instead.
```

axum/CHANGELOG.md
```diff
@@ -364,7 +364,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # 0.2.8 (07. October, 2021)
 
-- Document debugging handler type errors with "axum-debug" ([#372])
+- Document debugging handler type errors with "axum-macros" ([#372])
```
jplatte, line 367: Do you really want to rewrite the changelog of 0.2.8?

davidpdrsn: whoops! Nope that shouldn't change.

jplatte (approved):

Issue #726, opened 2022-01-25
Support opt-out of extra derived traits for rejections for `#[derive(FromRequest)]`


davidpdrsn:
See https://github.com/tokio-rs/axum/pull/718#discussion_r791713993

Issue #725, opened 2022-01-25
`#[derive(FromRequest)]` doesn't work on empty structs


davidpdrsn:
This fails

```rust
use axum_macros::FromRequest;

#[derive(FromRequest)]
struct Extractor {}
```

with

```
error[E0004]: non-exhaustive patterns: type `&ExtractorRejection` is non-empty
 --> tests/from_request/pass/named_empty.rs:4:10
  |
4 | #[derive(FromRequest)]
  |          ^^^^^^^^^^^ `ExtractorRejection` defined here
  |
  = help: ensure that all possible cases are being handled, possibly by adding wildcards or more match arms
  = note: the matched value is of type `&ExtractorRejection`
  = note: references are always considered inhabited
  = note: this error originates in the derive macro `FromRequest` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Doesn't work on

```rust
#[derive(FromRequest)]
struct Extractor();
```

either.

jplatte:
Interesting error message. Why is the rejection enum non-empty?

davidpdrsn:
I think rust gets confused because `self` is a reference and we do `match self {}`. I ran into it while building it but apparently didn't catch this case.

Kinda thinking we should special case empty and unit structs and use `Infallible` as the rejection.

jplatte:
> Kinda thinking we should special case empty and unit structs and use `Infallible` as the rejection.

Sounds good to me 👍🏼

PR #729, opened 2022-01-28
Support opt-out of extra derived traits for rejections for `#[derive(FromRequest)]`

davidpdrsn:
Makes it possible to write

```rust
#[derive(FromRequest)]
#[from_request(rejection_derive(!Display, !Error))]
struct Extractor {
    other: OtherExtractor,
}
```

Which opts out of the rejection implementing `Display` and `Error`.

The allowed syntax inside `rejection_derive` is intentionally quite limited. Can always be expanded in the future if we needed.

Fixes https://github.com/tokio-rs/axum/issues/726
Fixes https://github.com/tokio-rs/axum/issues/725


axum-macros/src/from_request/attr.rs
```diff
@@ -0,0 +1,202 @@
...
+        if lh.peek(kw::via) {
+            let via = input.parse::<kw::via>()?;
+            let content;
+            syn::parenthesized!(content in input);
+            content.parse().map(|path| Self::Via { via, path })
+        } else {
+            Err(lh.error())
+        }
+    }
+}
```
davidpdrsn, line 150: The duplication in this parsing code bothers me a little but I couldn't find a nicer solution that gave good error messages.

axum-macros/src/from_request.rs
```diff
@@ -270,7 +288,7 @@ fn extract_each_field_rejection(
         }
     };
 
-    let impl_display = {
+    let impl_display = if rejection_derive_opt_outs.derive_display() {
```
jplatte, line 291: Could use `bool::then` (plus `.transpose()?`) here.

axum-macros/src/from_request/attr.rs
jplatte: Seems fine to me as-is.

axum-macros/src/from_request/attr.rs
```diff
@@ -0,0 +1,202 @@
...
+        let mut debug = None;
+        let mut display = None;
+        let mut error = None;
+
+        while !input.is_empty() {
+            input.parse::<Token![!]>()?;
+
+            let lh = input.lookahead1();
+            if lh.peek(kw::Debug) {
+                debug = Some(input.parse()?);
```
jplatte, line 184: Should probably raise an error on multiple assignment here too? Could avoid the extra nesting by moving the assignment or error into the helper function.

jplatte (approved):

axum-macros/src/from_request.rs
davidpdrsn: Hm maybe. I think its fine as is.

davidpdrsn:
I fixed a few other things I discovered:

- Added docs for `rejection_derive`
- Multiple opt outs of the same trait like you mentioned
- Using both `via` and `rejection_derive`. Since `via` on the container changes it to use the "via extractors"'s rejection, and doesn't generate a few type, you cannot opt out of which traits are derived for it
- Better error message if opting out of `Debug` or `Display` with also opting out of `Error`. `std::error::Error` requires `Debug` and `Display` so the error message was slightly confusing

axum-macros/src/from_request.rs
```diff
@@ -313,11 +333,19 @@ fn extract_each_field_rejection(
...
+        })
+    } else {
+        None
+    };
+
+    let impl_debug = if rejection_derive_opt_outs.derive_debug() {
+        Some(quote! { #[derive(Debug)] })
+    } else {
+        None
     };
```
jplatte, line 345: Here you don't even need `.transpose()?` 😉

```suggestion
    let impl_debug = rejection_derive_opt_outs.derive_debug().then(|| {
        quote! { #[derive(Debug)] }
    });
```

But of course the current code is fine too.

jplatte (approved):

axum-macros/src/from_request.rs
davidpdrsn: Yeah thats definitely nicer!

Issue #731, opened 2022-01-30
Missing allow header in 405 Method Not Allowed replies


mdickopp:
### Version

axum v0.4.4
axum-core v0.1.1

### Platform

Debian Linux 12 (“bookworm”)
Linux feynman 5.15.0-3-amd64 <span>#</span>1 SMP Debian 5.15.15-1 (2022-01-18) x86_64 GNU/Linux
rustc 1.58.1 (db9d1b20b 2022-01-20)

### Description

When the server sends a reply with a 405 Method Not Allowed status, it should include an `allow` header, but doesn't.

#### Steps to reproduce

Run the axum examle program:

```rust
use axum::{routing::get, Router};

#[tokio::main]
async fn main() {
    let app = Router::new().route("/", get(|| async { "Hello, World!" }));
    axum::Server::bind(&"0.0.0.0:3000".parse().unwrap())
        .serve(app.into_make_service())
        .await
        .unwrap();
}
```

Send a HTTP/1.1 `DELETE` request for `/`. I used the the following command on the Linux command line:

```sh
printf 'DELETE / HTTP/1.1\r\n\r\n' | nc 127.0.0.1 3000
```

#### Observed behavior

The server replies with a 405 Method Not Allowed status (as expected), but does not send an `allow` header.

#### Expected behavior

The output contains the line:

```
allow: GET, HEAD
```

Web servers are expected to include an `allow` reply header when they send a 405 status reply (see RFC 7231, section 6.5.5).

davidpdrsn:
I wasn't aware of this. Thanks for reporting! Should be easy to fix since [`MethodRouter` knows exactly which methods it has routes for](https://github.com/tokio-rs/axum/blob/main/axum/src/routing/method_routing.rs#L477-L488).

mdickopp:
Thank you for fixing this so quickly. Works great for me!

PR #733, opened 2022-01-30
Set `Allow` header when returning `405 Method Not Allowed`

davidpdrsn:
This changes `MethodRouter` so it sets the `Allow` header when returning `405 Method Not Allowed`.

The header value is built up as a `BytesMut` to avoid allocations when creating the response.

The header is also set when the fallback is hit, unless the response generated by the fallback already contains the header. This should result in it doing the thing most of the time while allowing users to override it if they handle additional methods in the fallback.

Fixes https://github.com/tokio-rs/axum/issues/731


axum/src/docs/method_routing/fallback.md
```diff
@@ -51,3 +51,12 @@ async fn fallback_two() -> impl IntoResponse {}
...
 # };
 ```
+
+## Setting the `Allow` header
+
+By default `MethodRouter` will set the `Allow` header when returning `405 Method
+Not Allowed`. This is also done when the fallback is used unless the response
+generated by the fallback already sets the `Allow` header.
+
+This means if you use `fallback` to accept additional methods make sure you set
```
jplatte, line 61:
```suggestion
This means if you use `fallback` to accept additional methods, you should make sure you set
```

jplatte (approved):

axum/src/docs/method_routing/fallback.md
davidpdrsn: 84df7eaa38aa3 👍

PR #735, opened 2022-01-31
fix crates.io links

programatik29:
## Motivation

Some links in [crates.io](https://crates.io/crates/axum) page aren't working.

## Solution

Use absolute url for broken links instead of paths.


davidpdrsn (approved):
Thanks!

PR #736, opened 2022-01-31
update matchit to 0.4.6

ibraheemdev:


jplatte:
This only raises the minimum version, `cargo update` would get you the same version either way when using axum. What's the point?

davidpdrsn (approved):
Thanks!

davidpdrsn:
> Oh is this a followup to the release with a higher MSRV? Why not just yank that one then?

Exactly. I don't think a yank is necessary as this accomplishes the same thing 🤷

PR #751, opened 2022-02-08
Fix rejection status codes

davidpdrsn:
I think these status codes are more appropriate.


jplatte (approved):
I wasn't aware of the unprocessable entity status code, but after looking it up it seems appropriate.

PR #753, opened 2022-02-10
Update dependencies

davidpdrsn:
axum depended on an old version of sha-1 which could result in multiple versions being pulled in. Its an internal dependency so is safe to update.

Also updated dependencies in the examples while I was at it.

Also gonna backport this patch.


davidpdrsn:
Would like to avoid more duplicate deps so working with [tungstenite](https://github.com/snapview/tungstenite-rs/issues/262) and headers maintainers to get new releases.

jplatte (approved):

jplatte:
I guess maybe the reference should be to the COPYRIGHT file which includes the reason for `AND BSD-3-Clause`: https://github.com/hsivonen/encoding_rs/blob/master/COPYRIGHT

davidpdrsn:
The license checks are only warnings though so CI fails because of duplicate deps.

PR #755, opened 2022-02-12
Fix setting `content-length: 0` for HEAD responses with compression middleware

### LINKED ISSUE #747: Incorrect `content-length` header in reply to HEAD request when chunked transfer enconding is used
## Bug Report

This is a followup to #730. Unfortunately, I found a case in which it still does not work correctly.

### Version

axum main branch, commit 16ee83a

### Platform

Debian Linux 12 (“bookworm”)
Linux feynman 5.15.0-3-amd64 <span>#</span>1 SMP Debian 5.15.15-1 (2022-01-18) x86_64 GNU/Linux
rustc 1.58.1 (db9d1b20b 2022-01-20)

### Description

The `content-length` header in the reply to a simple `HEAD` request is set to 0 if the corresponding `GET` request uses `transfer-encoding: chunked`.

#### Steps to reproduce

Run the example program (which uses tower-http v0.2.1):

```rust
use tower_http::compression::{predicate::SizeAbove, CompressionLayer};

#[tokio::main]
async fn main() {
    let app = Router::new()
        .route("/", get(|| async { "Hello, World!" }))
        .layer(CompressionLayer::new().compress_when(SizeAbove::new(0)));
        .await
        .unwrap();
}
```

Send a HTTP/1.1 `HEAD` request for /. Set the `accept-encoding` request header to `gzip`. I used the the following command on the Linux command line:

```sh
printf 'HEAD / HTTP/1.1\r\naccept-encoding: gzip\r\n\r\n' | nc 127.0.0.1 3000
```

#### Observed behavior

The output contains the line:

```
content-length: 0
```

#### Expected behavior

The output contains no `content-length` header.

Note the response to the corresponding `GET` request uses chunked transfer encoding. Its headers look like this:

```
HTTP/1.1 200 OK
content-encoding: gzip
transfer-encoding: chunked
date: Sun, 06 Feb 2022 16:03:44 GMT
```

davidpdrsn:
Fixes #747

The bug was that applying a middleware to a method router would cause the response to pass through a `RouteFuture` twice and we'd set the `content-length` and strip the response body twice. That happened because `Router::layer` wrapped things in a `Route` then called `MethodRoute::route` which also applied a `Route`.

Solved it by having `RouteFuture::poll` set a response extension and not alter the response if that is present. Its a little unfortunate to add more overhead here but couldn't think of another solution since `Router` and `MethodRouter` need to be correct and make sense on their own, and needs to support general services that might not handle `HEAD` requests correctly.

I also removed `RouterFuture` so `Router`'s response is now `RouteFuture<RequestBody, Infallible>`. The additional indirection of `RouterFuture` hasn't proven to be useful so I think its safe to remove.

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -54,6 +54,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **fixed:** Set `Allow` header when responding with `405 Method Not Allowed` ([#733])
 - **fixed:** Correctly set the `Content-Length` header for response to `HEAD`
   requests ([#734])
+- **fixed:** Fix wrong `content-length` for `HEAD` requests to endpoints that returns chunked
+  responses ([#755])
 
 [#644]: https://github.com/tokio-rs/axum/pull/644
 [#665]: https://github.com/tokio-rs/axum/pull/665
@@ -62,6 +64,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#719]: https://github.com/tokio-rs/axum/pull/719
 [#733]: https://github.com/tokio-rs/axum/pull/733
 [#734]: https://github.com/tokio-rs/axum/pull/734
+[#755]: https://github.com/tokio-rs/axum/pull/755
 
 # 0.4.4 (13. January, 2022)
 
--- a/axum/src/routing/future.rs
+++ b/axum/src/routing/future.rs
@@ -1,21 +1,3 @@
 //! Future types.
 
-use crate::response::Response;
-use std::convert::Infallible;
-
 pub use super::{into_make_service::IntoMakeServiceFuture, route::RouteFuture};
-
-opaque_future! {
-    /// Response future for [`Router`](super::Router).
-    pub type RouterFuture<B> = RouteFuture<B, Infallible>;
-}
-
-impl<B> RouterFuture<B> {
-    pub(super) fn from_future(future: RouteFuture<B, Infallible>) -> Self {
-        Self::new(future)
-    }
-
-    pub(super) fn from_response(response: Response) -> Self {
-        Self::new(RouteFuture::from_response(response))
-    }
-}
--- a/axum/src/routing/method_routing.rs
+++ b/axum/src/routing/method_routing.rs
@@ -983,7 +983,7 @@ where
             ) => {
                 if $method == Method::$method_variant {
                     if let Some(svc) = $svc {
-                        return RouteFuture::from_future(svc.0.clone().oneshot($req))
+                        return RouteFuture::from_future(svc.oneshot_inner($req))
                             .strip_body($method == Method::HEAD);
                     }
                 }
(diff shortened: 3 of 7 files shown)


jplatte:
I don't think I really understand what went wrong here. `content-length` setting should only happen if there is no `content-length` yet, right?

davidpdrsn:
I understand that its unclear, honestly it wasn't clear to me either. I've done some more digging and think I understand whats going on now:

If you have a router like this:

```rust
Router::new()
    .route("/", get(|| async { "Hello, World!" }))
    .layer(CompressionLayer::new().compress_when(SizeAbove::new(0)))
```

run it with hyper and send it `HEAD /` with `accept-encoding: gzip` the flow will be:

- `Router` receives the request, does its routing, and passes it to the matching route
- That route has a `Compression` middleware applied but it doesn't change the request, just checks for `accept-encoding`
- The `MethodRouter` is called which yields a `RouteFuture`
- This route future is wrapped in `Compression`'s response future
- That response future is wrapped in `RouteFuture` again when it is returned from `Router`
- So the future stack is `RouteFuture<CompressionFuture<RouteFuture<Handler>>>`. Of course these types kind actually exist like that but hopefully you understand.
- The handler future yields a response as normal
- The first, inner most, `RouteFuture` sees that the request was a `HEAD` so it sets the content-length (since the body returned by the handler had an exact known size) and then sets the body to `Empty`
- The compression future removes the `content-length` (since you cannot have `transfer-encoding: chunked` + `content-length`)
- Now the response arrives at the outer most `RouteFuture` which first tries to sent the content-length but can't since the body doesn't have a know size. [That is `CompressionBody` doesn't impl `size_hint`.](https://github.com/tower-rs/tower-http/blob/master/tower-http/src/compression/body.rs#L193-L237) However this future will set the body to `Empty` again.
- hyper will see that there is no content-length and check if the body has a exact size, which `Empty` does and so hyper sets the content-length to 0.

I think the most correct thing is to omit the content-length since if the request is issued again with `GET` then it wont have a content-length, because it'll have `transfer-encoding: chunked`, since its compressed on the fly.

Phew that was a lot 😅 Hopefully it makes sense.

jplatte:
I see two bugs there:

* The compression layer should not do anything in this case since the body is actually empty. Maybe ignore an already-set content-length header when the method is HEAD / OPTIONS, or don't do anything on those methods even without the `compress_when` option.
* Hyper shouldn't be setting content-length when encountering `transfer-encoding: chunked`, if that combination is illegal.

davidpdrsn:
If you send a HEAD request the response is

```
HTTP/1.1 200 OK
content-encoding: gzip
date: Sun, 13 Feb 2022 08:40:12 GMT
```

with no body. Before this patch this response would have `content-length: 0` which is wrong since changing the method to `GET` wont return an empty response.

With this patch the response to `GET` is

```
HTTP/1.1 200 OK
content-encoding: gzip
transfer-encoding: chunked
date: Sun, 13 Feb 2022 08:41:14 GMT

21
��H�����/�IQ��J�
0
```

This is unchanged by this patch

jplatte:
> The compression middleware was configured with `.compress_when(SizeAbove::new(0))` which is a bit silly in the first place.

But `0` isn't above `0` so it should still not compress.. Is `SizeAbove` really `SizeGreaterOrEqual`?

> By default it wont compress empty bodies. But I agree its probably better that compression doesn't do anything on `HEAD` requests no matter what. I think thats a separate bug though.

> Hyper does the right thing afaik. Hyper calls `http_body::Body::size_hint` and sees an extract of 0 and sets `content-length: 0`. There is no `transfer-encoding` in the response. Compression actually doesn't set `transfer-encoding: chunked`, it leaves that up to hyper as hyper does the actual encoding of the body into the stream.

Okay, when and why does `transfer-encoding: chunked` get set then?

jplatte:
> I think thats a separate bug though.

Can you think of other reasonable middlewares / overall setups that would cause the same issue?

davidpdrsn:
> But 0 isn't above 0 so it should still not compress.. Is SizeAbove really SizeGreaterOrEqual?

That's true. I think that's a bug in the implementation. The docs says "above a certain size" so I don't think it's a breaking change to fix.

> Can you think of other reasonable middlewares / overall setups that would cause the same issue?

[`Decompression`](https://docs.rs/tower-http/latest/tower_http/decompression/index.html) probably does.

PR #756, opened 2022-02-13
WIP: Type safe routing

davidpdrsn:
Some things thats been bugging me about axum's routing for the longest time:

- There is no connection between the route you use in `Router::router` and the path you try and extract with `Path`. Requires you to make changes in multiple places.
- `Path` might try and extract captures that don't exist on the route
- By looking at just the handler function you don't know what the route will be. You have to find the places where its added to a router.

Attributes like `#[get("/users/:id")]` can fix that but they require some pretty gnarly parsing of the function arguments to make sure things are aligned. RA also isn't great with attributes so I'd like to avoid them.

I've been experimenting with another approach that I think is interesting:

```rust
use axum::{response::IntoResponse, Router};
use axum_extra::routing::{typed_path, RouterExt};
use axum_macros::TypedPath;
use serde::Deserialize;

// new `TypedPath` trait in axum_extra:
//
// pub trait TypedPath {
//     // the path with captures such as `/users/:id`
//     const PATH: &'static str;
//
//     // returns an actual request path with parameters interpolated such as `/users/1`
//     // that way you can create type safe links to routes that are known to exist
//     fn path(&self) -> Cow<'static, str>;
// }

#[tokio::main]
async fn main() {
    let app = Router::new()
        // `typed_path::get` requires a handler who's first argument implements `TypedPath`
        // so it can acces it's `PATH` const
        .with(typed_path::get(users_index).post(users_create))
        .with(typed_path::get(users_show));

        .await
        .unwrap();
}

// `#[derive(TypedPath)]` derives `TypedPath` and `FromRequest`
#[derive(TypedPath)]
#[typed_path("/users")]
struct UsersCollection;

// for paths with fields `TypedPath` implements `FromRequest` via `Path`
// which requires `Deserialize`
#[derive(Deserialize, TypedPath)]
#[typed_path("/users/:id")]
struct UsersMember {
    id: u32,
}

// regular handlers that take a `TypedPath` as their first argument
async fn users_index(_: UsersCollection) -> impl IntoResponse {
    "users#index"
}

async fn users_create(_: UsersCollection, _payload: String) -> impl IntoResponse {
    "users#create"
}

async fn users_show(UsersMember { id }: UsersMember) -> impl IntoResponse {
    format!("users#show: {}", id)
}
```

Advantages:

- The only macro is a `#[derive(_)]` of a known trait which RA tends to work well with.
- No macros on the handlers so RA still works great there.
- You get typed connection between the route and the handler.

@jplatte @programatik29 what do you think about this?


axum-extra/src/routing/typed_path.rs
```diff
@@ -0,0 +1,64 @@
...
+impl<P, B> HasRoutes<B> for TypedPathRouter<P, B> {
+    fn routes(self) -> Router<B> {
+        self.router
+    }
+}
+
+pub trait FirstElementIs<P> {}
+impl<P> FirstElementIs<P> for (P,) {}
+impl<P, T1> FirstElementIs<P> for (P, T1) {}
+impl<P, T1, T2> FirstElementIs<P> for (P, T1, T2) {}
```
jplatte, line 64: Nice! I think I might borrow this idea for the type safe routing I've recently implemented in [Conduit](https://gitlab.com/famedly/conduit) if and when I end up rewriting that to not have its own `Handler` trait (what I implemented relies on much more codegen though and isn't really suited for axum overall).

axum-extra/src/routing/typed_path.rs
```diff
@@ -0,0 +1,64 @@
...
+    where
+        H: Handler<T, B>,
+        T: FirstElementIs<P> + 'static,
+    {
+        self.router = self.router.route(P::PATH, routing::post(handler));
+        self
+    }
+}
+
+impl<P, B> HasRoutes<B> for TypedPathRouter<P, B> {
```
jplatte, line 55: I forgot that `HasRoutes` is a thing. I'm wondering whether I should just use that and `axum_extra`s `RouterExt` instead of having a new extension trait in Conduit 😄

axum-macros/src/lib.rs
```diff
@@ -385,6 +386,12 @@ pub fn debug_handler(_attr: TokenStream, input: TokenStream) -> TokenStream {
     return expand_attr_with(_attr, input, debug_handler::expand);
 }
 
+/// TODO
+#[proc_macro_derive(TypedPath, attributes(typed_path))]
+pub fn derive_typed_path(input: TokenStream) -> TokenStream {
```
jplatte, line 391: Hm, does it really make sense for this to be in `axum-macros`? I think it would be more sensible to create a separate proc-macro crate for it and have all the new functionality plus a re-export of this macro in `axum-extra` behind a feature flag.

programatik29:
I think this is a nice addition. Empowers users to inverse the logic so that instead of routes having handlers, now handlers have routes.

davidpdrsn:
Realized that by adding `typed_*` methods to `RouterExt` we can avoid having the `TypedPathRouter` entirely:

```rust
let app = Router::new()
    .typed_get(users_index)
    .typed_post(users_create)
    .typed_get(users_show)
    .typed_get(users_edit);
```

I think thats way nicer because you only have to use and learn `Router` and can add middleware/fallbacks in the same way.

axum-extra/src/routing/typed_path.rs
davidpdrsn: Yeah I think `HasRoutes` is pretty neat. Kinda wanna move it into axum and make `Router::merge` use it. Then we wouldn't need `RouterExt::with`

axum-macros/src/lib.rs
davidpdrsn: I would like to not have too many `axum-*` crates and it doesn't add any new deps. However re-exporting it from axum-extra is a good idea, as an optional feature.

davidpdrsn:
This is ready for review now :)

axum-extra/Cargo.toml
```diff
@@ -11,7 +11,9 @@ repository = "https://github.com/tokio-rs/axum"
 version = "0.1.2"
 
 [features]
-erased-json = ["serde", "serde_json"]
+default = []
```
jplatte, line 14: Why declare this explicitly? Isn't that a no-op?

axum-extra/src/routing/mod.rs
```diff
@@ -44,6 +167,94 @@ where
     {
         self.merge(routes.routes())
     }
+
+    #[cfg(feature = "typed-routing")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "typed-routing")))]
```
jplatte, line 172: If you activate `doc_auto_cfg` instead of `doc_cfg`, `doc(cfg)` attributes are implied and you only need to write them out if you want the docs show some other cfg expression.

axum-extra/src/routing/typed.rs
```diff
@@ -0,0 +1,171 @@
...
+impl_first_element_is!(T1, T2, T3, T4, T5, T6, T7);
+impl_first_element_is!(T1, T2, T3, T4, T5, T6, T7, T8);
+impl_first_element_is!(T1, T2, T3, T4, T5, T6, T7, T8, T9);
+impl_first_element_is!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10);
+impl_first_element_is!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11);
+impl_first_element_is!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12);
+impl_first_element_is!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13);
+impl_first_element_is!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14);
+impl_first_element_is!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15);
+impl_first_element_is!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16);
```
jplatte, line 171: No `all_tuples!`?

axum-extra/src/routing/typed.rs
```diff
@@ -0,0 +1,171 @@
...
+///
+/// async fn users_destroy(_: UsersCollection) { /* ... */ }
+///
+/// #
+/// # let app: Router<axum::body::Body> = app;
+/// ```
+///
+/// # Using `#[derive(TypedPath)]`
+///
+/// While `TypedPath` can be implemented manually its _highly_ recommended to derive it:
```
jplatte, line 67:
```suggestion
/// While `TypedPath` can be implemented manually, it's _highly_ recommended to derive it:
```

axum-extra/src/routing/typed.rs
```diff
@@ -0,0 +1,171 @@
...
+/// ```
+///
+/// The macro expands to:
+///
+/// - A `TypedPath` implementation.
+/// - A [`Display`] implementation that interpolates the captures. This can be used to, among other
+/// things, create links to known paths and have them verified statically.
+/// - A [`FromRequest`] implementation compatible with [`RouterExt::typed_get`],
+/// [`RouterExt::typed_post`], etc. This implementation uses [`Path`] and thus your struct must
+/// also implement [`serde::Deserialize`], unless its a unit struct.
```
jplatte, line 87:
```suggestion
/// also implement [`serde::Deserialize`], unless it's a unit struct.
```

axum-macros/src/typed_path.rs
```diff
@@ -0,0 +1,279 @@
...
+        ident,
+        generics,
+        fields,
+        ..
+    } = &item_struct;
+
+    if !generics.params.is_empty() || generics.where_clause.is_some() {
+        return Err(syn::Error::new_spanned(
+            generics,
+            "`#[derive(TypePath)]` doesn't support generics",
```
jplatte, line 17:
```suggestion
            "`#[derive(TypedPath)]` doesn't support generics",
```

axum-extra/src/routing/typed.rs
```diff
@@ -0,0 +1,171 @@
...
+/// struct UsersMember {
+///     id: u32,
+/// }
+/// ```
+///
+/// The macro expands to:
+///
+/// - A `TypedPath` implementation.
+/// - A [`Display`] implementation that interpolates the captures. This can be used to, among other
+/// things, create links to known paths and have them verified statically.
```
jplatte, line 84: I think it should be noted somewhere that the types of the fields of the struct need to implement `Display`, and that `Display` and `Deserialize` for those types need to agree on the string representation. (maybe `Serialize` should be used instead of `Display` internally?)

axum-macros/src/typed_path.rs
```diff
@@ -0,0 +1,279 @@
...
+        .collect()
+}
+
+enum Segment {
+    Capture(String, Span),
+    Static(String),
+}
+
+#[test]
+fn ui() {
```
jplatte, line 267: I think it would be nice to ignore this test on non-stable, but still allow it to be run using `cargo test -- --ignored` / `cargo test -- --include-ignored`:

```suggestion
#[test]
#[rustversion::attr(not(stable), ignore)]
fn ui() {
```

Then you shouldn't need the rustversion attributes inside the function.

axum-macros/tests/typed_path/fail/missing_capture.stderr
```diff
@@ -0,0 +1,14 @@
+error[E0027]: pattern does not mention field `id`
+ --> tests/typed_path/fail/missing_capture.rs:5:14
+  |
+5 | #[typed_path("/users")]
+  |              ^^^^^^^^ missing field `id`
+  |
+help: include the missing field in the pattern
+  |
+5 | #[typed_path("/users" { id })]
+  |                       ++++++
```
jplatte, line 10: This is a weird error message. I thought this case would be caught by the proc-macro directly instead of later by the compiler trying to instantiate `MyPath`.

axum-macros/tests/typed_path/fail/not_deserialize.stderr
```diff
@@ -0,0 +1,8 @@
+error[E0277]: the trait bound `for<'de> MyPath: serde::de::Deserialize<'de>` is not satisfied
+ --> tests/typed_path/fail/not_deserialize.rs:4:14
+  |
+4 | #[typed_path("/users/:id")]
+  |              ^^^^^^^^^^^^ the trait `for<'de> serde::de::Deserialize<'de>` is not implemented for `MyPath`
+  |
+  = note: required because of the requirements on the impl of `serde::de::DeserializeOwned` for `MyPath`
+  = note: required because of the requirements on the impl of `FromRequest<B>` for `axum::extract::Path<MyPath>`
```
jplatte, line 8: I think it would make a little more sense for the span here to be the call-site span (point to the derive, not the path inside the `typed_path` attribute).

axum-macros/tests/typed_path/fail/unit_with_capture.stderr
```diff
@@ -0,0 +1,5 @@
+error: Typed paths for unit structs cannot contain captures
+ --> tests/typed_path/fail/unit_with_capture.rs:5:14
+  |
+5 | #[typed_path("/users/:id")]
+  |              ^^^^^^^^^^^^
```
jplatte, line 5: Same here.

axum-macros/tests/typed_path/pass/named_fields_struct.rs
```diff
@@ -0,0 +1,25 @@
...
+
+    assert_eq!(MyPath::PATH, "/users/:user_id/teams/:team_id");
+    assert_eq!(
+        format!(
+            "{}",
+            MyPath {
+                user_id: 1,
+                team_id: 2
+            }
+        ),
```
jplatte, line 22:
```suggestion
        MyPath {
            user_id: 1,
            team_id: 2
        }.to_string(),
```

axum-macros/tests/typed_path/pass/tuple_struct.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#[derive(TypedPath, Deserialize)]
+#[typed_path("/users/:user_id/teams/:team_id")]
+struct MyPath(u32, u32);
+
+fn main() {
+    axum::Router::<axum::body::Body>::new().route("/", axum::routing::get(|_: MyPath| async {}));
+
+    assert_eq!(MyPath::PATH, "/users/:user_id/teams/:team_id");
+    assert_eq!(format!("{}", MyPath(1, 2)), "/users/1/teams/2");
```
jplatte, line 12:
```suggestion
    assert_eq!(MyPath(1, 2).to_string(), "/users/1/teams/2");
```

axum-macros/tests/typed_path/pass/unit_struct.rs
```diff
@@ -0,0 +1,13 @@
...
+#[derive(TypedPath)]
+#[typed_path("/users")]
+struct MyPath;
+
+fn main() {
+    axum::Router::<axum::body::Body>::new()
+        .route("/", axum::routing::get(|_: MyPath| async {}));
+
+    assert_eq!(MyPath::PATH, "/users");
+    assert_eq!(format!("{}", MyPath), "/users");
```
jplatte, line 12:
```suggestion
    assert_eq!(MyPath.to_string(), "/users");
```

axum/src/routing/mod.rs
```diff
@@ -88,6 +87,17 @@ where
     }
 }
 
+impl<B> fmt::Debug for Router<B> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Router")
```
jplatte, line 92: Could consider including `type_name::<B>()` here.

jplatte (commented):
One interesting issue I see with this is nested routers. Nesting a router that uses the new typed routes under a route that has its own captures wouldn't work.

axum-extra/src/routing/mod.rs
davidpdrsn: How do you do that?

I tried setting

```
#![cfg_attr(docsrs, feature(doc_auto_cfg, doc_cfg))]
```

and doing the standard cfg feature

```
#[cfg(feature = "typed-routing")]
pub use axum_macros::TypedPath;
```

and building the docs with

```
RUSTFLAGS="--cfg docsrs" cargo +nightly doc --no-deps
```

But the labels don't show up.

davidpdrsn: maybe one has to use `RUSTDOCFLAGS` instead :think

jplatte: We do basically that (though `doc_cfg` isn't used) in Ruma and it works fine. Can try in axum myself later today.

davidpdrsn: Yep that did it. 8ddf49501fff6

I'll update the rest of axum in a separate PR.

axum-extra/src/routing/typed.rs
jplatte: Maybe you should publish it as its own crate 😄

axum-extra/src/routing/typed.rs
davidpdrsn: Thats a good point! I'll add a note about that.

Using `Serialize` instead is interesting 🤔 I guess it would need to be serializer that only supports simple types like strings, numbers, and bools. Things that make sense in path segments and aren't nested.

davidpdrsn: Added in feb86f6bb367f

axum-extra/src/routing/typed.rs
```diff
@@ -80,11 +80,13 @@ use super::sealed::Sealed;
...
 ///
 /// - A `TypedPath` implementation.
-/// - A [`Display`] implementation that interpolates the captures. This can be used to, among other
-/// things, create links to known paths and have them verified statically.
 /// - A [`FromRequest`] implementation compatible with [`RouterExt::typed_get`],
 /// [`RouterExt::typed_post`], etc. This implementation uses [`Path`] and thus your struct must
 /// also implement [`serde::Deserialize`], unless it's a unit struct.
+/// - A [`Display`] implementation that interpolates the captures. This can be used to, among other
+/// things, create links to known paths and have them verified statically. Note that the
+/// [`Display`] implementation for each field must return something that's compatible with it's
```
jplatte, line 88:
```suggestion
/// [`Display`] implementation for each field must return something that's compatible with its
```

axum-macros/src/typed_path.rs
davidpdrsn: I think that makes sense but for some reason if I do

```rust
#[test]
#[rustversion::attr(not(nightly), ignore)]
fn ui() {
    let t = trybuild::TestCases::new();
    t.compile_fail("tests/typed_path/fail/*.rs");
    t.pass("tests/typed_path/pass/*.rs");
}
```

and then `cargo +nightly t -p axum-macros typed_path` the tests still runs. Don't see an issue on rustversion about this so not sure whats up.

axum-macros/tests/typed_path/fail/not_deserialize.stderr
davidpdrsn: Fixed!

axum-extra/src/routing/typed.rs
davidpdrsn: haha omg one day I really should learn english 😂

axum-macros/src/typed_path.rs
jplatte: Yeah, it needs to be `not(stable)` instead of `not(nightly)` 😄

davidpdrsn:
Tell me if this is too much but I had a thought this morning. Why only make the paths types, why not the http methods as well? Basically:

```rust
use axum::{response::IntoResponse, Router};
use axum_extra::routing::{Any, Get, OneOf, Post, RouterExt, TypedPath};
use serde::Deserialize;

#[tokio::main]
async fn main() {
    let app = Router::new()
        .typed_route(users_index)
        .typed_route(users_create)
        .typed_route(users_show)
        .typed_route(users_edit);

        .await
        .unwrap();
}

#[derive(TypedPath)]
#[typed_path("/users")]
struct UsersCollection;

#[derive(Deserialize, TypedPath)]
#[typed_path("/users/:id")]
struct UsersMember {
    id: u32,
}

#[derive(Deserialize, TypedPath)]
#[typed_path("/users/:id/edit")]
struct UsersEdit(u32);

async fn users_index(_: Get, _: UsersCollection) -> impl IntoResponse {
    "users#index"
}

async fn users_create(_: Post, _: UsersCollection, _payload: String) -> impl IntoResponse {
    "users#create"
}

async fn users_show(_: Any, UsersMember { id }: UsersMember) -> impl IntoResponse {
    format!("users#show: {}", id)
}

async fn users_edit(_: OneOf<(Get, Post)>, UsersEdit(id): UsersEdit) -> impl IntoResponse {
    format!("users#edit: {}", id)
}
```

I think this is pretty interesting but its probably also gonna be a little foreign to some people. Too much advanced type trickery?

A hunch I have is that it might be convenient for openapi spec generation since more of the data is in the handler itself, but thats just a hunch.

SpyrosRoum:
I think on a user's perspective it the signatures might start getting a bit too noise with both the http method and the typed path.
Without knowing much on the internals would it be possible to combine the two on something like a `Metadata` struct that would contain both the http method, the path, and any arguments from it?

niklasf:
Regarding the general idea: It looks like this approach could also enable typesafe reverse routing. Safely generating URLs that are guaranteed to be accepted by the router is a useful feature. Changed the path in some way? All links created using reverse routing are automatically kept in sync.

jplatte:
After thinking about it for the last days, I don't think the typed method stuff makes that much sense. In theory it could be nice for generating redirects but only with a decently more complex API (I really don't like the `_: Get` params).

I think typed paths are worthwhile because they not only solve the problem of parameter mismatches (by the way: what about `*rest` params?), but also the problem of forgetting to percent-encode parameters in redirects¹. The same doesn't seem to apply for methods.

¹ or even requests? there's some server-server APIs like Matrix federation where you could use the same path type for receiving and sending requests

jplatte:
> Oh I had forgotten about that. Currently the `Display` impl it generates doesn't URL the segments. Should it? 🤔

I think it should, yes. Why not?

axum-extra/src/routing/typed.rs
```diff
@@ -117,6 +117,28 @@ use super::sealed::Sealed;
 /// struct UsersMember(u32);
 /// ```
 ///
+/// ## Percent encoding
+///
+/// The generated [`Display`] implementation will automatically percent encoding parameters:
```
jplatte, line 122:
```suggestion
/// The generated [`Display`] implementation will automatically percent-encode parameters:
```

jplatte: Also should it be 'arguments' (or 'parameter values') instead of 'parameters'?

davidpdrsn: Yes thats true!

davidpdrsn: The way I remember it from my CS was "actual arguments" (both start with 'a') and "actual" meaning the actual values and then "format parameters" which are just the names without the values.

jplatte (approved):

PR #757, opened 2022-02-14
remove ecosystem project `Datafuse`,which already migrate to other web framework

zys864:


axum/Cargo.toml
```diff
@@ -40,7 +40,6 @@ serde = "1.0"
 serde_urlencoded = "0.7"
 sync_wrapper = "0.1.1"
 tokio = { version = "1", features = ["time"] }
-tokio-util = "0.6"
```
davidpdrsn, line 43: Oh wow we're not using this anymore. Doesn't seem like it. Although please file a separate PR that removes this. A bit weird to have it mixed in with this.

davidpdrsn (changes requested):

axum/Cargo.toml
zys864: `tokio-util` only exists in `Cargo.toml`. It never used in `.rs` file.

davidpdrsn: Yeah I get that. But please submit the removal in a separate PR. Its not related to ECOSYSTEM.md at.

PR #758, opened 2022-02-15
remove unused `axum`'s dependency: `tokio-util`

zys864:
## Motivation
In `axum` crate, the `tokio-util` dependency only exists in `Cargo.toml`, it never be used in lib (`.rs` file).

## Solution
remove `tokio-util` from `axum` dependencies list


davidpdrsn (approved):
Thanks!

PR #760, opened 2022-02-15
examples(jwt): Claims.exp represents a UTC timestamp

frobiac:
For the example to work, it should be an epoch value in the future.

See https://github.com/Keats/jsonwebtoken/blob/d8a33def00a61bab12aa5e94ff76d3043d4d474e/README.md?plain=1#L61

## Motivation

Running the curl commands given in the example heade did not validate the JWT.

## Solution

Adjust the Claims.exp value so it represents an epoch value in the future

### DIFF
--- a/examples/jwt/src/main.rs
+++ b/examples/jwt/src/main.rs
@@ -93,7 +93,8 @@ async fn authorize(Json(payload): Json<AuthPayload>) -> Result<Json<AuthBody>, A
     let claims = Claims {
         sub: "b@b.com".to_owned(),
         company: "ACME".to_owned(),
-        exp: 100000,
+        // Mandatory expiry time as UTC timestamp
+        exp: 2000000000, // May 2033
     };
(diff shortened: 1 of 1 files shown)


frobiac:
This should probably be made even clearer by calculating current time plus a given offset...

examples/jwt/src/main.rs
```diff
@@ -93,7 +93,8 @@ async fn authorize(Json(payload): Json<AuthPayload>) -> Result<Json<AuthBody>, A
     let claims = Claims {
         sub: "b@b.com".to_owned(),
         company: "ACME".to_owned(),
-        exp: 100000,
+        // Mandatory expiry time as UTC timestamp
+        exp: 2000000000, // May 2033
```
davidpdrsn, line 97: Hopefully something better has come along by 2033 😅

davidpdrsn (approved):
Thanks!

PR #767, opened 2022-02-17
Add IntoResponse impl for BytesMut and Chain

AlisCode:
Adds two `IntoResponse` implementations for convenience to work on `bytes` types

## Motivation

* Enlarging the surface of axum's API to work better with the `bytes` crate
* In my context, I want my service to output a file and append something to it,  `bytes::Chain` seemed like a good opportunity to do so.

## Solution

Two new implementations of IntoResponse : 

* `BytesMut` which freezes (zero-cost operation) to get a `Bytes` handle and use its IntoResponse's implementation. 
* `Chain<T,U>` where `T: Buf` and `U: Buf`, which iterates over every collects everything into a `Bytes`. Sadly [the implementation of `FromIterator`](https://docs.rs/bytes/1.1.0/src/bytes/bytes_mut.rs.html#1197-1201) seems to show that it's not exactly zero-cost.

### DIFF
--- a/axum-core/src/response/mod.rs
+++ b/axum-core/src/response/mod.rs
@@ -8,16 +8,22 @@ use crate::{
     body::{boxed, BoxBody},
     BoxError,
 };
-use bytes::Bytes;
+use bytes::{buf::Chain, Buf, Bytes, BytesMut};
 use http::{
     header::{self, HeaderMap, HeaderName, HeaderValue},
     StatusCode,
 };
 use http_body::{
     combinators::{MapData, MapErr},
-    Empty, Full,
+    Empty, Full, SizeHint,
 };
-use std::{borrow::Cow, convert::Infallible, iter};
+use std::{
+    borrow::Cow,
+    convert::Infallible,
+    iter,
+    pin::Pin,
+    task::{Context, Poll},
+};
 
 /// Type alias for [`http::Response`] whose body type defaults to [`BoxBody`], the most common body
 /// type used with axum.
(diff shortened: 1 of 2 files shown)


jplatte:
Could `hyper::body::aggregate` be used instead for the `Chain` case?

axum-core/src/response/mod.rs
```diff
@@ -292,6 +292,18 @@ impl IntoResponse for Bytes {
...
+    fn into_response(self) -> Response {
+        self.freeze().into_response()
+    }
+}
+
+impl<T: Buf, U: Buf> IntoResponse for Chain<T, U> {
+    fn into_response(self) -> Response {
+        self.into_iter().collect::<Bytes>().into_response()
+    }
+}
```
davidpdrsn, line 305: I thought that maybe a custom body would be more efficient. That way the data is only copied when the body is polled:

```suggestion
impl<T, U> IntoResponse for Chain<T, U>
where
    T: Buf + Unpin + Send + 'static,
    U: Buf + Unpin + Send + 'static,
{
    fn into_response(self) -> Response {
        let (first, second) = self.into_inner();
        let mut res = Response::new(boxed(ChainBody {
            first: Some(first),
            second: Some(second),
        }));
        res.headers_mut().insert(
            header::CONTENT_TYPE,
            HeaderValue::from_static(mime::APPLICATION_OCTET_STREAM.as_ref()),
        );
        res
    }
}

struct ChainBody<T, U> {
    first: Option<T>,
    second: Option<U>,
}

impl<T, U> http_body::Body for ChainBody<T, U>
where
    T: Buf + Unpin,
    U: Buf + Unpin,
{
    type Data = Bytes;
    type Error = Infallible;

    fn poll_data(
        mut self: Pin<&mut Self>,
        _cx: &mut Context<'_>,
    ) -> Poll<Option<Result<Self::Data, Self::Error>>> {
        if let Some(mut buf) = self.first.take() {
            let bytes = buf.copy_to_bytes(buf.remaining());
            return Poll::Ready(Some(Ok(bytes)));
        }

        if let Some(mut buf) = self.second.take() {
            let bytes = buf.copy_to_bytes(buf.remaining());
            return Poll::Ready(Some(Ok(bytes)));
        }

        Poll::Ready(None)
    }

    fn poll_trailers(
        self: Pin<&mut Self>,
        _cx: &mut Context<'_>,
    ) -> Poll<Result<Option<HeaderMap>, Self::Error>> {
        Poll::Ready(Ok(None))
    }

    fn is_end_stream(&self) -> bool {
        self.first.is_none() && self.second.is_none()
    }

    fn size_hint(&self) -> SizeHint {
        match (self.first.as_ref(), self.second.as_ref()) {
            (Some(first), Some(second)) => {
                let total_size = first.remaining() + second.remaining();
                SizeHint::with_exact(total_size as u64)
            }
            (Some(buf), None) => SizeHint::with_exact(buf.remaining() as u64),
            (None, Some(buf)) => SizeHint::with_exact(buf.remaining() as u64),
            (None, None) => SizeHint::with_exact(0),
        }
    }
}
```

axum/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **added:** `middleware::from_fn` for creating middleware from async functions.
   This previously lived in axum-extra but has been moved to axum ([#719])
+- **added:** Implement `IntoResponse` for `bytes::BytesMut` and `bytes::Chain<T,U>` ([#767])
```
davidpdrsn, line 12:
```suggestion
- **added:** Implement `IntoResponse` for `bytes::BytesMut` and `bytes::Chain<T, U>` ([#767])
```

Also have to add a reference for `[#767]` otherwise the link wont work.

davidpdrsn (changes requested):
Thanks! I had an idea regarding `Chain`. Let me know what you think :)

axum-core/src/response/mod.rs
jplatte: This seems great, but `ChainBody` seems like it should be part of `http-body` 🙂

davidpdrsn: Yeah it really should. I might try and upstream it.

davidpdrsn:
You can just ignore the cargo deny CI failure. I'll get that fixed at some point 😊

axum-core/src/response/mod.rs
AlisCode: I'm thinking that renaming `ChainBody` to `BytesChainBody` will avoid the possible confusion with two chained "HTTP Bodies"

davidpdrsn: Yeah sure. As long as not public the name doesn't really matter.

AlisCode:
Re-implemented `IntoResponse` for `Chain` using your custom body that I renamed to `BytesChainBody`, that should avoid potential confusion.

Also squashed commits to keep a clean history :)

axum/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **added:** `middleware::from_fn` for creating middleware from async functions.
   This previously lived in axum-extra but has been moved to axum ([#719])
+- **added:** Implement `IntoResponse` for `bytes::BytesMut` and `bytes::Chain<T, U>` ([#767])
```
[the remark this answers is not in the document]
AlisCode: Didn't notice that there was a changelog specific to subcrates, I'll move that there :)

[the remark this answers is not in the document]
jplatte: I know this won't be very helpful, but I don't care much either way 😅

davidpdrsn (approved):
LGMT! Thanks!

davidpdrsn:
> Also squashed commits to keep a clean history :)

We squash merge all PRs so you actually don't have to.

axum/CHANGELOG.md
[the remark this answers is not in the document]
davidpdrsn: @jplatte thats a first 😂

PR #765, opened 2022-02-16
Fix doc(cfg)s

jplatte:
## Motivation

Fixes #763

PR #777, opened 2022-02-22
Backport recent changes to 0.4.x

davidpdrsn:
This backports all the recent non-breaking changes to 0.4.x. There is a bunch of good stuff there that I think we should ship. This includes the new typed routing stuff in axum-extra, new middleware docs, and `middleware::from_fn` moved into axum.

I'm sorry that this PR isn't really reviewable because its so big. The only new code I had to add was in [`axum-extra/src/lib.rs`](https://github.com/tokio-rs/axum/pull/777/files#diff-4882a4828cf7b2f2c01deb617cf2e6cc3e7aca613737882700ec8bd7c6c25b65) to re-export `middleware::from_fn` with the old names they had in axum-extra previously.

There are changes to all the crates so when this is merged I'll do release PRs for each in order: 1) axum-core 2) axum 3) axum-macros 4) axum-extra.

In the future I think I'll be more aggressive with back porting stuff so its easier to review and manage.

I still don't have a timeline for 0.5 and wanna address the open issues first, in case they require breaking changes to fix.


jplatte:
#765 was technically a breaking change because the last commit from it moved some rejection types behind Cargo feature that weren't previously feature-gated.

davidpdrsn:
> #765 was technically a breaking change because the last commit from it moved some rejection types behind Cargo feature that weren't previously feature-gated.

Yes I did notice that. I removed those feature flags here though. See https://github.com/tokio-rs/axum/pull/777/files#diff-7a01bc014c59b4acd51a3ba6a0c64b099cf398565ec3dec2eb89bcd23dd19965, namely the absence of `cfg`s.

davidpdrsn:
Also removed the new features from https://github.com/tokio-rs/axum/pull/777/files#diff-2250e9f68bdd22e6fd3415ff68bd5a401011d765bf9779f6adbd89565c46562e, which are present on `main`.

jplatte (approved):

PR #778, opened 2022-02-22
Make `middleware::from_fn` response future public

davidpdrsn:
While doing https://github.com/tokio-rs/axum/pull/777 I noticed that the response future from `middleware::from_fn` wasn't public so the type wasn't nameable.

I would expect `#[deny(unreachable_pub)]` to catch this but it didn't. Might be a bug 🤔

### DIFF
--- a/axum/src/middleware/mod.rs
+++ b/axum/src/middleware/mod.rs
@@ -5,3 +5,9 @@
 mod from_fn;
 
 pub use self::from_fn::{from_fn, FromFn, FromFnLayer, Next};
+
+pub mod futures {
+    //! Future types.
+
+    pub use super::from_fn::ResponseFuture as FromFnResponseFuture;
+}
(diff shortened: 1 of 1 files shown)


axum/src/middleware/mod.rs
```diff
@@ -5,3 +5,9 @@
 mod from_fn;
 
 pub use self::from_fn::{from_fn, FromFn, FromFnLayer, Next};
+
+pub mod future {
+    //! Future types.
+
+    pub use super::from_fn::ResponseFuture as FromFnResponseFuture;
+}
```
davidpdrsn, line 13: Its rare that users actually care about the future type so tower relates crates commonly have a `futures` module like this. [Prior art: `tower::util::future`](https://docs.rs/tower/latest/tower/util/future/index.html).

jplatte (approved):

PR #782, opened 2022-02-22
Fix dependency versions in axum-extra

davidpdrsn:
Updating to axum-extra 0.1.3 might cause the build to fail if axum
wasn't also updated, since axum-extra 0.1.3 needs
`axum::middleware::from_fn`.

This fixes that by depending on the right patch versions for axum and
axum-macros.

Why can't the tooling detect stuff like this 😢


davidpdrsn:
I'm just gonna merge and publish this quickly when CI is green. Also gonna yank version 0.1.3 of axum-extra.

PR #785, opened 2022-02-22
feat(examples): add routing-match-http-method-for-any-path example

Oliboy50:
## Motivation

I spent 1 day trying to update my router to always return some `Response` when a specific `HTTP method` is used **for any path** (ℹ️ and I still didn't manage to make it work, see below)

## Solution

I suggest to add an `example` to show how to do it "the good way".

## Problem for the current solution

The current implementation using `Response::builder().body(...).unwrap()` demonstrate a bug in `axum` (which I can't fix from my userland even if I wrap things using `Box::pin(...)`) because we end up with the following compiler errors:

```shell
$ cargo check
    Checking example-routing-match-http-method-for-any-path v0.1.0 (/usr/local/var/www/oliver/axum/examples/routing-match-http-method-for-any-path)
error[E0308]: `if` and `else` have incompatible types
  --> examples/routing-match-http-method-for-any-path/src/main.rs:53:9
   |
46 |  /     if req.method() == Method::OPTIONS {
47 |  |         Ok(Response::builder()
   |  |_________-
48 | ||             .header("x-matched-from", req.uri().to_string())
49 | ||             .status(200)
50 | ||             .body(Body::from("matched OPTIONS method").boxed_unsync())
51 | ||             .unwrap())
   | ||______________________- expected because of this
52 |  |     } else {
53 |  |         Ok(next.run(req).await)
   |  |         ^^^^^^^^^^^^^^^^^^^^^^^ expected struct `hyper::error::Error`, found struct `axum::Error`
54 |  |     }
   |  |_____- `if` and `else` have incompatible types
   |
   = note: expected type `Result<Response<http_body::combinators::box_body::UnsyncBoxBody<_, hyper::error::Error>>, _>`
              found enum `Result<Response<http_body::combinators::box_body::UnsyncBoxBody<_, axum::Error>>, _>`
```

## Useless notes

- Using `warp` (instead of `axum`), matching for all paths for a specific HTTP method is as simple as using the `warp::SOME_METHOD()` filters (e.g. `warp::options()` if we want a custom response when the `OPTIONS` method is used).

### DIFF
--- a/examples/routing-match-http-method-for-any-path/Cargo.toml
+++ b/examples/routing-match-http-method-for-any-path/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "example-routing-match-http-method-for-any-path"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+axum = { path = "../../axum" }
+axum-extra = { path = "../../axum-extra" }
+tokio = { version = "1.0", features = ["full"] }
+tracing = "0.1"
+tracing-subscriber = { version="0.3", features = ["env-filter"] }
+tower = { version = "0.4", features = ["util", "filter"] }
+hyper = { version = "0.14", features = ["full"] }
--- a/examples/routing-match-http-method-for-any-path/src/main.rs
+++ b/examples/routing-match-http-method-for-any-path/src/main.rs
@@ -1,0 +1,64 @@
+//! Run with
+//!
+//! ```not_rust
+//! cargo run -p example-routing-match-http-method-for-any-path
+//! # listening on 3000
+//!
+//! curl -v -X OPTIONS http://localhost:3000/some_path
+//! # OPTIONS /some_path matched!
+//!
+//! curl -v -X GET http://localhost:3000/some_path
+//! # GET /some_path matched!
+//!
+//! curl -v -X OPTIONS http://localhost:3000/another_path
+//! # OPTIONS /another_path matched!
+//!
+//! curl -v -X GET http://localhost:3000/another_path
+//! # GET /another_path did not match anything!
+//! ```
+
+use axum::{
+    body::{Body, HttpBody},
+    http::{Method, Request},
+    middleware::{self, Next},
+    response::{IntoResponse, Response},
+    routing::get,
+    Router,
+};
+use std::net::SocketAddr;
+
+#[tokio::main]
+async fn main() {
+    let app = Router::new()
+        // matches "OPTIONS" method for any path
+        .layer(middleware::from_fn(intercept_options_method))
+        // matches "GET" method for "/some_path" only
+        .route("/some_path", get(get_some_path_handler));
+
+    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
+    println!("listening on {}", addr);
+    axum::Server::bind(&addr)
+        .serve(app.into_make_service())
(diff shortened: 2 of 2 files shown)


davidpdrsn:
I don't think we need an example for this. Its quite niche and the parts needed are already covered in the docs. I wish you'd have just asked in [the Discord](https://github.com/tokio-rs/tokio). Someone would have been able to help you very quickly.

What rust is complaining about is that the body's error type doesn't match. axum requires bodies to use `axum::Error`. You can use `axum::body::boxed` to convert it. However it seems you didn't share all your code, so I suspect something else is also wrong. I only see `Ok`s, whats the error type?

If you change the first half of the `if` to use `Err` then it works just fine:

```rust
async fn intercept_options_method<B>(req: Request<B>, next: Next<B>) -> impl IntoResponse {
    if req.method() == Method::OPTIONS {
        Err(Response::builder()
            .header("x-matched-from", req.uri().to_string())
            .status(200)
            .body(Body::from("OPTIONS matched!"))
            .unwrap())
    } else {
        Ok(next.run(req).await)
    }
}
```

However I would argue this is nicer:

```rust
async fn intercept_options_method<B>(req: Request<B>, next: Next<B>) -> impl IntoResponse {
    if req.method() == Method::OPTIONS {
        let headers = Headers([("x-matched-from", req.uri().to_string())]);
        let body = "OPTIONS matched!";
        Err((headers, body))
    } else {
        Ok(next.run(req).await)
    }
}
```

That uses [`Headers`](https://docs.rs/axum/latest/axum/response/struct.Headers.html) and the fact that `(Headers, &'static str)` implements `IntoResponse`.

I'll close this PR for now but you're welcome to ask follow up questions.

Oliboy50:
> However it seems you didn't share all your code, so I suspect something else is also wrong.

🤔 the full code of the example is in the PR, what do you mean?

Besides, using `Err` instead of `Ok` seems weird to me when I want to send a "correct" response (status 200) 🤔, is it still the "right way" to go?

davidpdrsn:
Hm sorry about that. My bad. I would expect Rust to give a different error. But the fix is the same:

```diff
--- a/examples/routing-match-http-method-for-any-path/src/main.rs
+++ b/examples/routing-match-http-method-for-any-path/src/main.rs
@@ -45,7 +45,7 @@ async fn main() {

 async fn intercept_options_method<B>(req: Request<B>, next: Next<B>) -> impl IntoResponse {
     if req.method() == Method::OPTIONS {
-        Ok(Response::builder()
+        Err(Response::builder()
             .header("x-matched-from", req.uri().to_string())
             .status(200)
             .body(Body::from("OPTIONS matched!").boxed_unsync())
```

It works because [`Result<T, E>` implements `IntoResponse`](https://github.com/tokio-rs/axum/blob/main/axum-core/src/response/mod.rs#L183-L194) if `T` and `E` do. [axum's error handling docs](https://docs.rs/axum/latest/axum/error_handling/index.html) have more details about what we consider "errors".

Oliboy50:
👌 no problem, thanks for your time

By the way, someone helped me to fix it too.

Instead of using:
```rust
.body(Body::from("OPTIONS matched!").boxed_unsync())
```

we can use:
```rust
.body(body::boxed(body::Full::from("OPTIONS matched!")))
```

and no more error with this one 🙆‍♂️

PR #788, opened 2022-02-24
axum-extra: bump version to 0.1.4

mgrachev:
## Motivation

Wrong release version in the repository.

## Solution

Bumped a version in `Cargo.toml` and `CHANGELOG.md`.

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+None.
+
+# 0.1.4 (22. February, 2021)
+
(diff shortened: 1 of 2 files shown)


davidpdrsn:
The `main` branch contains breaking changes and is working towards the next major version. Backports (like the recent release) is done from the [v0.4.x](https://github.com/tokio-rs/axum/tree/v0.4.x) branch which also contains the up to date changelog and crate versions. When we ship the next major we'll also merge the changelogs.

davidpdrsn:
So if we update the changelogs on main we should copy them from the v0.4.x not just add dates. For example no breaking changes were shipped in axum-extra 0.1.4

mgrachev:
@davidpdrsn Thanks for the clarification 🙂

davidpdrsn:
I've made GitHub releases for the subcrates https://github.com/tokio-rs/axum/releases. Hopefully that makes changes more clear.

Issue #789, opened 2022-02-24
Implement into Uri for TypedPath


norman784:
## Feature Request

Add ergonomic improvements to `TypedPath`

### Motivation

It feels so off to write `Redirect::temporary(MyPath.to_string().parse().unwrap())` instead of `Redirect::temporary(MyPath.into_uri())`.

### Proposal

Add and implement a new trait `IntoUri` to `TypedPath`.

### Alternatives

Maybe `Into<Uri>` works  also just fine, the only open question is if/how can fail to generate the `Uri`.

davidpdrsn:
Good call! I think we should implement `TryFrom<MyPath> for Uri` where the error type is [`InvalidUri`](https://docs.rs/http/0.2.6/http/uri/struct.InvalidUri.html).

PR #790, opened 2022-02-24
Easily convert typed paths into URIs

davidpdrsn:
`#[derive(TypedPath)]` will now also generate `TryFrom<_> for Uri` for
easily converting paths into URIs for use with `Redirect` and friends.

Fixes https://github.com/tokio-rs/axum/issues/789

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning].
 
 # Unreleased
 
+- **added:** Add `TypedPath::to_uri` for converting the path into a `Uri` ([#790])
 - **added:** Add type safe routing. See `axum_extra::routing::typed` for more details ([#756])
 - **breaking:** `CachedRejection` has been removed ([#699])
 - **breaking:** `<Cached<T> as FromRequest>::Rejection` is now `T::Rejection`. ([#699])
@@ -16,6 +17,7 @@ and this project adheres to [Semantic Versioning].
 [#699]: https://github.com/tokio-rs/axum/pull/699
 [#719]: https://github.com/tokio-rs/axum/pull/719
 [#756]: https://github.com/tokio-rs/axum/pull/756
+[#790]: https://github.com/tokio-rs/axum/pull/790
 
 # 0.1.2 (13. January, 2022)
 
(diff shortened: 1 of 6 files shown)


zys864:
The `CI / test-versions (stable) (pull_request)`'s command `cargo test --all --all-features --all-targets` failed due to the unexpected `not_send.stderr` which occurs on `stable-1.59`.

I test it on my machine. When I use `stable-1.58` version, the test command successes.

But `stable-1.59` may have update stderr. May be we should fix it,be the test will failed on older rust version.


Here is `stable-1.59` `not_send.stderr`'s `ACTUAL OUTPUT`
````
error: future cannot be sent between threads safely
 --> tests/debug_handler/fail/not_send.rs:4:1
  |
4 | async fn handler() {
  | ^^^^^ future returned by `handler` is not `Send`
  |
  = help: within `impl Future<Output = ()>`, the trait `Send` is not implemented for `Rc<()>`
note: future is not `Send` as this value is used across an await
 --> tests/debug_handler/fail/not_send.rs:6:13
  |
5 |     let rc = std::rc::Rc::new(());
  |         -- has type `Rc<()>` which is not `Send`
6 |     async {}.await;
  |             ^^^^^^ await occurs here, with `rc` maybe used later
7 | }
  | - `rc` is later dropped here
note: required by a bound in `check`
 --> tests/debug_handler/fail/not_send.rs:4:1
  |
4 | async fn handler() {
  | ^^^^^ required by this bound in `check`
````

davidpdrsn:
@zys864 thanks for looking into it! I did think the test failure seemed odd but forgot that a new rust came out yesterday. I'm gonna make a separate PR to fix that in a few hours (have meetings), or you're free to do it if you have the time.

davidpdrsn:
Hm I just thought that this actually is a breaking change. Someone could have implemented `TryFrom<_> for Uri` themselves which will conflict with this but I'm fine with shipping 0.2 of axum-extra and axum-macros.

jplatte (commented):
Should the conversion really be fallible? With parameters being url-encoded, the only case where I could see it failing to convert is if the route itself is invalid (in which case registering it would fail too).

davidpdrsn:
Ah yes that's true, doesn't need to be fallible!

davidpdrsn:
@jplatte I've added a `to_uri` method to the `TypedPath` trait. That way its not a breaking change and we don't have to change the macro, since we can parse the `Display` output. What do you think?

jplatte:
I like it! Changelog / docs still reference `TryFrom`, otherwise lgtm.

jplatte (approved):

PR #796, opened 2022-02-25
axum-macros: use fully qualified Result type

matze:
# Motivation

It is pretty common to define custom `Result<T>` type aliases however the typed_path macros generate `Result<T, E>` on the assumption they are imported from the std prelude. Having the type aliases defined causes errors like:

              ^^^^^^^^^
              |
              expected 1 generic argument
              help: remove this generic argument

## Solution

Generate fully qualified `std::result::Result` return types.

### DIFF
--- a/axum-macros/src/typed_path.rs
+++ b/axum-macros/src/typed_path.rs
@@ -97,7 +97,7 @@ fn expand_named_fields(ident: &syn::Ident, path: LitStr, segments: &[Segment]) -
         {
             type Rejection = <::axum::extract::Path<Self> as ::axum::extract::FromRequest<B>>::Rejection;
 
-            async fn from_request(req: &mut ::axum::extract::RequestParts<B>) -> Result<Self, Self::Rejection> {
+            async fn from_request(req: &mut ::axum::extract::RequestParts<B>) -> ::std::result::Result<Self, Self::Rejection> {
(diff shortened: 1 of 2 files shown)


davidpdrsn (approved):
Thanks! Guessed we missed these.

axum-macros/src/typed_path.rs
```diff
@@ -97,7 +97,7 @@ fn expand_named_fields(ident: &syn::Ident, path: LitStr, segments: &[Segment]) -
         {
             type Rejection = <::axum::extract::Path<Self> as ::axum::extract::FromRequest<B>>::Rejection;
 
-            async fn from_request(req: &mut ::axum::extract::RequestParts<B>) -> Result<Self, Self::Rejection> {
+            async fn from_request(req: &mut ::axum::extract::RequestParts<B>) -> std::result::Result<Self, Self::Rejection> {
```
jplatte, line 100:
```suggestion
            async fn from_request(req: &mut ::axum::extract::RequestParts<B>) -> ::std::result::Result<Self, Self::Rejection> {
```

jplatte: Everywhere else too, of course.

matze:
> By the way, it doesn't make this PR less valid but many people now prefer
> 
> ```rust
> type Result<T, E = MyError> = std::result::Result<T, E>;
> ```

Good point, yes.

Issue #770, opened 2022-02-18
Implement IntoResponse for Extension<T>


Foleychris:
The pattern of passing state into handlers via Request.extensions works very well across Axum, including passing state from middleware into a handler. Similarly, I would like to pass state back from a handler to a middleware. My thought was have Extension<T> populate Response.extension, which could then be extracted in middleware, and combinations similar to `Headers` today. 

For a more concrete example, in the case of session handling today, individual handlers are required to handle session cookies (https://github.com/tokio-rs/axum/blob/main/examples/sessions/src/main.rs#L54-L58). With this approach, they could return `(StatusCode, Extension<User>)` from the handler, and the middleware would responsible for the cookie lifecycle end to end (extracting if present, setting if needed, etc).

I would be willing to attempt the implementation of this, but may require some mentoring through the process. 

EDIT: There's a long personal journey in this thread, but the tl;dr is in https://github.com/tokio-rs/axum/issues/770#issuecomment-1045559125

Foleychris:
Just capturing my own thoughts here -- it looks like the approach used for the tuple type with headers won't work here. They implement `IntoResponse` like:

```
impl<H, T> IntoResponse for (StatusCode, H, T)
where
    H: IntoResponseHeaders,
    T: IntoResponse,
{
    fn into_response(self) -> Response {
       [...]
    }
}
```
and
```
impl<H, T> IntoResponse for (StatusCode, H, T)
where
    H: IntoResponseHeaders,
    T: IntoResponse,
{
    fn into_response(self) -> Response {
       [...]
    }
}
```

Ignoring the specifics of the implementation, I first tried doing something similar, defining an `IntoExtension` trait in axum-core, that I could then implement in axum:
```
pub trait IntoExtensions {
    fn into_extensions(self) -> http::Extensions;
}
```
and then attempt to implement for a tuple:
```
impl<E, T> IntoResponse for (E, T)
where
    E: IntoExtensions,
    T: IntoResponse,
{
    fn into_response(self) -> Response {
        todo!()
    }
}
```
but this quite reasonably fails:
> conflicting implementation for `(_, _)`

which makes sense, a type could implement both `IntoResponseHeaders` and `IntoExtensions` (it would probably be nonsense but...) and the complier rejects this. 

I'm definitely not familiar enough with the code base to say if this is a blocker. The only thing that immediately comes to mind is pulling `Extension<T>` into axum-core and using it directly in the `impl` with something like:

```
[...]

PR #797, opened 2022-02-27
Add `IntoResponseParts`

davidpdrsn:
Note: Design has changed quite a bit from the description here. See comments for details.

---

This is a proposal that changes how responses are made by introducing a new trait:

```rust
pub trait IntoResponseParts {
    fn into_response_parts(self, res: &mut ResponseParts);
}
```

Where `ResponseParts` is:

```rust
pub struct ResponseParts { ... }

impl ResponseParts {
    pub fn set_version(&mut self, version: Version) {}
    pub fn set_status(&mut self, status: StatusCode) {}
    pub fn insert_header<K, V>(&mut self, key: K, value: V) {}
    pub fn insert_extension<T>(&mut self, extension: T) {}
}
```

And these impls

```rust
// this makes `impl IntoResponse` work like before,
impl<T> IntoResponse for T
where
    T: IntoResponseParts,
{}

// for tuples
impl<T1> IntoResponseParts for (T1,)
where
    T: IntoResponseParts,
{}

impl<T1, T2> IntoResponseParts for (T1, T2)
where
    T1: IntoResponseParts,
    T2: IntoResponseParts,
{}

// etc...
```

This means you can return tuples of response parts like `(status, headers, body)` just like before, but the order no longer matters. So `(body, headers, status)` also works.

I've also implemented `IntoResponseParts` for `Extension` and `[(HeaderNameIsh, HeaderValueIsh); N]` so `(Extension(1), Extension(2), [("content-type", "text/plain")], "Hello, World!")` works!

Overall I'm quite happy with this approach. I think its more consistent since `Extension` works and I like that the order of things in tuples no longer matters. I also like that there is only trait to implement vs the previous `IntoResponseHeaders` and `IntoResponse`.

The only downside I see is that users might be confused by having to implement `IntoResponseParts` but use `IntoResponse`. We should help by clearly documenting the blanket impl and how that works.

### Things to note

- `IntoResponse` is now sealed, so users will have to implement `IntoResponseParts` instead.
- `http::Extensions` doesn't implement `IntoResponseParts` since we cannot merge extensions.
- `Response` and `http::response::Parts` implements `IntoResponse` and _not_ `IntoResponseParts`, again because we cannot merge extensions.
- `Result<T, E>` implements `IntoResponse` and _not_ `IntoResponseParts`. This makes `Result<impl IntoResponse, E>` work. Means you cannot have results in tuples of parts but thats probably fine.
- `Headers` has been removed. `[(HeaderNameIsh, HeaderValueIsh); N]` works instead and doesn't require a wrapper type.

Fixes https://github.com/tokio-rs/axum/issues/770

### TODO

- [x] Docs, including auditing existing docs and explaining more clearly more things compose
- [x] changelog

### Future improvements

- Move `Extension` into the root because it, like `Json`, is both an extractor and response part. I'll do that in a follow up PR. Will be re-exported from `axum::response` and `axum::extract` like `Json`. Same for `TypedHeader`.


jplatte:
Small overall comment after reading through (only) the description: What I meant in #770 was a trait for just the headers and extensions, i.e. the parts that can added onto. The impls would still work the same way as the ones for tuples containing `IntoResponseHeaders` right now (i.e. one optional status code at the start, response body or other `IntoResponse` type at the end).

axum-core/src/response/mod.rs
```diff
@@ -199,127 +373,125 @@ where
...
+        res.set_body(self);
     }
 }
 
-impl IntoResponse for &'static str {
-    #[inline]
-    fn into_response(self) -> Response {
-        Cow::Borrowed(self).into_response()
+impl IntoResponseParts for &'static str {
+    fn into_response_parts(self, res: &mut ResponseParts) {
```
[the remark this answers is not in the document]
davidpdrsn: Not sure if they really matter and they were applied inconsistently 🤷

axum/src/docs/response.md
```diff
@@ -2,27 +2,21 @@ Types and traits for generating responses.
...
-// We've already seen returning &'static str
-async fn plain_text() -> &'static str {
-    "foo"
-}
+// `()` gives an empty response
+async fn empty() {}
 
-// String works too and will get a `text/plain; charset=utf-8` content-type
-async fn plain_text_string(uri: Uri) -> String {
+// String get a `text/plain; charset=utf-8` content-type
```
jplatte, line 18: `get` should be `will get` as before, or maybe `gets`.

jplatte (commented):
Code looks good, but I'm still not entirely sure whether the arbitrary order of tuple elements is worth it considering the potential confusion / pitfalls around `Result` or opaque `impl IntoResponse` types in tuples¹. We'd also lose the static verification of only a single status code and body being set.

¹ in particular, I wouldn't be surprised if some people are currently writing handlers that call other handlers (where the fn signatures are likely `-> impl IntoResponse`) and override the status code or add headers

davidpdrsn:
I think I've cracked it. Now `IntoResponseParts` is only allowed to insert extensions or headers. It cannot touch the status or response.

Additionally I was able to tweak the impls such that:

- `(StatusCode, impl IntoResponse)` works like before. The status code overrides the one from the response.
- (T1, ..., Tn, impl IntoResponse) where T1, ..., Tn: IntoResponseParts` can be used to insert headers and extensions but _not_ the status
- (StatusCode, T1, ..., Tn, impl IntoResponse) where T1, ..., Tn: IntoResponseParts` can be used to insert headers and extensions and the status.

Thus a status code must be the first element in the tuple, and there can only be one status code. The body must be the last element. Anything in the middle must implement `IntoResponseParts` and cannot touch the status or body.

How that doesn't give conflicting impls is magical to me, but it works! `Result<impl IntoResponse, E>` works like before.

I think this addresses the footguns.

davidpdrsn:
Still have to update the docs but I'll do that if you think the code makes sense.

axum-core/src/response/into_response.rs
```diff
@@ -0,0 +1,507 @@
...
+    }
+}
+
+impl IntoResponse for Version {
+    fn into_response(self) -> Response {
+        let mut res = ().into_response();
+        *res.version_mut() = self;
+        res
+    }
+}
```
davidpdrsn, line 144: Kinda surprised no one has asked by `Version` didn't implement `IntoResponse` earlier 🤔

axum-core/src/response/into_response_parts.rs
```diff
@@ -0,0 +1,124 @@
...
+    K::Error: fmt::Display,
+    V: TryInto<HeaderValue>,
+    V::Error: fmt::Display,
+{
+    fn into_response_parts(self, res: &mut ResponseParts) {
+        for (key, value) in self {
+            res.insert_header(key, value);
+        }
+    }
+}
```
davidpdrsn, line 124: The rule with these impls is that if things can be appended to a response, then they should implement `IntoResponseParts` and not `IntoResponse`. That only applies to headers and extensions.

`Extension` implements `IntoResponseParts` in another file.

axum-macros/tests/debug_handler/fail/wrong_return_type.stderr
```diff
@@ -1,9 +1,10 @@
-error[E0277]: the trait bound `bool: IntoResponse` is not satisfied
+error[E0277]: the trait bound `bool: IntoResponseParts` is not satisfied
  --> tests/debug_handler/fail/wrong_return_type.rs:4:23
   |
 4 | async fn handler() -> bool {
-  |                       ^^^^ the trait `IntoResponse` is not implemented for `bool`
+  |                       ^^^^ the trait `IntoResponseParts` is not implemented for `bool`
   |
+  = note: required because of the requirements on the impl of `IntoResponse` for `bool`
```
davidpdrsn, line 7: Its quite unfortunate that this error mentions `IntoResponseParts` at the top and not `IntoResponse`. I suppose rust is confused by the blanket `impl<T> IntoResponse for T where T: IntoResponseParts`.

axum-core/src/extract/rejection.rs
```diff
@@ -15,11 +14,9 @@ impl BodyAlreadyExtracted {
...
 }
 
-impl crate::response::IntoResponse for BodyAlreadyExtracted {
-    fn into_response(self) -> crate::response::Response {
-        let mut res = Response::new(body::boxed(Full::from(Self::BODY)));
-        *res.status_mut() = StatusCode::INTERNAL_SERVER_ERROR;
-        res
+impl IntoResponse for BodyAlreadyExtracted {
+    fn into_response(self) -> Response {
+        (StatusCode::INTERNAL_SERVER_ERROR, Self::BODY).into_response()
```
davidpdrsn, line 19: Like this the response also gets `content-type: text/plain` which is probably more correct.

axum-core/src/response/into_response.rs
jplatte: Why would you ever want to override that? I feel like the http version should always be set by hyper.

axum-macros/tests/debug_handler/fail/wrong_return_type.stderr
jplatte: Yeah, same as with `IntoResponseHeaders` before. But why would we *need* the blanket impl? We would still accept more things with the remaining changes, without the blanket impl, right? And `Extension<T>` could independently implement `IntoResponse` so it's allowed as the last tuple element, or a "standalone" `impl IntoResponse` return value.

axum/src/extract/extension.rs
```diff
@@ -73,3 +75,12 @@ impl<T> Deref for Extension<T> {
...
 }
+
+impl<T> IntoResponseParts for Extension<T>
+where
+    T: Send + Sync + 'static,
+{
+    fn into_response_parts(self, res: &mut ResponseParts) {
+        res.insert_extension(self.0);
+    }
+}
```
davidpdrsn, line 86: I'm thinking if we should even implement `tower::Layer` for `Extension`? Then we wouldn't need `AddExtension` layer and users would only deal with `Extension`. But maybe thats odd? 🤔

jplatte: Oohh, nice idea! I don't really see why not. I'm personally going to continue using `ServiceBuilderExt::add_extension` either way, but as this would reduce axum's API surface a little bit, I'd be in favor.

axum-core/src/response/into_response_parts.rs
```diff
@@ -0,0 +1,124 @@
+use super::Response;
+use http::header::{HeaderMap, HeaderName, HeaderValue};
+use std::{convert::TryInto, fmt};
+
+/// Trait for adding headers and extensions to a response.
+///
+/// You generally shouldn't need to implement this trait manually. Its recommended instead to rely
```
jplatte, line 7:
```suggestion
/// You generally don't need to implement this trait manually. It's recommended instead to rely
```

axum-core/src/response/into_response_parts.rs
```diff
@@ -0,0 +1,124 @@
...
+/// Used with [`IntoResponseParts`].
+#[derive(Debug)]
+pub struct ResponseParts {
+    pub(crate) res: Result<Response, String>,
+}
+
+impl ResponseParts {
+    /// Insert a header into the response.
+    ///
+    /// If the header already exists it will be overwritten.
```
jplatte, line 25:
```suggestion
    /// If the header already exists, it will be overwritten.
```

axum-core/src/response/into_response.rs
```diff
@@ -0,0 +1,507 @@
...
+/// }
+///
+/// impl IntoResponse for MyError {
+///     fn into_response(self) -> Response {
+///         let body = match self {
+///             MyError::SomethingWentWrong => "something went wrong",
+///             MyError::SomethingElseWentWrong => "something else went wrong",
+///         };
+///
+///         // its often easier to implement `IntoResponse` by calling other implementations
```
jplatte, line 53:
```suggestion
///         // its often easiest to implement `IntoResponse` by calling other implementations
```

… since there is nothing "easier" relates to.

axum-extra/src/response/erased_json.rs
```diff
@@ -49,22 +48,24 @@ impl ErasedJson {
...
+                bytes,
+            )
+                .into_response(),
+            Err(err) => (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                [(
+                    header::CONTENT_TYPE,
+                    HeaderValue::from_static(mime::TEXT_PLAIN_UTF_8.as_ref()),
+                )],
+                err.to_string(),
```
jplatte, line 66: With `impl IntoResponse for String` already setting the `content-type` header, it is useless to set it here, no?

axum-macros/tests/debug_handler/fail/wrong_return_type.stderr
davidpdrsn: Oh yeah of course! We don't need the blanket impl at all!

axum-core/src/response/into_response.rs
davidpdrsn: Not sure that you need to actually. But tonic does set it, although I think hyper would handle that as well.

axum-extra/src/response/erased_json.rs
davidpdrsn: Ah yes true!

axum-core/src/response/into_response_parts.rs
```diff
@@ -4,7 +4,7 @@ use std::{convert::TryInto, fmt};
 
 /// Trait for adding headers and extensions to a response.
 ///
-/// You generally shouldn't need to implement this trait manually. Its recommended instead to rely
+/// You generally don't need to implement this trait manually. It's recommended instead to rely
```
jplatte, line 7: To be clear, I think this isn't the only instance where it was worded this way.

jplatte (approved):
That's all my nitpicks voiced.

jplatte:
Should we open a separate issue about whether or not to keep the `Version` impl? Or are you certain enough that it's useful?

davidpdrsn:
Yeah lets do that 👍

PR #800, opened 2022-02-28
Remove deprecated APIs

davidpdrsn:


jplatte (approved):
Commit message should say "deprecated" instead of "deprecations", otherwise LGTM.

PR #803, opened 2022-02-28
Move `TypedHeader` into root

davidpdrsn:
Like `Json`, `TypedHeader` is now both an extractor and a response so this moves it into the root and exports it from both `extract` and `response`.


axum/src/typed_header.rs
```diff
@@ -1,20 +1,20 @@
...
 ///
 /// In general, it's recommended to extract only the needed headers via `TypedHeader` rather than
 /// removing all headers with the `HeaderMap` extractor.
 ///
-/// # Example
-///
 /// ```rust,no_run
 /// use axum::{
-///     extract::TypedHeader,
+///     TypedHeader,
```
jplatte, line 17: I think oder rustfmt formatting rules this should now go on the same line as `Router` below. There's a nightly setting you can use to have rustfmt format doc codeblocks.

axum/src/typed_header.rs
```diff
@@ -31,7 +31,24 @@ use std::ops::Deref;
...
+/// };
+///
+/// async fn handler() -> impl IntoResponse {
+///     (
+///         TypedHeader(ContentType::text_utf8()),
+///         "Hello, World!",
+///     )
+/// }
+/// ```
+#[cfg(feature = "headers")]
```
jplatte, line 51: I think this should just be `doc(cfg)` since the whole module is already feature-gated like this (or should be).

axum/src/typed_header.rs
davidpdrsn: Ah nice! I didn't realize. I'll probably do another PR that formats the docs and merges imports. Not sure if that feature is stable in rustfmt yet.

axum/src/typed_header.rs
davidpdrsn: Does that actually make a difference somewhere? I tested it with `RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --no-deps` and it looks fine.

jplatte: No I don't think it makes a difference, it's just not so clear why there's a redundant `cfg` attribute there as-is.

jplatte (approved):

PR #804, opened 2022-02-28
Move `Extension` into root

davidpdrsn:
Same as https://github.com/tokio-rs/axum/pull/803 but for `Extension`.


axum/src/extension.rs
```diff
@@ -1,20 +1,20 @@
...
-/// # Example
+/// This is commonly used to share state across handlers.
 ///
 /// ```rust,no_run
 /// use axum::{
-///     extract::Extension,
-///     routing::get,
 ///     Router,
+///     Extension,
+///     routing::get,
```
jplatte, line 17:
```suggestion
///     routing::get,
///     Extension, Router,
```

jplatte (approved):

PR #806, opened 2022-03-01
Fix some `examples/*` little problems.

zys864:
## Motivation
Unused dependencies
Miss docs: `cargo run -p example...`

## Solution
Some examples use `println!("Playground: http://localhost:3000")` but `trace`, so I just delete `trace` and `tracing-subscriber` dependencies.
Some examples import `axum-extra`,but never use it, so I just delete it.
`examples/chat` use neither `trace` nor `prinltn`, so I add some `trace` behavior.
`/examples/async-graphql` miss docs `cargo run -p example...` liking other examples,so I add doc add the code top.

Here are lists:
- `examples/chat` deps `trace` and `tracing-subscriber` never be used. Add trace `debug` to `chat`
- remove `examples/reverse-proxy` unused dependencies `tracing-subscriber` and `trace`
- remove `examples/prometheus-metrics` unused dependency `axum-extra`
- remove `examples/print-request-response` unused dependency `axum-extra`
- remove `examples/async-graphql` unused dependencies `tracing-subscriber` and `trace`
- Add docs to `/examples/async-graphql`, just like other examples.


davidpdrsn (approved):
Thanks!

PR #811, opened 2022-03-01
Export `AddExtension` from `middleware`

davidpdrsn:
Keeps the root a bit more tidy.

Fixes https://github.com/tokio-rs/axum/issues/810


jplatte (commented):
Internally, maybe it would make sense for the service to be part of the `extension` module instead of its own 38 LOC module / file? I agree the public API makes sense like this though.

jplatte (approved):

PR #812, opened 2022-03-01
Using fully qualified path for Extension

ecton:
## Motivation

I was updating code after noticing new warnings of deprecations, and I tried to change my usage from AddExtensionLayer to `axum::Extension`, only to find that the type wasn't found.

## Solution

I updated the deprecation message to point to the fully qualified path. It appears this is re-exported in main, but not in 0.4.x. I'm not sure if this should be preferred to exporting it at the new location, but it seemed like the safer fix.

### DIFF
--- a/axum/src/add_extension.rs
+++ b/axum/src/add_extension.rs
@@ -14,7 +14,7 @@ use tower_service::Service;
 #[derive(Clone, Copy, Debug)]
 #[deprecated(
     since = "0.4.7",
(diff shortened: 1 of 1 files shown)


jplatte (approved):
I think it makes sense not to have it in the crate root in 0.4.x because it can only be used as a response part after some breaking changes that aren't release yet. Change LGTM.

davidpdrsn (approved):
Ah yes good catch. Thanks!

PR #813, opened 2022-03-01
Make `IntoResponseParts` more flexible

davidpdrsn:
I've been thinking that `IntoResponseParts` could be useful for adding cookies, flash messages, or more high level things to the response. Something like:

```rust
(
    Cookie::new("key", "value"),
    Cookie::new("key-2", "value-2"),
    flash::error("something went wrong"),
    response,
)
```

I think this is nicer than something like [tower-cookies](https://docs.rs/tower-cookies/latest/tower_cookies/index.html) which uses a middleware and a `Drop` implementation to set the cookies. This approach should compose more nicely and not require any middleware.

That made me think that `ResponseParts` probably isn't quite flexible enough since it only allows inserting headers and extensions. You can't check for existing values, or call any other methods on `HeaderMap` or `Extensions`.

So this makes the following changes:

- Add `extensions(_mut)` and `headers(_mut)` methods to `ResponseParts`. The `*_mut` methods users can `std::mem::take` them and replace the existing ones but I think thats fine as it makes things much more flexible.
- Add an associated error type to `IntoResponseParts`. I think handling errors, for example when converting things into headers, this way is cleaner. The new method signature `fn into_response_parts(self, res: ResponseParts) -> Result<ResponseParts, Self::Error>` is a bit of a mouthful but at least it forces the caller to check the result vs something like `fn(&mut ResponseParts) -> Result<(), _>` which makes it possible to discard the error.

### DIFF
--- a/axum-core/src/response/into_response_parts.rs
+++ b/axum-core/src/response/into_response_parts.rs
@@ -1,14 +1,22 @@
-use super::Response;
-use http::header::{HeaderMap, HeaderName, HeaderValue};
-use std::{convert::TryInto, fmt};
+use super::{IntoResponse, Response};
+use http::{
+    header::{HeaderMap, HeaderName, HeaderValue},
+    Extensions, StatusCode,
+};
+use std::{
+    convert::{Infallible, TryInto},
+    fmt,
+};
 
 /// Trait for adding headers and extensions to a response.
-///
-/// You generally don't need to implement this trait manually. It's recommended instead to rely
-/// on the implementations in axum.
 pub trait IntoResponseParts {
+    /// The type returned in the event of an error.
+    ///
+    /// This can be used to fallibly convert types into headers or extensions.
+    type Error: IntoResponse;
+
     /// Set parts of the response
-    fn into_response_parts(self, res: &mut ResponseParts);
+    fn into_response_parts(self, res: ResponseParts) -> Result<ResponseParts, Self::Error>;
 }
 
 /// Parts of a response.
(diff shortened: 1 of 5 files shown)


axum-core/src/response/into_response_parts.rs
```diff
@@ -116,9 +65,82 @@ where
...
 {
-    fn into_response_parts(self, res: &mut ResponseParts) {
+    type Error = TryIntoHeaderError<K::Error, V::Error>;
+
+    fn into_response_parts(self, mut res: ResponseParts) -> Result<ResponseParts, Self::Error> {
         for (key, value) in self {
-            res.insert_header(key, value);
+            let key = key.try_into().map_err(TryIntoHeaderError::key)?;
+            let value = value.try_into().map_err(TryIntoHeaderError::value)?;
+            res.headers_mut().insert(key, value);
```
jplatte, line 74: I think this should actually use `append` instead of `insert`, so it's possible to add multiple values for one header.

We should make sure never to set multiple content-types accidentally though.

davidpdrsn: Good point. I'd be a bit worried about always appending. There are probably lots of headers that we should add checks for 🤔

Maybe adding a separate `AppendHeaders` newtype is the safest while keeping insert as the default.

jplatte: I didn't mean there should runtime checks, but rather that the usage patterns in axum and examples should be reviewed.

I think a problem with insert by default is that then inserting via array directly has different behaviour from collecting the array into a HeaderMap and adding that to the response.

I wouldn't expect that to come up in practice usually, but for the eventual case where it comes up, it seems like it could result in a decent amount of confusion and/or serious bugs.

davidpdrsn: Yes thats true! I think `TypedHeader` should probably insert the first value and append any other values. I'll fix that in this PR as well.

jplatte (approved):

Issue #802, opened 2022-02-28
Should `http::Version` implement `IntoResponse`?


davidpdrsn:
In https://github.com/tokio-rs/axum/pull/797 I added `impl IntoResponse for http::Version` mostly because I figured why not. However we should decide if its actually useful, or if hyper always sets it for us.

davidpdrsn:
@seanmonstar what do you think? This is about having high level API so users can do

```rust
async fn handler() -> imp IntoResponse {
    (Version::HTTP_2, body)
}
```

Of course users can always set it by constructing a `http::Response` via the builder but is the high level API necessary or does hyper handle that automatically?

PR #815, opened 2022-03-02
Remove `IntoResponse` and `IntoResponseParts` impls for `Version`

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/802

Probably quite niche. We can always add the impls back if someone has a use case.


jplatte (approved):
Apart from stopping users shooting themselves in the foot for whatever reason, this should also make the `IntoResponse` implementor listing a little less scary 😄

PR #816, opened 2022-03-02
Add cookie management to axum-extra

davidpdrsn:
This adds extractors and responses for dealing with cookies to axum-extra. It uses the new `IntoResponseParts` APIs such that no middleware and `Arc<Mutex<_>>`s are required. This should make things more composable as there is no global "cookie manager" that all changes must go through. It just reads the `Cookie` header from the request and appends `Set-Cookie` headers to the response.

Example usage:

```rust
async fn set_cookie(mut jar: CookieJar) -> impl IntoResponse {
    jar.add(Cookie::new("key", "value"))
}

async fn get_cookie(jar: CookieJar) {
    dbg!(jar.get("key"));
}
```

All the heavy lifting is done by the [`cookie`](https://docs.rs/cookie/0.16.0/cookie/index.html) crate.

### DIFF
--- a/axum-extra/src/extract/cookie.rs
+++ b/axum-extra/src/extract/cookie.rs
@@ -403,3 +403,134 @@ impl<'a, K> Iterator for SignedCookieJarIter<'a, K> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use axum::{body::Body, http::Request, routing::get, Router};
+    use tower::ServiceExt;
+
+    #[tokio::test]
+    async fn set_get_remove_cookie() {
+        async fn set_cookie(mut jar: CookieJar) -> impl IntoResponse {
+            jar.add(Cookie::new("key", "value"))
+        }
+
+        async fn get_cookie(jar: CookieJar) -> impl IntoResponse {
+            jar.get("key").unwrap().value().to_owned()
+        }
+
+        async fn remove_cookie(mut jar: CookieJar) -> impl IntoResponse {
+            jar.remove(Cookie::named("key"))
+        }
+
+        let app = Router::<Body>::new()
+            .route("/set", get(set_cookie))
+            .route("/get", get(get_cookie))
+            .route("/remove", get(remove_cookie));
+
+        let res = app
+            .clone()
+            .oneshot(Request::builder().uri("/set").body(Body::empty()).unwrap())
+            .await
+            .unwrap();
+        let cookie_value = res.headers()["set-cookie"].to_str().unwrap();
+
+        let res = app
+            .clone()
+            .oneshot(
+                Request::builder()
+                    .uri("/get")
+                    .header("cookie", cookie_value)
+                    .body(Body::empty())
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+        let body = body_text(res).await;
+        assert_eq!(body, "value");
+
+        let res = app
+            .clone()
+            .oneshot(
+                Request::builder()
+                    .uri("/remove")
+                    .header("cookie", cookie_value)
+                    .body(Body::empty())
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+        assert!(res.headers()["set-cookie"]
+            .to_str()
+            .unwrap()
+            .contains("key=;"));
+    }
+
+    #[tokio::test]
+    async fn set_get_remove_signed_cookie() {
+        async fn set_cookie(mut jar: SignedCookieJar) -> impl IntoResponse {
+            jar.add(Cookie::new("key", "value"))
+        }
+
+        async fn get_cookie(jar: SignedCookieJar) -> impl IntoResponse {
+            jar.get("key").unwrap().value().to_owned()
+        }
+
+        async fn remove_cookie(mut jar: CookieJar) -> impl IntoResponse {
+            jar.remove(Cookie::named("key"))
+        }
+
+        let app = Router::<Body>::new()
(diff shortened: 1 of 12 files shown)


axum-extra/src/extract/cookie.rs
```diff
@@ -0,0 +1,405 @@
...
+use cookie_lib::SignedJar;
+use http::header::{COOKIE, SET_COOKIE};
+use std::{convert::Infallible, fmt, marker::PhantomData};
+
+pub use cookie_lib::{Cookie, Key};
+
+/// Extractor that grabs cookies from the request and managers the jar.
+///
+/// Note that methods like [`CookieJar::get`], [`CookieJar::remove`], etc returns a
+/// [`CookieJarDelta`]. This value _must_ returned from the handler as part of the response for the
```
jplatte, line 20:
```suggestion
/// [`CookieJarDelta`]. This value _must_ be returned from the handler as part of the response for the
```

axum-extra/src/extract/cookie.rs
```diff
@@ -0,0 +1,405 @@
...
+    response::{IntoResponse, IntoResponseParts, Response, ResponseParts},
+    Extension,
+};
+use cookie_lib::SignedJar;
+use http::header::{COOKIE, SET_COOKIE};
+use std::{convert::Infallible, fmt, marker::PhantomData};
+
+pub use cookie_lib::{Cookie, Key};
+
+/// Extractor that grabs cookies from the request and managers the jar.
```
jplatte, line 17:
```suggestion
/// Extractor that grabs cookies from the request and manages the jar.
```

axum-extra/src/extract/cookie.rs
```diff
@@ -0,0 +1,405 @@
...
+///
+/// Note that methods like [`CookieJar::get`], [`CookieJar::remove`], etc returns a
```
jplatte, line 19:
```suggestion
/// Note that methods like [`CookieJar::get`], [`CookieJar::remove`], etc return a
```

axum-extra/src/extract/cookie.rs
```diff
@@ -0,0 +1,405 @@
...
+    }
+}
+
+impl IntoResponse for CookieJarDelta {
+    fn into_response(self) -> Response {
+        (self, ()).into_response()
+    }
+}
+
+/// Extractor that grabs signed cookies from the request and managers the jar.
```
jplatte, line 197:
```suggestion
/// Extractor that grabs signed cookies from the request and manages the jar.
```

axum-extra/src/extract/cookie.rs
```diff
@@ -0,0 +1,405 @@
...
+            key,
+            _marker: PhantomData,
+        })
+    }
+}
+
+impl<K> SignedCookieJar<K> {
+    /// Get a cookie from the jar.
+    ///
+    /// If the cookie exists and its authenticity and integrity can be verified then it's plaintext
```
jplatte, line 308:
```suggestion
    /// If the cookie exists and its authenticity and integrity can be verified then its plaintext
```

axum-extra/src/extract/cookie.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// version is returned.
```
jplatte, line 309:
```suggestion
    /// form is returned.
```

jplatte: Or maybe just "then it is returned in plain text".

jplatte (commented):
Insteadbof having a separate delta type in the API, why not change `add` / `remove` to owned-`self` methods that return `Self` instead of that separate type, and implement `IntoResponseParts` for the jars directly? 

If the jar types have a `must_use` attribute, that should make it equally hard to misuse while making chaining possible and slightly simplifying the API surface. You also wouldn't have to have the jar be a mutable variable then.

jplatte:
Did you see my overall comment as part of the review?

> Instead of having a separate delta type in the API, why not change `add` / `remove` to owned-`self` methods that return `Self` instead of that separate type, and implement `IntoResponseParts` for the jars directly?
> 
> If the jar types have a `must_use` attribute, that should make it equally hard to misuse while making chaining possible and slightly simplifying the API surface. You also wouldn't have to have the jar be a mutable variable then.

davidpdrsn:
Ah no I missed that somehow. Good idea!

davidpdrsn:
@jplatte so much nicer now 😃

jplatte (approved):
Two more comments:

* The upstream `CookieJar`s have an `add_original` method that only modifies the jar if a cookie of the given name doesn't exist yet. Maybe that should be mirrored? (although the one thing I can immediately think of it being used for would be tracking cookies.. yuck)
* Adding an example for this would be nice

davidpdrsn:
Hm not sure. When extracting the jars I call `add_original` with headers from the request as [per the docs](https://docs.rs/cookie/latest/cookie/struct.CookieJar.html#method.add_original)

> This method is intended to be used to seed the cookie jar with cookies received from a client’s HTTP message.

jplatte:
Right. Let's leave it out for now then!

demurgos:
Hello!
Any idea when a new released will be released to crates.io? Cookie support is the last thing I need to migrate to Axum.

davidpdrsn:
Hopefully sometimes this month but no guarantees. Take a look at tower-cookies in the mean time. `main` is also fairly stable so if you wanna go bleeding edge and use that via a git dependency then thats probably fine.

PR #817, opened 2022-03-02
Implement `IntoResponseParts` for tuples

davidpdrsn:
This is quite a niche feature but worth having imo.

Currently you can use tuples of `IntoResponseParts` with up to 16 elements. This adds `IntoResponseParts` impls for tuples such that they can be infinitely nested and you can return more 16 response parts:

```rust
fn handler() -> impl crate::response::IntoResponse {
    (
        Extension(1),
        Extension(2),
        Extension(3),
        Extension(4),
        Extension(5),
        Extension(6),
        Extension(7),
        Extension(8),
        Extension(9),
        Extension(10),
        Extension(11),
        Extension(12),
        Extension(13),
        Extension(14),
        Extension(15),
        // these count as one but actually contain two parts
        (Extension(16), Extension(17)),
        "body",
    )
}
```

We have similar impls for [`FromRequest`](https://docs.rs/axum/latest/axum/extract/trait.FromRequest.html#impl-FromRequest%3CB%3E-for-()) as well.

### DIFF
--- a/axum-core/src/response/into_response_parts.rs
+++ b/axum-core/src/response/into_response_parts.rs
@@ -144,3 +144,32 @@ where
         }
     }
 }
+
+macro_rules! impl_into_response_parts {
+    ( $($ty:ident),* $(,)? ) => {
+        #[allow(non_snake_case)]
+        impl<$($ty,)*> IntoResponseParts for ($($ty,)*)
+        where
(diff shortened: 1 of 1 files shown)


jplatte (approved):
Might be useful for passing around a group of extensions 🤔

Also just doesn't hurt of course 🙂

PR #819, opened 2022-03-03
Move `HasRoutes` into axum

davidpdrsn:
I think `HasRoutes` in axum-extra is pretty neat. It allows users to build their own abstractions for building routes and have them work directly with axum. I've been thinking that for 0.5 we should promote the trait to live in axum.

This is partly driven by some talks @LucioFranco and I have been having about possibly moving tonic's routing to axum in a more public way. Tonic currently reinvents a bunch of middleware stuff that axum already handles. That'd also give users more flexibility in how they compose their gRPC services. Such a change could have tonic implement `HasRoutes` for its services so they'd work with `axum::Router::merge`. However there are no concrete plans so its mostly speculation on my part.

I don't see much downside to this. `Router` implements `HasRoutes` so everything works like it used to.

@jplatte what do you think?

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -247,14 +247,14 @@ where
     #[doc = include_str!("../docs/routing/merge.md")]
     pub fn merge<R>(mut self, other: R) -> Self
     where
-        R: HasRoutes<B>,
+        R: Into<Router<B>>,
     {
         let Router {
             routes,
             node,
             fallback,
             nested_at_root,
-        } = other.routes();
+        } = other.into();
 
         for (id, route) in routes {
             let path = node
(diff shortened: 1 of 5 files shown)


axum/src/routing/mod.rs
```diff
@@ -697,3 +700,73 @@ fn traits() {
...
+///
+/// let app = Router::new().merge(
+///     ToSameHandler::new(|| async {})
+///         .add_path("/foo")
+///         .add_path("/bar")
+///         .add_path("/baz")
+/// );
+/// # let _: Router<axum::body::Body> = app;
+/// ```
+pub trait HasRoutes<B> {
```
davidpdrsn, line 763: Any other suggestions for the name? I don't love `HasRoutes` but have no other ideas 🤔

LucioFranco: `MakeRouter`? lol

davidpdrsn: That's not too bad actually 🧐 I think I like that more.

jplatte (approved):
👍🏼

axum/src/routing/mod.rs
davidpdrsn: @jplatte do you have an opinion about the name? Otherwise I'm going with `MakeRouter` I think.

davidpdrsn: Could even use `Into<Router<B>>` instead of a dedicated trait.

jplatte: I don't care too much about the name. I can only add more suggestions, maybe you like one of them even more:

* `IntoRouter`
* `Into<Router>` (i.e. no new trait at all)

PR #821, opened 2022-03-03
Add note about order of extractors in the reverse-proxy example

FSMaxB:
## Motivation

When building a proxy handler based on the `reverse-proxy` example, I ran into an error because I put the `Request<Body>` extractor before my `Extension` extractors.

## Solution

Add a note to the `reverse-proxy` example that the order of the extractors is relevant and `Request<Body>` needs to come last.

### DIFF
--- a/examples/reverse-proxy/src/main.rs
+++ b/examples/reverse-proxy/src/main.rs
@@ -36,7 +36,12 @@ async fn main() {
         .unwrap();
 }
 
-async fn handler(Extension(client): Extension<Client>, mut req: Request<Body>) -> Response<Body> {
+async fn handler(
+    Extension(client): Extension<Client>,
+    // NOTE: Make sure to put the request extractor last because once the request
+    // is extracted, extensions can't be extracted anymore.
+    mut req: Request<Body>,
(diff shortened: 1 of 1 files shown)


examples/reverse-proxy/src/main.rs
```diff
@@ -36,7 +36,12 @@ async fn main() {
         .unwrap();
 }
 
-async fn handler(Extension(client): Extension<Client>, mut req: Request<Body>) -> Response<Body> {
+async fn handler(
+    Extension(client): Extension<Client>,
+    // NOTE: Make sure to put the request extractor last because one the request
+    // extracted, extensions can't be extracted anymore.
```
jplatte, line 42:
```suggestion
    // NOTE: Make sure to put the request extractor last because once the request
    // is extracted, extensions can't be extracted anymore.
```

FSMaxB: Fixed.

jplatte (approved):

davidpdrsn (approved):
It's also in the docs but can't hurt to put here as well.

PR #825, opened 2022-03-04
Add example for implementing `IntoResponseParts`

davidpdrsn:
Figure this might be good to have.

### DIFF
--- a/axum-core/src/response/into_response_parts.rs
+++ b/axum-core/src/response/into_response_parts.rs
@@ -9,6 +9,67 @@ use std::{
 };
 
 /// Trait for adding headers and extensions to a response.
+///
+/// # Example
+///
+/// ```rust
+/// use axum::{
+///     response::{ResponseParts, IntoResponse, IntoResponseParts, Response},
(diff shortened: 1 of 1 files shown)


axum-core/src/response/into_response_parts.rs
```diff
@@ -9,6 +9,67 @@ use std::{
...
+///             },
+///         }
+///
+///         Ok(res)
+///     }
+/// }
+///
+/// // Its also recommended to implement `IntoResponse` so `SetHeader` can be used on its own as
+/// // the response
+/// impl<'a> IntoResponse for SetHeader<'a> {
```
jplatte, line 52: Should there be a derive macro for this?

```rust
#[derive(IntoResponse)]
#[into_response(via(IntoResponseParts))]
```

jplatte (approved):

axum-core/src/response/into_response_parts.rs
davidpdrsn: Hm perhaps 🤔

PR #827, opened 2022-03-04
Add Host extractor

### LINKED ISSUE #826: Add host extractor
### Motivation

I'd like an easy way to see either the host section of the request.

### Proposal

An extractor to get the relevant information in a handler.

```rust
async fn foo(Host(host): Host) -> ...
```

### Alternatives

David mentioned on discord this is how warp does it https://docs.rs/warp/0.3.2/src/warp/filters/host.rs.html#1-96.

tbillington:
## Motivation

This pr closes #826, implementing the host extractor.

## Solution

1. Check the host portion of the `uri` to see if it's already been provided, returning if it has
2. Read from the `"host"` header value and return if it was present
3. Return an empty string if neither strategy yielded a value

### DIFF
--- a/axum/src/extract/host.rs
+++ b/axum/src/extract/host.rs
@@ -1,9 +1,8 @@
-use super::{FromRequest, RequestParts};
+use super::{FromRequest, RequestParts, rejection::{HostRejection, FailedToResolveHost}};
 use async_trait::async_trait;
-use std::{convert::Infallible};
 
 /// Extractor that extracts the host from a request.
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 pub struct Host(pub String);
 
 #[async_trait]
@@ -11,18 +10,26 @@ impl<B> FromRequest<B> for Host
 where
     B: Send,
 {
-    type Rejection = Infallible;
+    type Rejection = HostRejection;
 
     async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
+        if let Some(host) = req.headers().get(http::header::FORWARDED).and_then(|host| host.to_str().ok()) {
+            return Ok(Host(host.to_owned()));
+        }
+
+        if let Some(host) = req.headers().get("X-Forwarded-Host").and_then(|host| host.to_str().ok()) {
+            return Ok(Host(host.to_owned()));
+        }
+
+        if let Some(host) = req.headers().get(http::header::HOST).and_then(|host| host.to_str().ok()) {
+            return Ok(Host(host.to_owned()));
+        }
+
         if let Some(host) = req.uri().host() {
-            return Ok(Host(host.to_string()));
+            return Ok(Host(host.to_owned()));
         }
 
-        if let Some(Ok(host)) = req.headers().get("host").map(|host| host.to_str()) {
-            return Ok(Host(host.to_string()));
-        }
-
-        Ok(Host("".to_string()))
+        Err(HostRejection::FailedToResolveHost(FailedToResolveHost))
     }
 }
 
--- a/axum/src/extract/host.rs
+++ b/axum/src/extract/host.rs
@@ -1,7 +1,17 @@
-use super::{FromRequest, RequestParts, rejection::{HostRejection, FailedToResolveHost}};
+use super::{
+    rejection::{FailedToResolveHost, HostRejection},
+    FromRequest, RequestParts,
+};
 use async_trait::async_trait;
 
-/// Extractor that extracts the host from a request.
+const X_FORWARDED_HOST_HEADER_KEY: &'static str = "X-Forwarded-Host";
+
+/// Extractor that resolves the hostname of the request.
+///
+/// Hostname is resolved through the following, in order:
+/// - `X-Forwarded-Host` header
+/// - `Host` header
+/// - request target / URI
 #[derive(Debug, Clone)]
 pub struct Host(pub String);
 
@@ -15,11 +25,19 @@ where
     async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
         // todo: extract host from http::header::FORWARDED
 
-        if let Some(host) = req.headers().get("X-Forwarded-Host").and_then(|host| host.to_str().ok()) {
+        if let Some(host) = req
+            .headers()
+            .get(X_FORWARDED_HOST_HEADER_KEY)
+            .and_then(|host| host.to_str().ok())
+        {
             return Ok(Host(host.to_owned()));
         }
 
-        if let Some(host) = req.headers().get(http::header::HOST).and_then(|host| host.to_str().ok()) {
+        if let Some(host) = req
+            .headers()
+            .get(http::header::HOST)
+            .and_then(|host| host.to_str().ok())
+        {
             return Ok(Host(host.to_owned()));
         }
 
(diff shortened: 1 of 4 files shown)


axum/src/extract/host.rs
```diff
@@ -0,0 +1,59 @@
...
+#[async_trait]
+impl<B> FromRequest<B> for Host
+where
+    B: Send,
+{
+    type Rejection = Infallible;
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
+        if let Some(host) = req.uri().host() {
+            return Ok(Host(host.to_string()));
```
jplatte, line 18: Nitpick about the style: I would much prefer if all these `to_string()` calls on `str`s were `to_owned()` instead.

axum/src/extract/host.rs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        if let Some(Ok(host)) = req.headers().get("host").map(|host| host.to_str()) {
```
davidpdrsn, line 21: nit: I prefer this style

```suggestion
        if let Some(host) = req.headers().get(http::header::HOST).and_then(|host| host.to_str().ok()) {
```

axum/src/extract/host.rs
```diff
@@ -0,0 +1,59 @@
...
+            return Ok(Host(host.to_string()));
+        }
+
+        Ok(Host("".to_string()))
```
davidpdrsn, line 25: Hm I'm not sure what is right here. Returning an empty host feels odd 🤔 Maybe we should return a rejection instead. Then the user can use `Option<Host>` to explicitly make hosts optional.

axum/src/extract/host.rs
```diff
@@ -0,0 +1,59 @@
...
+    type Rejection = Infallible;
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
+        if let Some(host) = req.uri().host() {
+            return Ok(Host(host.to_string()));
+        }
+
+        if let Some(Ok(host)) = req.headers().get("host").map(|host| host.to_str()) {
+            return Ok(Host(host.to_string()));
+        }
```
tsing, line 23: Should this extractor also handles the `x-forwarded-host` header?

axum/src/extract/host.rs
```diff
@@ -0,0 +1,59 @@
...
+            Request::builder()
+                .header("host", "cats.fun")
+                .body(())
+                .unwrap(),
+        );
+        assert_eq!(
+            &Host::from_request(&mut req).await.unwrap().0,
+            "cats.fun"
+        );
+    }
```
davidpdrsn, line 58: I think we should port this test to use `TestClient` like the other tests. Then its a bit more true to life.

axum/src/extract/mod.rs
```diff
@@ -28,6 +28,8 @@ pub use self::{
     request_parts::{BodyStream, RawBody},
 };
 
+pub mod host;
```
davidpdrsn, line 31: Since this module only contains one type I think we should make the module private and `pub use` the type from here. Like we do with other extractors.

axum/src/extract/host.rs
```diff
@@ -0,0 +1,59 @@
+use super::{FromRequest, RequestParts};
+use async_trait::async_trait;
+use std::{convert::Infallible};
+
+/// Extractor that extracts the host from a request.
+#[derive(Debug, Clone, Default)]
```
davidpdrsn, line 6: Not sure this should implement `Default`. Whats the use case?

```suggestion
#[derive(Debug, Clone)]
```

davidpdrsn (changes requested):
Good stuff! Thanks for working on this 😃

axum/src/extract/host.rs
davidpdrsn: Yes I think so.

Prior art: actix-web looks at that https://docs.rs/actix-web/latest/actix_web/dev/struct.ConnectionInfo.html#method.host, as well as `Forwarded`.

That might involve a bunch of header parsing which can be hard so I'd be fine with adding this in a follow up PR.

tbillington: In the actix documentation they say they look at uri after all the headers, however I've done the opposite. If we're adopting the actix header suggestions should we also use the same ordering?

tbillington: I wonder is there some canonical/standard for resolving the host of a http request?

axum/src/extract/host.rs
tbillington: I'm unfamiliar with the codebase, do you have a file or test I should use as the current standard?

axum/src/extract/host.rs
tbillington: I like this, I initially wanted to return an Option but couldn't quite work out the types. I didn't realise the result could be converted into an option if that's how it was used in the fn def. What makes that work?

axum/src/extract/host.rs
tbillington: Whoops, good catch, accident from copy-pasting query extractor as a starting point.

axum/src/extract/mod.rs
tbillington: I've updated to use the same style, is that what you had in mind?

axum/src/extract/host.rs
tbillington: I've created an error type for this method, please check :)

axum/src/extract/host.rs
tbillington: 👍 sounds good

axum/src/extract/rejection.rs
```diff
@@ -65,6 +65,14 @@ define_rejection! {
     pub struct InvalidFormContentType;
 }
 
+define_rejection! {
+    #[status = BAD_REQUEST]
```
tbillington, line 69: I didn't really know which status to use, this probably isn't ideal

axum/src/extract/host.rs
davidpdrsn: https://github.com/tokio-rs/axum/blob/main/axum/src/routing/tests/mod.rs#L32-L63

axum/src/extract/rejection.rs
davidpdrsn: I think `Bad Request` is appropriate.

axum/src/extract/rejection.rs
```diff
@@ -65,6 +65,14 @@ define_rejection! {
...
+    #[body = "No host found in request"]
+    /// Rejection type used if the [`Host`](super::Host) extractor is unable to
+    /// resolve a host from it's supplied `RequestParts`.
```
davidpdrsn, line 72: nitpick:

```suggestion
    /// resolve a host.
```

axum/src/extract/host.rs
davidpdrsn: It would be cool if the [headers](https://docs.rs/headers/latest/headers) crate supported it but that doesn't seem to be the case.

axum/src/extract/host.rs
davidpdrsn: Looks good!

The `Option` and `Result` extract trick works because of https://github.com/tokio-rs/axum/blob/main/axum-core/src/extract/mod.rs#L217-L241

axum/src/extract/mod.rs
davidpdrsn: Yes!

axum/src/extract/host.rs
```diff
@@ -0,0 +1,64 @@
+use super::{FromRequest, RequestParts, rejection::{HostRejection, FailedToResolveHost}};
+use async_trait::async_trait;
+
+/// Extractor that extracts the host from a request.
```
davidpdrsn, line 4: I think we should document how we try to resolve the host, and in which order. [Similarly to what actix does.](https://docs.rs/actix-web/latest/actix_web/dev/struct.ConnectionInfo.html#method.host)

davidpdrsn (changes requested):
The code looks good!

Wanna add an entry to the [changelog](https://github.com/tokio-rs/axum/blob/main/axum/CHANGELOG.md) as well? Then we can merge this.

davidpdrsn (approved):
Thanks!

bluetech:
I don't think axum should trust `X-Forwarded-Host` blindly. The `X-Forwarded-*` headers are meant to be set by proxies, but a client is free to set them itself, so they should only be trusted when axum is running behind a proxy which sets or santitizes them. But this is circumstantial, especially `X-Forwarded-Host` is not set often. This will be a security issue if someone ends up using the `Host` extractor for something security-sensitive.

[Here's a recent article](https://adam-p.ca/blog/2022/03/x-forwarded-for/) about `X-Forwarded-For`, which is more complex/sensitive, but the gist is also relevant for `X-Forwarded-Host`.

tsing:
@bluetech I guess it also applies to the `Host` header as the a client is also free to set the `Host` header given the Axum is not running behind a proxy.

bluetech:
The client is supposed to set the Host header - it is expected. But the client is not supposed to set `X-Forwarded-Host` - if it does, that's unexpected. But it's still possible, so it must be considered.

bluetech:
My suggestion is to make it configurable, like [Django does](https://docs.djangoproject.com/en/4.0/ref/settings/#std:setting-USE_X_FORWARDED_HOST).

tbillington:
I'm struggling to see the practical implication.

- The application developer doesn't see which header the host came from
- The proxy will set x-forwarded-host if it is set up to, clobbering the user host in the process
- Fallsback to host header, which the user is supposed to set
- Optionally it could have been pulled from x-forwarded-host which the user set, but if we're already relying on them to inform us which host then does it matter?

tbillington:
Unless the user is somehow able to override the proxys set header, I don't know enough about proxy servers to know what's common/possible.

davidpdrsn:
@bluetech clients can also set the `Host` header to whatever they want.

```
curl localhost:3000 -H "Host: foobar"

[examples/hello-world/src/main.rs:25] &headers = {
    "host": "foobar",
    "user-agent": "curl/7.64.1",
    "accept": "*/*",
}
```

What makes `x-forwarded-host` different? Seems to me like you can never trust data from clients and have to be careful regardless which header you look at.

jplatte:
Hm, can JavaScript `fetch` set `x-forwarded-host` from the browser? Since it's checked first, that would allow to "override" the host set by the browser itself.

bluetech:
I started writing a longer explanation, but I found an article which does so already: https://portswigger.net/web-security/host-header/exploiting. Also https://portswigger.net/web-security/host-header.

davidpdrsn:
Still seems to me like devs are gonna have to validate the host header regardless.


@bluetech what do you think about [this](https://github.com/tokio-rs/axum/pull/827#issuecomment-1060330234):

> What makes x-forwarded-host different? Seems to me like you can never trust data from clients and have to be careful regardless which header you look at.

davidpdrsn:
I've spoken to the lead maintainer of actix-web and think we should keep things as they are now, but add a note to the docs about the security risks. I'll do that.

bluetech:
A warning is good, thanks.

I would still recommend following what Django does, namely [`ALLOWED_HOSTS`](https://docs.djangoproject.com/en/4.0/ref/settings/#allowed-hosts) (to validate `Host` - must be specified) and [`USE_X_FORWARDED_HOST`](https://docs.djangoproject.com/en/4.0/ref/settings/#std:setting-USE_X_FORWARDED_HOST) (to opt into using `X-Forwarded-Host` when it's safe - default false). Because a lot of people won't read the docs and won't bother to validate the host which will expose them to attacks.

I also looked at what Ruby on Rails does -- they once trusted `X-Forwarded-Host` but made some changes (I'm not sure what exactly... I'm less familiar with rails) - see https://github.com/rails/rails/issues/29893.

I look to Django and Ruby on Rails because they're the most popular web frameworks and have had to deal with every security issue under the sun, so "just do what they do" is a decent heuristic.

davidpdrsn:
Axum doesn't have something similar to Rails' `url_for`. Our closest thing is `Redirect` but still requires using piecing things together themselves. Axum doesn't implicitly use the host behind the scenes. To me this makes it seem like a different thing. That seems like the crux of Rails' issues.

bluetech:
Right, I guess it depends on what is the usecase for the Host extractor in the first place.

PR #829, opened 2022-03-04
Replace set_var usage in examples

jplatte:
## Motivation

`set_var`'s safety is at odds with `std` calling into libc which can read the environment without taking a lock at basically any point. Rust is moving towards deprecating the ability to call it in safe code, or the whole function (which approach wins is not yet decided AFAIK).

## Solution

Don't use `set_var`. It's possible to supply a default value to the tracing_subscriber `EnvFilter` if `RUST_LOG` isn't set in different ways (I have taken the easiest approach I could find).

Side note: The examples are a bit inconsistent about whether they

* Initialize `tracing_subscriber` with a `RUST_LOG` fallback value
* Initialize `tracing_subscriber` without `RUST_LOG` fallback
* Don't initialize `tracing_subscriber` at all

Should we initialize it the same way everywhere?

### DIFF
--- a/examples/chat/Cargo.toml
+++ b/examples/chat/Cargo.toml
@@ -10,4 +10,4 @@ futures = "0.3"
 tokio = { version = "1", features = ["full"] }
 tower = { version = "0.4", features = ["util"] }
 tracing = "0.1"
-tracing-subscriber = { version="0.3", features = ["env-filter"] }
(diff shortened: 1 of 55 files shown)


davidpdrsn (approved):
Awesome! Thanks for fixing this.

It would be nice if the examples were consistent but shouldn't block merging this.

PR #831, opened 2022-03-05
ci: Set CARGO_TERM_COLOR = always to get colored logs

jplatte:


jplatte:
Before: https://github.com/tokio-rs/axum/runs/5428599104?check_suite_focus=true#step:6:1
After: https://github.com/tokio-rs/axum/runs/5432508235?check_suite_focus=true#step:6:1

davidpdrsn (approved):
😍

Issue #678, opened 2021-12-30
Better support for overlapping parameters


ibraheemdev:
Currently routes like `/:a/:b` and `/:a` are supported, but only if they have a common prefix (`/:a`). This means a route like `/:a/:b` conflicts with `/:c`. Having the user forced to change the parameter named doesn't feel very nice.

cc ibraheemdev/matchit#13

davidpdrsn:
I've spent a bunch of time working on this and while its possible it also adds a bunch of extra complexity.

The solution is to normalize the params so routes like `/:a/foo/:b` would become `/:axum_param_0/foo/:axum_param/2`. So split the path on `/` and rename params based on their depth. You then also need to store the mapping back to the original param name so `Path` can access it.

Opaque nested services is also fun since you need to carry the param mapping forward via a request extension so inner routers can access it. You also need to be careful that inner routers don't use the same normalized param names as outer routers which, since inner routers are created first.

All these things are do-able but I don't feel they're worth the additional complexity. I haven't heard anyone else ask about this so I'd consider this an edge case. Ideally it should be fixed in matchit.

PR #833, opened 2022-03-05
Refactor storing URL params in extensions

davidpdrsn:
I've been working on https://github.com/tokio-rs/axum/issues/678 but did refactoring of path params first. Thought I'd submit that as its own PR to make reviewing easier.

### DIFF
--- a/axum/src/routing/url_params.rs
+++ b/axum/src/routing/url_params.rs
@@ -1,0 +1,43 @@
+use crate::util::{ByteStr, PercentDecodedByteStr};
+use http::Extensions;
+use matchit::Params;
+
+pub(crate) enum UrlParams {
+    Params(Vec<(ByteStr, PercentDecodedByteStr)>),
+    InvalidUtf8InPathParam { key: ByteStr },
+}
+
+pub(super) fn insert_url_params(extensions: &mut Extensions, params: Params) {
+    if let Some(UrlParams::InvalidUtf8InPathParam { .. }) = extensions.get() {
+        // nothing to do here since an error was stored earlier
+        return;
+    }
+
+    let params = params
+        .iter()
+        .filter(|(key, _)| !key.starts_with(super::NEST_TAIL_PARAM))
+        .map(|(key, value)| (key.to_owned(), value.to_owned()))
+        .map(|(k, v)| {
+            if let Some(decoded) = PercentDecodedByteStr::new(v) {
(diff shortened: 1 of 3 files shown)


axum/src/routing/url_params.rs
```diff
@@ -0,0 +1,43 @@
...
+        .map(|(k, v)| {
+            if let Some(decoded) = PercentDecodedByteStr::new(v) {
+                Ok((ByteStr::new(k), decoded))
+            } else {
+                Err(ByteStr::new(k))
+            }
+        })
+        .collect::<Result<Vec<_>, _>>();
+
+    match (extensions.get_mut(), params) {
```
jplatte, line 29: Is there any reason to look up the extension a second time here instead of just beginning the function with `let extension = extensions.get_mut();` and basing the short-circuiting on that?

axum/src/routing/url_params.rs
```diff
@@ -0,0 +1,43 @@
...
+
+pub(super) fn insert_url_params(extensions: &mut Extensions, params: Params) {
+    if let Some(UrlParams::InvalidUtf8InPathParam { .. }) = extensions.get() {
+        // nothing to do here since an error was stored earlier
+        return;
+    }
+
+    let params = params
+        .iter()
+        .filter(|(key, _)| !key.starts_with(super::NEST_TAIL_PARAM))
```
jplatte, line 18: This seems a bit weird, will this special-cased parameter name go away with the next PR that normalizes path params?

axum/src/routing/url_params.rs
davidpdrsn: Nope. I just missed it

axum/src/routing/url_params.rs
davidpdrsn: This has to do with nested opaque services. Its implemented as a wildcard route like `/path/*NEST_TAIL_PARAM`. So [`NEST_TAIL_PARAM`](https://github.com/tokio-rs/axum/blob/main/axum/src/routing/mod.rs#L101) is reserved by axum and shouldn't pollute the path params the user might see if using `Path<HashMap<String, String>>`.

jplatte: I see. I think it could use being more obviously special in that case, something like `__private__axum_nest_tail_param`.

jplatte (approved):

PR #839, opened 2022-03-07
Add security note about `extract::Host`

davidpdrsn:
As discussed in https://github.com/tokio-rs/axum/pull/827 users should take care to validate the value extracted by `Host`. This adds a note about that to the docs.

### DIFF
--- a/axum/src/extract/host.rs
+++ b/axum/src/extract/host.rs
@@ -13,8 +13,8 @@ const X_FORWARDED_HOST_HEADER_KEY: &str = "X-Forwarded-Host";
 /// - `Host` header
 /// - request target / URI
 ///
-/// Note that user agents can set `X-Forwarded-Host` and `Host` headers to arbitrarily so make sure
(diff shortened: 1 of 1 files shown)


axum/src/extract/host.rs
```diff
@@ -4,14 +4,17 @@ use super::{
...
+const X_FORWARDED_HOST_HEADER_KEY: &str = "X-Forwarded-Host";
 
 /// Extractor that resolves the hostname of the request.
 ///
 /// Hostname is resolved through the following, in order:
 /// - `X-Forwarded-Host` header
 /// - `Host` header
 /// - request target / URI
+///
+/// Note that user agents can set `X-Forwarded-Host` and `Host` headers to arbitrarily so make sure
```
jplatte, line 16: Should be either of "to arbitrary values" or "arbitrarily", not a mix of both 😉

jplatte (approved):

PR #840, opened 2022-03-08
Added example project todos: SQLX + Tera + PostgreSQL.

Jerrody:
I learned backend for about 2-3 days, so I don't have any experience. Did a pet project and thought it would be a good example for `axum`.

Current issues:
- The first account registration is very slow, but other registrations are fast, I don't understand what the problem is.

P.S.
I'm new to the backend, so my code (I'm sure) might be of poor quality. If you fix something, make a code review and point out errors, I will be very grateful.

### DIFF
--- a/examples/todos_sqlx_postgres_tera/Cargo.toml
+++ b/examples/todos_sqlx_postgres_tera/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "todos_sqlx_postgres_tera"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+axum = "0.4"
+serde = { version = "1", features = ["derive"] }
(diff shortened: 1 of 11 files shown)


davidpdrsn (changes requested):
I'm sorry but I think don't think we will accept this example. Some reasons:

- Crates in this repo cannot depend on `tower-cookies` because it also depends on axum. Then we'll get dependency issues with making breaking changes in this repo, which we're currently doing.
- What the example covers is already covered by other examples. We already have sqlx and templating examples. We don't have a tera example but I don't think its relevant to make examples for every lib under the sun. Each example creates more work the maintainers.
- I don't think requiring users to run migrations is appropriate. The examples should focus on the axum specific bits. Which exact tables you have isn't important.
- User's also tend to think of examples as "the axum approved way of doing things", even if that isn't their goal. Another reason to keep the examples small.

You're welcome to host the repo yourself and add a link to it in the "Project showcase" [here](https://github.com/tokio-rs/axum/blob/main/ECOSYSTEM.md).

Issue #830, opened 2022-03-04
Tests for truly nested routing?


jplatte:
After the discussion in #714, I'm wondering whether all those variants (the updated behavior as of #824) work the same way when the inner router is wrapped in another service, and we hit the `Err` branch of [`try_downcast::<Router<B>, _>(svc)`](https://github.com/tokio-rs/axum/blob/f045f2356c16cd31c257ca3f41a9600d8ae64b01/axum/src/routing/mod.rs#L193). Do the tests cover this scenario already?

davidpdrsn:
In general it seems we need better test coverage for nested opaque routes. This one fails:

```rust
#[tokio::test]
async fn nest_at_capture() {
    let api_routes = Router::new().route("/:b", get(|| async {})).boxed_clone();

    let app = Router::new().nest("/:a", api_routes);

    let client = TestClient::new(app);

    let res = client.get("/foo/bar").send().await;
    assert_eq!(res.status(), StatusCode::OK);
}
```

It gets 404.

PR #841, opened 2022-03-08
Fix nesting of opaque services that paths that contain params

davidpdrsn:
While working on https://github.com/tokio-rs/axum/issues/830 I discovered that nesting opaque services at paths that contain params was totally broken. Things like

```rust
let api_routes = Router::new().route("/:b", get(|| async {})).boxed_clone();
let app = Router::new().nest("/:a", api_routes);
```

We would try and strip the prefix as `/:a` but thats a placeholder that should match anything. It shouldn't be used as the literal prefix.

This required some changes to [`StripPrefix`](https://github.com/tokio-rs/axum/blob/main/axum%2Fsrc%2Frouting%2Fstrip_prefix.rs). I also added a bunch more tests and some property tests to get better coverage against panics.

### DIFF
--- a/axum/src/routing/strip_prefix.rs
+++ b/axum/src/routing/strip_prefix.rs
@@ -148,7 +148,7 @@ mod tests {
             $name:ident,
             uri = $uri:literal,
             prefix = $prefix:literal,
-            expected = $expected:expr $(,)?
+            expected = $expected:expr,
         ) => {
             #[test]
             fn $name() {
@@ -159,7 +159,7 @@ mod tests {
         };
     }
 
-    test!(empty, uri = "/", prefix = "/", expected = Some("/"));
+    test!(empty, uri = "/", prefix = "/", expected = Some("/"),);
 
     test!(
         single_segment,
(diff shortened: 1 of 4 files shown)


davidpdrsn:
Honestly surprised no one ran into this bug before. I guess users are either nesting `ServeDir` or `Router`.

davidpdrsn:
I swear I had the tests passing locally but I must have fat fingered something. I'll get it fixed.

axum/src/routing/strip_prefix.rs
```diff
@@ -35,43 +34,402 @@ where
...
+        uri = "/a/b",
+        prefix = "/a",
+        expected = Some("/b")
+    );
+
+    test!(
+        multi_segment_2,
+        uri = "/b/a",
+        prefix = "/a",
+        expected = None,
```
jplatte, line 234: Nitpick, would make sense to be consistent about trailing commas after `expected = …` 😉

jplatte (approved):

PR #842, opened 2022-03-08
Fix routing for opaque nested services

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/830

This brings nesting of opaque services inline with nesting `Router`s, except for one case which is explained in a test. Also added a bunch more comments to the prefix stripping so I'll still be able to understanding it two days from now 😅

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -236,8 +236,8 @@ where
             }
             // otherwise we add a wildcard route to the service
             Err(svc) => {
-                let path = if path == "/" {
-                    format!("/*{}", NEST_TAIL_PARAM)
+                let path = if path.ends_with('/') {
+                    format!("{}*{}", path, NEST_TAIL_PARAM)
                 } else {
                     format!("{}/*{}", path, NEST_TAIL_PARAM)
                 };
--- a/axum/src/routing/strip_prefix.rs
+++ b/axum/src/routing/strip_prefix.rs
@@ -42,8 +42,7 @@ where
 }
 
 fn strip_prefix(uri: &Uri, prefix: &str) -> Option<Uri> {
-    let path_and_query: http::uri::PathAndQuery = if let Some(path_and_query) = uri.path_and_query()
-    {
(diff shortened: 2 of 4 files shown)


axum/src/routing/mod.rs
```diff
@@ -236,8 +236,8 @@ where
             }
             // otherwise we add a wildcard route to the service
             Err(svc) => {
-                let path = if path == "/" {
-                    format!("/*{}", NEST_TAIL_PARAM)
+                let path = if path.ends_with('/') {
+                    format!("{}*{}", path, NEST_TAIL_PARAM)
```
davidpdrsn, line 240: Before we would add a double slash to `nest("/foo/", _)`.

axum/src/routing/strip_prefix.rs
```diff
@@ -334,6 +350,13 @@ mod tests {
         expected = Some("/"),
     );
 
+    test!(
+        param_13,
+        uri = "/a/a",
+        prefix = "/a/",
+        expected = Some("/a"),
+    );
```
davidpdrsn, line 358: Case that wasn't handled properly before.

axum/src/routing/tests/nest.rs
```diff
@@ -391,7 +433,22 @@ nested_route_test!(nest_3, nest = "", route = "/a/", expected = "/a/");
 nested_route_test!(nest_4, nest = "/", route = "/", expected = "/");
 nested_route_test!(nest_5, nest = "/", route = "/a", expected = "/a");
 nested_route_test!(nest_6, nest = "/", route = "/a/", expected = "/a/");
-nested_route_test!(nest_7, nest = "/a", route = "/", expected = "/a");
+
+// This case is different for opaque services.
+//
+// The internal route becomes `/a/*__private__axum_nest_tail_param` which, according to matchit
+// doesn't match `/a`. However matchit detects that a route for `/a/` is exists and so it issues a
```
jplatte, line 440:
```suggestion
// doesn't match `/a`. However matchit detects that a route for `/a/` exists and so it issues a
```

jplatte (approved):

PR #861, opened 2022-03-16
Add example for consuming request body in middleware and extractor

davidpdrsn:
Gotten a few questions lately about how to consume the request body in middleware and extractors. So this adds an example for it.

### DIFF
--- a/examples/consume-body-in-extractor-or-middleware/src/main.rs
+++ b/examples/consume-body-in-extractor-or-middleware/src/main.rs
@@ -1,0 +1,97 @@
+//! Run with
+//!
+//! ```not_rust
+//! cargo run -p example-consume-body-in-extractor-or-middleware
+//! ```
+
+use axum::{
+    async_trait,
+    body::{self, BoxBody, Bytes, Full},
+    extract::{FromRequest, RequestParts},
+    http::{Request, StatusCode},
+    middleware::{self, Next},
+    response::{IntoResponse, Response},
+    routing::post,
+    Router,
+};
+use std::net::SocketAddr;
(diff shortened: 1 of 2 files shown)


examples/consume-body-in-extractor-or-middleware/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+    type Rejection = Response;
+
+    async fn from_request(req: &mut RequestParts<BoxBody>) -> Result<Self, Self::Rejection> {
+        let request = Request::from_request(req)
+            .await
+            .map_err(|err| err.into_response())?;
+
+        let request = buffer_request_body(request).await?;
+
+        *req = RequestParts::new(request);
```
davidpdrsn, line 70: This business with recreating the `RequestParts` is necessary because `RequestParts::body_mut` returns `Option<&mut B>` and not `&mut Option<B>`. That means once you've taken the body, you cannot put it back. I'm thinking we should change that for 0.5.

jplatte: Seems reasonable.

jplatte (commented):
With how the code is layed out, I find that I have to jump around a lot to understand things. Can you move the extractor next to the handler function?

davidpdrsn:
@jplatte Done!

examples/consume-body-in-extractor-or-middleware/src/main.rs
davidpdrsn: I'll do that in a separate PR then 😊

jplatte (approved):

PR #868, opened 2022-03-17
Make status codes for `JsonRejection` more precise

### LINKED ISSUE #865: HTTP Status on Invalid JSON Request
When I send a request where the `content-type` is not `application/json` I get the answer `415 Unsupported Media Type` and if the JSON parsing fails I get `422 Unprocessable Entity`

Two issues:

1. This is not documented correctly: https://docs.rs/axum/latest/axum/struct.Json.html In this documentation both of them would result in `400 Bad Request`

2. `422 Unprocessable Entity` seems to be part of an HTTP extension intended for WebDAV. Is it appropriate to use this in a non-WebDAV server?

3. The description of `422 Unprocessable Entity` https://datatracker.ietf.org/doc/html/rfc4918#section-11.2 says that this should be used for syntactically correct, but semantically wrong requests. I also get that with a syntactically wrong JSON like `}`

davidpdrsn:
Fixes #865

### DIFF
--- a/axum/src/extract/rejection.rs
+++ b/axum/src/extract/rejection.rs
@@ -9,10 +9,24 @@ pub use axum_core::extract::rejection::*;
 #[cfg(feature = "json")]
 define_rejection! {
     #[status = UNPROCESSABLE_ENTITY]
+    #[body = "Failed to deserialize the JSON body into the target type"]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    /// Rejection type for [`Json`](super::Json).
+    ///
+    /// This rejection is used if the request body is syntactically valid JSON but couldn't be
+    /// serialized into the target type.
+    pub struct JsonDataError(Error);
+}
+
+#[cfg(feature = "json")]
+define_rejection! {
+    #[status = BAD_REQUEST]
     #[body = "Failed to parse the request body as JSON"]
     #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
     /// Rejection type for [`Json`](super::Json).
-    pub struct InvalidJsonBody(Error);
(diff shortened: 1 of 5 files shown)


axum/src/extract/rejection.rs
```diff
@@ -9,10 +9,24 @@ pub use axum_core::extract::rejection::*;
 #[cfg(feature = "json")]
 define_rejection! {
     #[status = UNPROCESSABLE_ENTITY]
+    #[body = "Failed to deserialize the JSON body into the target type"]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    /// Rejection type for [`Json`](super::Json).
+    ///
+    /// This rejection is used if the request body is syntactically valid JSON but couldn't be
+    /// serialized into the target type.
```
jplatte, line 17:
```suggestion
    /// deserialized into the target type.
```

axum/src/extract/rejection.rs
```diff
@@ -9,10 +9,24 @@ pub use axum_core::extract::rejection::*;
...
+
+#[cfg(feature = "json")]
+define_rejection! {
+    #[status = BAD_REQUEST]
     #[body = "Failed to parse the request body as JSON"]
     #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
     /// Rejection type for [`Json`](super::Json).
-    pub struct InvalidJsonBody(Error);
+    ///
+    /// This rejection is used if the request body is didn't contain syntactically valid JSON.
```
jplatte, line 28:
```suggestion
    /// This rejection is used if the request body didn't contain syntactically valid JSON.
```

jplatte (approved):

PR #869, opened 2022-03-18
Make `RequestParts::body_mut` return `&mut Option<B>`

davidpdrsn:
So users can swap the body, which isn't possible with an `Option<&mut B>`.

### DIFF
--- a/axum-core/src/extract/mod.rs
+++ b/axum-core/src/extract/mod.rs
@@ -204,7 +204,7 @@ impl<B> RequestParts<B> {
     /// Gets a mutable reference to the request body.
     ///
     /// Returns `None` if the body has been taken by another extractor.
-    // this returns `&mut Option<B>` rather than `Option<&mut B>` such that users can swap it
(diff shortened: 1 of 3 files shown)


axum-core/src/extract/mod.rs
```diff
@@ -204,8 +204,9 @@ impl<B> RequestParts<B> {
     /// Gets a mutable reference to the request body.
     ///
     /// Returns `None` if the body has been taken by another extractor.
-    pub fn body_mut(&mut self) -> Option<&mut B> {
-        self.body.as_mut()
+    // this returns `&mut Option<B>` rather than `Option<&mut B>` such that users can swap it
```
jplatte, line 207: Missing period and I think "swap it" is a bit unclear, maybe something like this?

```suggestion
    // this returns `&mut Option<B>` rather than `Option<&mut B>` such that users can use it to set the body.
```

jplatte (approved):

PR #870, opened 2022-03-18
Exclude sqlx example

davidpdrsn:
Getting an ICE https://github.com/tokio-rs/axum/runs/5602144413?check_suite_focus=true. Will look into that later, can't reproduce locally.


davidpdrsn:
huh it appears to be because of something in nightly. Changed CI to run `check` on beta.

PR #872, opened 2022-03-18
chore: add single page application example

ttys3:
## Motivation

I see https://github.com/tokio-rs/axum/discussions/867, think we can add an example here

## Solution

a simple SPA with 3 page routes, wihout api request (to avoid frontend code complex)

image

### DIFF
--- a/examples/single-page-application/Cargo.toml
+++ b/examples/single-page-application/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "example-single-page-application"
+version = "0.1.0"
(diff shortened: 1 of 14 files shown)


davidpdrsn (changes requested):
I'm actually working adding a dedicated SPA helper to axum-extra. So I don't think we need this example.

Also not interesting in having javascript in the repo that I have to maintain 😊

ttys3:
OK, I'll keep it under  https://github.com/ttys3/axum-single-page-application-example

PR #881, opened 2022-03-23
Make CI Workflow faster

TaKO8Ki:
## Motivation

This pull request disables incremental compilation on CI.

> CI builds often are closer to from-scratch builds, as changes are typically much bigger than from a local edit-compile cycle. For from-scratch builds, incremental adds an extra dependency-tracking overhead. It also significantly increases the amount of IO and the size of ./target, which make caching less effective.

ref: https://matklad.github.io/2021/09/04/fast-rust-builds.html

## Solution

Add `CARGO_INCREMENTAL: 0` to CI.yml.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -2,6 +2,7 @@ name: CI
 
 env:
   CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: 0
 
 on:
   push:


jplatte:
The rust-cache action we use does this already. We don't need to set it separately.

TaKO8Ki:
@jplatte Oh. I understand.

PR #885, opened 2022-03-24
add ezsockets to community maintained ecosyste

gbaranski:


stanbar (changes requested):
As a rule of thumb always prefer "... that ..." over "..., which ...".

Whenever you can, skip "that" completely. 
```
- [ezsockets](https://github.com/gbaranski/ezsockets): Easy to use WebSocket library that integrates with Axum.
```

davidpdrsn (approved):

PR #889, opened 2022-03-30
Accept &str for Redirect

### LINKED ISSUE #888: Redirects should support URL fragments
### Version

```
├── axum v0.4.8
│   ├── axum-core v0.1.2
```

### Platform

`Darwin ... 21.4.0 Darwin Kernel Version 21.4.0: Mon Feb 21 20:35:58 PST 2022; root:xnu-8020.101.4~2/RELEASE_ARM64_T6000 arm64`

### Description

I tried this code:

```
pub async fn handler() -> Result<Redirect> {
  let location = "/path/after/login#access_token=blah";
  let location = Uri::from_str(location).unwrap();
  Ok(Redirect::temporary(location))
}
```

I expected to have the browser redirected to the proper URL, including the URL fragment.

Instead, the `Uri` parsing strips the fragment so the browser is redirected to `/path/after/login` without the access token.

I understand this is the intended behavior from the `http::Uri` but, if so, axum's `Redirect` should probably accept a `url::Url` or just a string instead of a `Uri`.

For the time being, I'll just create an `axum::Response` myself and set the `Location` header manually.

emschwartz:
## Motivation

`http::Uri` strips URL fragments, which are important for redirects, for example for login flows with access tokens in the fragment.

## Solution

As discussed in https://github.com/tokio-rs/axum/issues/888, this changes the `Redirect` struct to accept a `&str` instead of a `Uri`.

Resolves #888

### DIFF
--- a/axum-extra/src/extract/cookie.rs
+++ b/axum-extra/src/extract/cookie.rs
@@ -45,7 +45,7 @@ pub use cookie_lib::{Cookie, Key};
 ///             // the updated jar must be returned for the changes
 ///             // to be included in the response
 ///             jar.add(Cookie::new("session_id", session_id)),
-///             Redirect::to("/me".parse().unwrap()),
+///             Redirect::to("/me"),
 ///         ))
 ///     } else {
 ///         Err(StatusCode::UNAUTHORIZED)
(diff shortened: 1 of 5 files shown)


jplatte:
CI fails because you didn't update the example in the docs for the `Redirect` type.

davidpdrsn (approved):
Thanks!

PR #898, opened 2022-04-01
axum-extra: re-export the Expiration and SameSite structs from the co…

paolobarbolini:
## Motivation

Makes it possible to set `SameSite` cookies without having to import the `cookie` crate

## Solution

Re-exports it. While at it I also re-exported `Expiration`


paolobarbolini:
Done

davidpdrsn (approved):

davidpdrsn:
Thanks!

Issue #896, opened 2022-03-31
Add Private Cookies to Axums Cookie Extension


genusistimelord:
### Motivation

I would like to have the possibility to use Private cookies along side regular cookies

### Proposal

Could make it so a function can be called to implement a Signed private cookie from the signed cookie jar or regular Cookies jar or both.

### Alternatives

Create its own Structure with its own features that only does private. But make sure it is possible to call private with public cookies together.

PR #900, opened 2022-04-01
`PrivateCookieJar` for managing private cookies

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/896

Its pretty much identical to `SignedCookieJar` but private.

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -32,7 +32,7 @@ axum-macros = { path = "../axum-macros", version = "0.2", optional = true }
 serde = { version = "1.0", optional = true }
 serde_json = { version = "1.0.71", optional = true }
 percent-encoding = { version = "2.1", optional = true }
-cookie-lib = { package = "cookie", version = "0.16", features = ["percent-encode", "signed"], optional = true }
+cookie-lib = { package = "cookie", version = "0.16", features = ["percent-encode", "signed", "private"], optional = true }
 
 [dev-dependencies]
 axum = { path = "../axum", version = "0.5", features = ["headers"] }
--- a/axum-extra/src/extract/cookie.rs
+++ b/axum-extra/src/extract/cookie.rs
@@ -1,6 +1,6 @@
 //! Cookie parsing and cookie jar management.
 //!
-//! See [`CookieJar`] and [`SignedCookieJar`] for more details.
+//! See [`CookieJar`], [`SignedCookieJar`], and [`PrivateCookieJar`] for more details.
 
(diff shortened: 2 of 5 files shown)


axum-extra/Cargo.toml
```diff
@@ -32,7 +32,7 @@ axum-macros = { path = "../axum-macros", version = "0.2", optional = true }
 serde = { version = "1.0", optional = true }
 serde_json = { version = "1.0.71", optional = true }
 percent-encoding = { version = "2.1", optional = true }
-cookie-lib = { package = "cookie", version = "0.16", features = ["percent-encode", "signed"], optional = true }
+cookie-lib = { package = "cookie", version = "0.16", features = ["percent-encode", "signed", "private"], optional = true }
```
jplatte, line 35: Maybe pass through the features instead? I.e. have a `signed-cookies` feature that activates `SignedCookieJar` and `cookie-lib/signed`, equivalent for `private` / `PrivateCookieJar`.

axum-extra/src/extract/cookie.rs
```diff
@@ -15,6 +15,9 @@ use http::{
 };
 use std::{convert::Infallible, fmt, marker::PhantomData};
 
+mod private;
+
+pub use self::private::PrivateCookieJar;
```
davidpdrsn, line 20: Gonna move the plaintext and signed versions to their own mods in a follow up PR

jplatte (commented):
Since the underlying cookie library seems to have some support for this using `CookieJar::private_jar` and such, should we maybe have the ability to manage some plain, some signed / private cookies using a single jar?

axum-extra/Cargo.toml
davidpdrsn: I played around with some time ago but couldn't find a decent way of passing in the `Key`. You can get it manually via an extension and pass it in when passing `.private_jar(key)` but didn't like that too much 🤷

davidpdrsn: and yeah did think about adding separate features for enabling signed or private jars since it pulls in a few deps but thats a breaking change and we just shipped a new major yesterday :P

davidpdrsn: yeah thats true :P

jplatte (approved):
Raised all questions I could think off, I'll you decide how to proceed with crate features and stuff like that.

jplatte:
Oh, I didn't realize auto-merge was enabled. Well, any changes can still be done in separate PRs 🙂

PR #901, opened 2022-04-02
Add `Field::chunk`

itstristanb:
Added the function `chunk` for `exum::extract's multipart Field` to give a clearer option to use rather than try_next.

## Motivation

I needed a way to stream large files from a frontend file uploader directly into a file. These files can be 10gb+ in size, and need to be paused at any time. Though this is doable with try_next, I felt that it was clearer from the user's side to have chunk.

## Solution

By implementing the `chunk` function for `axum::extract's multipart Field`, readability of code and nicer access the underlying multer::Field's `chunk` funciton is allowed.

## Change log
1. Added chunk function to multipart field
2. Changed to doc comment for clarification of usage, and changed `&self` to `&mut self`
3. Fixed formatting
4. Cleaned up nits and correct example to reflex best practices
5. Removed last unwrap

### DIFF
--- a/axum/src/extract/multipart.rs
+++ b/axum/src/extract/multipart.rs
@@ -147,21 +147,23 @@ impl<'a> Field<'a> {
     ///
     /// When the field data has been exhausted, this will return [`None`].
     ///
-    /// This does the exact same thing as the `Stream` impl and is for convenience
+    /// Note this does the same thing as `Field`'s [`Stream`] implementation.
     ///
-    /// # Examples
+    /// # Example
     ///
     /// ```
     /// use axum::{
     ///    extract::Multipart,
     ///    routing::post,
+    ///    response::IntoResponse,
     ///    Router,
     /// };
+    /// use http::StatusCode;
     ///
-    /// async fn upload(mut multipart: Multipart) {
-    ///     while let Some(mut field) = multipart.next_field().await.unwrap() {
+    /// async fn upload(mut multipart: Multipart) -> impl IntoResponse {
+    ///     while let Some(mut field) = multipart.next_field().await.map_err(|err| (StatusCode::BAD_REQUEST, err.to_string()))? {
     ///         while let Some(chunk) = field.chunk().await.unwrap() {
-    ///             println!("Chunk: {:?}", chunk);
+    ///             println!("received {} bytes", chunk.len());
     ///         }
     ///     }
     /// }
--- a/axum/src/extract/multipart.rs
+++ b/axum/src/extract/multipart.rs
@@ -162,7 +162,7 @@ impl<'a> Field<'a> {
     ///
     /// async fn upload(mut multipart: Multipart) -> impl IntoResponse {
     ///     while let Some(mut field) = multipart.next_field().await.map_err(|err| (StatusCode::BAD_REQUEST, err.to_string()))? {
-    ///         while let Some(chunk) = field.chunk().await.unwrap() {
+    ///         while let Some(chunk) = field.chunk().await {
     ///             println!("received {} bytes", chunk.len());
     ///         }
     ///     }
--- a/axum/src/extract/multipart.rs
+++ b/axum/src/extract/multipart.rs
@@ -143,8 +143,32 @@ impl<'a> Field<'a> {
         self.inner.text().await.map_err(MultipartError::from_multer)
     }
     
-    // Stream a chunk of the field data.
-    pub async fn chunk(self) -> Result<Option<Bytes>, MultipartError> {
+    /// Stream a chunk of the field data.
+    ///
+    /// When the field data has been exhausted, this will return [`None`].
+    ///
+    /// This does the exact same thing as the `Stream` impl and is for convenience
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use axum::{
+    ///    extract::Multipart,
+    ///    routing::post,
+    ///    Router,
+    /// };
+    ///
+    /// async fn upload(mut multipart: Multipart) {
+    ///     while let Some(mut field) = multipart.next_field().await.unwrap() {
+    ///         while let Some(chunk) = field.chunk().await.unwrap() {
+    ///             println!("Chunk: {:?}", chunk);
+    ///         }
+    ///     }
+    /// }
+    ///
+    /// let app = Router::new().route("/upload", post(upload));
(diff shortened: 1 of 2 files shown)


rmxbalanque (approved):
Looks good to me.

axum/src/extract/multipart.rs
```diff
@@ -142,6 +142,11 @@ impl<'a> Field<'a> {
     pub async fn text(self) -> Result<String, MultipartError> {
         self.inner.text().await.map_err(MultipartError::from_multer)
     }
+    
+    // Stream a chunk of the field data.
```
davidpdrsn, line 146: This should be a doc comment.

I think we should also include

> When the field data has been exhausted, this will return `None`.

It's in multer's docs as well.

I think we should also mention that this does the exact same as the `Stream` implementation. Otherwise that isn't clear.

axum/src/extract/multipart.rs
```diff
@@ -142,6 +142,11 @@ impl<'a> Field<'a> {
...
+    pub async fn chunk(self) -> Result<Option<Bytes>, MultipartError> {
```
davidpdrsn, line 147: I think this needs to take `&mut self` otherwise you can only access the first chunk.

davidpdrsn (changes requested):
Makes sense to add as convenience even though it's the same as the `Stream` impl.

axum/src/extract/multipart.rs
itstristanb: Yup make sense. This has been added along with an example

axum/src/extract/multipart.rs
itstristanb: Was debating this as well. Updated it to use `&mut self`

itstristanb:
@davidpdrsn missed a formatting issue in the initial pr cuasing cli to fail. it is fixed now but needs further approval.

Thanks.

axum/src/extract/multipart.rs
```diff
@@ -142,6 +142,38 @@ impl<'a> Field<'a> {
     pub async fn text(self) -> Result<String, MultipartError> {
         self.inner.text().await.map_err(MultipartError::from_multer)
     }
+
+    /// Stream a chunk of the field data.
+    ///
+    /// When the field data has been exhausted, this will return [`None`].
+    ///
+    /// This does the exact same thing as the `Stream` impl and is for convenience
```
davidpdrsn, line 150: I think this reads a bit nicer

```suggestion
    /// Note this does the same thing as `Field`'s [`Stream`] implementation.
```

axum/src/extract/multipart.rs
```diff
@@ -142,6 +142,38 @@ impl<'a> Field<'a> {
...
+    ///
+    /// # Examples
```
davidpdrsn, line 152: Since there is just one example

```suggestion
    /// # Example
```

axum/src/extract/multipart.rs
```diff
@@ -142,6 +142,38 @@ impl<'a> Field<'a> {
...
+    /// use axum::{
+    ///    extract::Multipart,
+    ///    routing::post,
+    ///    Router,
+    /// };
+    ///
+    /// async fn upload(mut multipart: Multipart) {
+    ///     while let Some(mut field) = multipart.next_field().await.unwrap() {
+    ///         while let Some(chunk) = field.chunk().await.unwrap() {
+    ///             println!("Chunk: {:?}", chunk);
```
davidpdrsn, line 164: I think the data is likely to be binary. So perhaps just printing the length is better 🤔 

```suggestion
    ///             println!("received {} bytes", chunk.len());
```

axum/src/extract/multipart.rs
```diff
@@ -142,6 +142,38 @@ impl<'a> Field<'a> {
...
+    /// ```
+    /// use axum::{
+    ///    extract::Multipart,
+    ///    routing::post,
+    ///    Router,
+    /// };
+    ///
+    /// async fn upload(mut multipart: Multipart) {
+    ///     while let Some(mut field) = multipart.next_field().await.unwrap() {
+    ///         while let Some(chunk) = field.chunk().await.unwrap() {
```
davidpdrsn, line 163: We should handle the errors and not use `unwrap`. Users tend to copy/paste docs verbatim so need to make sure we lead by example.

I think `.map_err(|err| (StatusCode::BAD_REQUEST, err.to_string())?` is fine.

davidpdrsn (changes requested):
A few more nitpicks.

Wanna add a changelog entry as well?

axum/src/extract/multipart.rs
itstristanb: fixed

axum/src/extract/multipart.rs
itstristanb: changed

axum/src/extract/multipart.rs
itstristanb: fixed

axum/src/extract/multipart.rs
itstristanb: Agreed

axum/src/extract/multipart.rs
```diff
@@ -142,6 +142,40 @@ impl<'a> Field<'a> {
...
+    ///    extract::Multipart,
+    ///    routing::post,
+    ///    response::IntoResponse,
+    ///    Router,
+    /// };
+    /// use http::StatusCode;
+    ///
+    /// async fn upload(mut multipart: Multipart) -> impl IntoResponse {
+    ///     while let Some(mut field) = multipart.next_field().await.map_err(|err| (StatusCode::BAD_REQUEST, err.to_string()))? {
+    ///         while let Some(chunk) = field.chunk().await.unwrap() {
```
davidpdrsn, line 165: One last `unwrap` missing here.

davidpdrsn (changes requested):

davidpdrsn:
I've cleaned up the docs just now so should be good now 😊

I've also noticed that we use `unwrap`s [here](https://docs.rs/axum/latest/axum/extract/struct.Multipart.html#example). Sorry if you saw that and based your original commit on that. I'm gonna remove those unwraps as well.

davidpdrsn (approved):

davidpdrsn:
Thanks!

PR #903, opened 2022-04-03
Add "stream to file" example

davidpdrsn:
A few people have asked how to stream requests bodies to a file without buffering everything in memory. So I figure its worth adding an example for. Basically just putting together some combinators that exist already.

### DIFF
--- a/examples/stream-to-file/src/main.rs
+++ b/examples/stream-to-file/src/main.rs
@@ -57,7 +57,9 @@ async fn show_form() -> Html<&'static str> {
         r#"
         <!doctype html>
         <html>
-            <head></head>
+            <head>
+                <title>Upload something!</title>
+            </head>
             <body>
                 <form action="/" method="post" enctype="multipart/form-data">
                     <div>
--- a/examples/stream-to-file/src/main.rs
+++ b/examples/stream-to-file/src/main.rs
@@ -1,0 +1,115 @@
+//! Run with
+//!
+//! ```not_rust
+//! cargo run -p example-stream-to-file
+//! ```
+
+use axum::{
+    body::Bytes,
+    extract::{BodyStream, Multipart, Path},
+    http::StatusCode,
+    response::{Html, Redirect},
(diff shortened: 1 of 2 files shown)


examples/stream-to-file/src/main.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// Handler that returns HTML for a multipart form.
+async fn show_form() -> Html<&'static str> {
+    Html(
+        r#"
+        <!doctype html>
+        <html>
+            <head></head>
+            <body>
```
jplatte, line 61: If you want a minimal simple but valid document, you need a `<title>`, but you don't need to wrap things in `<html>`, `<head>` and `<body>`.

examples/stream-to-file/src/main.rs
```diff
@@ -0,0 +1,115 @@
...
+        </html>
+        "#,
+    )
+}
+
+// Handler that accepts a multipart form upload and streams each field to a file.
+async fn accept_form(mut multipart: Multipart) -> Result<Redirect, (StatusCode, String)> {
+    while let Some(field) = multipart.next_field().await.unwrap() {
+        let file_name = if let Some(file_name) = field.file_name() {
+            file_name.to_owned()
```
jplatte, line 84: Is the `.to_owned()` necessary?

davidpdrsn: Yeah because we move `field` into `stream_to_file` while `file_name` is borrowed from `field`:

```
error[E0505]: cannot move out of `field` because it is borrowed
  --> examples/stream-to-file/src/main.rs:91:35
   |
85 |         let file_name = if let Some(file_name) = field.file_name() {
   |                                                  ----------------- borrow of `field` occurs here
...
91 |         stream_to_file(file_name, field)
   |         --------------            ^^^^^ move out of `field` occurs here
   |         |
   |         borrow later used by call
```

jplatte (approved):

PR #911, opened 2022-04-06
feat: add response::Error

npmccallum:
This type makes for efficient use of the `?` operator when in a function
with multiple return error types that all implement `IntoResponse`.

Signed-off-by: Nathaniel McCallum <nathaniel@profian.com>

### DIFF
--- a/axum-core/src/response/error.rs
+++ b/axum-core/src/response/error.rs
@@ -1,0 +1,222 @@
+use super::{IntoResponse, Response};
+use crate::BoxError;
+
+use bytes::{buf::Chain, Buf, Bytes, BytesMut};
+use http::header::{HeaderMap, HeaderName, HeaderValue};
+use http::StatusCode;
+use http_body::combinators::{MapData, MapErr};
+use http_body::{Empty, Full};
+use std::borrow::Cow;
+use std::convert::{Infallible, TryInto};
+
+/// An [IntoResponse]-based error type
+///
+/// All types which implement [IntoResponse] can be converted to an [Error].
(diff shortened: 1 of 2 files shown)


jplatte:
Ah, that's unfortunate. I wonder whether there's a way around it 🤔

davidpdrsn:
I think this is an interesting idea as well!

Can you give some examples of how you imagine people would use it? Both for error types that you own (and can implement `IntoResponse` for) and types you don't (like `anyhow`).

Specifically it needs to beat `.map_err(IntoResponse::into_response)?` to be justified imo.

npmccallum:
@davidpdrsn We have, for example, different traits for different kinds of backends doing different functions. Each of them have their own error types, all of which `impl IntoResponse`. However, because there isn't one mega error type (it isn't appropriate for our usage) it means the `?` operator doesn't work and we have to do `.map_err()` all over the place. This causes the flow of our (complex) logic to get hidden between the error-handling overhead.

Therefore, we need one common `Error` type which can work with the `?` operator.

davidpdrsn:
So do you then implement `From<Error>` for each of your error types and return `Result<_, Error>` from handlers?

npmccallum:
I have [added a separate](https://github.com/tokio-rs/axum/pull/921) PR for a simpler approach. I'll leave this here so we can compare approaches.

davidpdrsn:
I think we should go with https://github.com/tokio-rs/axum/pull/921 instead as its simpler and more flexible. I'll close this for now.

PR #913, opened 2022-04-06
Remove unnecessary default features for axum from axum-extra

Fxxxlei:
axum-extra does not need any features from axum.
I'd like to use axum-extra without including axum's default features.


jplatte (approved):

davidpdrsn (approved):
Thanks!

PR #918, opened 2022-04-08
axum-extra:re-export `time::OffsetDatetime` and `time::Duration` from…

sinyo-matu:
… `cookie` crate

## Motivation 
the `Expiration` which already be re-exported from `cookie` needs `time::OffsetDatetime` and `time::Duration` for its construction sometimes. and `time::Duration` also required by `Cookie`'s `max_age` api.
So I think it would nice to add a re-export, then users can avoid `time` crate as a dependency.

## Solution
re-export `time::OffsetDate` and  `time::Duration` from `cookie` crate.

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning].
 
 # Unreleased
 
+- **added:** Re-export `time::OffsetDatetime` and `time::Duration` from the `cookie` crate.
 - **fixed:** Don't depend on axum with default features enabled ([#913])
 
 [#913]: https://github.com/tokio-rs/axum/pull/913
--- a/axum-extra/src/extract/cookie.rs
+++ b/axum-extra/src/extract/cookie.rs
@@ -18,7 +18,10 @@ use std::{convert::Infallible, fmt, marker::PhantomData};
 mod private;
 
 pub use self::private::PrivateCookieJar;
-pub use cookie_lib::{Cookie, Expiration, Key, SameSite};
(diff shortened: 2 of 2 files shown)


axum-extra/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning].
 
 # Unreleased
 
+- **added:** Re-export `time::OffsetDatetime` and `time::Duration` from the `cookie` crate.
```
davidpdrsn, line 10: This needs a reference link to the PR, like the line just below.

davidpdrsn (changes requested):
I suppose this isn't too bad but I'm generally not a fan of re-exports. I just makes things more cloudy and we often get issues from people confused about where certain types comes from (looking at you `axum::Server`). Re-exporting stuff from a dependency of a dependency just feels wrong...

Whats wrong with doing `cargo add time`?

sinyo-matu:
> Whats wrong with doing `cargo add time`?

Someone who want to use  the re-exported `Expiration`(like me) should need `time::Duration`.
And if they are just right on hand may be more nature I thought.

But you are right! re-export dependency of a dependency is a bad idea.
Thanks for your time! Closed.

davidpdrsn:
I'm not necessarily saying its a bad idea. I'm just not sure where to draw the line 🙂

sinyo-matu:
> I'm not necessarily saying its a bad idea. I'm just not sure where to draw the line 🙂

Thanks for your kindness. You are not saying that. I just realized it as myself. 
Maybe the dependency of a dependency could be a line as you said.

PR #920, opened 2022-04-08
fix `route` module clippy warning

zys864:
## Motivation

clippy warning
## Solution
adopt clippy advices

remove `Redirect::permanent(&format!("{}/", req.uri().to_string()))`'s uneeded `to_string()`, because `Uri` already impl `Display` 

remove `Redirect::permanent(&req.uri().to_string().strip_suffix('/').unwrap())`'s `&`, because the  reference be immediately dereferenced by compiler


davidpdrsn (approved):
Thanks!

PR #921, opened 2022-04-08
feat: add response::Error

npmccallum:
This type makes for efficient use of the `?` operator when in a function
with multiple return error types that all implement `IntoResponse`.

Signed-off-by: Nathaniel McCallum <nathaniel@profian.com>


npmccallum:
This PR is similar to https://github.com/tokio-rs/axum/pull/911 but is simpler. Note that to achieve this simplicity, `Error` must not implement `IntoResponse`. Otherwise, we collide with other blanked implementations.

davidpdrsn:
Can this be defined outside of axum-core? I would like to avoid expanding its API unless necessary.

jplatte:
I'm pretty sure this can't be defined elsewhere, that `IntoResponse` impl would violate the orphan rule.

jplatte (commented):
I really like this, but if we decide to have it, it would be good do have more docs before merging.

I also wonder whether this should have a more specific name and/or whether there should be a `Result` alias for it.

npmccallum:
@jplatte I like the idea of a `Result` alias. I don't have strong opinions about the name. However, having `Error` with an associated `Result` alias is a pretty common Rust idiom.

Can you tell me what additional docs you're looking for?

jplatte:
Before putting more effort into this, I think it makes sense to hear what David thinks again.

In terms of docs, I am thinking of

* mentioning this type / the Result alias if it's added in the main crate's docs
* adding an example to the type's docs

npmccallum:
@jplatte I already added an example in the `Result` alias docs. I'll wait for @davidpdrsn before making any new pushes.

davidpdrsn:
I think it's looking good but I wanna play around with it a bit first to see. I'll do that during the weekend and let you know 😊

npmccallum:
@davidpdrsn Any update?

davidpdrsn:
Not yet. I'll let you know.

axum-core/src/response/mod.rs
```diff
@@ -17,3 +17,52 @@ pub use self::{
...
+///     Err((StatusCode::NOT_FOUND, "not found"))?;
+///     Err(StatusCode::BAD_REQUEST)?;
+///     Ok("ok")
+/// }
+///
+/// fn main() {
+///     let response: Response = foo().into_response();
+/// }
+/// ```
+pub type Result<T> = std::result::Result<T, Error>;
```
davidpdrsn, line 45: This way its a full substitute for `std::result::Result`:

```suggestion
pub type Result<T, E = Error> = std::result::Result<T, E>;
```

axum-core/src/response/mod.rs
```diff
@@ -17,3 +17,52 @@ pub use self::{
...
+        match self {
+            Ok(ok) => ok.into_response(),
+            Err(err) => err.0,
+        }
+    }
+}
+
+/// An [IntoResponse]-based error type
+///
+/// See [Result] for more details.
```
davidpdrsn, line 59:
```suggestion
/// An [`IntoResponse`]-based error type.
///
/// See [`Result`] for more details.
```

axum-core/src/response/mod.rs
```diff
@@ -17,3 +17,52 @@ pub use self::{
...
+///
+/// ```no_run
+/// use axum_core::response::{IntoResponse, Response, Result};
+/// use http::StatusCode;
+///
+/// fn foo() -> Result<&'static str> {
+///     Err((StatusCode::NOT_FOUND, "not found"))?;
+///     Err(StatusCode::BAD_REQUEST)?;
+///     Ok("ok")
+/// }
```
davidpdrsn, line 39: I think example should show something slightly more realistic. Perhaps calling a few other fallible functions with different error types. I don't expect users actually do `Err((StatusCode::NOT_FOUND, "not found"))?` in their own code 😅

axum-core/src/response/mod.rs
```diff
@@ -17,3 +17,52 @@ pub use self::{
...
+/// All types which implement [IntoResponse] can be converted to an [Error].
+/// This makes it useful as a general error type for functions which combine
+/// multiple distinct error types but all of which implement [IntoResponse].
+///
+/// For example, note that the error types below differ. However, both can be
+/// used with the [Result], and therefore the `?` operator, since they both
+/// implement [IntoResponse].
+///
+/// ```no_run
+/// use axum_core::response::{IntoResponse, Response, Result};
```
davidpdrsn, line 32: Examples should use `axum`, not `axum-core`, as thats what users are most likely to depend on. That'll also require re-exporting this from `axum` which we should do imo.

```suggestion
/// use axum::response::{IntoResponse, Response, Result};
```

axum-core/src/response/mod.rs
```diff
@@ -17,3 +17,52 @@ pub use self::{
...
+///
+/// fn foo() -> Result<&'static str> {
+///     Err((StatusCode::NOT_FOUND, "not found"))?;
+///     Err(StatusCode::BAD_REQUEST)?;
+///     Ok("ok")
+/// }
+///
+/// fn main() {
+///     let response: Response = foo().into_response();
+/// }
```
davidpdrsn, line 43: I think we can just skip this.

```suggestion
```

davidpdrsn (changes requested):
I personally prefer having a single error enum made with `thiserror` and a manual implementation of `IntoResponse` but for more adhoc stuff this might be nice.

I think the name `Error` is too broad though, it also conflicts with `axum::Error`. What about `ErrorResponse` or `ResponseForError`?

jplatte:
I like the suggestions of `ErrorResponse`, in which case maybe the result alias should also have a more unique name (`ResultResponse`)?

npmccallum:
@jplatte @davidpdrsn I'll leave the decision up to you guys. However, `ErrorResponse` and `ResultResponse` seem very non-Rust idiomatic to me. The Rust idiom is to name types `Error` and `Result` within their own module. For example: `std::io::Error` and `std::io::Result`.

jplatte:
The thing is that `ErrorResponse` isn't really an error in the traditional sense (it doesn't implement `std::error::Error`). The fact that `ResultResponse` is a type alias for `Result<_, ErrorResponse>` is basically an implementation detail necessary to make `?` work currently – we could change this in a breaking release one the try-related traits are stable and can be implemented for ones own types.

davidpdrsn:
I'm updating this branch now in case you didn't notice. Hopefully I haven't overwritten anything you've done locally.

You're right about std having multiple error types but I see that as kind of a special case, and honestly not very user friendly. I have heard from several users who have been confused about types with the same names, even though they were in different modules. `Json` for example used to two types. One that was an extractor and one that was a response. That was changed in axum 0.2 I believe.

The error types in std also implement `std::error::Error` which makes them more error-like I suppose. This error type wouldn't implement that trait.

npmccallum:
My intuition is that less than half of all types used as an error in `Result` within the Rust ecosystem actually implement `std::error::Error` (there are a variety of reasons for this, many of which caused the creation of crates like `anyhow` and `thiserror`). So I don't see that as a reason to break the idiom here.

And while I agree there is a minor paper cut having two types with the same name in two different modules, I think the greater sin is diverging from ecosystem-wide convention. Anyone who learns Rust will hit the `Result` naming convention pretty immediately. Diverging from that pattern therefore feels weird to me.

davidpdrsn:
I think keeping the `Result` name is probably fine, thats quite common. But I think we should stick with `ErrorResponse`.

jplatte:
I don't care too much either way.

npmccallum:
I'm good with the above compromise.

davidpdrsn:
Sweet! I'm fixing up the PR now 😊

npmccallum:
I am too. :)

npmccallum:
I have rebased and made the discussed changes.

npmccallum:
@davidpdrsn I may have accidentally done a force push seconds after your push.

axum-core/src/response/mod.rs
npmccallum: See the latest commit. Is this what you had in mind?

axum-core/src/response/mod.rs
npmccallum: Fixed.

davidpdrsn:
Why 😭 I had made a bunch of nitpicky docs changes that I might just force push on top of yours then.

npmccallum:
@davidpdrsn I'm so sorry! Do you still have the commit locally?

davidpdrsn:
Yep! I'll look at it tomorrow 😊

davidpdrsn (approved):
image

jplatte (approved):
Small nitpick about the changelogs, otherwise looks great!

Issue #925, opened 2022-04-09
Allow response multiple headers with the same name 


sinyo-matu:
### Motivation
People may need to response multiple headers with the same name.
In my case it's Set-Cookie.
on `axum` version 0.4.x it worked. but not on version 0.5.x

### Proposal
```diff
impl<K, V, const N: usize> IntoResponse for [(K, V); N]
where
    K: TryInto<HeaderName>,
    K::Error: fmt::Display,
    V: TryInto<HeaderValue>,
    V::Error: fmt::Display,
{
    fn into_response(self) -> Response {
        let mut res = ().into_response();

        for (key, value) in self {
            let key = match key.try_into() {
                Ok(key) => key,
                Err(err) => {
                    return (StatusCode::INTERNAL_SERVER_ERROR, err.to_string()).into_response()
                }
            };

            let value = match value.try_into() {
                Ok(value) => value,
                Err(err) => {
                    return (StatusCode::INTERNAL_SERVER_ERROR, err.to_string()).into_response()
                }
            };
-            res.headers_mut().insert(key, value);
+            res.headers_mut().append(key, value);
        }

        res
    }
}
```
this worked. but I'm not sure is there any drawback on this change.

### Alternatives
using `axum_extra`'s `CookieJar` can solve this

davidpdrsn:
I think overriding headers is the correct default. Internally in axum we use `[("content-type", foo)]` in a bunch of places, which we couldn't do if headers were append since you can only have one `content-type`. However we could provide an `AppendHeaders` newtype that does what you want.

sinyo-matu:
> However we could provide an AppendHeaders newtype that does what you want.

Yeah, that is enough for me.

PR #927, opened 2022-04-09
Add `AppendHeaders`

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/925


jplatte (commented):
This can be part of axum rather than axum-core, no?

davidpdrsn:
I thought so at first but the constructors for `TryIntoHeaderError` are private. So either those have to be made public or this goes into axum-core. I chose the latter but don't feel strongly about it though. Figured since `AppendHeaders` has a very small API its probably fine.

jplatte:
I see. I think it would be fine to have those constructors public, but not clearly better than having `AppendHeaders` in axum-core.

jplatte (approved):

jtk18:
Why the constant array? Can this be a vec instead? Thank you!

davidpdrsn:
I don't specifically remember. We probably could use any `IntoIterator`.

jtk18:
That would make it much easier to work with. Thank you for the great addition!

PR #936, opened 2022-04-16
Fix trailing redirection with query parameters

calavera:
## Motivation

When the request URI matches a route that needs a trailing slash, or has an extra trailing slash, the redirect URI is not generated correctly.

This change adds or removes a trailing slash to the path part of the URI, instead of the full URI, preserving query parameters during redirection.

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -497,13 +497,19 @@ where
 
         match self.node.at(&path) {
             Ok(match_) => self.call_route(match_, req),
-            Err(MatchError::MissingTrailingSlash) => RouteFuture::from_response(
-                Redirect::permanent(&format!("{}/", req.uri())).into_response(),
-            ),
-            Err(MatchError::ExtraTrailingSlash) => RouteFuture::from_response(
-                Redirect::permanent(req.uri().to_string().strip_suffix('/').unwrap())
-                    .into_response(),
-            ),
+            Err(MatchError::MissingTrailingSlash) => {
+                let new_uri = req.uri().to_string().replace(&path, &format!("{}/", &path));
+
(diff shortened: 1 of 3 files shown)


jplatte (commented):
`str::replace` replaces all matches of a pattern, so `/foo?bar=/foo` would get rewritten to `/foo/?bar=/foo/`. Please use something like `.find('?')` + `.insert(idx, '/')` instead.

davidpdrsn:
There are also methods to take apart the path and query. Can't quite put it back together as easily but I think that's safer to use. We do that already in `StripPrefix` afaik.

calavera:
@jplatte, @davidpdrsn I've updated it to make it safer

axum/src/routing/mod.rs
```diff
@@ -512,6 +519,19 @@ where
...
 
+fn replace_trailing_slash(uri: &Uri, new_path: &str) -> Uri {
+    let mut new_path_and_query = new_path.to_string();
+    if let Some(query) = uri.query() {
+        new_path_and_query.push('?');
+        new_path_and_query.push_str(query);
+    }
+
+    let mut parts = uri.clone().into_parts();
+    parts.path_and_query = new_path_and_query.parse().ok();
```
davidpdrsn, line 530: Feels to me like this should be an `unwrap`

```suggestion
    parts.path_and_query = new_path_and_query.parse().unwrap();
```

davidpdrsn (changes requested):
Thanks for noticing this! I had one minor comment.

calavera:
> Thanks for noticing this! I had one minor comment.

Updated

davidpdrsn (approved):

PR #937, opened 2022-04-16
todos Example: Use tokio::sync::RwLock instead of std::sync::RwLock

sidkshatriya:
todos Example: Use `tokio::sync::RwLock` instead of `std::sync::RwLock`

This is because `std::sync::RwLock` will block the OS thread if it cannot acquire the lock [1] [2] -- and we don't want that in an async system, ideally.

[1] https://doc.rust-lang.org/std/sync/struct.RwLock.html#method.read
[2] https://doc.rust-lang.org/std/sync/struct.RwLock.html#method.write

### DIFF
--- a/examples/todos/src/main.rs
+++ b/examples/todos/src/main.rs
@@ -22,12 +22,8 @@ use axum::{
     Json, Router,
 };
 use serde::{Deserialize, Serialize};
-use std::{
-    collections::HashMap,
-    net::SocketAddr,
-    sync::{Arc, RwLock},
-    time::Duration,
-};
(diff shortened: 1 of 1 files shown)


davidpdrsn:
Actually using a sync lock here is fine since we are not doing any IO while holding the lock and the critical section is very short. See https://docs.rs/tokio/1.17.0/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

Making this an async lock would just add additional overhead.

sidkshatriya:
Ok, fair enough.

Issue #933, opened 2022-04-13
Allow falling back to index page in `SpaRouter` when serving  on `/`


dnaka91:
### Motivation

Wanted to replace our current solution in [Trunk](https://github.com/thedodd/trunk) for SPA apps with the new `SpaRouter` from `axum-extra`. This would allow us to drop our current implementation, done with `hyper-staticfile`.

Unfortunately, I noticed that when serving the `SpaRouter` at the root `/`, it doesn't fall back to `index.html` anymore.

In Trunk the usual setup is to have the index page and additional assets all in the same folder. So without custom config it would always result in `SpaRouter::new("/", "dist")`. That makes the router consider everything to be a 404 as it is considered inside the _assets_ folder.

Instead, we need the router to fall back to the `index.html`, same as when using the router on anything else but the root.

### Proposal

The `SpaRouter` is technically correct in its implementation, but I see the root `/` rather as a special case. Therefore, I see 2 solutions.

1. Check that the serve route is `/` and enable the fallback to `index.html` on all missing files.

2. Add a new option that allows to override the 404 logic, always returning the `index.html` on missing files, no matter what.

davidpdrsn:
Are you saying you're serving assets from the root path? So like `/style.css` rather than `/assets/style.css`? I've never done that (rails doesn't) but some people have brought it up. I suppose it could make sense to support.

dnaka91:
Exactly that. In our case, we compile a project to wasm, including other web-related files. Then we output everything in one single folder and serve that one locally. So the assets and the `index.html` are all together in one single folder (plus sub-folders eventually), and represent the whole frontend app.

I already had a look at the PR you prepared so quickly, still have to test this out. While looking at the code, I realized that my initial statement is a bit wrong.

We don't just need to serve from the root, but technically need this _“always fall back to index.html”_ behavior pretty much everywhere.

So if the root changes through a user flag, it changes the whole root of where the wasm app is served. But the problem remains, that we have the index.html and all assets together and the same root where the actual app is served. Therefore, we technically move the whole root.

I hope it isn't asked too much, to turn this feature into a flag, that can be passed to the `SpaRouter` and allows to unconditionally fall back to `index.html`?

davidpdrsn:
> We don't just need to serve from the root, but technically need this “always fall back to index.html” behavior pretty much everywhere.

Isn't that what my change in the PR does?

```rust
Router::new()
    .route("/foo", get(|| async { "GET /foo" }))
[...]

PR #941, opened 2022-04-17
Support serving assets at root with `SpaRouter`

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/933

@dnaka91 are you able to test if this works for you?

### DIFF
--- a/axum-extra/src/routing/spa.rs
+++ b/axum-extra/src/routing/spa.rs
@@ -50,6 +50,20 @@ use tower_service::Service;
 /// - `GET /some/other/path` will serve `index.html` since there isn't another
 ///   route for it
 /// - `GET /api/foo` will serve the `api_foo` handler function
+///
+/// # Serving assets from the root
+///
+/// Assets can also be served from the root, i.e. not nested `/assets` or similar:
+///
+/// ```
+/// use axum_extra::routing::SpaRouter;
+/// use axum::{Router, routing::get};
+///
+/// let spa = SpaRouter::new("/", "dist");
+///
+/// let app = Router::new().merge(spa);
+/// # let _: Router<axum::body::Body> = app;
+/// ```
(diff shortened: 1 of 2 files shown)


axum-extra/src/routing/spa.rs
```diff
@@ -153,18 +167,40 @@ where
     HandleError<Route<B, io::Error>, F, T>:
         Service<Request<B>, Response = Response, Error = Infallible>,
     <HandleError<Route<B, io::Error>, F, T> as Service<Request<B>>>::Future: Send,
-    B: HttpBody + Send + 'static,
+    B: HttpBody + Default + Send + 'static,
```
davidpdrsn, line 170: This makes it a breaking change. Don't think there is a way around it since we need to first call the assets service and if that 404s then call the index service.

jplatte:
Is there still some life in [the tower-http PR for `ServeDir` fallback](https://github.com/tower-rs/tower-http/pull/243)? If that is finished, this would be a super thin layer on top (to the point that I wouldn't add an axum-extra dependency for it, though it might be nice for discoverability).

davidpdrsn:
Yep I haven't forgotten about that yet! One day I'd like to merge these two things but I thought that getting something into axum-extra would probably be easier.

axum-extra/src/routing/spa.rs
jplatte: Didn't you want to publish a breaking release of axum-extra soon anyways? :)

davidpdrsn: yep I do :)

jplatte: Oh this thread was 3 days old.. I think I arrived here from a notification though? Weird 😄

daheige (approved):

davidpdrsn:
Closing. See https://github.com/tokio-rs/axum/issues/933#issuecomment-1110686456.

PR #945, opened 2022-04-19
WIP: OpenAPI generator proof-of-concept

abonander:
A different approach than #459 with a simpler API. See `examples/openapi-test` for expected usage.

I chose to have the OpenAPI typedefs in-crate instead of using `okapi` as it's actually not that complex and we can use more efficient typedefs for the specific use-case (like having `&'static str` everywhere for strings).

One annoying thing is that `#[route]` changes the function signature which doesn't seem to play well with IDE support. I'm not sure how to avoid that while still being able to attach the documentation string to the function lvalue.

Also `route` and `RouteHandler` are probably the wrong names but it's what came to mind.

TODO: handle path parameters (change `/foo/:id` to `/foo/{id}`), fill out API

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,5 +4,7 @@ members = [
     "axum-core",
     "axum-extra",
     "axum-macros",
+    "axum-openapi",
+    "axum-openapi-macros",
     "examples/*",
 ]
--- a/axum-openapi-macros/Cargo.toml
+++ b/axum-openapi-macros/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "axum-openapi-macros"
+version = "0.1.0"
+edition = "2021"
+authors = ["Austin Bonander <austin@launchbadge.com>"]
+
+[lib]
+proc-macro = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+syn = { version = "1.0.91", features = ["full"] }
+quote = "1.0.18"
+proc-macro2 = "1.0.37"
--- a/axum-openapi-macros/src/json_body.rs
+++ b/axum-openapi-macros/src/json_body.rs
@@ -1,0 +1,29 @@
+use proc_macro2::TokenStream;
+
+use std::mem;
+
+pub fn expand(mut input: syn::DeriveInput) -> syn::Result<TokenStream> {
+    let docs = dbg!(super::collect_docs(&input.attrs))?;
+
+    let shadowed_ident = quote::format_ident!("__{}", input.ident);
+
+    let name = mem::replace(&mut input.ident, shadowed_ident);
+    let shadowed_ident = &input.ident;
+    let description = docs.iter().map(|s| s.trim());
+
+    Ok(quote::quote! {
+        impl axum_openapi::JsonBody for #name {
+            fn description() -> &'static str {
+                concat!(#(#description, "\n"),*).trim()
+            }
+
+            fn json_schema() -> axum_openapi::__macro_reexport::schemars::schema::RootSchema {
+                #[derive(axum_openapi::__macro_reexport::schemars::JsonSchema)]
+                #[schemars(crate = "axum_openapi::__macro_reexport::schemars")]
+                #input
+
+                axum_openapi::__macro_reexport::schemars::schema_for!(#shadowed_ident)
+            }
+        }
+    })
+}
--- a/axum-openapi-macros/src/lib.rs
+++ b/axum-openapi-macros/src/lib.rs
@@ -1,0 +1,53 @@
+use proc_macro::TokenTree;
+use std::mem;
+
+use proc_macro2::TokenStream;
+use quote::ToTokens;
+use syn::{Attribute, DeriveInput, FnArg, Signature, Visibility};
+use syn::parse::{Parse, ParseStream};
+
+mod json_body;
+mod route;
+
+#[proc_macro_attribute]
+pub fn route(args: proc_macro::TokenStream, input: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    let input_ = input.clone();
+    let handler_fn = syn::parse_macro_input!(input_ as route::HandlerFn);
+
+    syn_result(route::expand(handler_fn))
+}
+
+#[proc_macro_derive(JsonBody, attributes(doc, schemars, serde, validate))]
+pub fn derive_json_body(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    let input = syn::parse_macro_input!(input as DeriveInput);
+
+    syn_result(json_body::expand(input))
+}
+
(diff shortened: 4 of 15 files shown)


davidpdrsn:
Hey! Exciting that you've started working on this! I've done a couple of attempts, my latest being the branch `openapi-attempt-3` which I pushed just now. I haven't been able to get very far with my attempts though so great that you're giving it a go!

Like I wrote [here](https://github.com/tokio-rs/axum/issues/50#issuecomment-986040352) I still think there is value in a macro free solution and since I haven't fully explored that yet I'm not sure we could merge something that used macros. However that shouldn't hold you back! You're of course free to publish this as a standalone crate and work on it independently. Although I am squatting `axum-openapi` on crates.io and would like to hold on to it for now.

Maybe some day I'll realize that it cannot be done in an elegant way without macros and then we can move this into the axum repo.

abonander:
The primary point of the macros is to capture doc-comments which I think is far and away the most ergonomic way to write the docstrings. With IntelliJ-Rust, you get syntax highlighting for Markdown in doc-comments which makes them very pleasant to write. `derive(JsonBody)` has a lot of value here, although I kind of wish `derive(schemars::JsonSchema)` captured doc-comments on the object itself and not just its properties; it might be worth opening a PR over there for that.

However, I'm not so hot on the `#[route]` attribute macro for reasons I already mentioned. I could see a declarative macro like the following instead: 

```rust
router.route(
    "/foo/bar",
    route! {
        /// Foo
        /// 
        /// Bar
        /// Baz
        post: foo_bar
    }
)
```

which wouldn't require any annotation on the handler functions themselves, along with a trait-based approach for capturing the handler fn args and return type.

I'm working on this for use in LaunchBadge projects as we have an urgent need for a good API documentation solution. Right now, our frontend devs have to ask backend devs how to call the APIs they create, or just get used to reading the backend code themselves. The only other decent option I can think of would be for us to hand-write the docs with help from something like https://github.com/airtasker/spot.

We don't really want to own the `axum-openapi` crate though as we've already got our hands full with SQLx, and would rather it live here if we can arrive at a solution that you're happy with.

davidpdrsn:
> The primary point of the macros is to capture doc-comments which I think is far and away the most ergonomic way to write the docstrings

Thats a good point. The API for that I had in mind was `.route("/users/:id", get(users_show).description("Get a user by ID"))`. Though it certainly isn't as ergonomic as doc comments.

> We don't really want to own the axum-openapi crate though as we've already got our hands full with SQLx, and would rather it live here if we can arrive at a solution that you're happy with.

I understand that. I don't know when I'll be able to continue working on my ideas for openapi. Had some personal things going on and haven't been able to do much coding outside work.

abonander:
Huh, so it seems like `derive(schemars::JsonSchema)` *does* capture doc comments on the object itself. So the `derive(JsonBody)` is redundant there, actually. I think I still prefer if `axum_openapi` can somehow directly provide the derive, though, and I have a small issue with how `schemars` processes newlines in doc comments: https://github.com/GREsau/schemars/issues/38#issuecomment-1102992564

I could rework this to get rid of the attribute macro, though.

jplatte:
`.route("/users/:id", get(users_show).description("Get a user by ID"))` seems like something that a functional proc-macro could expand to :)

abonander:
That's the idea, yeah. I think I can get away with using a `macro_rules!` macro though.

davidpdrsn:
Yeah thats my main reason for wanting macro free APIs, so users can more easily build on top of it.

abonander:
In this case, the macro wouldn't be generating any code that the user couldn't hand-write. It just puts a nicer interface on it.

davidpdrsn:
Right. Thats fine.

abonander:
@davidpdrsn updated proof-of-concept with no proc macros besides `schemars::JsonSchema` (keeping `axum-openapi-macros` for the time being).

abonander:
Also thinking it might be convenient to have something like:
```rust
router! {
    "/foo/bar": {
        // FooBar!
        post: foo_bar,
    }
}
```

abonander:
@davidpdrsn @jplatte can I get some feedback on this approach before I put a bunch of work into filling the API out?

davidpdrsn:
I'll try and make time for it this week.

tasn:
We would so love having this in. Just got a customer report an issue that would have been caught if we had it. :)

Thanks a lot @abonander for working on it. 🙏🏻

lfn3:
I took a stab at something similar at around the same time, but I started with the path conversion parts. @abonander there's some code (and tests!) [here](https://github.com/lfn3/axum-open-api/blob/main/src/lib.rs#L361) which implements the path format (`:` -> `{}`) conversion. There's also code to extract and transform the routes into okapi PathItems.

davidpdrsn:
I just posted an update [here](https://github.com/tokio-rs/axum/issues/50#issuecomment-1132027216) thats relevant for this as well.

We probably wont be merging this since I still wanna explore the approach of https://github.com/tokio-rs/axum/pull/459. Unfortunately I don't know when I'll get around to that. Until then I'm afraid I don't have the bandwidth to take this on.

abonander:
That's unfortunate, because I don't think you're going to get a better UX with proper docstrings using an approach like #459. It sounds like the concept wasn't even fully fleshed out yet so I don't think I'd be able to help out there.

My goal with this PR was to get something ready to use sooner rather than later, because the lack of good options for API documentation is giving us trouble. `airtasker/spot` is a decent stopgap but still requires duplicating all the API structures over to Typescript, and some of our engineers have expressed discontent at having to do that.

I understand the hesitancy to utilize macros here but like we already talked about, they don't generate any code that you couldn't hand-write. The macros aren't even that complex internally and don't recurse more than necessary. Writing the tests for it, it integrates fine with IDEs, or at least IntelliJ-Rust.

The proc-macro crate can go away, I just kept it around for the time being in case I wanted to experiment more with it.

I suppose LB could just own the crate and call it `lb-axum-openapi` or something, but if/when you get around to implementing your vision, that'll be two competing solutions for the same problem, which I think would cause unnecessary confusion.

davidpdrsn:
> It sounds like the concept wasn't even fully fleshed out yet so I don't think I'd be able to help out there.

What I had was a POC so yeah it's not fully not fleshed out.

> I suppose LB could just own the crate and call it lb-axum-openapi or something, but if/when you get around to implementing your vision, that'll be two competing solutions for the same problem, which I think would cause unnecessary confusion.

I'll leave that decision up to you but I agree it's unfortunate.

I think I will close this PR for now since I don't think merging it into this repo is the best way forward. That doesn't mean I don't think this could be a viable solution or starting point though.

But thank you for your work! I hope you or someone else will pick up the work and finish it 😊

tasn:
@abonander, we are VERY interested in having something (even if suboptimal). I think it's better to infer things from types, but we would take whatever. :)

Is there a way we can follow your progress?

PR #949, opened 2022-04-19
Move private and signed cookies behind feature flags

davidpdrsn:
Since private and signed cookies pulls in additional dependencies I think it makes sense to have them behind feature flags.

This also moves the signed cookie code into its own module, which makes the diff look larger. Reviewing commit by commit might be easier.


axum-extra/Cargo.toml
```diff
@@ -15,6 +15,8 @@ default = []
 erased-json = ["serde_json", "serde"]
 typed-routing = ["axum-macros", "serde", "percent-encoding"]
 cookie = ["cookie-lib"]
+cookie-signed = ["cookie-lib/signed"]
+cookie-private = ["cookie-lib/private"]
```
jplatte, line 19:
```suggestion
cookie-signed = ["cookie", "cookie-lib/signed"]
cookie-private = ["cookie", "cookie-lib/private"]
```

jplatte (approved):

davidpdrsn:
Once https://github.com/tokio-rs/axum/pull/941 is done I think we can merge this and ship a new major for axum-extra.

PR #950, opened 2022-04-19
Implement `IntoResponse` for `Response<()>` and `response::Parts`

neoeinstein:
Add a simple implementation to use `Parts` and `Response<()>` as a source for `IntoResponse`.

This doesn't fit neatly into the `IntoResponseParts` paradigm because `Parts` already contains a status code. The implementation itself is rather trivial.

## Motivation

I have a library where I generate some generic template responses using `http::response::Builder` and wanted to be able to make it one step easier to use those templates (often built with a body of `()`, and then substituted with the intended body) as responses. This makes it so that the template can be built and specified next to the body to be placed into the response.

## Solution

Add impls for `Parts` and `Response<()>`. The new `Response<()>` impl discards the existing `()` body and uses the provided body to construct the response and then moves the headers and status into the new response.


jplatte:
I think the `Parts` one might make sense, but

* it should be done in the macro below, in an equivalent manner to
  `impl<R, $($ty,)*> IntoResponse for (StatusCode, $($ty),*, R)`
* it should propagate response extensions as well

I think the `Response<X>` one is too much of a footgun though.

neoeinstein:
Yep. The `Extensions` and `Version` were both things I looked at, but it's difficult to merge in the extensions as mentioned in that linked issue, as there's no way to iterate and extend at the moment.

I can see `Response<X>` being a potential footgun. It's a bit difficult because I can't provide an impl for `http::response::Response<R> where R: IntoResponse`, as that would conflict with the existing one that expects a fully formed `http_body::Body`. This was my workaround for that, but even just having `Parts` would make it one step easier.

axum-core/src/response/into_response.rs
```diff
@@ -396,6 +396,27 @@ where
     }
 }
 
+impl<R> IntoResponse for (http::response::Parts, R)
+where
+    R: IntoResponse,
+{
+    fn into_response(self) -> Response {
+        let (parts, res) = self;
+        (parts.status, parts.headers, res).into_response()
```
davidpdrsn, line 405: Should we be replacing the version here? If not, why? The general pattern we try to follow is that things further to the left in the tuple takes precedence.

We did previously talk about `Version` and decided to just let hyper handle it or require the user to manually build a response. See https://github.com/tokio-rs/axum/issues/802.

I'm not sure what the right call is here.

jplatte: We could request that the http crate explicitly specifies that the default value is `HTTP_11` and add an assert for it 😄

neoeinstein: > I think we should probably also merge the headers with `HeaderMap::extend`.

`IntoResponseParts` does the right thing here. If there are any headers in the `HeaderMap`, then the headers from `R::into_response()` will be extended with those headers.

davidpdrsn: Ah yes I see thats right!

davidpdrsn:
I think this looks good as is but I don't think we should merge it until https://github.com/hyperium/http/pull/546 has landed.

davidpdrsn:
I think this looks good as is but I don't think we should merge it until https://github.com/hyperium/http/pull/546 has landed.

daheige (approved):

davidpdrsn (approved):
I think this is good to go now :)

jplatte (approved):
Changes look good either way.

davidpdrsn:
Good idea! I filed https://github.com/tokio-rs/axum/issues/979. I'll look into it.

PR #952, opened 2022-04-21
Implement an alternative SPA solution

dnaka91:
This is an experiment, trying to solve the issues faced in #933.

## Motivation

The current `SpaRouter` has some limitations for certain use cases and this PR tries a different approach to allow more flexibility in how the routing is set up.

## Solution

This solution provides an extensions on top of the `Router` through `SpaRouterExt`. It adds a new methods that sets up a combination of `ServeDir` and `ServeFile`. In its very basic form it is equivalent to a `ServeDir` to provide assets from a directory. Based on the additional boolean flags, it allows to define fallbacks for to the `index.html`, as usually needed by Single Page Applications on the frontend.

The logic is as follows in regards to fallbacks:

- NO global, NO assets: Same as using `ServeDir` on the given route to serve assets.
- WITH global, NO assets: Same as first, but with a fallback to the `index.html` for **any** unmatched route. This excludes the assets directory itself, as a missing file is not considered unmatched.
- NO global, WITH assets: Same as first, but falls back to the `index.html` **if** some requested path **inside** the assets directory was missing.
- WITH global, WITH assets: Same as first and a combination of the previous. Falls back to `index.html` for **any** unmatched route  **and** any missing file inside the assets directory.

This solution is currently very similar to #941, but uses an extension trait and allows to set an option to apply a global fallback.

I would like to get away from the fallback, and, if this solution is considered good by the Axum team, would split it up a bit more. My idea for this is something like the following:

```rust
    Router::new()
        .spa_assets("/assets", "dist", true)
        .fallback(spa_index("dist"))
```

Here the `spa_assets` would come from an extension trait and only handle logic for the assets dir serving itself. So only the `ServeDir` + option to fall back to the index file on missing assets paths.

And the `spa_index` would be a public function which is just a shorthand for `ServeFile` with the file being `dist/index.html` in this case.

**Update**:

I changed the implementation to be split up, already. Please see the test cases for the new variants. Basically, no option flags anymore. After all, serving assets from a directory without any fallback is just the same as using `ServeDir` directly.

Therefore, `spa_assets` always enables the `index.html` fallback. And a global fallback can be configured with the `.fallback(spa_index("dist"))` call.

I wonder if it would be more natural to instead go a level lower (to the tower service level), or at least closer. What I mean is to have 2 distinct types `ServeSpaIndex` and `ServeSpaDir` which are just very thin wrappers around the `tower-http` types.

- `ServeSpaIndex` would be just a little shorthand that limits to `GET` requests (eventually check for the right mime type too), and accepts the assets directory, so it will join that folder with the `index.html`.
- `ServeSpaDir` would be the same as what the extension trait does, but as a type instead.

Also, these two types would not implement the `handle_error` part, so the user has to handle it. That gives more freedom by letting the user decide what to do with I/O errors. And, having it as a type again, it'd allow to apply additional layers/middleware on top.

### DIFF
--- a/axum-extra/src/routing/spa.rs
+++ b/axum-extra/src/routing/spa.rs
@@ -1,7 +1,8 @@
 use axum::{
     body::{Body, HttpBody},
     error_handling::HandleError,
-    response::Response,
+    handler::Handler,
+    response::{IntoResponse, Response},
     routing::{get_service, Route},
     Router,
 };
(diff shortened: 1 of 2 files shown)


axum-extra/src/routing/spa.rs
```diff
@@ -201,15 +202,173 @@ where
...
+            let res = serve_assets.call(req).await.into_response();
+
+            if res.status() == StatusCode::NOT_FOUND {
+                serve_index.call(req_clone).await.into_response()
+            } else {
+                res
+            }
+        };
+
+        self.nest(route, handler.into_service())
```
davidpdrsn, line 233: This wont work if `route` is `/`. See https://github.com/tokio-rs/axum/issues/933#issuecomment-1104819601

axum-extra/src/routing/spa.rs
```diff
@@ -201,15 +202,173 @@ where
     }
 }
 
+pub trait SpaRouterExt {
+    fn spa_assets(self, route: &str, path: impl AsRef<Path>) -> Self;
+}
```
davidpdrsn, line 207: I don't think an extension trait is right for this. I think its nicer to use `Router::merge` since it accepts `impl Into<Router>`. That means users can build their own types with lots of configuration/builder-methods and have it merged directly into an axum router.

daheige (approved):

axum-extra/src/routing/spa.rs
dnaka91: Agreed. At first, I thought it'd be nice, but once I built it… it felt not _Axum-ish_ 😅. Will update this later today (if I find the time).

axum-extra/src/routing/spa.rs
dnaka91: Ah right, I forgot about this. Will see how I can rewrite this with a `.merge`. Does that mean, in the original `SpaRouter` the `.fallback` call is not propagated to the _outer_ router?
So the fallback only applies to the inner SPA router rather than registering a global fallback in the outer one?

daheige:
Generally speaking, this design looks good. It is recommended to set a single route. At the same time, if you can provide relevant methods for global routing or support this design, you can add unit test.


------------------ Original ------------------
From: Dominik Nakamura ***@***.***&gt;
Date: Sun,Apr 24,2022 9:29 AM
To: tokio-rs/axum ***@***.***&gt;
Cc: heige ***@***.***&gt;, Comment ***@***.***&gt;
Subject: Re: [tokio-rs/axum] Implement an alternative SPA solution (PR #952)





 
@dnaka91 commented on this pull request.
 
 
In axum-extra/src/routing/spa.rs:
 &gt; +                *clone.method_mut() = req.method().clone(); +                *clone.uri_mut() = req.uri().clone(); +                *clone.headers_mut() = req.headers().clone(); +                clone +            }; + +            let res = serve_assets.call(req).await.into_response(); + +            if res.status() == StatusCode::NOT_FOUND { +                serve_index.call(req_clone).await.into_response() +            } else { +                res +            } +        }; + +        self.nest(route, handler.into_service())  
Ah right, I forgot about this. Will see how I can rewrite this with a .merge. Does that mean, in the original SpaRouter the .fallback call is not propagated to the outer router?
 So the fallback only applies to the inner SPA router rather than registering a global fallback in the outer one?

axum-extra/src/routing/spa.rs
davidpdrsn: You'll end up with the same design I had 😅 at least I don't see how it's different

davidpdrsn:
Closing. See https://github.com/tokio-rs/axum/issues/933#issuecomment-1110686456.

PR #959, opened 2022-04-24
add  rest and grpc example

NateLing:
This is my first PR, Add rest and grpc example.

### DIFF
--- a/examples/hybrid/src/hybrid.rs
+++ b/examples/hybrid/src/hybrid.rs
@@ -1,0 +1,71 @@
+use std::{future::Future, pin::Pin, task::Poll};
+
+use axum::body::BoxBody;
+use futures::ready;
+use hyper::{Body, Request, Response};
+use pin_project::pin_project;
+use tower::Service;
+
+#[derive(Clone)]
+pub struct HybridService<Web, Grpc> {
+    web: Web,
+    grpc: Grpc,
+}
+
+impl<Web, Grpc> HybridService<Web, Grpc> {
+    pub fn new(web: Web, grpc: Grpc) -> Self {
+        HybridService { web, grpc }
+    }
+}
+
+impl<Web, Grpc, Error> Service<Request<Body>> for HybridService<Web, Grpc>
+where
+    Web: Service<Request<Body>, Response = Response<BoxBody>, Error = Error>,
+    Grpc: Service<Request<Body>, Response = Response<BoxBody>, Error = Error>,
+{
+    type Error = Error;
+    type Future = HybridFuture<Web::Future, Grpc::Future>;
+    type Response = Response<BoxBody>;
+
+    fn poll_ready(
+        &mut self,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        Poll::Ready(if let Err(err) = ready!(self.web.poll_ready(cx)) {
+            Err(err)
+        } else {
+            ready!(self.web.poll_ready(cx))
+        })
+    }
+
+    fn call(&mut self, req: Request<Body>) -> Self::Future {
+        let hv = req.headers().get("content-type").map(|x| x.as_bytes());
+
+        if let Some(val) = hv && b"application/grpc".eq(val)  {
+						return HybridFuture::Grpc(self.grpc.call(req));
+				}
+
+        HybridFuture::Web(self.web.call(req))
+    }
+}
+
+#[pin_project(project = HybridFutureProj)]
+pub enum HybridFuture<WebFuture, GrpcFuture> {
+    Web(#[pin] WebFuture),
+    Grpc(#[pin] GrpcFuture),
+}
+
+impl<WebFuture, GrpcFuture, Output> Future for HybridFuture<WebFuture, GrpcFuture>
+where
+    WebFuture: Future<Output = Output>,
+    GrpcFuture: Future<Output = Output>,
+{
+    type Output = Output;
+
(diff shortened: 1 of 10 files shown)


examples/hybrid/src/main.rs
```diff
@@ -0,0 +1,53 @@
+#![feature(let_chains)]
```
davidpdrsn, line 1: We cannot use unstable features in examples.

examples/hybrid/src/hybrid.rs
```diff
@@ -0,0 +1,71 @@
...
+            Err(err)
+        } else {
+            ready!(self.web.poll_ready(cx))
+        })
+    }
+
+    fn call(&mut self, req: Request<Body>) -> Self::Future {
+        let hv = req.headers().get("content-type").map(|x| x.as_bytes());
+
+        if let Some(val) = hv && b"application/grpc".eq(val)  {
```
davidpdrsn, line 44: This should use `header_value.starts_with("application/grpc")` since some clients add `+...` at the end.

examples/hybrid/src/hybrid.rs
```diff
@@ -0,0 +1,71 @@
...
+
+        HybridFuture::Web(self.web.call(req))
+    }
+}
+
+#[pin_project(project = HybridFutureProj)]
+pub enum HybridFuture<WebFuture, GrpcFuture> {
+    Web(#[pin] WebFuture),
+    Grpc(#[pin] GrpcFuture),
+}
```
davidpdrsn, line 56: We can just use `futures::future::Either`. It should be identical to this.

examples/hybrid/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+
+mod grpc;
+mod web;
```
davidpdrsn, line 4: Honestly I think we should just put everything in this file. Splitting things across multiple files just makes the example harder to read. I think most people read examples in the browser which makes it hard to switch between multiples files.

examples/hybrid/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+use std::{convert::Infallible, net::SocketAddr};
+
+use axum::response::Response;
+use hyper::{server::conn::AddrStream, service::make_service_fn};
+use tower::ServiceExt;
+use tracing::info;
+use tracing_subscriber::FmtSubscriber;
+
+pub type BoxError = Box<dyn std::error::Error + Send + Sync>;
+pub type Result<T> = std::result::Result<T, BoxError>;
```
davidpdrsn, line 17: `BoxError` also exists in axum as `axum::BoxError` and I don't think we should have a `Result` alias in the example. It wont be used in many places and will just confuse people.

examples/hybrid/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+
+fn make_hybrid_service<Web, Grpc>(web: Web, grpc: Grpc) -> hybrid::HybridService<Web, Grpc> {
+    hybrid::HybridService::new(web, grpc)
+}
```
davidpdrsn, line 21: Not sure we need a function that just calls `hybrid::HybridService::new`. Can just call `hybrid::HybridService::new` directly.

examples/hybrid/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    // initialize tracing
+    let subscriber = FmtSubscriber::builder()
+        .with_max_level(tracing::Level::DEBUG)
+        .finish();
+
+    tracing::subscriber::set_global_default(subscriber)?;
```
davidpdrsn, line 30: This should be changed to the same tracing setup as the other examples.

examples/hybrid/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+    let service = make_hybrid_service(web, grpc);
+
+    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
+    info!("listening on {}", addr);
+    hyper::Server::bind(&addr)
+        .tcp_keepalive(Some(std::time::Duration::from_secs(360)))
+        .serve(make_service_fn(move |_conn: &AddrStream| {
+            let hybrid_service = service.clone();
+            async move { Ok::<_, Infallible>(hybrid_service) }
+        }))
```
davidpdrsn, line 49: I think we should use `.serve(tower::make::Shared::new(hybrid_service))` instead. Its the same but way shorter.

examples/hybrid/src/web/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+// basic handler that responds with a static string
+async fn root() -> &'static str {
+    "Hello, World!"
+}
+
+pub struct ApiService;
+
+impl ApiService {
+    pub fn build() -> axum::Router {
+        axum::Router::new().route("/", get(root).post(root))
```
davidpdrsn, line 12: I think we should just put this line in `main`. No reason to have `ApiService` since it doesn't have any state and only have inherent methods.

davidpdrsn (changes requested):
Thanks! I think something like this makes sense to have but I had a bunch of comments.

As a general rule of thumb we should keep the examples as simple as possible and focus on the stuff that is unique to integrating things with axum. Everything else is likely to just confuse people.

davidpdrsn:
@NateLing why did you close this?

NateLing:
@calavera  sorry, i dont know why it closed. i will fix an retry to commit

PR #969, opened 2022-04-26
Add ignore .DS_Store

NateLing:
Here is what the [.DS_Store](https://en.wikipedia.org/wiki/.DS_Store).

I need clean the `.DS_Store` before commit. Add to gitignore could resolve this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 /target
 Cargo.lock
+
+.DS_Store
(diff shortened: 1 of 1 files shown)


davidpdrsn:
You should add this to your global .gitiginore then you don't need to add it for each repo. That's what I do.

I guess this doesn't hurt though.

.gitignore
```diff
@@ -1,2 +1,4 @@
 /target
 Cargo.lock
+
```
davidpdrsn, line 3:
```suggestion
```

PR #962, opened 2022-04-24
Careful! Add utf8_unckecked feature

NateLing:
Like this comment:

```rust
// `ByteStr` can only be constructed from strings which are always valid
// utf-8 so this wont panic.
std::str::from_utf8(&self.0).unwrap()
```

We know it always valid, But axum promise `This crate uses #![forbid(unsafe_code)] to ensure everything is implemented in 100% safe Rust.` 

So, i think we can add a feature that user can use `unsafe` by themself.

PR #971, opened 2022-04-26
Replace `ByteStr` with `Arc<str>`

davidpdrsn:
## Motivation

Our internal `ByteStr` utility is basically just an `Arc<str>` in disguise. So we probably should just use `Arc<str>` instead.

## Solution

Use `Arc<str>`.

I benchmarked this using the setup posted [here](https://github.com/tokio-rs/axum/pull/962#issuecomment-1109343311) and saw a small increase in performance. Not very significant though.

### DIFF
--- a/axum/src/extract/path/de.rs
+++ b/axum/src/extract/path/de.rs
@@ -1,10 +1,10 @@
 use super::{ErrorKind, PathDeserializationError};
-use crate::util::{ByteStr, PercentDecodedByteStr};
+use crate::util::PercentDecodedStr;
 use serde::{
     de::{self, DeserializeSeed, EnumAccess, Error, MapAccess, SeqAccess, VariantAccess, Visitor},
     forward_to_deserialize_any, Deserializer,
 };
-use std::any::type_name;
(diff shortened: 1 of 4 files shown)


jplatte:
I kind of liked the idea of a utf8-validated `Bytes` instance. Maybe it could be upstreamed into the `bytes` crate and then used here?

davidpdrsn:
In general maybe but for our use case I'm not sure why its better. How is it different from `Arc<str>`? We're never mutating it or anything.

NateLing:
`Arc\<str\>`  is better then add a feature, Users  needn't to think too much to get the benefits.

jplatte (approved):

jplatte:
I thought we created these from `Bytes` and so they'd share the ref counter. But I now see that's not the case, in which case I agree `Arc<str>` is better 👍🏼

davidpdrsn:
> I thought we created these from Bytes and so they'd share the ref counter

Yeah exactly. Its strings all the way down 😅

PR #972, opened 2022-04-26
Optimize zero memory variable

NateLing:


NateLing:
@jplatte @davidpdrsn   `()` `None` means `b"null"`,   `vec![()]` means "b"[null]"",  this pr optimize `()` and `None`, so we can save 128bit memory.

davidpdrsn:
How about

```rust
struct Foo

impl Serialize for Foo {
  ...
}
```

the size of `Foo`s is 0 but you cannot just turn that into `null`.

NateLing:
@davidpdrsn yep, It's wrong.

PR #973, opened 2022-04-26
axum: Version 0.5.4

davidpdrsn:


jplatte (approved):

paolobarbolini:
Could you also release axum-extra so that we can avoid the duplicate dependency version on tower-http :smiley: ?

davidpdrsn:
Ah yes I didn't realize it depended on tower-http as well. Can do tomorrow.

PR #977, opened 2022-04-29
feat: allow TypedHeaderRejection pattern matching

rvolosatovs:
Signed-off-by: Roman Volosatovs <rvolosatovs@riseup.net>

## Motivation

Currently it is not possible to use `TypedHeaderRejection` in `match` statements to ergonomically handle the error cases.

## Solution

Expose `name` and `reason` fields, such that pattern matching would work, e.g.: https://github.com/profianinc/drawbridge/blob/b2370859603b2e508767408ab1a2f33060521224/crates/type/src/meta.rs#L61-L76

### DIFF
--- a/axum/src/typed_header.rs
+++ b/axum/src/typed_header.rs
@@ -109,8 +109,8 @@ where
 #[cfg(feature = "headers")]
 #[derive(Debug)]
 pub struct TypedHeaderRejection {
-    name: &'static http::header::HeaderName,
-    reason: TypedHeaderRejectionReason,
+    pub name: &'static http::header::HeaderName,
+    pub reason: TypedHeaderRejectionReason,
 }
 
 impl TypedHeaderRejection {


davidpdrsn:
There are accessor methods for `name` and `reason`. If the fields we public users would be able to create their own `TypedHeaderRejection`s which we don't want.

image

jplatte:
Could make the fields public but add `#[non_exhaustive]`. That would allow for destructuring but not creating an instance yourself. I'm not sure what's the point though.

rvolosatovs:
Well, the issue is that `TypedHeaderRejectionReason` is not `PartialEq`, so it's not possible to "just" do a `e.reason() == TypedHeaderRejectionReason::Missing`, for example.

I suppose this is good enough alternative:
```rust
        let size = match req.extract().await {
            Ok(TypedHeader(ContentLength(size))) => Some(size),
            Err(e) if matches!(e.reason(), TypedHeaderRejectionReason::Missing) => None,
            Err(e) => return Err(e),
        };
```

davidpdrsn:
I just looked at it and `TypedHeaderRejectionReason` cannot be `PartialEq` because `headers::Error` isn't `PartialEq`, which makes sense I guess. So you gotta do what you wrote here https://github.com/tokio-rs/axum/pull/977#issuecomment-1113193143 but thats also fine imo.

davidpdrsn:
I'd consider that a nitpick. I think its fine as is.

Issue #979, opened 2022-04-29
Implement `IntoResponse for (Parts | Request<()>, $(impl IntoResponseParts)+, impl IntoResponse)`


davidpdrsn:
As a natural extension of https://github.com/tokio-rs/axum/pull/950. Suggested by @jplatte [here](https://github.com/tokio-rs/axum/pull/950#issuecomment-1113456902).

PR #980, opened 2022-04-29
Implement `IntoResponse for (Parts | Request<()>, $(impl IntoResponseParts)+, impl IntoResponse)`

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/979

### DIFF
--- a/axum-core/src/response/into_response.rs
+++ b/axum-core/src/response/into_response.rs
@@ -486,21 +486,10 @@ macro_rules! impl_into_response {
             R: IntoResponse,
         {
             fn into_response(self) -> Response {
-                let (
-                    http::response::Parts {
-                        status,
-                        version: _,
-                        headers,
-                        extensions,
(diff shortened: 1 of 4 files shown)


axum-core/src/response/into_response.rs
```diff
@@ -478,6 +478,58 @@ macro_rules! impl_into_response {
...
             }
         }
+
+        #[allow(non_snake_case)]
+        impl<R, $($ty,)*> IntoResponse for (http::response::Parts, $($ty),*, R)
+        where
+            $( $ty: IntoResponseParts, )*
+            R: IntoResponse,
+        {
+            fn into_response(self) -> Response {
```
jplatte, line 488: Why is this implementation so much more complicated than the one for `(http::response::Parts, R)`? 🤔

jplatte (approved):

PR #981, opened 2022-04-30
Ignore target directories in individual examples

crodjer:
## Motivation

These get created with `rust-analyzer`, which pollutes the git status
output.

For example:
```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        examples/chat/target/
        examples/sqlx-postgres/target/
```

`cargo run -p example-chat` from `/example` doesn't cause this though, which explains why this may not have been an issue faced by anyone else.

## Solution

Add `/examples/*/target` to `gitignore`. Have been using `.git/info/exclude` for this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /target
 /examples/target
+/examples/*/target
 Cargo.lock
 .DS_Store


renancloudwalk:
Can you share with us how to reproduce this? 

Built examples are being generated at `axum/axum/examples/target/...` and not on the example folder itself in my case.

So I'm not able to simulate this issue

crodjer:
It seems to happen for me when I open an example main file in my editor
(Neovim), which triggers rust-analyzer.

davidpdrsn (approved):
I suppose I didn't get this because I change rust-analyzer's target dir to a shared location.

LGMT!

PR #982, opened 2022-04-30
import utoipa openapi example code

idiomatic-rust-com:
Hi,

This code example provides solution for this problem https://github.com/tokio-rs/axum/issues/50

(It's not my code, I asked @[juhaku](https://github.com/juhaku) for permission)

Best regards,
Michal


davidpdrsn:
Not sure we need this hear since utopia already has a similar example https://github.com/juhaku/utoipa/tree/master/examples/todo-axum

idiomatic-rust-com:
No problem. I was thinking that it may be useful.

PR #988, opened 2022-05-03
Fix driving readiness for `example-rest-grpc-multiplex`

davidpdrsn:
This fixes how readiness is handled in `example-rest-grpc-multiplex`.


examples/rest-grpc-multiplex/src/multiplex_service.rs
```diff
@@ -40,14 +55,15 @@ where
...
+        // drive readiness for each inner service and record which is ready
         loop {
             match (self.rest_ready, self.grpc_ready) {
                 (true, true) => {
                     return Ok(()).into();
                 }
                 (false, _) => {
                     ready!(self.rest.poll_ready(cx))?;
-                    self.rest_ready = false;
+                    self.rest_ready = true;
```
davidpdrsn, line 66: This would result in an infinite loop 🤦

jplatte (approved):

PR #990, opened 2022-05-03
`Path` extractor works with `Deserialize` impls using `&str`

geigerzaehler:
Before this change the extractor `Path<Test>` would fail if the `Deserialize` implementation of `Test` was calling
`Deserializer::deserialize_str()`.

Now we use `Visitor::visit_borrowed_str()` instead of `Visitor::visit_str()` which is also recommended in the [guide to implement a deserializer][1].

[1]: https://serde.rs/impl-deserializer.html

### DIFF
--- a/axum/src/extract/path/de.rs
+++ b/axum/src/extract/path/de.rs
@@ -88,7 +88,7 @@ impl<'de> Deserializer<'de> for PathDeserializer<'de> {
                 .got(self.url_params.len())
                 .expected(1));
         }
-        visitor.visit_str(&self.url_params[0].1)
+        visitor.visit_borrowed_str(&self.url_params[0].1)
     }
 
     fn deserialize_unit<V>(self, visitor: V) -> Result<V::Value, Self::Error>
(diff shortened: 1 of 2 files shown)


davidpdrsn:
Would you mind also writing an integration test in https://github.com/tokio-rs/axum/blob/main/axum/src/extract/path/mod.rs so we are sure it actually works?

geigerzaehler:
> Would you mind also writing an integration test in https://github.com/tokio-rs/axum/blob/main/axum/src/extract/path/mod.rs so we are sure it actually works?

Of course. I added some tests in https://github.com/tokio-rs/axum/pull/990/commits/93b71a868604f23d8b2744565f95e599baa41311

jplatte (commented):
I thought we've been percent-decoding path params automatically for a while. How could this work then?

davidpdrsn (approved):
I'm not a serde expert but it looks good to me and tests are passing so looks good to me! I also confirmed that `str_reference_deserialize` does indeed fail without this patch.

jplatte:
Do we have tests for percent decoding of path params though? What happens when those are combined with borrowing this way?

jplatte (approved):

davidpdrsn:
@geigerzaehler Thanks!

PR #732, opened 2022-01-30
Rework middleware docs

davidpdrsn:
[Rendered 🖼](https://github.com/tokio-rs/axum/blob/middleware-docs/axum/src/docs/middleware.md)

Fixes https://github.com/tokio-rs/axum/issues/696

PR #993, opened 2022-05-04
fix link about backpressure from routing section

### LINKED ISSUE #992: Documentation has broken link from Routing, a reference to backpressure, to middleware section
## Bug Report

On the latest version of the docs (0.5.4 as of writing) there is a link that needs to be updated from https://docs.rs/axum/0.5.4/axum/struct.Router.html, 

```
"See the [Routing to services and backpressure](https://docs.rs/axum/0.5.4/index.html#routing-to-services-and-backpressure) module for more details."
```

This section does not exist at the index, and instead was moved to this section on middleware https://docs.rs/axum/0.5.4/axum/middleware/index.html#routing-to-servicesmiddleware-and-backpressure

### Version

0.5.4

cpdean:
## Motivation

There was a section moved but links to it were not updated

fixes #992

## Solution

tested locally with `cargo doc` found both the section name and the location had changed but it should be good now.

I believe it was introduced in https://github.com/tokio-rs/axum/pull/732

### DIFF
--- a/axum/src/docs/routing/route.md
+++ b/axum/src/docs/routing/route.md
@@ -164,7 +164,7 @@ Routing to arbitrary services in this way has complications for backpressure
 ([`Service::poll_ready`]). See the [Routing to services and backpressure] module
 for more details.
 
-[Routing to services and backpressure]: ../index.html#routing-to-services-and-backpressure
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
Thanks!

PR #998, opened 2022-05-05
[DO NOT MERGE] Add security

NateLing:
I made a module that deals with security, it's not a formal proposal.

I would like to get some guidance from here.

1. Is the code location in the right place?
2. Is it worth continuing?
3. Other suggestions?

I have referenced the following articles:
[X-XSS-Protection](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection)
[Rust XSS Guide](https://www.stackhawk.com/blog/rust-xss-guide-examples-and-prevention/)

### DIFF
--- a/axum/src/middleware/mod.rs
+++ b/axum/src/middleware/mod.rs
@@ -4,6 +4,7 @@
 
 mod from_extractor;
 mod from_fn;
+mod security;
 
 pub use self::from_extractor::{from_extractor, FromExtractor, FromExtractorLayer};
 pub use self::from_fn::{from_fn, FromFn, FromFnLayer, Next};
--- a/axum/src/middleware/security.rs
+++ b/axum/src/middleware/security.rs
@@ -1,0 +1,130 @@
+use crate::{body::Body, http::Request, response::Response};
+use futures_util::future::BoxFuture;
+use headers::{ HeaderMap, HeaderName, HeaderValue};
+use std::task::{Context, Poll};
+use tower::{Layer, Service};
+
+pub(crate) trait Policy {
+    const HEADER_NAME: &'static str;
+    fn header(&self) -> (HeaderName, HeaderValue);
+}
+
+pub(crate) struct Security(HeaderMap);
+
+impl Security {
+    pub(crate) fn new() -> Self {
+        Self(HeaderMap::new())
+    }
+    pub(crate) fn with(mut self, policy: impl Policy) -> Self {
+        let (key, value) = policy.header();
+        if self.0.contains_key(&key) {
+            let _ = self.0.remove(&key);
+        }
+        self.0.insert(key, value);
+        self
+    }
+}
+
+enum XssFilter {
+    Disable,
+    Enable,
+    ModeBlock,
+}
+
+impl Policy for XssFilter {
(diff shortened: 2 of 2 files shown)


davidpdrsn:
Like I said in https://github.com/tokio-rs/axum/discussions/985 I don't believe this is something that needs to be built in to axum. I also don't know enough about these kinds to offer much guidance.

NateLing:
> Like I said in https://github.com/tokio-rs/axum/discussions/985 I don't believe this is something that needs to be built in to axum. I also don't know enough about these kinds to offer much guidance.

Ok, I think such a sample code isn't enough to make it feel like it's worth digging into.  Perhaps the best place to go is to make a more detailed sample.  The reason why I am obsessed with advancing security work is closely related to my work experience.  Every day I need to submit the project to the black box testing department to scan the pages and interfaces. Some security problems can be solved by Nginx configuration, and some are solved by modifying the front-end and back-end code.  I would like to simplify these tedious details and expect to control the security policies of Document and RESTApi in a selective way (loose, default and strict).

ttys3:
With `from_fn`, this can easily implemented with 3 LOC

```rust
pub async fn add_security_headers<B>(mut req: Request<B>, next: Next<B>) -> impl IntoResponse {
    let mut response = next.run(req).await;
    let headers = response.headers_mut();
    headers.insert("x-xss-protection", "1; mode=block".parse().unwrap());
    response
}
```

NateLing:
@ttys3  Thanks for your feedback, I put this just to give an introduction to what I intend to do, but of course there's more to it than that. 

Here is an incomplete list:

[X-XSS-Protection](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection)	Prevents some reflected XSS attacks.
[X-Content-Type-Options](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options)	Prevents client sniffing of MIME type.
[X-Frame-Options](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options)
[Strict-Transport-Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)	Enforces strict use of HTTPS.
[Expect-CT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect-CT)	Enables certificate transparency.	
[Referrer-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy)	Enables referrer policy.	
[X-DNS-Prefetch-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-DNS-Prefetch-Control)	Controls browser DNS prefetching.	
[Permissions-Policy](https://github.com/w3c/webappsec-permissions-policy/blob/a45df7b237e2a85e1909d7f226ca4eb4ce5095ba/permissions-policy-explainer.md)	Allows or block browser features.	

In my idea, input filtering should be implemented, pseudo code:

```rust

#[derive(Deserialize)]
struct User{
    #[filter(sql_injection)]
    name:String
}

#[derive(Deserialize)]
struct Blog{
    title:String,
    #[filter(xss)]
    rich_text:Option<String>
}

```

Issue #994, opened 2022-05-04
Make `TypedPath` support `Result` and `Option`


ahcodedthat:
### Motivation

When path extraction fails, I want to send an empty 404 response with no error message. The default error message exposes implementation details of my application to clients, and axum normally sends status code 400 instead of 404 in this situation.

With ordinary extractors like `Path`, this is simple: take an `Option<Path<_>>` or `Result<Path<_>, PathRejection>` instead of simply `Path<_>`, and then handle the failure as needed. But this is not allowed when using `TypedPath` because `Option` and `Result` don't implement it:

```
error[E0277]: the trait bound `std::option::Option<MyTypedPathStruct>: TypedPath` is not satisfied
  --> src/main.rs:87:5
   |
87 |             .typed_get(my_handler_fn)
   |              ^^^^^^^^^ the trait `TypedPath` is not implemented for `std::option::Option<MyTypedPathStruct>`
   |
   = note: required because of the requirements on the impl of `FirstElementIs<std::option::Option<MyTypedPathStruct>>` for `(std::option::Option<MyTypedPathStruct>, Extension<Arc<web_axum::Context>>)`
note: required by a bound in `typed_get`
  --> …/.cargo/registry/src/github.com-1ecc6299db9ec823/axum-extra-0.3.0/src/routing/mod.rs:36:12
   |
36 |         T: FirstElementIs<P> + 'static,
   |            ^^^^^^^^^^^^^^^^^ required by this bound in `typed_get`
```

### Proposal

Please implement `TypedPath` for `Option` and/or `Result`.

### Alternatives

* Adding a layer that removes all `text/plain` bodies from error responses, parses the error messages, adjusts status codes, and so on.
* Not using `TypedPath` at all.

davidpdrsn:
Sounds good! I'm not quite sure what needs to be done but I have hunch its more about adding new impls for `FirstElementIs` than `TypedPath`.

In the same vein I think we should consider adding a way to use a different rejection such that you don't need to use `Option`/`Result` in each handler. Perhaps something the impls `From<PathRejection> for YourNewRejection`. I think thats a separate feature though.

PR #1001, opened 2022-05-05
Support `Option` and `Result` in typed paths

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/994

Makes this work:

```rust
#[typed_path("/users/:id")]
struct UsersShow {
    id: String,
}

// this previously wouldn't compile when used with `Router::typed_*`
async fn option_handler(_: Option<UsersShow>) {}

async fn result_handler(_: Result<UsersShow, PathRejection>) {}
```

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning].
 
 # Unreleased
 
-- None.
+- **fixed:** `Option` and `Result` are now supported in typed paths ([#1001])
+
+[#1001]: https://github.com/tokio-rs/axum/pull/1001
 
 # 0.3.0 (27. April, 2022)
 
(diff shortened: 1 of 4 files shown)


axum-extra/CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning].
 
 # Unreleased
 
-- None.
+- **fixed:** `Option` and `Result` are now supported in typed paths ([#1001])
```
jplatte, line 10: This sounds like it's about fields inside the type that derives `TypedPath`. Maybe sth. like this? 

```suggestion
- **fixed:** `Option` and `Result` are now supported in typed path route handler parameters ([#1001])
```

axum-macros/CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- **fixed:** `Option` and `Result` are now supported in typed paths ([#1001])
```
jplatte, line 10: Same here 😉

jplatte (approved):

PR #1011, opened 2022-05-08
Update static file server example

davidpdrsn:
Serving static files has gotten easier lately thanks to `SpaRouter` and `ServeDir::fallback` so think it makes sense to show those in the static file server example.

### DIFF
--- a/examples/static-file-server/src/main.rs
+++ b/examples/static-file-server/src/main.rs
@@ -4,10 +4,14 @@
 //! cd examples && cargo run -p example-static-file-server
 //! ```
 
-use axum::{routing::get, Router};
-use std::{convert::Infallible, io, net::SocketAddr};
-use tower::{make::Shared, ServiceExt};
-use tower_http::services::ServeDir;
+use axum::{
+    http::StatusCode,
+    response::IntoResponse,
+    routing::{get, get_service},
+    Router,
+};
+use std::{io, net::SocketAddr};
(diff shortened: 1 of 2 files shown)


examples/static-file-server/src/main.rs
```diff
@@ -19,22 +20,31 @@ async fn main() {
...
-                    format!("Unhandled internal error: {}", error),
-                )
-            }),
+    // `SpaRouter` is the easiest way to serve assets at a nested route like `/assets`
+    // let app = Router::new()
+    //     .route("/foo", get(|| async { "Hi from /foo" }))
+    //     .merge(axum_extra::routing::SpaRouter::new("/assets", "."));
+
+    // for serving assets directly at the root you can use `tower_http::services::ServeDir`
+    // with a `Router` as the fallback
```
jplatte, line 29: Why this way around, instead of `ServeDir` as the fallback of a router? (that seems more common to me, that you'd want routes added in code to take preference over files)

davidpdrsn: Something like

```rust
    let app = Router::new()
        .route("/foo", get(|| async { "Hi from /foo" }))
        .fallback(axum::routing::get_service(ServeDir::new(".")).handle_error(...));
```

works too. Maybe thats easier 🤔

davidpdrsn: Yeah it much nicer. You can use `.into_make_service` and `.layer` :)

jplatte (approved):

PR #1012, opened 2022-05-08
Support customizing rejections for `#[derive(TypedPath)]`

davidpdrsn:
Previously `#[derive(TypedPath)]` would always use `PathRejection` as the rejection type. Your only way to customize that was using `Result<YourPath, PathRejection>` in the handler, which is annoying if you're using it in multiple places.

This makes it possible to use your own rejection instead:

```rust
#[typed_path("/users/:id", rejection(UsersMemberRejection))]
struct UsersMember {
    id: String,
}

struct UsersMemberRejection;

// Your rejection type must implement `From<PathRejection>`.
//
// Here you can grab whatever details from the inner rejection
// that you need
impl From<PathRejection> for UsersMemberRejection {
    fn from(rejection: PathRejection) -> Self {
        // ...
    }
}

// And your rejection must implement `IntoResponse`, like all rejections
impl IntoResponse for UsersMemberRejection {
        // ...
    }
}
```

### DIFF
--- a/axum-extra/src/routing/typed.rs
+++ b/axum-extra/src/routing/typed.rs
@@ -165,7 +165,7 @@ use http::Uri;
 /// // Your rejection type must implement `From<PathRejection>`.
 /// //
 /// // Here you can grab whatever details from the inner rejection
-/// // that you need
+/// // that you need.
 /// impl From<PathRejection> for UsersMemberRejection {
 ///     fn from(rejection: PathRejection) -> Self {
 ///         # UsersMemberRejection
@@ -173,7 +173,7 @@ use http::Uri;
 ///     }
 /// }
 ///
-/// // And your rejection must implement `IntoResponse`, like all rejections
+/// // Your rejection must implement `IntoResponse`, like all rejections.
 /// impl IntoResponse for UsersMemberRejection {
 ///     fn into_response(self) -> Response {
 ///         # ().into_response()
@@ -193,16 +193,8 @@ use http::Uri;
 /// #[typed_path("/users", rejection(UsersCollectionRejection))]
 /// struct UsersCollection;
 ///
+/// #[derive(Default)]
 /// struct UsersCollectionRejection;
-///
-/// // Since there are no path params the rejection isn't created via `From<PathRejection>` but
-/// // instead via `Default`
-/// impl Default for UsersCollectionRejection {
-///     fn default() -> Self {
-///         # UsersCollectionRejection
-///         // ...
-///     }
-/// }
 ///
 /// impl IntoResponse for UsersCollectionRejection {
 ///     fn into_response(self) -> Response {
--- a/axum-extra/src/routing/typed.rs
+++ b/axum-extra/src/routing/typed.rs
@@ -140,12 +140,85 @@ use http::Uri;
 /// );
 /// ```
 ///
+/// ## Customizing the rejection
+///
+/// By default the rejection used in the [`FromRequest`] implemetation will be [`PathRejection`].
+///
+/// That can be customized using `#[typed_path("...", rejection(YourType))]`:
+///
+/// ```
+/// use serde::Deserialize;
+/// use axum_extra::routing::TypedPath;
+/// use axum::{
+///     response::{IntoResponse, Response},
+///     extract::rejection::PathRejection,
+/// };
+///
+/// #[derive(TypedPath, Deserialize)]
+/// #[typed_path("/users/:id", rejection(UsersMemberRejection))]
+/// struct UsersMember {
+///     id: String,
+/// }
+///
+/// struct UsersMemberRejection;
+///
+/// // Your rejection type must implement `From<PathRejection>`.
+/// //
+/// // Here you can grab whatever details from the inner rejection
+/// // that you need
+/// impl From<PathRejection> for UsersMemberRejection {
+///     fn from(rejection: PathRejection) -> Self {
+///         # UsersMemberRejection
+///         // ...
+///     }
+/// }
+///
+/// // And your rejection must implement `IntoResponse`, like all rejections
+/// impl IntoResponse for UsersMemberRejection {
+///     fn into_response(self) -> Response {
+///         # ().into_response()
+///         // ...
+///     }
+/// }
+/// ```
+///
+/// The `From<PathRejection>` requirement only applies if your typed path is a struct with named
+/// fields or a tuple struct. For unit structs your rejection type must implement `Default`:
+///
+/// ```
+/// use axum_extra::routing::TypedPath;
+/// use axum::response::{IntoResponse, Response};
+///
+/// #[derive(TypedPath)]
+/// #[typed_path("/users", rejection(UsersCollectionRejection))]
+/// struct UsersCollection;
+///
+/// struct UsersCollectionRejection;
+///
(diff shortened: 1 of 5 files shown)


NateLing:
So excited,  I am not sure could  implement like this?

```rust
#[typed_path("/users/:id", rejection("the path param for id `{0}` is not a number"))]
struct UsersMember {
    id: String,
}
```

davidpdrsn:
@NateLing Not quite. `Path` has more failure cases than that and you'd have to handle all them:

```rust
impl From<PathRejection> for UsersMemberRejection {
    fn from(rejection: PathRejection) -> Self {
        match rejection {
            PathRejection::FailedToDeserializePathParams(inner) => match inner.into_kind() {
                ErrorKind::WrongNumberOfParameters { got, expected } => ...,
                ErrorKind::ParseErrorAtKey { key, value, expected_type } => ...,
                ErrorKind::ParseErrorAtIndex { index, value, expected_type } => ...,
                ErrorKind::ParseError { value, expected_type } => ...,
                ErrorKind::InvalidUtf8InPathParam { key } => ...,
                ErrorKind::UnsupportedType { name } => ...,
                ErrorKind::Message(_) => ...,
                _ => ...,
            },
            PathRejection::MissingPathParams(inner) => ...,
            _ => ...,
        }
    }
}
```

axum-extra/src/routing/typed.rs
```diff
@@ -140,12 +140,85 @@ use http::Uri;
...
+/// struct UsersMember {
+///     id: String,
+/// }
+///
+/// struct UsersMemberRejection;
+///
+/// // Your rejection type must implement `From<PathRejection>`.
+/// //
+/// // Here you can grab whatever details from the inner rejection
+/// // that you need
```
jplatte, line 168:
```suggestion
/// // that you need.
```

axum-extra/src/routing/typed.rs
```diff
@@ -140,12 +140,85 @@ use http::Uri;
...
+/// struct UsersCollectionRejection;
+///
+/// // Since there are no path params the rejection isn't created via `From<PathRejection>` but
+/// // instead via `Default`
+/// impl Default for UsersCollectionRejection {
+///     fn default() -> Self {
+///         # UsersCollectionRejection
+///         // ...
+///     }
+/// }
```
jplatte, line 205: I think in reality people will just derive `Default`.

```suggestion
/// #[derive(Default)]
/// struct UsersCollectionRejection;
```

axum-macros/src/typed_path.rs
```diff
@@ -18,52 +18,78 @@ pub(crate) fn expand(item_struct: ItemStruct) -> syn::Result<TokenStream> {
...
+    rejection: Option<syn::Path>,
+}
+
+impl Parse for Attrs {
+    fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
+        let path = input.parse()?;
+
+        let _ = input.parse::<Token![,]>();
+
+        let rejection = if input.parse::<kw::rejection>().is_ok() {
```
jplatte, line 51: This `is_ok()` failing could mean both 'end of input' and 'found another token'. I think you should use `input.lookahead1()`. Also it's a bit weird to require the path to come first..

axum-macros/src/typed_path.rs
```diff
@@ -314,6 +372,33 @@ enum Segment {
...
+    if let Some(rejection) = rejection {
+        let path_rejection = path_rejection();
+        quote! {
+            .map_err(|rejection| {
+                <#rejection as ::std::convert::From<#path_rejection>>::from(rejection)
+            })
+        }
+    } else {
+        quote! {}
+    }
```
jplatte, line 399: `Option<TokenStream>` can also be interpolated, so this can be

```suggestion
    rejection.map(|rejection| {
        let path_rejection = path_rejection();
        quote! {
            .map_err(|rejection| {
                <#rejection as ::std::convert::From<#path_rejection>>::from(rejection)
            })
        }
    })
```

(or you can keep the function signature and add `.unwrap_or_default()` to that snippet)

axum-macros/src/typed_path.rs
```diff
@@ -314,6 +372,33 @@ enum Segment {
...
+        <::axum::extract::Path<Self> as ::axum::extract::FromRequest<B>>::Rejection
+    }
+}
+
+fn rejection_assoc_type(rejection: &Option<syn::Path>) -> TokenStream {
+    if let Some(rejection) = rejection {
+        quote! { #rejection }
+    } else {
+        path_rejection()
+    }
```
jplatte, line 386: `match` is one line shorter 😉 😅

axum-macros/src/typed_path.rs
davidpdrsn: You mean like this d289e7f?

> Also it's a bit weird to require the path to come first..

Why? I like that if you only specify the path you just do `#[typed_path("/foo")]`. I think thats the common case so nice that its short.

axum-macros/src/typed_path.rs
```diff
@@ -18,52 +18,80 @@ pub(crate) fn expand(item_struct: ItemStruct) -> syn::Result<TokenStream> {
...
+
+        let lh = input.lookahead1();
+        let rejection = if lh.peek(kw::rejection) {
+            input.parse::<kw::rejection>()?;
+            let content;
+            syn::parenthesized!(content in input);
+            Some(content.parse()?)
+        } else {
+            None
+        };
```
jplatte, line 59:
```suggestion
        } else if lh.is_empty() {
            None
        } else {
            return Err(lh.error());
        };
```

davidpdrsn: But `rejection(...)` is optional.

jplatte: That's what the `is_empty()` branch is for: If there is no further input after the (optional) comma (by the way: it will probably work to omit is even when `rejection(...)` follows), rejection will be `None`.

jplatte: I guess the most correct solution would be to check `input.is_empty()` before attempting to parse the comma, and if the input isn't empty, just parse a comma plus the `rejection` stuff without lookahead, but also without any `is_ok()` or other checks (instead just `?`s to error out if something unexpected follows).

davidpdrsn: ah of course yes! I somehow missed the empty branch.

axum-macros/src/typed_path.rs
```diff
@@ -18,52 +18,82 @@ pub(crate) fn expand(item_struct: ItemStruct) -> syn::Result<TokenStream> {
...
+        let rejection = if lh.peek(kw::rejection) {
+            input.parse::<kw::rejection>()?;
+            let content;
+            syn::parenthesized!(content in input);
+            Some(content.parse()?)
+        } else if lh.is_empty() {
+            None
+        } else {
+            return Err(lh.error());
+        };
```
jplatte, line 61: Okay so this should be fully correct 😅

Otherwise `#[typed_path("foo" rejection(Rejection))]` would work I think.

```suggestion
        let rejection = if input.is_empty() {
            None
        } else {
            let _: Token![,] = input.parse()?;
            let _: kw::rejection = input.parse()?;

            let content;
            syn::parenthesized!(content in input);
            Some(content.parse()?)
        };
```

jplatte (approved):

PR #1013, opened 2022-05-10
Make docs around extracting `Request` more explicit

davidpdrsn:
Handlers like

```rust
async fn broken(
    request: Request<Body>,
    Extension(state): Extension<State>,
) {
    // ...
}
```

are very subtly broken since extracting `Request<Body>` will remove all extensions so `Extension<State>` will fail. The solution is to extract the request last. There is no way around this currently since `http::Extensions` cannot be cloned 😞

The [docs did mention this previously](https://docs.rs/axum/latest/axum/extract/index.html#applying-multiple-extractors) but I think it could be called out more explicitly.

### DIFF
--- a/axum/src/docs/extract.md
+++ b/axum/src/docs/extract.md
@@ -211,7 +211,7 @@ async fn also_broken(
 
 Also keep this in mind if you extract or otherwise consume the body in
 middleware. You either need to not extract the body in handlers or make sure
-your middleware reinserts the body using [`RequestParts::body_mut`] so its
+your middleware reinserts the body using [`RequestParts::body_mut`] so it's
 available to handlers.
 
 # Optional extractors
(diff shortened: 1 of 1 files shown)


axum/src/docs/extract.md
```diff
@@ -136,31 +136,83 @@ async fn get_user_things(
...
+    body_json: Json<Value>,
+    request: Request<Body>,
+) {
+    // ...
+}
+```
+
+Also keep this in mind if you extract or otherwise consume the body in
+middleware. You either need to not extract the body in handlers or make sure
+your middleware reinserts the body using [`RequestParts::body_mut`] so its
```
jplatte, line 214:
```suggestion
your middleware reinserts the body using [`RequestParts::body_mut`] so it's
```

jplatte (approved):
One typo, otherwise looks good.

PR #1022, opened 2022-05-10
Add protocol field to WebSocket

### LINKED ISSUE #1020: WebSocket subprotocols negotiation
## Feature Request

Axum has support for specifying the supported WebSocket subprotocols when handling a WS request

```rust
ws.protocols(["cloudevents.cbor", "cloudevents.json"]).on_upgrade(handle_socket);
```

However, there is currently no way for the `handle_socket` function to know which protocol was chosen as the result of the client-server negotiation.

### Proposal

The simplest solution seems to be adding a new field to the [axum::extract::ws::WebSocket](https://docs.rs/axum/latest/axum/extract/ws/struct.WebSocket.html) struct called `protocol: Option<String>` (with a corresponding getter function `WebSocket::protocol -> Option<&str>`).

### Alternatives

I submitted this problem on [Discord](https://discord.com/channels/500028886025895936/870760546109116496/973633756990697542) and this seems to be the simplest solution.

MatteoJoliveau:
Signed-off-by: Matteo Joliveau <matteojoliveau@gmail.com>

Closes #1020 

I used `String::from_utf8_lossy` because I didn't want to add complex error handling for a corner case (non-UTF8 characters in an HTTP header) that, for this particular use-case, is pretty rare. I would have loved to use a `Cow` instead of a `String` inside of `WebSocket` to avoid an allocation, but I had a bit of a hard time fighting with lifetimes so I eventually conceded.

Feedbacks and suggestions welcome!

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -207,12 +207,20 @@ impl WebSocketUpgrade {
         let on_upgrade = self.on_upgrade;
         let config = self.config;
 
+        let protocol = self
+            .protocol
+            .as_ref()
+            .map(|header| String::from_utf8_lossy(header.as_bytes()).to_string());
+
         tokio::spawn(async move {
             let upgraded = on_upgrade.await.expect("connection upgrade failed");
             let socket =
                 WebSocketStream::from_raw_socket(upgraded, protocol::Role::Server, Some(config))
                     .await;
-            let socket = WebSocket { inner: socket };
+            let socket = WebSocket {
+                inner: socket,
+                protocol,
+            };
             callback(socket).await;
         });
 
@@ -309,6 +317,7 @@ fn header_contains<B>(req: &RequestParts<B>, key: HeaderName, value: &'static st
 #[derive(Debug)]
 pub struct WebSocket {
     inner: WebSocketStream<Upgraded>,
+    protocol: Option<String>,
 }
 
 impl WebSocket {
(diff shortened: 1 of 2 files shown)


axum/src/extract/ws.rs
```diff
@@ -331,6 +340,11 @@ impl WebSocket {
     pub async fn close(mut self) -> Result<(), Error> {
         self.inner.close(None).await.map_err(Error::new)
     }
+
+    /// Return the selected WebSocket subprotocol, if one has been chosen
+    pub fn protocol(&self) -> Option<&str> {
```
davidpdrsn, line 345: I think we should just return `&HeaderValue` and let the user deal with whether they care about non utf-8 or not.

davidpdrsn (changes requested):

axum/src/extract/ws.rs
```diff
@@ -331,6 +337,11 @@ impl WebSocket {
     pub async fn close(mut self) -> Result<(), Error> {
         self.inner.close(None).await.map_err(Error::new)
     }
+
+    /// Return the selected WebSocket subprotocol, if one has been chosen
```
jplatte, line 341:
```suggestion
    /// Return the selected WebSocket subprotocol, if one has been chosen.
```

davidpdrsn (commented):
Nice!

jplatte (approved):

MatteoJoliveau:
@davidpdrsn changelog updated!

davidpdrsn (approved):

davidpdrsn:
Thanks!

Issue #1021, opened 2022-05-10
Unclear how to resolve: Invalid URL: Wrong number of parameters. Expected 1 but got 2


EdmundsEcho:
I generated the above error at runtime by incorrectly using the `Path` repeated times in the same handler.

### The code snippets

Here are the code snippets that compile but are "bound to fail" at runtime:

```rs
// the Route
	.route("/drive/:auth_provider/:project_id", get(authorize::handle))

// the handler
pub async fn handle(
    Path(drive_provider): Path<String>,
    Path(project_id): Path<String>,
) { ... }
```
Note: Using `String` is a simplified version of the example; using different types does not prevent the error.

### The `url` requested

```
http://localhost:3099/drive/google/dfsdfsd
```

### The error returned in the response

```http
Invalid URL: Wrong number of parameters. Expected 1 but got 2
```

### How to interpret the error
Quoting @davidpdrsn 
> Path<String> makes it look for 1 dynamic param in the URL, but since you have two in the router it fails with expected 2, got 1. The 2 being from /:foo/:bar and the 1 being from Path<String> (1 type param) 

### The issue

There are many other ways to interpret this message :))

### Proposed solution
Somehow clarify the message with what often is the cause given the following:

1.  It is a error generated by a malformed path (that for whatever reason was not 404) 
2.  Suggest using `#[debug_handler]` 
3.  Use the runtime context to generate an error that is more explicit about "what received 1, and what is 2"
     * the <xyz> trait for `Path<String>` was called
     * the Request included 2 `Path` parameters

... indicates a misunderstanding of how to use `Path`.

PR #1023, opened 2022-05-11
Improve `Path` error when its extracted multiple times

davidpdrsn:
This makes the error returned when trying to extract multiple path params with `async fn foo(Path<String>, Path<String>)` and hints at using `async fn foo(Path<(String, String)>)` or a struct.

Fixes https://github.com/tokio-rs/axum/issues/1021

### DIFF
--- a/axum/src/extract/path/mod.rs
+++ b/axum/src/extract/path/mod.rs
@@ -66,8 +66,7 @@ use std::{
 /// ```
 ///
 /// Path segments also can be deserialized into any type that implements
-/// [`serde::Deserialize`]. Path segment labels will be matched with struct
-/// field names.
+/// [`serde::Deserialize`]. This includes tuples and structs:
 ///
 /// ```rust,no_run
 /// use axum::{
@@ -78,6 +77,7 @@ use std::{
 /// use serde::Deserialize;
 /// use uuid::Uuid;
 ///
+/// // Path segment labels will be matched with struct field names
 /// #[derive(Deserialize)]
 /// struct Params {
 ///     user_id: Uuid,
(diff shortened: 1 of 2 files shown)


axum/src/extract/path/mod.rs
```diff
@@ -368,14 +386,13 @@ impl IntoResponse for FailedToDeserializePathParams {
...
             | ErrorKind::InvalidUtf8InPathParam { .. }
-            | ErrorKind::WrongNumberOfParameters { .. }
             | ErrorKind::ParseError { .. }
             | ErrorKind::ParseErrorAtIndex { .. }
             | ErrorKind::ParseErrorAtKey { .. } => (
                 StatusCode::BAD_REQUEST,
                 format!("Invalid URL: {}", self.0.kind),
             ),
-            ErrorKind::UnsupportedType { .. } => {
+            ErrorKind::WrongNumberOfParameters { .. } | ErrorKind::UnsupportedType { .. } => {
```
davidpdrsn, line 395: `WrongNumberOfParameters` is a programmer error so it shouldn't be a `4xx` imo. I'd categorize this as a bug fix and not a breaking change but I guess thats up for debate.

axum/src/extract/path/mod.rs
```diff
@@ -319,11 +329,19 @@ impl fmt::Display for ErrorKind {
...
             ErrorKind::InvalidUtf8InPathParam { key } => write!(f, "Invalid UTF-8 in `{}`", key),
-            ErrorKind::WrongNumberOfParameters { got, expected } => write!(
-                f,
-                "Wrong number of parameters. Expected {} but got {}",
-                expected, got
-            ),
+            ErrorKind::WrongNumberOfParameters { got, expected } => {
+                write!(
+                    f,
+                    "Wrong number of type parameters to `Path`. Expected {} but got {}",
```
jplatte, line 335:
```suggestion
                    "Wrong number of path arguments for `Path`. Expected {} but got {}",
```

If it says parameters, expected and got are the wrong way around I think.

axum/src/extract/path/mod.rs
jplatte: I agree with the categorization as a fix.

jplatte (approved):

PR #1028, opened 2022-05-12
Dont link to deprecated item with different name

kosta:
## Motivation

Clicking a link should lead to the item I clicked on.

## Solution

Link to the item that the text suggests should be linked to.


kosta:
Tested locally.

davidpdrsn (approved):
Thanks!

PR #1031, opened 2022-05-15
Add `axum_extra::extract::Form`

davidpdrsn:
This adds `axum_extra::extract::Form` which uses [serde_html_form](https://crates.io/crates/serde_html_form) and  supports multi-value items.

Its an exact copy for `axum::extract::Form` except that it calls serde_html_form.


axum-extra/Cargo.toml
```diff
@@ -18,6 +18,7 @@ cookie = ["cookie-lib"]
 cookie-signed = ["cookie", "cookie-lib/signed"]
 cookie-private = ["cookie", "cookie-lib/private"]
 spa = ["tower-http/fs"]
+form = ["serde", "serde_html_form"]
```
jplatte, line 21:
```suggestion
form = ["serde", "serde_html_form"]
spa = ["tower-http/fs"]
```

axum-extra/src/extract/form.rs
```diff
@@ -0,0 +1,124 @@
...
+
+/// Extractor that deserializes `application/x-www-form-urlencoded` requests
+/// into some type.
+///
+/// `T` is expected to implement [`serde::Deserialize`].
+///
+/// # Differences from `axum::extract::Form`
+///
+/// This extractor uses [`serde_html_form`] under-the-hood which supports multi-value items. These
+/// are sent by HTML checkboxes and multi selects. Those values can be collected into a `Vec`.
```
jplatte, line 23:
```suggestion
/// are sent by multiple `<input>` attributes of the same name (e.g. checkboxes) and `<select>`s with the `multiple` attribute. Those values can be collected into a `Vec` or other sequential container.
```

axum-extra/src/extract/form.rs
```diff
@@ -0,0 +1,124 @@
...
+
+            Ok(Form(value))
+        }
+    }
+}
+
+mod has_content_type {
+    include!(concat!(
+        env!("CARGO_MANIFEST_DIR"),
+        "/../axum/src/extract/has_content_type.rs"
```
jplatte, line 87: I think this will either fail publishing, or fail to build when pulled from crates.io 😄

jplatte (changes requested):

axum-extra/src/extract/form.rs
davidpdrsn: ah thats too bad 😜 I've removed it.

axum-extra/src/extract/form.rs
```diff
@@ -0,0 +1,134 @@
...
+
+/// Extractor that deserializes `application/x-www-form-urlencoded` requests
+/// into some type.
+///
+/// `T` is expected to implement [`serde::Deserialize`].
+///
+/// # Differences from `axum::extract::Form`
+///
+/// This extractor uses [`serde_html_form`] under-the-hood which supports multi-value items. These
+/// are sent by multiple `<input>` attributes of the same name (e.g. checkboxes) and `<select>`s with the `multiple` attribute. Those values can be collected into a `Vec` or other sequential container.
```
jplatte, line 23: Maybe break across multiple lines again ;)

jplatte (approved):

Issue #1019, opened 2022-05-10
debug_handler: Raise a warning if Request is used as non-final extractor


jplatte:
### Motivation

From time to time, somebody introduces a runtime bug by trying to extract parts of the `Request` after it was already consumed through a `Request` handler argument.

### Proposal

Raise a warning or an error when encountering `http::Request` as an extractor other than the last one in a handler annotated with #[debug_handler]. We could use [proc-macro2-diagnostics](https://crates.io/crates/proc-macro2-diagnostics) to have a warning on Nightly and an error on Stable automatically. (since the proc_macro diagnostics API is still unstable)

### Alternatives

\-

Issue #1018, opened 2022-05-10
debug_handler: Raise a warning for multiple Path extractors


jplatte:
### Motivation

From time to time, people try to use multiple `Path` instead of a single one with a tuple inside when they want to extract multiple distinct parameters from the path.

### Proposal

Raise a warning or an error when encountering multiple `Path` extractors in a handler annotated with `#[debug_handler]`. We could use [proc-macro2-diagnostics](https://crates.io/crates/proc-macro2-diagnostics) to have a warning on Nightly and an error on Stable automatically. (since the `proc_macro` diagnostics API is still unstable)

### Alternatives

\-

PR #1035, opened 2022-05-15
Check `Request` and `Path` in `debug_handler`

davidpdrsn:
I don't care much for emitting warnings on nightly 🤷

Fixes https://github.com/tokio-rs/axum/issues/1019
Fixes https://github.com/tokio-rs/axum/issues/1018

### DIFF
--- a/axum-macros/src/debug_handler.rs
+++ b/axum-macros/src/debug_handler.rs
@@ -4,6 +4,8 @@ use syn::{parse::Parse, spanned::Spanned, FnArg, ItemFn, Token, Type};
 
 pub(crate) fn expand(attr: Attrs, item_fn: ItemFn) -> syn::Result<TokenStream> {
     check_extractor_count(&item_fn)?;
+    check_request_last_extractor(&item_fn)?;
+    let check_path_extractor = check_path_extractor(&item_fn);
 
     let check_inputs_impls_from_request = check_inputs_impls_from_request(&item_fn, &attr.body_ty);
     let check_output_impls_into_response = check_output_impls_into_response(&item_fn);
@@ -11,6 +13,7 @@ pub(crate) fn expand(attr: Attrs, item_fn: ItemFn) -> syn::Result<TokenStream> {
 
     let tokens = quote! {
         #item_fn
+        #check_path_extractor
         #check_inputs_impls_from_request
         #check_output_impls_into_response
         #check_future_send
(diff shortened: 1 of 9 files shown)


axum-macros/tests/debug_handler/fail/request_not_last.stderr
```diff
@@ -0,0 +1,13 @@
...
+5 | async fn handler(_: Request<Body>, _: Extension<String>) {}
+  |                  ^^^^^^^^^^^^^^^^
+
+warning: unused imports: `body::Body`, `extract::Extension`, `http::Request`
+ --> tests/debug_handler/fail/request_not_last.rs:1:12
+  |
+1 | use axum::{body::Body, extract::Extension, http::Request};
+  |            ^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^
+  |
+  = note: `#[warn(unused_imports)]` on by default
```
jplatte, line 13: To avoid spurious warnings like this, it would be nice if we could emit `syn::Error`s inside the original fn instead of replacing it entirely. Applies to existing errors as well I think.

axum-macros/src/debug_handler.rs
```diff
@@ -60,6 +63,67 @@ fn check_extractor_count(item_fn: &ItemFn) -> syn::Result<()> {
...
+            .map(|(_, arg, _)| {
+                syn::Error::new_spanned(
+                    arg,
+                    "Multiple parameters must be extracted with a tuple \
+                    `Path<(_, _)>` or a struct `Path<YourParams>`, not by applying \
+                    multiple `Path<_>` extractors",
+                )
+                .to_compile_error()
+            })
+            .collect()
```
jplatte, line 121: Instead of emitting multiple `compile_error!`s here, it might be better to build up one `syn::Error` from the path extractors using `syn::Error::combine` (i.e. one time this error message, then "other path extractor here") and turn that into a `compile_error!`.

davidpdrsn: Do you mean something like:

```rust
let mut errors = path_extractors
    .into_iter()
    .map(|(_, arg, _)| {
        syn::Error::new_spanned(
            arg,
            "Multiple parameters must be extracted with a tuple \
            `Path<(_, _)>` or a struct `Path<YourParams>`, not by applying \
            multiple `Path<_>` extractors",
        )
    })
    .collect::<Vec<_>>();
let mut error = errors.remove(0);
error.extend(errors); // extend calls `syn::Error::combine`
error.to_compile_error()
```

because that doesn't change the output.

jplatte (approved):

axum-macros/src/debug_handler.rs
jplatte: I guess the output might be different on nightly? Anyways, if it's always separate errors on stable even with `Error::combine`, then I guess it doesn't make a lot of sense.

Issue #940, opened 2022-04-17
Query Parameters With Arrays And Suggested Work Around Dooes Not Work


bonedaddy:
# Overview

Using the latest version of axum, the [workaround](https://github.com/tokio-rs/axum/issues/434#issuecomment-954889152) for using arrays in query parameters no longer works. Using it gives the following error:

```
missing generics for trait `FromRequest` expected 1 generic argument
```

Changing to the following allowed the code to be compiled:

```rust
pub struct Qs<T>(pub T);

#[axum::async_trait]
impl<B, T> FromRequest<B> for Qs<T>
where
    T: serde::de::DeserializeOwned,
    B: std::marker::Send
{
    type Rejection = Infallible;

    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
        // TODO: error handling
        match req.uri().query() {
            Some(params) => {
                Ok(Self(serde_qs::from_str(params).unwrap()))
            }
            None => {
                Ok(Self(serde_qs::from_str("").unwrap()))
            }
        }
    }
}
```

Unfortunately this still results in the same error "invalid type expected a sequence".

# Suggestions

This kind of functionality is absolutely required for any modern day web framework. Not having this kind of functionality out of the box makes reasoning to use a particular web framework hard. Granted axum is fantastic, and is pre v1 so all hope is not loss yet.

So my suggestions:

* Add an example, and test for query parameters with arrays
* Add this feature in before v1 is released
* Before v1 indicate in the readme that arrays in query parameters does not work

# Code Example

```rust
#[derive(Debug, Serialize, Deserialize)]
pub struct FetchV1ObligationsLtvs {
    #[serde(skip_serializing_if = "Option::is_none")]
    pub authorities: Option<Vec<String>>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub user_farms: Option<Vec<String>>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub account_addresses: Option<Vec<String>>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub per_page: Option<i64>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub page: Option<i64>,
}

pub struct Qs<T>(pub T);

#[axum::async_trait]
impl<B, T> FromRequest<B> for Qs<T>
where
    T: serde::de::DeserializeOwned,
    B: std::marker::Send
{
    type Rejection = Infallible;

[...]

Issue #504, opened 2021-11-13
Consider using serde_qs instead of serde_urlencoded for deserializing Query


theowenyoung:
### Motivation

Hi, for query string like  `key[]=value1&key[]=value2`,  `serde_urlencoded` will not respect it as an array, could we consider using [serde_qs](https://github.com/samscott89/serde_qs) instead of serde_urlencoded for deserializing Query?

https://github.com/tokio-rs/axum/blob/c7b8813c4715afd0a6fdf0872545a516bd8e1924/axum/src/extract/query.rs#L60

### Proposal

Just using `serde_qs::from_str<T>(raw_query_string)` instead of `serde_urlencoded`

### Alternatives

Yes, now I can get the raw query string to parse it.

davidpdrsn:
There is no standard way to serialize arrays in query strings, therefore `serde_urlencoded` doesn't support it. Therefore I think its also the right choice for axum. Note that warp and actix-web also uses `serde_urlencoded`.

See https://github.com/tokio-rs/axum/issues/434 for more details and [a workaround](https://github.com/tokio-rs/axum/issues/434#issuecomment-954898159).

Issue #434, opened 2021-10-29
Support for Vec in Query?


davidpdrsn:
### Discussed in https://github.com/tokio-rs/axum/discussions/433

<div type='discussions-op-text'>

<sup>Originally posted by **Vesafary** October 29, 2021</sup>
Hi! 
I've been struggling with this for a while, and didn't get it to work. 
Looking through the docs and examples also didn't give me the answer.

What I'm looking for is support for one of the following:
```
<url>?id=1&id=2&id=3
<url>?id=[1, 2, 3]
```

I've tried several options like
```
struct IdQuery {
    ids: Vec<uuid::Uuid>,
}
async fn get_ids(filter: Query<IdQuery>)
```
```
struct IdQuery {
    id: uuid::Uuid,
}
async fn get_ids(filter: Query<Vec<IdQuery>>)
```
Of course I can implement it myself using the RawQuery extractor, but it seems like fairly basic usage so I was wondering whether I did something wrong.

Cheers!</div>

ezracelli:
For the curious, the patch of `serde_urlencoded` mentioned above is now maintained as a separate crate: https://crates.io/crates/serde_html_form

Other relevant issues in `serde_urlencoded` with its maintainer's reasoning as to why this isn't natively supported:

- https://github.com/nox/serde_urlencoded/issues/6
- https://github.com/nox/serde_urlencoded/issues/75

jplatte:
Yes, and we have `Query` and `Form` extractors in axum-extra that use it.

flo-at:
> `id=[1,2,3]` already works with `serde_urlencoded` if you add a small helper function along the lines of
> 
> ```rust
> fn deserialize_array<'de, D, T>(deserializer: D) -> Result<Vec<T>, D::Error>
> where
>     D: Deserializer<'de>,
>     T: Deserialize<'de>,
> {
>     let s = String::deserialize(deserializer)?;
>     let v = todo!("parse the bracketed comma-separated items")
>     Ok(v)
> }
> ```
> 
> and use it with `#[serde(deserialize_with = "deserialize_arrray")]`.

This does not work for strings though. At least I haven't found a way to make it work. The problem is that at that point, the string `s` is already url-decoded, so any `[`, `]`, or `,` in the array element string makes it impossible to decide which comma is a delimiter and which is part of the string element.

The workaround with the custom Extractor that uses serde_qs is not perfect either because it's not possible to apply this to only one field but only the whole struct afaik.

What's left is:
[...]

PR #1038, opened 2022-05-17
feat: use serde_qs to parse query

suikammd:
## Motivation

In some daily works, I need to parse vector in query which serde_urlencoded not support

## Solution

Change to serde_qs which support parsing vector in query and fully compatible with serde_urlencoded

### DIFF
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -12,14 +12,14 @@ repository = "https://github.com/tokio-rs/axum"
 
 [features]
 default = ["form", "http1", "json", "matched-path", "original-uri", "query", "tower-log"]
-form = ["serde_urlencoded"]
+form = ["serde_qs"]
 http1 = ["hyper/http1"]
 http2 = ["hyper/http2"]
 json = ["serde_json"]
 matched-path = []
 multipart = ["multer"]
 original-uri = []
-query = ["serde_urlencoded"]
+query = ["serde_qs"]
 tower-log = ["tower/log"]
 ws = ["tokio-tungstenite", "sha-1", "base64"]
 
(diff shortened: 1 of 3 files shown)


davidpdrsn:
I'm sorry but we wont merge this. We're sticking with serde_urlencoded for the time being since it just implements [the standard](https://url.spec.whatwg.org/#application/x-www-form-urlencoded). There is unfortunately no standard for passing collections via query params. See https://github.com/tokio-rs/axum/issues/434 https://github.com/tokio-rs/axum/issues/504 and https://github.com/tokio-rs/axum/issues/940 for more details and discussion.

It also seems serde_qs gained axum support recently https://github.com/samscott89/serde_qs/pull/59

jplatte:
If this is just about arrays, we could solve this with a `Query` in axum-extra the same way as `Form` though, right?

That is, with `serde_html_form`¹ in a `Query` extractor, this would work:

```rust
#[derive(Deserialize)]
struct Params {
    #[serde(rename = "foo[]")]
    foo: Vec<String>,
}

async fn my_handler(qry: Query<Params>) {}
```

¹ context: fork of `serde_urlencoded` I wrote to support multiple occurrences of the same item specifically

davidpdrsn:
> If this is just about arrays, we could solve this with a Query in axum-extra the same way as Form though, right?

Yep we could do that. I considered it earlier but didn't think it was as useful as `Form`. Maybe I was wrong 😅

jplatte:
[axum-extra 0.3.3](https://github.com/tokio-rs/axum/releases/tag/axum-extra-v0.3.3) contains a `Query` type that uses `serde_html_form`, with that you should be able to deserialize lists from a query string with repetitions of the same field name.

PR #1041, opened 2022-05-17
Add `axum_extra::extract::Query`

jplatte:
Same thing as #1031, but for query string stuff regardless of request method.


davidpdrsn (approved):
Noice!

Turbo87:
I just came across this alternative `Query` impl and I was wondering: is there is a downside to making this implementation the default in `axum` itself? are there plans to eventually do that?

PR #1043, opened 2022-05-18
Implement `Default` for `Extension`

nolanderc:
## Motivation

When using an `Extension` we often have to wrap the inner value in an `Arc` (and some lock if we need mutability). If we then want to use our extension we need to write out `Extension<Arc<Mutex<T>>>` each time we want to use that `T` in a handler. This becomes tedious very quickly. Thus we introduce a type alias for that extension:

```rust
type MyExtension = Extension<Arc<Mutex<...>>>;

// which is used like:
async fn handler(ext: MyExtension) -> ... { ... }
```

With that type alias, it would be really nice if we could easily construct the `Arc<Mutex<...>>` type just by calling `MyExtension::default()`. However, `Extension<T>` does not implement `Default`, even though it is just a simple newtype wrapper around the inner `T`.

## Solution

We just add an additional `#[derive(Default)]` to the `Extension` struct.

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- **added:** Implement `Default` for `Extension` ([#1043])
 
 # 0.5.6 (15. May, 2022)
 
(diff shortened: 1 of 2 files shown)


davidpdrsn (commented):
Makes sense!

Wanna add a note to the changelog as well?

nolanderc:
> Wanna add a note to the changelog as well?

Done!

davidpdrsn (approved):
Thanks!

PR #1045, opened 2022-05-19
change `RoutId` from `AtomU32` to `AtomU64/AtomU32` according to `cfg`

zys864:
## Motivation
Tokio use `AtomU64` and `Atomu32` according to  macros `cfg_has_atomic_u64` and `cfg_not_has_atomic_u64!`.

We could use these on axum `RouteId`

## Solution
change `RoutId` from `AtomU32` to `AtomU64/AtomU32` accroding to `cfg`.

At the same time ,abolish `RouteId` boundray cheking.


davidpdrsn:
But why? Do you have more than `u32::MAX` routes?

zys864:
No. But every time, Axum have a new route, it will check that it is less than `u32::MAX`.  Is It really need it? 

In early Axum,`RoutId` use Atomu64 and don't check `MAX`. Above of all, it just a advice😅.

jplatte:
I don't see what's problematic with that check. The first clippy fixing commit seems useful though, do you want to create a separate PR with just that?

davidpdrsn:
Thanks for the advice but I think we'll just leave it as is. I think the check is problematic either and its only executed during setup so unless you're adding routes while serving requests it probably wont matter.

3d7df15fa5716f2e265a665a57c6e3aac348c6a3 looks good though. I agree that a separate PR for that would be cool.

PR #1046, opened 2022-05-19
Remove `util::try_downcast` Duplicated code fragment

dengzhq:
## Motivation
When I look at https://github.com/tokio-rs/axum/blob/main/axum-core/src/body.rs source code through idea，idea show the function try_downcast :Duplicated code fragment (12 lines long)，so i find them，and try to solve it

## Solution
Remove the try_downcast function and test case in the axum crate, and modify the related dependencies

### DIFF
--- a/axum-core/src/body.rs
+++ b/axum-core/src/body.rs
@@ -20,7 +20,8 @@ where
     try_downcast(body).unwrap_or_else(|body| body.map_err(Error::new).boxed_unsync())
 }
 
-pub(crate) fn try_downcast<T, K>(k: K) -> Result<T, K>
+/// Returns some mutable reference to the inner value if it is of type T, or Returns Err
+pub fn try_downcast<T, K>(k: K) -> Result<T, K>
 where
     T: 'static,
     K: Send + 'static,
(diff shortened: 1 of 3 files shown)


axum-core/src/body.rs
```diff
@@ -20,7 +20,8 @@ where
     try_downcast(body).unwrap_or_else(|body| body.map_err(Error::new).boxed_unsync())
 }
 
-pub(crate) fn try_downcast<T, K>(k: K) -> Result<T, K>
+/// Returns some mutable reference to the inner value if it is of type T, or Returns Err
+pub fn try_downcast<T, K>(k: K) -> Result<T, K>
```
davidpdrsn, line 24: Its intentionally duplicated. We don't want `try_downcast` to be part of axum-core's public API.

davidpdrsn (changes requested):

axum-core/src/body.rs
dengzhq: I'm not quite sure whether really need to fix it.
I found the previous commit and issue about the origin of axum-core,it is used as the foundation for the rest of axum crates.
As a developer, Duplicated Code is a code smell in my experience. Because when you want to change them, multiple codes need to be changed. 
there are some common items (eg: try_downcast, macro_rules! composite_rejection) in both axum and axum-core.
Are there plans to refactor Duplicated Code in the future?  Or what do you think about this

davidpdrsn: Duplicated code isn't always bad. Its a trade-off. In this case we choose a little bit of duplication over adding these to the public API.

davidpdrsn:
Like mentioned I don't think we'll do this so I'll close for now.

PR #1047, opened 2022-05-20
Implement path extraction for untagged enums.

alvra:
This PR implements path extraction for untagged enums.

I'm not sure if this this is the right approach though, so any feedback is appreciated.

This enables "flattened" enums:
```rust
#[derive(serde::Deserialize, Debug)]
#[serde(untagged)]
enum Main {
    A(A),
    B(B),
}

#[derive(serde::Deserialize, Debug)]
enum A {
    A1,
    A2,
}

#[derive(serde::Deserialize, Debug)]
enum B {
    B1,
    B2,
}
```

### DIFF
--- a/axum/src/extract/path/de.rs
+++ b/axum/src/extract/path/de.rs
@@ -837,20 +837,20 @@ mod tests {
 
     #[test]
     fn test_untagged_enum() {
-        #[derive(serde::Serialize, serde::Deserialize, PartialEq, Eq, Debug)]
+        #[derive(Deserialize, PartialEq, Eq, Debug)]
         #[serde(untagged)]
         enum Main {
             A(A),
             B(B),
         }
 
-        #[derive(serde::Serialize, serde::Deserialize, PartialEq, Eq, Debug)]
+        #[derive(Deserialize, PartialEq, Eq, Debug)]
         enum A {
             A1,
             A2,
         }
 
-        #[derive(serde::Serialize, serde::Deserialize, PartialEq, Eq, Debug)]
+        #[derive(Deserialize, PartialEq, Eq, Debug)]
         enum B {
             B1,
             B2,
--- a/axum/src/extract/path/de.rs
+++ b/axum/src/extract/path/de.rs
@@ -56,7 +56,6 @@ impl<'de> PathDeserializer<'de> {
 impl<'de> Deserializer<'de> for PathDeserializer<'de> {
     type Error = PathDeserializationError;
 
-    unsupported_type!(deserialize_any, "any");
     unsupported_type!(deserialize_bytes, "bytes");
     unsupported_type!(deserialize_option, "Option<T>");
     unsupported_type!(deserialize_identifier, "identifier");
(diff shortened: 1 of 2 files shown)


axum/src/extract/path/de.rs
```diff
@@ -79,6 +78,18 @@ impl<'de> Deserializer<'de> for PathDeserializer<'de> {
...
+    fn deserialize_any<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        if self.url_params.len() != 1 {
+            return Err(PathDeserializationError::wrong_number_of_parameters()
+                .got(self.url_params.len())
+                .expected(1));
+        }
+        visitor.visit_str(&self.url_params[0].1)
```
davidpdrsn, line 90: Hm I'm wondering if this could break something else 🤔 Not very familiar with these kinds of deserializer details.

@jplatte do you know?

axum/src/extract/path/de.rs
```diff
@@ -823,4 +834,38 @@ mod tests {
...
+        assert_eq!(
+            Main::deserialize(PathDeserializer::new(&url_params)).unwrap(),
+            Main::A(A::A2)
+        );
+
+        let url_params = create_url_params(vec![("a", "B2")]);
+        assert_eq!(
+            Main::deserialize(PathDeserializer::new(&url_params)).unwrap(),
+            Main::B(B::B2)
+        );
```
davidpdrsn, line 869: Thanks for adding a test! Would you wanna add an integration test as well [here](https://github.com/tokio-rs/axum/blob/main/axum/src/extract/path/mod.rs#L412)? Then we'll know that it actually works end-to-end.

alvra: Done!

axum/src/extract/path/de.rs
jplatte: It doesn't make a lot of sense to me to require there to be exactly one url param for `deserialize_any`. I think deserializing to a free-form thing like `serde_json::Value` should probably get you a sequence of all the values. A sequence of key-value pair like `deserialize_any` on a query string could make sense as well, but here the names are implicit so I'm leaning towards just values.

In any case, no longer erroring in `deserialize_any` isn't going to break anything, but deciding for one behavior of that function means that other reasonable implementations can no longer be added backwards-compatibly.

Overall, I don't think we should merge this. Instead of `serde(flatten)`, `serde(from)` / `serde(try_from)` with an intermediate type (that contains the variants of both inner enums in the example case) would be a cleaner and more efficient way of solving OP's problem. `serde(flatten)` is a kind of ugly thing in general, it's slow and produces very bad error messages when deserialization fails.

alvra:
@jplatte Thanks for looking into this. I'm not very familiar with serde internals, but I was hoping to avoid using an itermediary enum.

Btw, this is not using `serde(flatten)` but `serde(untagged)`.

By "deciding for one behavior" do you mean deserializing as a string (`visit_str`)? Is there a way to implement this so that future extensions are possible?

jplatte:
Err yes, sorry. I actually meant `serde(untagged)`, though the same problems affects `serde(flatten)` too, IIRC.

jplatte:
> By "deciding for one behavior" do you mean deserializing as a string (`visit_str`)? Is there a way to implement this so that future extensions are possible?

Yes, exactly. There is no way to implement this such that future extensions are possible (outside of having different behavior for how many url params there are, which would probably be rather confusing).

To clarify: The way `untagged` deserialization works is that the input is first deserialized to a catch-all type similar to `serde_json::Value`, where the deserializer has to chose the representation (as opposed to the type being deserialized driving deserializion which is usually the case). Then deserialization from that immediate representation is attempted for the inner types of the enum in order of declaration.

If we were to merge this PR, `serde(flatten)` on an enum containing a `Vec` variant could not be made to work for the one-element case. At the same time, the behavior of deserializing to a catch-all type would be surprising.

alvra:
I understand. Thanks for your time!

PR #1048, opened 2022-05-22
Automatically handle `http_body::LengthLimitError`

davidpdrsn:
This adds special handling of `http_body::LengthLimitError` to `FailedToBufferBody` such that this just works and returns `413 Payload Too Large` if the limit is exceeded.

tower-http's new `RequestBodyLimitLayer` uses that so limiting both requests with a known length and streaming is just:

```rust

Router::new()
    .route("/", post(|body: Bytes| { ... }))
    .layer(RequestBodyLimitLayer::new(1024));
```

---

I think changing `ContentLengthLimit` to use `http_body::Limited` would be good but we cannot do that without breaking changes since it requires `T: FromRequest<B>` but we have to change that to `T: FromRequest<Limited<B>>`.

EDIT: Actually I don't think we can do that since you cannot change the body type in a `RequestParts` with only an `&mut RequestParts`. You'd need an owned value 🤔

### DIFF
--- a/axum-core/src/extract/rejection.rs
+++ b/axum-core/src/extract/rejection.rs
@@ -56,6 +56,11 @@ define_rejection! {
     #[status = PAYLOAD_TOO_LARGE]
     #[body = "Failed to buffer the request body"]
     /// Encountered some other error when buffering the body.
+    ///
+    /// This can  _only_ happen when you're using [`tower_http::limit::RequestBodyLimitLayer`] or
+    /// otherwise wrapping request bodies in [`http_body::Limited`].
+    ///
+    /// [`tower_http::limit::RequestBodyLimitLayer`]: https://docs.rs/tower-http/0.3/tower_http/limit/struct.RequestBodyLimitLayer.html
     pub struct LengthLimitError(Error);
 }
 
--- a/axum-core/src/extract/rejection.rs
+++ b/axum-core/src/extract/rejection.rs
@@ -2,6 +2,7 @@
 
 use crate::response::{IntoResponse, Response};
 use http::StatusCode;
+use http_body::LengthLimitError;
 use std::fmt;
 
 /// Rejection type used if you try and extract the request body more than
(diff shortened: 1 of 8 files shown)


axum-core/src/extract/rejection.rs
```diff
@@ -28,12 +29,74 @@ impl fmt::Display for BodyAlreadyExtracted {
...
-define_rejection! {
-    #[status = BAD_REQUEST]
-    #[body = "Failed to buffer the request body"]
-    /// Rejection type for extractors that buffer the request body. Used if the
-    /// request body cannot be buffered due to an error.
-    pub struct FailedToBufferBody(Error);
+/// Rejection type for extractors that buffer the request body. Used if the
+/// request body cannot be buffered due to an error.
+// TODO: in next major for axum-core make this a #[non_exhaustive] enum so we don't need the
+// additional indirection
```
jplatte, line 35: Couldn't we do this now? The only thing I know you can do with the current definition you couldn't with an enum is matching on it with a pattern of `FailedToBufferBody { .. }` and probably `FailedToBufferBody(..)`, which is a useless operation that might be disallowed entirely in the future to make changes from struct-with-only-private-fields to enum strictly okay.

axum-core/src/extract/rejection.rs
```diff
@@ -28,12 +29,74 @@ impl fmt::Display for BodyAlreadyExtracted {
...
+#[derive(Debug)]
+pub struct FailedToBufferBody(FailedToBufferBodyInner);
+
+impl FailedToBufferBody {
+    /// Check if the body failed to be buffered because a length limit was hit.
+    ///
+    /// This can  _only_ happen when you're using [`tower_http::limit::RequestBodyLimitLayer`] or
```
jplatte, line 42:
```suggestion
    /// This can _only_ happen when you're using [`tower_http::limit::RequestBodyLimitLayer`] or
```

axum-core/src/extract/rejection.rs
```diff
@@ -28,12 +29,74 @@ impl fmt::Display for BodyAlreadyExtracted {
...
+    /// otherwise wrapping request bodies in [`http_body::Limited`].
+    ///
+    /// [`tower_http::limit::RequestBodyLimitLayer`]: https://docs.rs/tower-http/latest/tower_http/limit/struct.RequestBodyLimitLayer.html
```
jplatte, line 45:
```suggestion
    /// [`tower_http::limit::RequestBodyLimitLayer`]: https://docs.rs/tower-http/0.3/tower_http/limit/struct.RequestBodyLimitLayer.html
```

axum-core/src/extract/rejection.rs
```diff
@@ -28,12 +29,74 @@ impl fmt::Display for BodyAlreadyExtracted {
...
+        }
+    }
+}
+
+impl crate::response::IntoResponse for FailedToBufferBody {
+    fn into_response(self) -> crate::response::Response {
+        match self.0 {
+            FailedToBufferBodyInner::Unknown(err) => (
+                http::StatusCode::BAD_REQUEST,
+                format!(concat!("Failed to buffer the request body", ": {}"), err),
```
jplatte, line 75:
```suggestion
                format!("Failed to buffer the request body: {}", err),
```

axum-core/src/extract/rejection.rs
```diff
@@ -28,12 +29,74 @@ impl fmt::Display for BodyAlreadyExtracted {
...
+            )
+                .into_response(),
+            FailedToBufferBodyInner::LengthLimitError(err) => (
+                StatusCode::PAYLOAD_TOO_LARGE,
+                format!(concat!("Failed to buffer the request body", ": {}"), err),
```
jplatte, line 80:
```suggestion
                format!("Failed to buffer the request body: {}", err),
```

axum-core/src/extract/rejection.rs
davidpdrsn: I think you're right 😊 I've changed it to an enum.

axum/src/extract/content_length_limit.rs
```diff
@@ -29,6 +29,9 @@ use std::ops::Deref;
 /// ```
 ///
 /// This requires the request to have a `Content-Length` header.
+///
+/// If you want to limit the size of request bodies without requiring a `Content-Length` header
```
jplatte, line 33:
```suggestion
/// If you want to limit the size of request bodies without requiring a `Content-Length` header,
```

jplatte (approved):

PR #1049, opened 2022-05-22
Use `impl IntoResponse` less in docs

davidpdrsn:
Previously we used `impl IntoResponse` quite a lot in the docs to avoid having to write the full return type. However we've seen several cases on Discord and Twitter where people who are new to Rust gets confused about that. People sometimes thinks it allows returning different types along the lines of

```rust
async fn handle() -> impl IntoResponse {
    if something() {
        "foo"
    } else {
        StatusCode::BAD_REQUEST
    }
}
```

Which unfortunately wont work 😕 Writing out the full type should make the issue clear.

So I think we're actually doing people a disservice using `impl IntoResponse` as much as we do. In general I think our docs should cater to beginners perhaps more than other crates, since web frameworks are likely to be the first big crate a newcomer picks up.

So this changes the docs to spell out the return type where it is appropriate. I also added a section about calling `.into_response()` to more easily return different types.

### DIFF
--- a/axum-core/src/response/into_response_parts.rs
+++ b/axum-core/src/response/into_response_parts.rs
@@ -57,7 +57,10 @@ use std::{
 /// }
 ///
 /// // We can now return `SetHeader` in responses
-/// async fn handler() -> impl IntoResponse {
+/// //
+/// // Note that returning `impl IntoResponse` might be easier if the response has many parts to
+/// // it. The return type is written out here for clarity.
+/// async fn handler() -> (SetHeader<'static>, SetHeader<'static>, &'static str) {
 ///     (
 ///         SetHeader("server", "axum"),
 ///         SetHeader("x-foo", "custom"),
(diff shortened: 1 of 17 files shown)


jplatte (approved):

axum-core/src/response/into_response_parts.rs
```diff
@@ -57,7 +57,10 @@ use std::{
 /// }
 ///
 /// // We can now return `SetHeader` in responses
-/// async fn handler() -> impl IntoResponse {
+/// //
+/// // Note that returning `impl IntoResponse` might be easier if the response has many parts to
+/// // it. The return type is written out here for clarity.
+/// async fn handler() -> (SetHeader<'static>, SetHeader<'static>, &'static str) {
 ///     (
```
bsodmike, line 64: @davidpdrsn doesn't the Impl IntoReponse interface recurse over the tuple essentially calling `.into_reponse()` on each tuple element?

davidpdrsn: It does yes.

Issue #1057, opened 2022-05-26
Possible bug in documentation of axum::extract::Path


fynzhak:
### Version

axum v0.5.6
├── axum-core v0.2.4

### Platform

Windows 10 64-bit

### Description

The documentation of extract::Path says that I should be able to extract all path parameters as a ```Path<Vec<(String, String)>>```.
However, when I try this, my handler responds with a status code of 500 and a response body of 
```Unsupported type '(alloc::string::String, alloc::string::String)'```
It seems that I should in fact be using a ```Path<Vec<String>>```.

I've put together a small sample to illustrate.

```rust
use std::net::SocketAddr;

use axum::{Router, routing::get, extract::Path};
use tower_http::trace::TraceLayer;
use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};

#[tokio::main]
async fn main() {
    tracing_subscriber::registry()
        .with(tracing_subscriber::EnvFilter::new(
            std::env::var("RUST_LOG")
                .unwrap_or_else(|_| "tower_http=debug".into()),
        ))
        .with(tracing_subscriber::fmt::layer())
        .init();

    let app = Router::new()
        .route("/users/:user_id/team/:team_id", get(params_vec))
        .layer(TraceLayer::new_for_http());

    let addr = SocketAddr::from(([127, 0, 0, 1], 9009));
    println!("listening on {}", addr);
    axum::Server::bind(&addr)
        .serve(app.into_make_service())
        .await
        .unwrap();
}

async fn params_vec(
    Path(params): Path<Vec<(String, String)>>,
) {
    println!(">>> params_vec: {params:?}");
}

async fn params_vec_ok(
    Path(params): Path<Vec<String>>,
) {
    println!(">>> params_vec: {params:?}");
}
```

Is this a bug in the documentation, the Path extractor, or am I  just doing something totally wrong?

Thanks!

davidpdrsn:
Oh whoops! Thanks for noticing. It'll be fixed in https://github.com/tokio-rs/axum/pull/1059

PR #1059, opened 2022-05-27
Support `Path<Vec<(String, String)>>`

davidpdrsn:
We mentioned using `Path<Vec<(String, String)>>` to get key/value pairs in [the docs](https://docs.rs/axum/latest/axum/extract/struct.Path.html#example) but it didn't work at all 🤦 Was probably just wishful thinking 😅

However this makes it work!

Fixes https://github.com/tokio-rs/axum/issues/1057

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 # Unreleased
 
 - **added:** Implement `Default` for `Extension` ([#1043])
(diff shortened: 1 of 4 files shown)


jplatte (approved):
It looks like `path/de.rs` doesn't have a test for the (probably more important) `Vec<String>`, but otherwise LGTM.

davidpdrsn:
@jplatte it well hidden but we have https://github.com/tokio-rs/axum/blob/main/axum/src/extract/path/de.rs#L707-L711

PR #1072, opened 2022-06-08
Add `AsyncReadBody`

davidpdrsn:
## Motivation

I think having an easy way to respond with an `AsyncRead` is useful. Its currently possible by combining things from the ecosystem but that requires you know about tokio-util.

I'm also thinking about adding a `FileResponse` that'll automatically set `content-type` and related headers. That'll require an `AsyncReadBody` which also makes sense on its own.

## Solution

Add `AsyncReadBody` which is a body created from a `tokio::io::AsyncRead`.

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -20,6 +20,7 @@ cookie-private = ["cookie", "cookie-lib/private"]
 form = ["serde", "serde_html_form"]
 query = ["serde", "serde_html_form"]
 spa = ["tower-http/fs"]
+async-read-body = ["tokio-util/io"]
 
 [dependencies]
 axum = { path = "../axum", version = "0.5", default-features = false }
@@ -27,6 +28,7 @@ bytes = "1.1.0"
 http = "0.2"
 mime = "0.3"
 pin-project-lite = "0.2"
+tokio = "1.19"
 tower = { version = "0.4", default_features = false, features = ["util"] }
 tower-http = { version = "0.3", features = ["map-response-body"] }
 tower-layer = "0.3"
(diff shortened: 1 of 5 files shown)


axum-extra/Cargo.toml
```diff
@@ -20,13 +20,15 @@ cookie-private = ["cookie", "cookie-lib/private"]
 form = ["serde", "serde_html_form"]
 query = ["serde", "serde_html_form"]
 spa = ["tower-http/fs"]
+async-read-body = ["tokio-util/io"]
```
jplatte, line 23: Alphabetical sorting please (but maybe after `default`, potentially newline-separated?) 🙏🏼

axum-extra/src/body/async_read_body.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// use tokio::fs::File;
+    ///
+    /// async fn cargo_toml() -> Result<Response, (StatusCode, String)> {
+    ///     let file = File::open("Cargo.toml")
+    ///         .await
+    ///         .map_err(|err| {
+    ///             (StatusCode::NOT_FOUND, format!("File not found: {}", err))
+    ///         })?;
+    ///
+    ///     let headers = [(CONTENT_TYPE, "text/x-toml")];
```
jplatte, line 39: I think we should use `TypedHeader` here, just because it's cool it can be used in responses 😉

davidpdrsn: [`ContentType`](https://docs.rs/headers/latest/headers/struct.ContentType.html#) doesn't have a `toml` constructor and neither does `mime`. So not sure thats possible.

jplatte: Oh 🙁

The `mime` crate should really have a `mime!` macro that does some equivalent of `const { $string_lit.parse::<Mime>().unwrap() }`.

jplatte (approved):

axum-extra/src/body/async_read_body.rs
davidpdrsn: Yeah that would be cool

PR #1080, opened 2022-06-09
Add `axum_extra::response::FileResponse`

davidpdrsn:
For easily responding with the contents of a file, with `content-type` inferred from the path.

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -18,6 +18,7 @@ cookie = ["cookie-lib"]
 cookie-private = ["cookie", "cookie-lib/private"]
 cookie-signed = ["cookie", "cookie-lib/signed"]
 erased-json = ["serde_json", "serde"]
+file-response = ["async-read-body", "mime_guess", "httpdate"]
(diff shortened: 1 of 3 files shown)


NateLing:
I have reservations about this kind of pr, similar to sap and file server, we usually use nginx server for processing in production practice. We usually combine file resources with cdn services.

davidpdrsn:
@NateLing we got this question https://github.com/tokio-rs/axum/discussions/608 which this simplifies a lot.

I don't believe everyone uses nginx and cdns.

jplatte:
I think the fact that this doesn't handle `If-Modified-Since` is a sign that this is the wrong abstraction. For cases like #608, wouldn't it be possible to somehow dynamically create a `ServeFile` or sth. like that?

davidpdrsn:
Hm maybe 🤔

davidpdrsn:
Lets just punt on this.

I would like to have some way for `IntoResponse` to access request headers for stuff like this. Inspecting the `Accept` header is also something I've needed a few times. But thats a bigger thing and probably not possible without breaking changes.

Issue #1068, opened 2022-06-07
How to implement wrapper function that implements Handler?


muusbolla:
I don't think this is a bug or issue per se, but I would still love some help and the documentation around Handler is very sparse. It says "it should just work if you have the correct type", which is true for most free functions, but not for my use case. And the recommended #[debug_handler] macro also doesn't work for my use case.

Like https://github.com/tokio-rs/axum/issues/1052, I was inspired by https://www.fpcomplete.com/blog/axum-hyper-tonic-tower-part4/ to create a combined Tonic/Axum server. However I want to take it a step further by just having a single implementation for each RPC (the GRPC handler) and having Axum endpoints be wrappers around the GRPC handlers. For the Axum endpoint, this involves accepting JSON, deserializing into the GRPC type, calling the Tonic handler, then re-serializing the result back to JSON. This works fine with a manually crafted wrapper function like this:
```rust
struct MyHandler;

#[async_trait]
impl Echo for MyGrpcHandler {
    async fn echo(
        &self,
        request: tonic::Request<EchoRequest>,
    ) -> Result<tonic::Response<EchoReply>, tonic::Status> {
        Ok(tonic::Response::new(EchoReply {
            message: format!("Echoing back: {}", request.get_ref().message),
        }))
    }
}

async fn EchoHandlerWrapper(Json(body): Json<EchoRequest>) -> Result<Json<EchoReply>, HybridError> {
    let handler = MyGrpcHandler {};
    let r = handler.echo(tonic::Request::new(body)).await;
    match r {
        Ok(r) => Ok(Json(r.into_inner())),
        Err(e) => Err(HybridError(e)) // wrap GRPC errors into HTTP errors, not relevant to this issue
    }
}

#[tokio::main]
async fn main() {
    let addr = SocketAddr::from(([0, 0, 0, 0], 3000));
    let axum_make_service = axum::Router::new()
        .route("/Echo", axum::handler::any(EchoHandlerWrapper))
        .into_make_service();
    let grpc_service = tonic::transport::Server::builder()
        .add_service(EchoServer::from_arc(arcHandler))
        .into_service();
    let hybrid_make_service = hybrid(axum_make_service, grpc_service);
    let server = hyper::Server::bind(&addr).serve(hybrid_make_service);
    if let Err(e) = server.await {
        eprintln!("server error: {}", e);
    }
}
```

[...]

PR #1082, opened 2022-06-10
Add generic JSON-over-GRPC wrapper to examples/rest-grpc-multiplex

muusbolla:
Resolves https://github.com/tokio-rs/axum/issues/1068 and https://github.com/tokio-rs/axum/discussions/1069.

Add a function to `examples/rest-grpc-multiplex` that provides a generic wrapper to expose a HTTP/JSON frontend for any gRPC call. For the provided example code, this provides a HTTP endpoint with the following behavior:
```
POST localhost:3000/Hello
Header: content-type "application/json"
Body: { "name": "muusbolla" }

RESPONSE 200 OK
Header: content-type "application/json"
Body: { "message": "Hello muusbolla, my name is HAL 9000." }
```

## Motivation

Make it easy to provide an API that is compatible with both gRPC and JSON, that uses a single backing implementation, with minimal additional boilerplate required for each new endpoint. This can be very helpful for users to create backwards/forwards-compatible APIs.

## Solution
For context - the following is a specific implementation of a JSON wrapper around the `GrpcServiceImpl::say_hello` function that is compatible with the axum::Handler trait:
```rust
async fn json_wrap_hello(Json(body): Json<HelloRequest>) -> Result<Json<HelloReply>, GrpcErrorAsJson> {
    let handler = GRPC_SERVICE.get().unwrap();
    let r = handler.say_hello(tonic::Request::new(body)).await;
    match r {
        Ok(r) => Ok(Json(r.into_inner())),
        Err(e) => Err(GrpcErrorAsJson(e))
    }
}

#[tokio::main]
async fn main() {
    let rest = Router::new()
        .route("/Hello", axum::routing::any(json_wrap_hello);
}
```
The json_wrap_grpc function included in this PR is a genericization of the above code, making it possible to wrap any GRPC handler into an axum::Handler with a single function call.

It was quite a hassle to make the generic version compile (I was butting heads with https://github.com/rust-lang/rust/issues/36582 and/or https://github.com/rust-lang/rust/issues/41078), so I am hoping that this can be helpful to others who might want to do something similar.

### DIFF
--- a/examples/rest-grpc-multiplex/src/main.rs
+++ b/examples/rest-grpc-multiplex/src/main.rs
@@ -100,8 +100,8 @@ async fn main() {
 
     // build the rest service
     let rest = Router::new()
-    .route("/", get(web_root))
-    .route("/Hello", axum::routing::any(json_wrap_grpc(GrpcServiceImpl::say_hello)));
+        .route("/", get(web_root))
+        .route("/Hello", axum::routing::any(json_wrap_grpc(GrpcServiceImpl::say_hello)));
 
     // build the grpc service
     let grpc = GreeterServer::from_arc(GRPC_SERVICE.get().unwrap().clone());
--- a/examples/rest-grpc-multiplex/src/main.rs
+++ b/examples/rest-grpc-multiplex/src/main.rs
@@ -4,13 +4,17 @@
 //! cd examples && cargo run -p example-rest-grpc-multiplex
 //! ```
 
-use self::multiplex_service::MultiplexService;
-use axum::{routing::get, Router};
+use self::multiplex_service::{GrpcErrorAsJson, MultiplexService};
+use axum::{extract::Json, routing::get, Router};
+use once_cell::sync::OnceCell;
 use proto::{
     greeter_server::{Greeter, GreeterServer},
     HelloReply, HelloRequest,
 };
+use std::future::Future;
 use std::net::SocketAddr;
+use std::pin::Pin;
+use std::sync::Arc;
 use tonic::{Response as TonicResponse, Status};
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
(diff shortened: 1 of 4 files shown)


examples/rest-grpc-multiplex/src/main.rs
```diff
@@ -32,19 +37,59 @@ impl Greeter for GrpcServiceImpl {
...
+/// 2. calls the gRPC RPC implementation
+/// 3. serializes the response back to JSON
+fn json_wrap_grpc<'a, 'r, F, ReqT, ResT> (grpc_impl_func: F)
+    -> impl FnOnce(Json<ReqT>)
+        -> Pin<Box<dyn Future<Output = Result<Json<ResT>, GrpcErrorAsJson>> + Send + 'a>> + Clone + Send + Sized + 'static
+where
+    F: FnOnce(&'r GrpcServiceImpl, tonic::Request<ReqT>)
+        -> Pin<Box<dyn Future<Output = Result<tonic::Response<ResT>, tonic::Status>> + Send + 'r>> + Clone + Send + Sync + 'static,
+    for<'de> ReqT: serde::Deserialize<'de> + Send + 'a,
+    ResT: serde::Serialize
```
davidpdrsn, line 67: I appreciate what you're trying to do but we cannot merge something this complicated as an example. That is just gonna confuse people. So unless you're able to find a simpler solution I don't think we can merge this.

davidpdrsn (changes requested):

NateLing:
To clarify here, I integrated tonic in axum, which is mainly used to solve the internal communication of the cluster. But I think, your example is helpful for wanting to access rpc through rest interface. Before, I referred to some spring projects, which can implement an interface to receive http, grpc and websocket access at the same time through annotations, but I don't have time to study these details now. When I have time, I'd like to try if I can do it based on axum and tonic.

muusbolla:
> To clarify here, I integrated tonic in axum, which is mainly used to solve the internal communication of the cluster. But I think, your example is helpful for wanting to access rpc through rest interface. Before, I referred to some spring projects, which can implement an interface to receive http, grpc and websocket access at the same time through annotations, but I don't have time to study these details now. When I have time, I'd like to try if I can do it based on axum and tonic.

If there was an implementation that was codegen-based, or somehow had access to the routing of tonic and could automatically generate routes based on that, that would be really nice.

examples/rest-grpc-multiplex/src/main.rs
muusbolla: Yeah, I understand. Now that I've made the PR at least people who are looking to do this can find it from the discussion

PR #1083, opened 2022-06-10
Add methods to easily run extractors from middleware

davidpdrsn:
Lately we've had a few people ask about how run extractors from middleware. This was previously possible by building your own `RequestParts` and then running the extractor. However the fact that people had to ask means we probably should have an easier way of doing it.

So this adds a `RequestExt` trait with helper methods for doing that. It also adds `RequestParts::into_parts` so you can get the `http::request::Parts` back and rebuild the request.

## TODO

- [ ] changelog

### DIFF
--- a/axum-core/src/extract/mod.rs
+++ b/axum-core/src/extract/mod.rs
@@ -176,6 +176,28 @@ impl<B> RequestParts<B> {
         Ok(req)
     }
 
+    /// Get the underlying `http::request::Parts`.
+    pub fn into_parts(self) -> http::request::Parts {
+        let Self {
+            method,
+            uri,
+            version,
+            headers,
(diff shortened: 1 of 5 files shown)


davidpdrsn:
Actually. I'm wondering if something like

```rust
async fn middleware(
    Extension(foo): Extension<Foo>,
    request: Request<B>,
    next: Next<B>,
) -> Response {
    // use `foo`
    next.run(request).await
}
```

Could work. So extending `middleware::from_fn` so it can run extractors automatically. Then we wouldn't need any new API. I'm gonna explore that.

davidpdrsn:
Superseded by https://github.com/tokio-rs/axum/pull/1088

Issue #1053, opened 2022-05-25
Adding nested fallbacks is harder than it should be


davidpdrsn:
We currently don't support nesting routers that have fallbacks. This panics:

```rust
let api_routes = Router::new()
    .route("/v1/foo", get(|| async {}))
    .fallback((|| async { "api fallback" }).into_service());

let app = Router::new()
    .nest("/api", api_routes)
    .fallback((|| async { "general fallback" }).into_service());
```

This is because while `Router::nest` takes `T: Service` it attempts to downcast that into a `Router`. If that succeeds it takes a different code path that simply moves routes from one router to another while adding the prefix. However that means we cannot have two `fallbacks` since we only have one `Router` in the end.

A possible workaround is to use another `Router` as a fallback:

```rust
let api_routes = Router::new()
    .route("/v1/foo", get(|| async {}));

let app = Router::new()
    .nest("/api", api_routes)
    .fallback(
        Router::new()
            .nest("/api", (|| async { "api fallback" }).into_service())
            .fallback((|| async { "general fallback" }).into_service()),
    );
```

While this works its not very obvious and it would be great if multiple nested fallbacks just worked.

Haven't thought through how to make it work but I believe it requires https://github.com/ibraheemdev/matchit/issues/18

lilyball:
In the short term, I believe that simply treating the nested router as an opaque service in the case where it has a fallback would work. You cannot currently add a nested router that has a fallback, so that approach can't possibly break any code. This would only need to be done if the nested router actually has a fallback, otherwise the current behavior is fine.

davidpdrsn:
That's a good idea actually 🤔 might get us most of the way there.

lilyball:
I don't know what the expected time frame for https://github.com/ibraheemdev/matchit/issues/18 is. If it will be a while, another potential workaround would be to detect the nested router w/fallback, don't merge it, but do convert all of its own routes to be wrapped in `StripPrefix` (instead of wrapping the whole router with that). Then `route_layer` itself could detect that the service for the nested route is a `Router` and call `route_layer` on that nested `Router` instead of applying the layer around it (like how it does for `Endpoint::MethodRouter`).

If I correctly understand the way this works, that would have a similar effect to merging the nested router with ibraheemdev/matchit#18 fixed, and since it's pushing the `StripPrefix` out to the nested routes that would preserve the full path for route layers.

davidpdrsn:
[...]

PR #1086, opened 2022-06-11
Update matchit and fix nesting inconsistencies

davidpdrsn:
While updating to the new matchit I ran into a bunch of inconsistencies regarding `nest`. Specifically cases where nesting a `Router` or an opaque service had different behavior. That is also something people have asked about on discord before.

This addresses that by always nesting services opaque in `nest`, i.e. we don't attempt to downcast it to a `Router` and change the behavior based on that. This makes the behavior more consistent and simplifies the implementation.

---

Old, outdated, description:

While updating to the new matchit I ran into a bunch of inconsistencies regarding `nest`. Specifically cases where nesting a `Router` or an opaque service had different behavior. That is also something people have asked about on discord before.

This addresses that by splitting `nest` into
- `nest`: Which only accepts `Router`s.
- `nest_service`: Which accepts any opaque `Service`.

Once this is merged I'm gonna change `fallback` to work how I described in https://github.com/tokio-rs/axum/issues/1053#issuecomment-1138994642. If we didn't add `nest_service` that would introduce yet another inconsistency so wanted to get this in first.


axum/src/routing/mod.rs
```diff
@@ -419,24 +415,8 @@ where
...
-                };
-
-                let matched_path = format!("{}{}", previous, matched_path);
-                matched_path.into()
-            } else {
-                Arc::clone(matched_path)
-            };
-            req.extensions_mut().insert(MatchedPath(matched_path));
+            req.extensions_mut()
+                .insert(MatchedPath(Arc::clone(matched_path)));
```
davidpdrsn, line 419: This change means that nested opaque services will only see the matched path from their point of view.

i.e. with

```rust
Router::new().nest_service(
    "/a",
    Router::new().route("/:foo", ...)
)
```

the inner router will see the matched path as `/:foo` and _not_ `/a/:foo`. That is consistent with how `Uri` works and why we have `OriginalUri`.

axum/src/routing/mod.rs
```diff
@@ -455,17 +435,6 @@ where
             Endpoint::Route(inner) => inner.call(req),
         }
     }
-
-    fn panic_on_matchit_error(&self, err: matchit::InsertError) {
-        if self.nested_at_root {
-            panic!(
-                "Invalid route: {}. Note that `nest(\"/\", _)` conflicts with all routes. Use `Router::fallback` instead",
-                err,
-            );
```
davidpdrsn, line 464: matchit now allows `/*` and `/whatever` so this failure goes away 🥳

axum/src/docs/routing/nest.md
```diff
@@ -103,18 +73,49 @@ the prefix stripped:
...
+let nested_router = Router::new().route("/users", get(|| async {}));
+
+let nested_service = Router::new().route("/app.js", get(|| async {}));
+
+async fn fallback() {}
+
+let app = Router::new()
+    .nest("/api", nested_router)
+    .nest_service("/assets", nested_service)
+    // the fallback is not called for request starting with `/bar` but will be
```
jplatte, line 109:
```suggestion
    // the fallback is not called for request starting with `/assets` but will be
```

axum/src/docs/routing/nest.md
```diff
@@ -103,18 +73,49 @@ the prefix stripped:
...
+    // called for requests starting with `/foo` if `nested_router` doesn't have
```
jplatte, line 110:
```suggestion
    // called for requests starting with `/api` if `nested_router` doesn't have
```

axum/src/docs/routing/nest.md
```diff
@@ -103,18 +73,49 @@ the prefix stripped:
...
+    // a matching route
+    .fallback(fallback.into_service());
+# let _: Router = app;
+```
+
+Note that you would normally use [`tower_http::services::ServeDir`] for serving
+static files and thus not calling `nest_service` with a `Router`.
```
jplatte, line 117:
```suggestion
static files and thus not call `nest_service` with a `Router`.
```
This fixes the grammar, although I'm still not actually clear on what this sentence wants to tell me 😅

davidpdrsn: The example makes a router that with a `/app.js` route. So this sentence is mostly so people don't assume that that is how you server static files.

In general I think the use cases for using `nest_service` with `Router`s are limited but thats what we wanna show here to point out how the fallback behavior is different.

axum/src/routing/mod.rs
davidpdrsn: Ah wait a sec. I reverted this in https://github.com/tokio-rs/axum/pull/1086/commits/657a0893b51f581b0090747cfdbe95070d126483 because I assumed it was a bug. Had forgotten about this comment. Actually not sure which behavior is better. I'll think about it!

davidpdrsn:
Soon I'll do a more thorough PR description+comments that details the changes a bit better.

axum/src/routing/mod.rs
davidpdrsn: I can't quite decide what the best behavior is here so think we should keep it the way it worked for 0.5.

axum/src/docs/routing/nest.md
davidpdrsn: I've tweaked the example a bit and hopefully made things more clear.

davidpdrsn:
@jplatte This is ready for review now. I hope the docs and changelog makes the changes clear but let me know if you have questions :)

axum/src/docs/routing/nest.md
```diff
@@ -103,18 +73,79 @@ the prefix stripped:
...
 # async {
 # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
 # };
 ```
 
+# Differences between `nest` and `nest_service`
+
+When [fallbacks] are called differs between `nest` and `nested_service`. Routers
+nested with `nest` will delegate to the outer router's fallback if they don't
+have a matching route, whereas `nested_service` will not.
```
lilyball, line 95: This makes it sound like nested routers won't call their own fallback if one is configured. I'm hoping it really means "if the nested router does not have an explicitly-configured fallback".

For perspective, as a user, a fallback is not a "route", and routers don't have "default fallbacks", routers are just collections of routes with an optional fallback, and there's some default behavior in the case where no fallback is configured.

davidpdrsn: A service nested via `Router::nest_service` will not call so the outer routers fallback. If the request matches the prefix then it goes into the opaque nested service and doesn't leave again. This is because axum cannot tell whether the response from the service was from a route or a fallback.

If you need that you gotta use `Router::nest`, which doesn't allow opaque services and instead only `Router`s.

With https://github.com/tokio-rs/axum/pull/1161 we'll support fallbacks on nested routers so I'm not sure there is much reason to call `nest_service` with a `Router`.

lilyball: I'm talking specifically about what this says about `Router::nest`. It says "Routers nested with `nest` will delegate to the outer router's fallback if they don't have a matching route". I suppose that with this PR alone nested routers don't allow fallbacks so this paragraph is okay, but #1161 leaves this language intact.

Given that, I guess #1161 needs to update these docs to better explain fallback behavior with `nest` (and to fix the example of adding a fallback to the nested router as it's currently done via `nest_service`).

davidpdrsn: Yep that's right. 1161 is still a draft and has a todo for updating the docs.

davidpdrsn:
I've been thinking about this and believe the right move is _not_ delegate to fallbacks on outer `Router`s for nested services (like mentioned [here](https://github.com/tokio-rs/axum/pull/1161#issuecomment-1189330754)). This makes use cases like @lilyball describes more annoying but I believe its the right move overall as it makes nesting more consistent.

If you're nesting a lot of `Router`s and want the same fallbacks on all of them I think writing something like this isn't too bad

```rust
fn router() -> Router {
    Router::new().fallback(...)
}

fn make_some_routes() {
    router().route("/", ...)
}
```

I've made the changes here and will fix the merge conflicts now so we can start the review process.

axum-extra/src/routing/resource.rs
```diff
@@ -136,34 +125,6 @@ where
         self.route(&path, delete(handler))
     }
 
-    /// Nest another route at the "member level".
-    ///
-    /// The routes will be nested at `/{resource_name}/:{resource_name}_id`.
-    pub fn nest<T>(mut self, svc: T) -> Self
```
jplatte, line 142: This change seems unrelated to the matchit upgrade, how does it relate?

axum/CHANGELOG.md
```diff
@@ -12,6 +12,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Use `axum::middleware::from_extractor` instead ([#1077])
 - **breaking:** Allow `Error: Into<Infallible>` for `Route::{layer, route_layer}` ([#924])
 - **breaking:** `MethodRouter` now panics on overlapping routes ([#1102])
+- **breaking:** Nested `Router`s will no longer delegate to the outer `Router`'s
+  fallback. Instead you must explicitly set a fallback on the inner `Router` ([#1086])
```
jplatte, line 16: Oh, we actually had this previously? That makes it a little more unfortunate that we now don't support it. Why was it previously possible without a `nest` / `nest_service` split, but is no longer (or is it just removed so `nest` doesn't do magic `Router` special casing)?

axum/CHANGELOG.md
```diff
@@ -12,6 +12,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **breaking:** The request `/foo/` no longer matches `/foo/*rest`. If you want
+  to match `/foo/` you have to add a route specifically for that ([#1086])
+- **breaking:** Path params for wildcard routes no longer include the prefix
+  `/`. e.g. `/foo.js` will match `/*filepath` with a value of `foo.js`, _not_
+  `/foo.js` ([#1086])
```
jplatte, line 21: I think we should separate out these runtime-breaking changes from the rest of the changes, move them to the top and put a big warning in front that these *must* be read when upgrading.

I've been burned by silent breaking changes being burried in the middle of the changelog in the context of actix-web, and I would really like axum to do better.

axum/src/docs/routing/nest.md
```diff
@@ -103,18 +73,73 @@ the prefix stripped:
...
-    }));
+    .nest("/bar", nested_router);
 # async {
 # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
 # };
 ```
 
+# Fallbacks
+
+If a request matches the prefix but the nested router doesn't have matching
```
jplatte, line 93:
```suggestion
If a request matches the prefix but the nested service doesn't have a matching
```

or

```suggestion
When nesting a router, if a request matches the prefix but the nested router doesn't have a matching
```

axum/src/docs/routing/nest.md
```diff
@@ -103,18 +73,73 @@ the prefix stripped:
...
+route the outer fallback will _not_ be called:
```
jplatte, line 94:
```suggestion
route, the outer fallback will _not_ be called:
```

axum/src/extract/matched_path.rs
```diff
@@ -139,33 +140,45 @@ mod tests {
             )
         }
 
-        let app = Router::new()
+        let app: _ = Router::new()
```
jplatte, line 143: What is this for?

axum/src/routing/mod.rs
```diff
@@ -177,7 +171,7 @@ where
         let mut node =
             Arc::try_unwrap(Arc::clone(&self.node)).unwrap_or_else(|node| (*node).clone());
         if let Err(err) = node.insert(path, id) {
-            self.panic_on_matchit_error(err);
+            panic!("Invalid route {:?}. {}", path, err);
```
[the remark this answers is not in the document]
jplatte: Also should the `.` be a `:`?

axum/src/routing/mod.rs
```diff
@@ -200,63 +194,22 @@ where
...
-                    // doesn't mean something is nested at root in _this_ router
-                    // thus we don't need to propagate that
-                    nested_at_root: _,
-                } = router;
-
-                if let Fallback::Custom(_) = fallback {
-                    panic!("Cannot nest `Router`s that has a fallback");
-                }
+        let path = if path.ends_with('/') {
+            format!("{}*{}", path, NEST_TAIL_PARAM)
```
jplatte, line 198:
```suggestion
            format!("{path}*{NEST_TAIL_PARAM}")
```

axum/src/routing/mod.rs
```diff
@@ -200,63 +194,22 @@ where
...
-
-                self = self.route(&path, strip_prefix::StripPrefix::new(svc, prefix));
-            }
+        // `/*rest` is not matched by `/` so we need to also register a router at the
+        // prefix itself. Otherwise if you were to nest at `/foo` then `/foo` itself
+        // wouldn't match, which it should
+        self = self.route(prefix, svc.clone());
+        if !prefix.ends_with('/') {
+            // same goes for `/foo/`, that should also match
+            self = self.route(&format!("{}/", prefix), svc);
```
jplatte, line 212:
```suggestion
            self = self.route(&format!("{prefix}/"), svc);
```

lilyball:
> I've been thinking about this and believe the right move is not delegate to fallbacks to outer Routers for nested services (like mentioned [here](https://github.com/tokio-rs/axum/pull/1161#issuecomment-1189330754)). This makes use cases like @lilyball describes more annoying but I believe its the right move overall as it makes nesting more consistent.

I still think this is a footgun, where people are going to nest routers and then configure a root fallback and forget to configure the same fallback on every level. Even if they make a `fn route() -> Router` they may simply forget to use it in one of the nested routers. The expectation of a `fn route() -> Router` also makes it more complicated to do things like move the nested router generation into other modules.

This also means that splitting up your route definitions into nested groups means it doesn't merge all the routes into a single route set anymore, which was previously being done as an "optimization" and so I assume there is in fact a measurable impact there.

I really don't like that this new design means the best performance is to skip nesting routers and instead set a bunch of routes with identical prefixes.

Have you given any thought to having a separate `RouteSet` type that can have routes added to it like a `Router` but no fallback? Then `Router` can have a `nest_routes()` method that does the same merging of routes into the same route set that it was doing when nesting routers before, and there's no complication over how to handle fallback behavior.

davidpdrsn:
> I really don't like that this new design means the best performance is to skip nesting routers and instead set a bunch of routes with identical prefixes.

I'm fine with that. I still think it's the right trade-off. 

> Have you given any thought to having a separate RouteSet type that can have routes added to it like a Router but no fallback?  Then Router can have a nest_routes() method that does the same merging of routes into the same route set that it was doing when nesting routers before, and there's no complication over how to handle fallback behavior

I have not but like to not complicate the routing more than it is already. The purpose of those is make nesting simpler so adding more methods goes against that.

axum/src/extract/matched_path.rs
davidpdrsn: I don't remember but not necessary :P

axum/src/routing/mod.rs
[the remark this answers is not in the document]
davidpdrsn: Yeah I think it should be.

axum/CHANGELOG.md
davidpdrsn: > or is it just removed so nest doesn't do magic Router special casing

yeah its because of this. We had it previously but only if you passed a `Router` to `nest`. It wasn't documented either and quite magical I think. Thats what I was getting at [here](https://github.com/tokio-rs/axum/pull/1086#issuecomment-1195323750). I agree its unfortunate but think its the right move overall.

axum/CHANGELOG.md
davidpdrsn: I agree. I wanna re-organize the changelog for 0.6 into the different areas that changed and give examples of how to migrate code using 0.5. That will also include calling out subtle changes like this. I think I'll leave it as is and do the re-org in a separate PR when we're closer to 0.6.

axum-extra/src/routing/resource.rs
davidpdrsn: `Resource::{nest, nest_collection}` would accept a `Service` and call `Router::nest`. If you passed a `Router` to `nest_collection` and also had a route for `edit` like so

```rust
Resource::named("users")
    .edit(|| async move {})
    .nest_collection(
        Router::new().route("/featured", get(|| async move {})),
    );
```

That would work because `Router::nest` would downcast the nested service to a router and move over the routes.

However now that we don't downcast anymore you're essentially doing this:

```rust
Router::new()
    .route("/users/:id/edit", get(|| async {}))
    .route("/users/*rest", Router::new());
```

Which is a conflict. `/users/1/edit` could match the edit route with `id = 1` or could match the wildcard with `rest = 1/edit`.

The same would happen previously if you didn't pass a `Router` but some opaque service.

It might be possible to work around this by complicating `Resource` a bunch but I kinda just wanna remove `Resource::nest` and `Resource::nest_collection`. Instead people can convert the `Resource` into a `Router` and add whatever additional routes they want:

```rust
let users = Resource::named("users").index(|| async { "users#index" });
let router = Router::from(users);
router.route("/users/:id/tweets", get(|| async {}));
```

axum/CHANGELOG.md
davidpdrsn: I made an issue about this https://github.com/tokio-rs/axum/issues/1247

PR #1089, opened 2022-06-12
fix missing assets running websockets example

yshing:
## Motivation

Checking out the websockets example recently. When running `cargo run` under `examples/websockets`, the example app isn't serving any assets.

image

## Solution

Make the assets folder relative to `CARGO_MANIFEST_DIR`.

### DIFF
--- a/examples/websockets/src/main.rs
+++ b/examples/websockets/src/main.rs
@@ -32,12 +32,13 @@ async fn main() {
         ))
         .with(tracing_subscriber::fmt::layer())
         .init();
-
+    let mut assets_dir = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    assets_dir.push("assets");
     // build our application with some routes
     let app = Router::new()
         .fallback(
             get_service(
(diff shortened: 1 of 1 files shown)


jplatte:
I think this makes sense in the context of the example, wdyt @davidpdrsn?

examples/websockets/src/main.rs
```diff
@@ -32,12 +32,13 @@ async fn main() {
         ))
         .with(tracing_subscriber::fmt::layer())
         .init();
-
+    let mut assets_dir = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    assets_dir.push("assets");
```
davidpdrsn, line 36:
```suggestion
    let mut assets_dir = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("assets");
```

davidpdrsn (commented):
Yeah I think it makes sense! I've been bitten by this myself while testing stuff :P

davidpdrsn (approved):

PR #1090, opened 2022-06-15
More robust asset paths in examples

davidpdrsn:
https://github.com/tokio-rs/axum/pull/1089 for the other examples that read assets.

### DIFF
--- a/examples/low-level-rustls/src/main.rs
+++ b/examples/low-level-rustls/src/main.rs
@@ -11,7 +11,14 @@ use hyper::server::{
     conn::{AddrIncoming, Http},
 };
 use rustls_pemfile::{certs, pkcs8_private_keys};
-use std::{fs::File, io::BufReader, net::SocketAddr, pin::Pin, sync::Arc};
+use std::{
+    fs::File,
+    io::BufReader,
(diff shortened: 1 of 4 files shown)


examples/low-level-rustls/src/main.rs
```diff
@@ -30,8 +37,8 @@ async fn main() {
         .init();
 
     let rustls_config = rustls_server_config(
-        "examples/tls-rustls/self_signed_certs/key.pem",
-        "examples/tls-rustls/self_signed_certs/cert.pem",
+        PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("self_signed_certs/key.pem"),
```
jplatte, line 40: Maybe use platform-native path separators while you're at it?

```suggestion
        PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("self_signed_certs").join("key.pem"),
```

davidpdrsn: Oh whoops sorry. I'll fix those!

davidpdrsn: There we go!

jplatte (approved):

PR #1093, opened 2022-06-15
Add `JsonLines` extractor and response

davidpdrsn:
This adds `JsonLines` which is a newline delimited JSON extractor and response. Its partly inspired by [`actix_web_lab::respond::NdJson`](https://docs.rs/actix-web-lab/0.16.1/actix_web_lab/respond/struct.NdJson.html).

I had a use case for this today and figured it makes sense to upstream.


axum-extra/src/ndjson.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+pin_project! {
+    #[project = InnerProj]
+    enum Inner<S> {
+        Response {
+            #[pin]
+            stream: S,
+        },
+        Extrator {
```
jplatte, line 76:
```suggestion
        Extractor {
```

axum-extra/src/lib.rs
```diff
@@ -10,11 +10,12 @@
 //! Name | Description | Default?
 //! ---|---|---
 //! `async-read-body` | Enables the `AsyncReadBody` body | No
-//! `cookie` | Enables the `CookieJar` extractor | No
 //! `cookie-private` | Enables the `PrivateCookieJar` extractor | No
 //! `cookie-signed` | Enables the `SignedCookieJar` extractor | No
+//! `cookie` | Enables the `CookieJar` extractor | No
```
jplatte, line 15: Why did you move this down?

FSMaxB:
Just a small comment about the naming. `NdJson` doesn't really explain what it is, I suggest calling it `JsonLines` instead. This would also be in line with https://jsonlines.org/

davidpdrsn:
@FSMaxB Good point! I had only heard of "newline delimited json" but seeing as json lines appears to be thing I like that name more.

axum-extra/src/lib.rs
davidpdrsn: Hm probably just vim sorting being weird

axum-extra/src/ndjson.rs
davidpdrsn: I'll fix the compile errors this will give later :)

axum-extra/src/lib.rs
```diff
@@ -67,6 +68,9 @@ pub mod extract;
 pub mod response;
 pub mod routing;
 
+#[cfg(feature = "json-lines")]
+pub mod json_lines;
```
jplatte, line 72: In `axum`, only the `Json` type is exported at the root, not the `json` module. Does this deviate from that because there's more types to expose?

jplatte (approved):
I've only skimmed but it looks good from that.

axum-extra/src/lib.rs
davidpdrsn: Exactly!

PR #1095, opened 2022-06-15
Implement `IntoResponse` for `Form`

davidpdrsn:
I figure this makes sense to support though not the most useful thing ever.

Required moving the `form` module to the root and re-exporting `Form` from both `extract` and `response`.


jplatte:
So, what's the use case? 😄

jplatte (approved):

PR #1098, opened 2022-06-17
Bump MSRV to 1.56 and update to the 2021 edition

davidpdrsn:
[CI builds are failing](https://github.com/tokio-rs/axum/runs/6939841098?check_suite_focus=true) because a dependency updated to the latest `hashbrown` which uses the 2021 edition. That isn't available in our MSRV which is 1.54. I've been reluctant to bump our MSRV without a good reason but IMO this is a good reason. Not quite sure how else to resolve it, without pinning to specific versions of crates which is a hassle.

1.56 came up in October 2021 so thats like 8 months which I think is reasonable.

I don't consider this a breaking change.

All the changes were done with search and replace. There are probably some places in the code that could be moved to newer features but we can do that in follow up PRs when we come across them.


davidpdrsn:
We could do something like https://github.com/tower-rs/tower/pull/670 but I still bumping our MSRV is fine

jplatte (approved):

PR #1099, opened 2022-06-17
Remove explicit TryFrom, TryInto imports

jplatte:
Followup to #1098.


examples/reverse-proxy/src/main.rs
```diff
@@ -14,7 +14,7 @@ use axum::{
     Router,
 };
 use hyper::{client::HttpConnector, Body};
-use std::{convert::TryFrom, net::SocketAddr};
+use std::{ net::SocketAddr};
```
davidpdrsn, line 17:
```suggestion
use std::net::SocketAddr;
```

Issue #1097, opened 2022-06-17
MethodRouter should panic on overlapping routes


davidpdrsn:
`get(foo).get(bar)` should panic similarly to what `Router` does.

davidpdrsn:
Fixed in https://github.com/tokio-rs/axum/pull/1102

PR #1102, opened 2022-06-18
Panic on overlapping routes in `MethodRouter`

davidpdrsn:
Makes `get(foo).get(bar)` panic, similarly to `Router::route`.

Fixes https://github.com/tokio-rs/axum/issues/1097


jplatte (commented):
What about `get(foo).head(bar)`? We probably want that to work, no?

davidpdrsn:
It does work already but I added tests just to make sure. It works because we only override the corresponding service so doing `get(foo)` doesn't set the `head` service. Instead we call `get` for both `GET` and `HEAD` requests here https://github.com/tokio-rs/axum/blob/main/axum/src/routing/method_routing.rs#L1079-L1080.

jplatte (approved):

PR #1108, opened 2022-06-20
axum: Version 0.5.9

davidpdrsn:

PR #1109, opened 2022-06-20
Use minimal dependency versions for MSRV tests on CI

davidpdrsn:
In an attempt to fix CI for https://github.com/tokio-rs/axum/pull/1108


jplatte:
I'd be surprised if this worked tbqh, considering min-versions AFAIK still operates recursively and thus only works if all dependendencies have working minimal version constraints.

davidpdrsn:
Yeah seems to just result in new errors https://github.com/tokio-rs/axum/runs/6966043238?check_suite_focus=true 😢

davidpdrsn:
@jplatte It should build now 🤞 I ended up adding an internal crate that depends on crates that don't build if we only pull in otherwise minimal versions. From reading https://github.com/rust-lang/cargo/issues/5657 it seemed like the only viable solution.

Issue #1115, opened 2022-06-23
Idea: Split FromRequest trait into three (like the Fn traits)


jplatte:
`FromRequest` with a shared-reference parameter would mostly be useful for middleware or other places where you have a request that you want to pull some bits out of while knowing you're not going to modify it for the final handler.
`FromRequestMut` would be used for all but the last extractor and `FromRequestOnce` would be used for the last one.

I think this could even obviate the need for the `RequestParts` type.

I'll mark as hard because I think it requires changes to all of the crates, including both `macro_rules!` and proc-macro code.

davidpdrsn:
I think this sounds very promising. I'm gonna explore it!

davidpdrsn:
I give this a quick shut but ran into some issues

```rust
trait IntoResponse {}
impl IntoResponse for () {}

struct Request<B>(B);

trait FromRequestOnce<B>: Sized {
    type Rejection: IntoResponse;

    fn from_request_once(req: Request<B>) -> Result<Self, Self::Rejection>;
}

trait FromRequestMut<B>: FromRequestOnce<B> + Sized {
    fn from_request_mut(req: &mut Request<B>) -> Result<Self, Self::Rejection>;
}

// I think this impl would be nice, otherwise you have to manually implement both traits all the
// time
impl<B, T> FromRequestOnce<B> for T
where
    T: FromRequestMut<B>,
{
    // what type should this be? `Rejection` is defined in `FromRequestOnce` which is the trait
    // we're implementing, but we're delegating to `FromRequestMut` which doesn't have it...
    type Rejection = T::Rejection;

    fn from_request_once(mut req: Request<B>) -> Result<Self, Self::Rejection> {
        T::from_request_mut(&mut req)
    }
}

// error[E0275]: overflow evaluating the requirement `<() as FromRequestOnce<B>>::Rejection == _`
impl<B> FromRequestMut<B> for () {
    fn from_request_mut(req: &mut Request<B>) -> Result<Self, Self::Rejection> {
        todo!()
    }
}
```

jplatte:
Hm, I wonder whether the trait hierarchy is even needed, maybe the blanket impls are enough. In that case I think the solution would be to just have `type Rejection` on all of the traits.

davidpdrsn:
Yeah that was my first though as well but that leads to conflicting impls:

```rust
trait FromRequestOnce<B> {
    type Rejection: IntoResponse;

[...]

PR #1121, opened 2022-06-26
Only allow last extractor to mutate the request

davidpdrsn:
Don't worry too much about the large diff. The vast majority of the changes is from code that previously was a `macro_rules` macro now being expanded and committed.

---

Fixes https://github.com/tokio-rs/axum/issues/1115

This is an implementation of the design I suggested [here](https://github.com/tokio-rs/axum/issues/1115#issuecomment-1166462990).

All these changes are to address the long standing issue that this fails at
runtime:

```rust
async fn handle(
    req: Request<Body>,
    ext: Extension<Foo>,
) {}
```

It fails because `Request<Body>` removes all the extensions so `Extension<Foo>`
fails. The solution is to swap the lines so `Request<Body>` runs last.

With this change that becomes a compile error:

```
error[E0277]: the trait bound `fn(Request<Body>, Extension<Foo>) -> impl Future<Output = ()> {handler}: Handler<_, _>` is not satisfied
   --> hello-world/src/main.rs:13:44
    |
13  |     let app = Router::new().route("/", get(handler));
    |                                        --- ^^^^^^^ the trait `Handler<_, _>` is not implemented for `fn(Request<Body>, Extension<Foo>) -> impl Future<Output = ()> {handler}`
    |                                        |
    |                                        required by a bound introduced by this call
    |
```

`#[debug_handler]` will give a more precise error.

It also makes multiple body extractors a compile error:

```rust
async fn handle(
    _: Json<User>,
    _: String,
) {}
```

It works by changing `FromRequest` to

```rust
// new type paramter `R`!
pub trait FromRequest<R, B> {
    type Rejection: IntoResponse;

    async fn from_request(req: &mut RequestParts<R, B>) -> Result<Self, Self::Rejection>;
}
```

In practice `R` is either `Mut` or `Once` which are zero sized marker types. Not stoked
about the names 🤷 

If you have a `RequestParts<Mut, _>` you cannot remove all the extensions, all
headers, or the body. I.e. the extractor is safe to run multiple times. If you
have a `RequestParts<Once, _>` then you can consume all the headers, extensions,
or the body.

So if your extractor needs to consume the body you implement `FromRequest<Once, _>`
otherwise you implement `FromRequest<R, _>`.

Then the `Handler` impls all now require the last extractor to implement
`FromRequest<Once, _>` and all the first to implement `FromRequest<Mut, _>`.
Thus if you have multiple extractors that consume the body that function wont
implement `Handler`.

## What to focus on when reviewing

Most changes are very mechanical. Just adding an `R` type param to extractors
that don't care about the body and `Once` if they do.

So reviewing should be focused on `axum-core/src/extract/mod.rs` and
specifically which methods `RequestParts<Mut, _>` and `RequestParts<Once, _>`
gives you access to. All other changes fall from that.

## Why is the diff so big?!

The diff is so large because we couldn't use a declarative macro to make all the
`Handler` impls, since we have to generate trait bounds in the macro. So I wrote
a little internal script that generates the code with `quote` and writes it to a
file.

## `#[derive(FromRequest)]`

I removed `#[derive(FromRequest)]` for now since making it work with the new
setup is a lot of work.

For example how should this

```rust
#[derive(FromRequest)]
struct Foo {
    one: HeaderMap,
    two: String,
}
```

implement `FromRequest`? To work it must implement `FromRequest<Once, _>`
because otherwise it cannot extract `String`. But how can the macro know that?

I suppose we could do

```rust
#[derive(FromRequest)]
#[from_request(mut/once)]
struct Foo {
    one: HeaderMap,
    two: String,
}
```

but we'd have to be careful not to allow

```rust
#[derive(FromRequest)]
#[from_request(once)]
struct Foo {
    one: Bytes,
    two: String,
}
```

We could do that in a similar way by requiring only the last field to implement
`FromRequest<Once, _>`.

All this is a lot of work that I didn't wanna do in this PR. I might work on it
after this is merged but I don't think it should block the release.

## TODO

- [ ] Update the changelog. It should describe how to update from the old setup
  to the new.


axum-core/src/extract/mod.rs
```diff
@@ -211,27 +206,118 @@ impl<B> RequestParts<B> {
...
-    pub fn extensions_mut(&mut self) -> &mut Extensions {
-        &mut self.extensions
+    pub fn insert_extension<T>(&mut self, extension: T)
+    where
+        T: Send + Sync + 'static,
+    {
+        self.extensions.insert(extension);
     }
 
+    // TODO(david): probably want insert_header and append_header
```
davidpdrsn, line 221: I actually think having a `remove_extension` method is also fine, since you cannot remove all extensions with such a method. Then we could make `WebSocketUpgade` implement `FromRequest<Ref, _>`.

We probably don't need `remove_header` since they you'll be able to remove all the headers, and cloning headers is always possible and cheap. You can't always clone extensions.

axum-core/src/extract/mod.rs
```diff
@@ -7,18 +7,90 @@
...
 
 mod request_parts;
 mod tuple;
 
+/// Maker type used to signify that an extractor can be run multiple times.
+///
+/// See [`FromRequest`] for more details.
+// TODO(david): naming
+#[derive(Debug, Clone, Copy)]
+pub struct Mut(Infallible);
```
davidpdrsn, line 26: Not a fan of this name. Should come up with a better one.

davidpdrsn:
Gonna mark this as a draft until https://github.com/tokio-rs/axum/pull/1086 is merged.

davidpdrsn:
I think given how hard it was to review https://github.com/tokio-rs/axum/pull/1155 I'll close this and break it up into smaller PRs. Those individual PRs wont compile but I think it'll be easier to review and discuss if the changes are smaller. Then once we're happy with the lower level changes we can fix all the downstream errors.

PR #1123, opened 2022-06-26
Make `Router` faster to clone

davidpdrsn:
## Motivation

`Router` is cloned once for every request in `IntoMakeService::call` so ideally it should be cheap to do so.

## Solution

Wrap the `Node` in a `Arc` and rebuild it when calling `Router::router`. This is fine since it only happens during startup.

I did a simple benchmark where I cloned a `Router` with 1000 routes and this brought the time down from 176.916µs to 92.698µs. So almost 50% 📈

I also attempted to `Arc` the other fields but since they're `!Sync` the `Arc` becomes `!Send` which is a no-go. Using `SyncWrapper` doesn't help as that requires exclusive access which we cannot get through an `Arc`.

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -64,7 +64,7 @@ impl RouteId {
 /// The router type for composing handlers and services.
 pub struct Router<B = Body> {
     routes: HashMap<RouteId, Endpoint<B>>,
-    node: Node,
+    node: Arc<Node>,
     fallback: Fallback<B>,
     nested_at_root: bool,
 }
@@ -73,7 +73,7 @@ impl<B> Clone for Router<B> {
     fn clone(&self) -> Self {
         Self {
             routes: self.routes.clone(),
-            node: self.node.clone(),
+            node: Arc::clone(&self.node),
             fallback: self.fallback.clone(),
             nested_at_root: self.nested_at_root,
         }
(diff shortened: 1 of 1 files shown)


axum/src/routing/mod.rs
```diff
@@ -162,9 +162,11 @@ where
             Err(service) => Endpoint::Route(Route::new(service)),
         };
 
-        if let Err(err) = self.node.insert(path, id) {
+        let mut node = (&*self.node).clone();
```
jplatte, line 165:
```suggestion
        let mut node = `Arc::try_unwrap(self.node).unwrap_or_else(|node| (*node).clone());
```

axum/src/routing/mod.rs
```diff
@@ -211,12 +214,12 @@ where
                     panic!("Cannot nest `Router`s that has a fallback");
                 }
 
-                for (id, nested_path) in node.route_id_to_path {
-                    let route = routes.remove(&id).unwrap();
-                    let full_path: Cow<str> = if &*nested_path == "/" {
+                for (id, nested_path) in node.route_id_to_path.iter() {
```
jplatte, line 217:
```suggestion
                for (id, nested_path) in &node.route_id_to_path {
```

jplatte (approved):

Issue #1122, opened 2022-06-26
Panicking inside "replace_trailing_slash" function


shamblesides:
### Version
0.5.7

### Platform
Linux prod 5.10.0-9-amd64 #1 SMP Debian 5.10.70-1 (2021-09-30) x86_64 GNU/Linux

### Description
Axum is panicking here: https://github.com/tokio-rs/axum/blob/main/axum/src/routing/mod.rs#L532

```
panicked at 'called `Result::unwrap()` on an `Err` value: InvalidUriParts(InvalidUri(SchemeMissing))
```

davidpdrsn:
I'm curious whether this is a bug in the `http` crate. I've asked in Discord.

niklasf:
Was this bringing down the server, or was unwinding caught?

davidpdrsn:
It wouldn't bring down the server. Just close the connection without sending a response.

The fix has been released in 0.5.10 🎉

shamblesides:
FYI this was indeed bringing down my server, but I guess this is probably because I was aborting upon panicking

PR #1124, opened 2022-06-27
Fix possible panic when doing trailing slash redirect

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1122

It panicked because we hit this https://github.com/hyperium/http/blob/master/src/uri/mod.rs#L248-L250 and did an `unwrap`. So rather than panicking we send the request to the fallback. We cannot issue a redirect since we couldn't add the path so guess there isn't much else we can do.

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -500,29 +500,35 @@ where
 
         match self.node.at(&path) {
             Ok(match_) => self.call_route(match_, req),
-            Err(MatchError::MissingTrailingSlash) => {
-                let new_uri = replace_trailing_slash(req.uri(), &format!("{}/", &path));
+            Err(err) => {
+                let mut fallback = match &self.fallback {
+                    Fallback::Default(inner) => inner.clone(),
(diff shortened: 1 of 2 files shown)


axum/src/routing/mod.rs
```diff
@@ -500,29 +500,35 @@ where
...
-            Err(MatchError::ExtraTrailingSlash) => {
-                let new_uri = replace_trailing_slash(req.uri(), path.strip_suffix('/').unwrap());
+                let new_uri = match err {
+                    MatchError::MissingTrailingSlash => {
+                        replace_path(req.uri(), &format!("{}/", &path))
+                    }
+                    MatchError::ExtraTrailingSlash => {
+                        replace_path(req.uri(), path.strip_suffix('/').unwrap())
+                    }
+                    MatchError::NotFound => return fallback.call(req),
```
jplatte, line 516:
```suggestion
                    MatchError::NotFound => None,
```

davidpdrsn: Oh yeah of course!

jplatte (approved):

PR #1130, opened 2022-06-30
TryFrom<http::Method> for MethodFilter

NicolaLS:
## Motivation
I am currently switching from tide to axum in a project which also uses http-types to dynamicaly set methods. When I wanted to change that to `axum::routing::MethodFilter`, I thought that it would be nice to just have a `TryFrom<http::Method> `for the MethodFilter, since axum already uses http and there is no unclear semantics going from e.g. `http::Method::GET` to `MethodFilter::GET`

## Solution
I just added a very simple `TryFrom<Method> for MethodFilter` implementation that maps the intersecting methods from both `Method` and `MethodFillter` appropriately and just returns a result if there is no method in `MethodFilter` for the given `http::Method `

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -16,7 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `/foo/`). That is no longer supported and such requests will now be sent to
   the fallback. Consider using `axum_extra::routing::RouterExt::route_with_tsr`
   if you want the old behavior ([#1119])
+- **added** Implement `TryFrom<http:: Method>` for `MethodFilter` and use new `NoMatchingMethodFilter` error in case of failure ([#1130])
 
+[#1130]: https://github.com/tokio-rs/axum/pull/1130
 [#1077]: https://github.com/tokio-rs/axum/pull/1077
 [#1102]: https://github.com/tokio-rs/axum/pull/1102
 [#1119]: https://github.com/tokio-rs/axum/pull/1119
--- a/axum/src/routing/method_filter.rs
+++ b/axum/src/routing/method_filter.rs
@@ -61,7 +61,60 @@ impl TryFrom<Method> for MethodFilter {
             Method::POST => Ok(MethodFilter::POST),
             Method::PUT => Ok(MethodFilter::PUT),
             Method::TRACE => Ok(MethodFilter::TRACE),
-            other => Err(NoMatchingMethodFilter { method: other}),
+            other => Err(NoMatchingMethodFilter { method: other }),
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn from_http_method() {
+        assert_eq!(
+            MethodFilter::try_from(Method::DELETE).unwrap(),
+            MethodFilter::DELETE
+        );
+
+        assert_eq!(
+            MethodFilter::try_from(Method::GET).unwrap(),
+            MethodFilter::GET
+        );
+
+        assert_eq!(
+            MethodFilter::try_from(Method::HEAD).unwrap(),
+            MethodFilter::HEAD
+        );
+
+        assert_eq!(
+            MethodFilter::try_from(Method::OPTIONS).unwrap(),
+            MethodFilter::OPTIONS
+        );
+
+        assert_eq!(
+            MethodFilter::try_from(Method::PATCH).unwrap(),
+            MethodFilter::PATCH
+        );
+
+        assert_eq!(
+            MethodFilter::try_from(Method::POST).unwrap(),
+            MethodFilter::POST
+        );
+
+        assert_eq!(
+            MethodFilter::try_from(Method::PUT).unwrap(),
+            MethodFilter::PUT
+        );
+
+        assert_eq!(
+            MethodFilter::try_from(Method::TRACE).unwrap(),
+            MethodFilter::TRACE
+        );
+
+        assert!(MethodFilter::try_from(http::Method::CONNECT)
+            .unwrap_err()
+            .to_string()
+            .contains("CONNECT"));
     }
 }
--- a/axum/src/routing/method_filter.rs
+++ b/axum/src/routing/method_filter.rs
@@ -1,6 +1,6 @@
+use std::{fmt, fmt::{Debug, Formatter}};
 use bitflags::bitflags;
 use http::Method;
-use axum_core::Error;
 
 bitflags! {
     /// A filter that matches one or more HTTP methods.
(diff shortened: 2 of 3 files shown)


axum/src/routing/tests/mod.rs
```diff
@@ -720,3 +720,15 @@ async fn limited_body_with_streaming_body() {
...
+async fn from_http_method() {
+    assert_eq!(MethodFilter::try_from(Method::DELETE).unwrap(), MethodFilter::DELETE);
+    assert_eq!(MethodFilter::try_from(Method::GET).unwrap(), MethodFilter::GET);
+    assert_eq!(MethodFilter::try_from(Method::HEAD).unwrap(), MethodFilter::HEAD);
+    assert_eq!(MethodFilter::try_from(Method::OPTIONS).unwrap(), MethodFilter::OPTIONS);
+    assert_eq!(MethodFilter::try_from(Method::PATCH).unwrap(), MethodFilter::PATCH);
+    assert_eq!(MethodFilter::try_from(Method::POST).unwrap(), MethodFilter::POST);
+    assert_eq!(MethodFilter::try_from(Method::PUT).unwrap(), MethodFilter::PUT);
+    assert_eq!(MethodFilter::try_from(Method::TRACE).unwrap(), MethodFilter::TRACE);
+}
```
davidpdrsn, line 734: This doesn't need to be async.

```suggestion
#[test]
fn from_http_method() {
    assert_eq!(MethodFilter::try_from(Method::DELETE).unwrap(), MethodFilter::DELETE);
    assert_eq!(MethodFilter::try_from(Method::GET).unwrap(), MethodFilter::GET);
    assert_eq!(MethodFilter::try_from(Method::HEAD).unwrap(), MethodFilter::HEAD);
    assert_eq!(MethodFilter::try_from(Method::OPTIONS).unwrap(), MethodFilter::OPTIONS);
    assert_eq!(MethodFilter::try_from(Method::PATCH).unwrap(), MethodFilter::PATCH);
    assert_eq!(MethodFilter::try_from(Method::POST).unwrap(), MethodFilter::POST);
    assert_eq!(MethodFilter::try_from(Method::PUT).unwrap(), MethodFilter::PUT);
    assert_eq!(MethodFilter::try_from(Method::TRACE).unwrap(), MethodFilter::TRACE);
}
```

axum/src/routing/method_filter.rs
```diff
@@ -21,3 +23,21 @@ bitflags! {
         const TRACE =   0b100000000;
     }
 }
+
+impl TryFrom<Method> for MethodFilter {
+    type Error = Error;
```
davidpdrsn, line 28: I don't think we should use `axum::Error` here. Its meant as an opaque error that we use wrap 3rd party errors in. I think we should make a new error type instead. Maybe `UnsupportedMethod` but that sounds quite broad 🤔 wdyt?

davidpdrsn (changes requested):
Thanks!

axum/src/routing/method_filter.rs
davidpdrsn: > Where do I define it ? right there where the macro & trait impl is ?

Yeah I think thats fine. Keeping the code close together makes sense here.

> Can I just do a pub struct `UnsupportedMethod(String)` with a new() constructor ?

I'd do

```rust
#[derive(Debug)]
struct UnsupportedMethod {
    // this is private since users shouldn't be allowed to construct this type.
    // only axum should be able to do that
    // we also don't want a constructor for the same reason
    method: http::Method,
}

impl UnsupportedMethod {
    // should probably provide an accessor though
    fn method(&self) -> &http::Method { ... }
}

impl fmt::Display for UnsupportedMethod {
    // return some readable error in `fmt`
}

// don't need to override source because this isn't wrapper other errors
impl std::error::Error for UnsupportedMethod {}
```

axum/src/routing/tests/mod.rs
davidpdrsn: You can also just click "Commit suggestion" to accept it, rather than typing it yourself.

axum/src/routing/method_filter.rs
```diff
@@ -21,3 +23,40 @@ bitflags! {
         const TRACE =   0b100000000;
     }
 }
+
+#[derive(Debug)]
+pub struct UnsupportedMethod {
+    method: http::Method,
+}
```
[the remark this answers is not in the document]
NicolaLS: we'll only give out a reference anyway so maybe a `&'a str` and just do `other => Err(UnsMet { method: other.as_str()})` is sufficient ?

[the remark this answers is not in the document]
davidpdrsn: I don't mean the data inside the struct, I specifically mean the name `UnsupportedMethod`. That name doesn't hint at `MethodFilter` which I think it should. "Unsupported method" sounds like you called a route with an unsupported method, but thats not what this error is for.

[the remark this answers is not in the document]
NicolaLS: oh the name, I thought you meant the type.. maybe _**rejected**_,  _**unsupported**_,  _**cause**_, _**failed_with**_ ?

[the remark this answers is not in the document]
davidpdrsn: How about `NoMatchingMethodFilter`? Its a mouthful but pretty precise.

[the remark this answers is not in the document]
NicolaLS: Yes sounds good, I was also thinking about something like `NotInDomain` `DomainMiss` but yours is definitely better ^^

davidpdrsn (changes requested):
Almost there! Wanna add a line to `axum/CHANGELOG.md` as well?

axum/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `/foo/`). That is no longer supported and such requests will now be sent to
   the fallback. Consider using `axum_extra::routing::RouterExt::route_with_tsr`
   if you want the old behavior ([#1119])
+- **added** Implement `TryFrom<http:: Method>` for `MethodFilter` and use new `NoMatchingMethodFilter` error in case of failure ([#1130])
```
davidpdrsn, line 19: You need to add a link reference thing for `([#1130])` as well.

davidpdrsn: I wish

axum/src/routing/method_filter.rs
```diff
@@ -21,3 +23,40 @@ bitflags! {
...
     }
 }
+
+#[derive(Debug)]
+pub struct NoMatchingMethodFilter {
+    method: http::Method,
+}
+
+impl NoMatchingMethodFilter {
+    pub fn method(&self) -> &http::Method {
```
davidpdrsn, line 33: Also need docs here.

davidpdrsn (changes requested):
One last thing :)

axum/src/routing/method_filter.rs
NicolaLS: thank you for all the help ! Just pushed but struggled hard to get a decent doc so I didn't do _Examples_ and I'm not sure if I did something wrong on the intra-doc links or its just my IDE

Issue #1132, opened 2022-06-30
Reverse proxied web socket extraction can panic


dwalker109:
### Version

axum v0.5.9
axum-core v0.2.6

### Platform

x86_64 GNU/Linux

### Crates

axum

### Description

With a misconfigured NGINX reverse proxy, extracting the websocket causes a panic.

I used a simple websocket route handler which uses the `WebSocketUpgrade` extractor. This works normally when accessing the Axum server directly. 

Also, when accessed via an NGINX proxy_pass with correct config like this it it works fine:

```
    location /ws/ {
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        proxy_pass_request_headers on;
        proxy_http_version 1.1;
        proxy_pass  http://127.0.0.1:8081/ws/;
    }
```

However, if the `proxy_http_version 1.1;` line is omitted trying to establish the connection panics:

```
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', /Users/me/.cargo/registry/src/github.com-1ecc6299db9ec823/axum-0.5.9/src/extract/ws.rs:277:69
```

I don't know enough about websockets to understand why this happened or if it is just one of those things, but perhaps the messaging could be clearer.

I'd be happy to tackle the bug if it was decided this was something to be fixed.

davidpdrsn:
Thanks for noticing! The panic will be fixed in https://github.com/tokio-rs/axum/pull/1135 so it instead returns a rejection. I'm not exactly sure why hyper couldn't upgrade the request. Try opening an issue with [hyper](https://github.com/hyperium/hyper).

dwalker109:
Will do - and thanks for jumping on this so quickly, incredible stuff 💯.

PR #1135, opened 2022-07-01
Return rejection if `WebSocket` request couldn't be upgraded

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1132

Rather than panicking we'll return a rejection. Unfortunately hyper doesn't give any details about why the connection couldn't be upgraded.

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -275,7 +275,7 @@ where
         let on_upgrade = req
             .extensions_mut()
             .remove::<OnUpgrade>()
-            .ok_or(MissingUpgradeState)?;
+            .ok_or(ConnectionNotUpgradable)?;
 
         let sec_websocket_protocol = req.headers().get(header::SEC_WEBSOCKET_PROTOCOL).cloned();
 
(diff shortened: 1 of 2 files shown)


jplatte (commented):
I suggest somebody who actually knows stuff about WebSockets should review this.

axum/src/extract/ws.rs
```diff
@@ -267,14 +267,15 @@ where
...
+        let sec_websocket_key = req
+            .headers_mut()
+            .remove(header::SEC_WEBSOCKET_KEY)
+            .ok_or(WebSocketKeyHeaderMissing)?;
 
-        let on_upgrade = req.extensions_mut().remove::<OnUpgrade>().unwrap();
+        let on_upgrade = req
+            .extensions_mut()
+            .remove::<OnUpgrade>()
+            .ok_or(MissingUpgradeState)?;
```
davidpdrsn, line 278: @seanmonstar How do you recommend axum deals with `OnUpgrade` missing? We got this issue https://github.com/tokio-rs/axum/issues/1132 recently from a user who ran into this.

This code is inspired by warp which just logs if the extension is missing (https://github.com/seanmonstar/warp/blob/master/src/filters/ws.rs#L154).

Is is possible to get more details from hyper about why the upgrade failed?

davidpdrsn:
@jplatte gentle ping (:

jplatte:
I still don't think I'm the right person to review this though 😅

davidpdrsn:
In that case I think I'll just yolo it. Its a pretty safe change and given what sean said it should be correct.

PR #1144, opened 2022-07-04
Re-export `debug_handler` from axum

### LINKED ISSUE #1143: Should we re-export `debug_handler` from axum itself?
I've been thinking that we should consider re-exporting `debug_handler` from axum itself. So its more directly accessible and highlighted more clearly in the docs. Should probably be behind an on-by-default feature so you can avoid the dependency on `syn`.

davidpdrsn:
Fixes #1143

### DIFF
--- a/axum-macros/Cargo.toml
+++ b/axum-macros/Cargo.toml
@@ -8,7 +8,7 @@ license = "MIT"
 name = "axum-macros"
 readme = "README.md"
 repository = "https://github.com/tokio-rs/axum"
-version = "0.2.3" # remember to also bump the version that axum depends on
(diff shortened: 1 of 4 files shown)


axum-macros/Cargo.toml
```diff
@@ -8,7 +8,7 @@ license = "MIT"
 name = "axum-macros"
 readme = "README.md"
 repository = "https://github.com/tokio-rs/axum"
-version = "0.2.3"
+version = "0.2.3" # remember to also bump the version that axum depends on
```
jplatte, line 11: And axum-extra?

jplatte (approved):

PR #1147, opened 2022-07-08
Fix typos in RequestParts docstrings

dspicher:
## Motivation

Two docstrings are missing a "to".

## Solution

Add them.


davidpdrsn (approved):
Thanks!

PR #1152, opened 2022-07-10
Support any middleware response that implements `IntoResponse`

davidpdrsn:
Rather than requiring middleware to return `Response<_>` I think we should allow any `IntoResponse`. Makes things a little more flexible and consistent with handler functions.

### DIFF
--- a/axum/src/handler/mod.rs
+++ b/axum/src/handler/mod.rs
@@ -42,7 +42,7 @@ use crate::{
     routing::IntoMakeService,
 };
 use http::Request;
-use std::{fmt, future::Future, marker::PhantomData, pin::Pin};
+use std::{convert::Infallible, fmt, future::Future, marker::PhantomData, pin::Pin};
 use tower::ServiceExt;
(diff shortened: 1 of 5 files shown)


axum/src/handler/mod.rs
```diff
@@ -284,15 +283,13 @@ where
     }
 }
 
-impl<S, T, ReqBody, ResBody> Handler<T, ReqBody> for Layered<S, T>
+impl<S, T, ReqBody> Handler<T, ReqBody> for Layered<S, T>
 where
-    S: Service<Request<ReqBody>, Response = Response<ResBody>> + Clone + Send + 'static,
-    S::Error: IntoResponse,
+    S: Service<Request<ReqBody>, Error = Infallible> + Clone + Send + 'static,
+    S::Response: IntoResponse,
```
davidpdrsn, line 289: Think I'll just change this here as well since its touching the same code. Handler's wrapped in middleware should be infallible just like everything else.

Issue #1110, opened 2022-06-20
Middleware that change the request body have poor usability


davidpdrsn:
We've heard from several people who try to use `tower_http::limit::RequestBodyLimitLayer` that the implications are hard to understand.

For example breaking your `Router` into functions is quite common. Something like:

```rust
#[tokio::main]
async fn main() {
    let app = my_routes();

    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
    axum::Server::bind(&addr)
        .serve(app.into_make_service())
        .await
        .unwrap();
}

fn my_routes() -> Router {
    Router::new()
}
```

Adding `RequestBodyLimitLayer` like so

```rust
let app = my_routes().layer(RequestBodyLimitLayer::new(1024));
```

should just work but produces a rather confusing error:

```
    Checking example-hello-world v0.1.0 (/Users/davidpdrsn/dev/major/axum/examples/hello-world)
error[E0277]: the trait bound `Route: tower_service::Service<Request<http_body::limited::Limited<_>>>` is not satisfied
  --> hello-world/src/main.rs:13:27
   |
13 |     let app = my_routes().layer(RequestBodyLimitLayer::new(1024));
   |                           ^^^^^ the trait `tower_service::Service<Request<http_body::limited::Limited<_>>>` is not implemented for `Route`
   |
   = help: the following implementations were found:
             <Route<B, E> as tower_service::Service<Request<B>>>
   = note: required because of the requirements on the impl of `tower_service::Service<Request<_>>` for `RequestBodyLimit<Route>`
```

The solution is instead to write

```rust
fn my_routes() -> Router<http_body::Limited<Body>> {
    Router::new()
}
```

but that isn't obvious and ideally shouldn't be necessary.

We should think about ways to improve that experience.

davidpdrsn:
I think this is closely related to https://github.com/tokio-rs/axum/pull/1154. If axum had its own body type, and a way to construct that from any `B: http_body::Body` then we could use that everywhere, even if you applied middleware.

That is something we'll have to explore once the development of hyper 1.0 (and hyper-utils) is further along. Hyper 1.0 will require a breaking release of axum anyway, so I think we should wait with this until we're closer to that happening.

I'm gonna remove this from the 0.6 milestone.

avdb13:
why does `ContentLengthLimit` exist in `Axum` when we have `tower_http::limit::RequestBodyLimitLayer` in `Tower`, on which `Axum` relies anyway?
[...]

Issue #1142, opened 2022-07-02
Missing extensions are runtime errors rather than compile errors


davidpdrsn:
Attempting to extract `Extension<Foo>` without having `.layer(Extension(foo))` will result in a runtime error, rather than a compile error. That has been a long standing problem since 0.1. We don't have a good solution and ergonomic solution to it yet but I think its something we should experiment with for 0.6.

davidpdrsn:
I have been thinking about this and have a proposal now.

The reason its hard in the first place is that `Router::route` takes any `T: Service`. So even if we added some state generic to `Router` it wouldn't be able to verify that the routes uses that same state type, as it just sees a generic `T: Service`.

My proposal is to change `Router` so instead its API looks like this

```rust
#[tokio::main]
async fn main() {
    let mut app = Router::new(AppState {});

    // can also consider having `Router::with_state(...)` and then `Router::new()` just being `with_state(())`

    // add a route for `GET /` that goes to `handler`
    app.get("/", handler);

    // add routes for both `GET /foo` and `POST /foo`
    app.get("/foo", handler).post(handler);

    // this wouldn't work
    // the thing returned by `get` doesn't allow adding routes with new paths
    // just new methods
    // this we can still debate but its not strictly relevant to this issue
    // app.get("/something", handler).post("/something-else", bar);

    axum::Server::bind(...)
        .serve(app.into_make_service())
        .await
        .unwrap();
}

// the specific state our app needs
#[derive(Clone)]
struct AppState {}

async fn handler() {}
```

Now `Router::{get, post, etc}` can take the handler itself as the argument without any indirection.

Next add a new `S` (for state) type param to handler and accept the state in `call`:

```rust
pub trait Handler<S, T, B = Body>: Clone + Send + Sized + 'static {
    type Future: Future<Output = Response> + Send + 'static;

    fn call(self, state: S, req: Request<B>) -> Self::Future;
}
```

The signature of `Router::get` would be

```rust
impl<B> Router<B> {
    pub fn get<H, S, T>(mut self, path: &str, handler: T) -> Self
    where
        H: Handler<S, T, B>
    {
        // ...
    }
}
```

Next the same state is present in `RequestParts`

```rust
impl<S, B> RequestParts<S, B> {
    // only immutable access to the state
    pub fn state(&self) -> &S {
        &self.state
    }
}
```

This changes `FromRequest` to

```rust
#[async_trait]
pub trait FromRequest<S, B>: Sized {
[...]

PR #1154, opened 2022-07-11
Remove request body type param

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1110

This removes the request body type param from everything (`Router`, `FromRequest`, `RequestParts`, etc...). The goal being to improve usability and solve issues like https://github.com/tokio-rs/axum/issues/1110.

The general approach is to implement `Service` like this:

```rust
impl<B> Service<Request<B>> for Router
where
    B: http_body::Body,
{
    fn call(&mut self, req: Request<B>) -> Self::Future {
        let req = req.map(|body| hyper::Body::wrap_body(body));
        // ...
    }
}
```

This allows us to accept any `impl http_body::Body` on the outside but convert that to a `hyper::Body` on the inside.

We need to accept any `impl http_body::Body` instead of only `hyper::Body` to support middleware that change the request body (such as [`RequestBodyLimit`](https://docs.rs/tower-http/latest/tower_http/limit/struct.RequestBodyLimit.html)).

I also considered using `BoxBody` rather than `hyper::Body` but that means users cannot extract `Request<hyper::Body>` which I think would confuse many users.

Note that `hyper::Body:wrap_body` isn't shipped yet. I have made [a PR to hyper](https://github.com/hyperium/hyper/pull/2913) that adds that. If hyper decides not to merge that then our only option is to use `BoxBody`. We need that to construct a `hyper::Body` from a generic `impl http_body::Body`.

This is also a win in another elsewhere because it removes a type param from `FromRequest` and `RequestParts`. In https://github.com/tokio-rs/axum/pull/1121 and a WIP fix for https://github.com/tokio-rs/axum/issues/1142 I'm adding two more type params to those types. So by removing the body we'll only have two type params instead of three.

### TODO

- [ ] Update docs
- [ ] Update changelog, including guide on how to upgrade


axum-core/src/body.rs
```diff
@@ -3,7 +3,10 @@
 use crate::{BoxError, Error};
 use bytes::Bytes;
 use bytes::{Buf, BufMut};
-use http_body::Body;
+use http_body::Body as _;
+
+#[doc(no_inline)]
+pub use hyper::Body;
```
davidpdrsn, line 9: Note this makes hyper a public dependency of axum-core. I think thats a little unfortunate but its required to make `impl FromRequest for Request<hyper::Body>` work. I think thats an okay trade-off.

davidpdrsn: It also does mean that any crate that depends on axum-core now also depends on hyper. Thats pretty unfortunate 🤔 But yeah not sure there is a way around it.

axum-core/src/body.rs
```diff
@@ -55,7 +58,7 @@ where
 // THE SOFTWARE.
 pub(crate) async fn to_bytes<T>(body: T) -> Result<Bytes, T::Error>
 where
-    T: Body,
+    T: http_body::Body,
```
davidpdrsn, line 61: Note to self: Wo can remove this method and just use the one in hyper.

axum-core/src/body.rs
davidpdrsn: One solution would be to keep the type param on `FromRequest` and `RequestParts` but that would mean they get three type params in 0.6 😬

davidpdrsn: Perhaps its also worth considering adding our own body type to `axum-core`, to avoid the dependency on hyper. I believe hyper's `Body` will be moved into hyper-utils for 1.0 and we don't wanna have a public dependency on hyper-utils. So we'll probably need our own body type eventually.

axum-core/src/extract/mod.rs
```diff
@@ -60,37 +61,41 @@ mod tuple;
...
-    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection>;
+    async fn from_request(req: &mut RequestParts) -> Result<Self, Self::Rejection>;
 }
 
 /// The type used with [`FromRequest`] to extract data from requests.
 ///
 /// Has several convenience methods for getting owned parts of the request.
 #[derive(Debug)]
-pub struct RequestParts<B> {
+pub struct RequestParts {
```
jplatte, line 77: Err, why is there a `request_parts` module, but `RequestParts` is defined elsewhere?

jplatte (commented):
I never used a non-default body type so I don't really have a good overview of what kinds of things this makes harder / pessimizes (if any). If this doesn't have any serious downsides I'd be in favor of course :)

davidpdrsn:
The main downside is that it locks us into one body type for extractors. I think a lot depends on the hyper/hyper-util split for their 1.0 and where stuff ends up. We cannot publicly depend on the body from hyper-util and hyper probably won't have a single body type.

davidpdrsn:
I'm gonna close this. All this is gonna have to change anyway when we get closer to hyper 1.0 so I'd rather deal with it then. See https://github.com/tokio-rs/axum/issues/1110#issuecomment-1222265042.

This PR is also too large to review so if/when we do this we have to do it in smaller chunks.

PR #1155, opened 2022-07-11
Add type safe state extractor

### LINKED ISSUE #1142: Missing extensions are runtime errors rather than compile errors

davidpdrsn:
This fixes #1142 by introducing a type safe `State` extractor. Attempting to
extract states that haven't seen set results in compile errors, rather than
runtime errors.

Example usage:

```rust
#[derive(Clone)]
struct AppState {}

async fn handler(
    State(app_state): State<AppState>,
) {
    ...
}

let app = Router::with_state(AppState { ... }).route("/", get(handler));
```

So the state is set with `Router::with_state` and extracted with `State`.

# How it works

This works by changing `Router::route` to accept a `MethodRouter`, rather than
any `Service`. `MethodRouter`, `FromRequest`, `RequestParts`, and `Handler` then
all have a new type parameter for the state extracted with `State`.

All extractors just requires some generic state `S` but `State` fixes that to be
some specific type.

Most of the changes in this PR comes from having to add the new parameter
everywhere. I'm dreading all the merge conflicts but oh well...

# Extracting substates

Substates can be extracted by implementing `From`:

```rust
#[derive(Clone)]
struct AppState {
    sub: Substate,
}

#[derive(Clone)]
struct Substate {}

impl From<AppState> for Substate { ... }

async fn handler(
    // this works even through we give an `AppState` to `Router::with_state`
    State(substate): State<Substate>,
) {
    ...
}

let app = Router::with_state(AppState { ... }).route("/", get(handler));
```

I think this gives most of the benefits of `Extension` while being type safe.

## State defaults to `()`

By default the state type param is `()`. So unless you're explicitly using state
you shouldn't have to worry about it:

```rust
// state is `()` and request body is `Body`
fn routes() -> Router {
    Router::new()
}

// same for `Handler` and `MethodRouter`
```

Note that `S` does not default to `()` on `FromRequest` and `RequestParts`. I'd be worried that library authors would accidentally bind their extractors to `()` rather than a generic `S`. So therefore these types requires specifying the state type.

# `MethodRouter` and `Handler` as `Service`

Only `Router` accepts the state in its constructor so only it can implement
`Service` directly.

However `MethodRouter` and `Handler` needs the state provided to implement
`Service`. Providing that is done like so:

```rust
let method_router_service = get(|_: State<String>| async {})
    // we have to provide the state with `.with_state(...)`
    .with_state(String::new());

// this also implements `Service` since the state is `()`
let method_router_service = get(|| async {});

// same for handlers
let handler_service = (|_: State<String>| async {})
    // provide the state and get something that implements `Service`
    .with_state(String::new());

// there is also a method to get a `MakeService`
let handler_make_service = (|_: State<String>| async {})
    .with_state(String::new())
    .into_make_service();

let handler_service = (|| async {})
    // this method exists on an extension trait thats only implemented
    // for Handlers where the state is `()`
    .into_service();
```

# Known limitations

No good way to provide the state automatically to nested `Router`s. You have to
something like this:

```rust
Router::with_state(state.clone()).nest("/api", Router::with_state(state));
```

Given all the discussions [here](https://github.com/tokio-rs/axum/pull/1086) I
think this is the best we can do.

Middleware cannot extract states since `Router::layer` just receives some `L:
Layer`. It doesn't know anything about the state. We cannot pull the same trick
I did with `Router::route` since there are many ways to create `Layer`s.

This setup might not be a great fit for optional state that might only exist in
some environments. I think those use cases are rare so IMO we can live with
this. `Extension` still works well for such use cases.

Same goes for routing to generic `Service`s, there is no good way to access the
state. I think thats fine because. In my experience you're either routing to
handlers (which know about state) or `Service`s like `ServeDir` which don't need
the state.

# Follow-ups

These are follow-up improvements I intend to make once this is merged:

- Support telling `#[debug_handler]` which state you use. Currently it has to
  assume you're using `()`.
- Same goes for `#[derive(FromRequest)]`.
- Change cookie extractors to use get the signing key via a substate rather than
  extensions, since its required.
- Write a new `#[derive(State)]` macro which adds the `From` impls for all
  fields in a struct and adds `FromRequest` impls that delegate to `State`. This
  should take care of all the boilerplate if you have a lot of state.
- Think about allowing middlware to extract state via `Extension`.


jplatte:
I'm not a fan of this change. It increases complexity of both the API surface and the implementation while only serving some use cases. Notably, the key for `SignedCookieJar` still uses `Extension` and I don't think there's an easy way to migrate it to this new extractor.

davidpdrsn:
I think sharing state across all/most handlers is a very common use case. It's not specific to cookies and haven't thought about migrating that. But I still think it makes sense.

Have a look at all the examples. Those got way nicer and more type safe.

I do agree re adding more complexity but if/when we remove the request body type param I think that balances things out.

jplatte:
> It's not specific to cookies and haven't thought about migrating that.

I didn't want to say cookies are a special case. In general, I feel like extractors that expect some sort of "configuration" in an extension currently are not going to be easy to integrate into this new pattern. The same applies to routes not tied to a particular service (e.g. a metrics route implementation in its own library), I think.

> Have a look at all the examples. Those got way nicer and more type safe.

I had a look. They certainly did get more type safe, but how are they way nicer?

Thinking more about this, here is another pattern I've used before that wouldn't integrate well with this: I used to work on a webservice that integrated with a bunch of external services (i.e. stuff like S3). For some routes, having these configured was necessary for them to do their work. I set up the server such that setting specific env vars would add specific extensions to the app-level `ServiceBuilder`. This allowed me to do simple dev builds without spinning up loads of docker containers if I only wanted to test new functionality that didn't depend on external services; calling one of the routes that needed it would result in an internal server error but everything else worked fine.

davidpdrsn:
I actually think requiring the state used with private and signed cookies to implement `From <AppState> for cookie::Key {}` is kinda nice. That also avoids runtime errors because you cannot forget the key.

> I had a look. They certainly did get more type safe, but how are they way nicer?

I meant the type safety made them way nicer 😛 

> Thinking more about this, here is another pattern I've used before that wouldn't integrate well with this

So you just wouldn't call those routes during development because the extensions would be missing?

Hm yeah not sure how to port that to this pattern 🤔 Maybe your state could have options for those configs that are `Some` if the env vars exist. And then an extractor for the configs that fails if they're `None`. 

But extensions aren't going away either so you could also just keep doing what you did. Any solutions I've heard for checking that extensions exist quickly dissolve in hlists which I really wanna avoid.

Maybe if `Router` and `FromRequest` used `()` by default for the state? Then I don't think the changes here would even be visible outside of some new methods.

jplatte:
> Maybe if `Router` and `FromRequest` used `()` by default for the state? Then I don't think the changes here would even be visible outside of some new methods.

Yeah, I guess this could be a good compromise. Though it does break my proposed `FromRequestOnce` design that doesn't need the weird dummy `Mut` / `Once` type param, which would be possible if we were to merge #1154 but not this. I admit this is a pretty weak motivation.

jplatte:
Nested routers can't inherit the state somehow, can they? That was actually something I considered a bit ugly when seeing it – `fn admin_routes() -> Router` becoming `fn admin_routes(state: SharedState) -> Router<SharedState>` in the `key-value-store` example.

davidpdrsn:
Yep that is annoying for sure. I did actually have a previous version where the state would be lazily applied so you didn't have to pass it around. But it was pretty complicated so I abandoned it.

davidpdrsn:
Alright this is basically done but I'll leave it as a draft until https://github.com/tokio-rs/axum/pull/1086 and https://github.com/tokio-rs/axum/pull/1121 has been merged. I would like to only fix the merge conflicts once.

davidpdrsn:
Alright this is ready for review now. I apologies for the large diff. This required touching pretty much all the files. The good news is that most changes are mechanical.

I think these are the files to pay most attention to:

- [`axum-core/src/extract/mod.rs`](https://github.com/tokio-rs/axum/pull/1155/files#diff-6bbc3220d3ec9f22ec90857669e8b82b0d4970bcc9dd6f0ca266cb5a8c16a468): These are the changes that cascades everywhere.
- [`axum/src/handler/mod.rs`](https://github.com/tokio-rs/axum/pull/1155/files#diff-d1ca66f6fa5355cd5c702c84c777a15f42910fb29caf860e2fac8a44d4eeaf8e): How handlers are converted into services.
- [`axum/src/extract/state.rs`](https://github.com/tokio-rs/axum/pull/1155/files#diff-4c32c4f402d416d7b1a0ae867a64c5d3965874df789fb2820e5f840b290591fe): The new extractor that makes it all work.
- [`axum/src/routing/mod.rs`](https://github.com/tokio-rs/axum/pull/1155/files#diff-3612ab48dcc0ce7483ee90b591848663098ecd84f64b2d616be1138e5cf40b36): The new `route`/`route_service` and  `fallback`/`fallback_service` methods and how the state is passed to merged routers.
- [`axum/src/routing/method_routing.rs`](https://github.com/tokio-rs/axum/pull/1155/files#diff-0ac534f5bbddb174f7788ec49e9319ccb032977b748a162f32ea3a8f3866d70d): How the state is provided to standalone `MethodRouter`s and when `MethodRouter` implements `Service`
- - [`axum/CHANGELOG.md`](https://github.com/tokio-rs/axum/pull/1155/files#diff-75b1551ef8a04f62139a37a08fbdbeeda76bd7a7dc055571b10e369a61d8bf1d): Does it make sense.

davidpdrsn:
Also note the "known limitations" and "follow-ups" I posted in the top comment. I'll make issues for the follow-ups once this is merged.

jplatte:
You wrote above that you want to merge #1121 first, has that plan changed?

In any case, I'm still not really happy about this change... I just don't see this composing well and I feel like it's also a major divergence from heavily leaning into tower.

davidpdrsn:
> You wrote above that you want to merge #1121 first, has that plan changed?

Oh I had forgotten about that. I thought this PR was more thoroughly baked so wanted to get it in first and don't wanna do both at the same time due to merge conflicts.

> I just don't see this composing well

Do you have some examples of where it might be problematic? Should I update the cookie extractors in the this branch (or a PR into this branch) so we can see what that looks like?

> and I feel like it's also a major divergence from heavily leaning into tower.

I don't think anything is lost with this change:
- You can still call `route_service` to call any service.
- If you never use a `State` extractor your state will be `()` and everything will implement `Service`, like it used it.

davidpdrsn:
I went ahead and ported the cookie extractors in https://github.com/tokio-rs/axum/pull/1250. Works pretty well I think.

jplatte:
Yeah, I understand that that does. But middleware isn't going to be able to interact with the type safe state, right?

davidpdrsn:
Accessing state from middleware isn't quite as elegant but definitely possible. I added some docs in babb073a8bb2e

davidpdrsn:
So, how do we move forward? Should I ask around for feedback from others?

jplatte:
I think that's a good idea, yeah. In particular I would be curious whether people maintaining axum apps of various sizes expect to use this over `Extension` after upgrading to 0.6. The only axum app I'm currently maintaining is very small and isn't using `Extension` at all 😅

davidpdrsn:
Alright. I've posted on [the rust language discord](https://discord.com/channels/273534239310479360/335502747409317928/1008739309861482506) and [twitter](https://twitter.com/DavidPdrsn/status/1559183808339083270).

NateLing:
I think i can put the db pool and some env variable into state , It would be great to be able to access the request when constructing the state. I will share the use case I conceived later.

ekanna:
I strongly believe we should avoid runtime errors where ever it's possible. The main reason we code in rust is it will give us compile time guarantees. I love this "Type Safe State Extractor".

davidpdrsn:
> It would be great to be able to access the request when constructing the state. I will share the use case I conceived later.

@NateLing that's not really possible since the state is set when creating the router. Might require a middleware.

yanns:
Being typesafe is great!
I was just being curious is the new parameter would make some code more complex or not.
I took the time to use this branch in a little application:
```
axum = { git = "https://github.com/tokio-rs/axum", branch = "router-state-extractor" }
```
At the end, the number of changes was very minimal:
- adding the `S` parameter to some `FromRequest` implementations
- adding a `()` state in `RequestParts::new(request, ());`

The rest of the application worked unchanged. 💯

jplatte:
> `RequestParts::new(request, ())`

@davidpdrsn Maybe keep `RequestParts::new` as before and add `RequestParts::with_state`?

davidpdrsn:
> Maybe keep RequestParts::new as before and add RequestParts::with_state?

Oh yeah good idea! That also matches `Router`. I'll change that!

yanns:
Additional comment: I've migrated some `Extension` to `State`, and this was straightforward. Awesome work!

112RG:
I am very new to rust. But I do have a question. I am using SeaOrm with connection pooling. How would I go about using the State? My code for an example `Extension(ref db): Extension<DatabaseConnection>,` ? Is this supported by AppState?

davidpdrsn:
@112RG Yes. Have a look at https://github.com/tokio-rs/axum/blob/router-state-extractor/examples/tokio-postgres/src/main.rs. It'll be very similar to that.

davidpdrsn:
We didn't get much feedback but people generally seem to be liking this.

jplatte:
Yeah I was hoping for more feedback, but whatever. I've started a proper review. I'm still sad that because of this, my `FromRequestOnce` idea needs to take on a pretty ugly form, but I've actually seen some ideas for language changes thrown around somewhat recently that would allow it to work in the future.

axum-extra/src/extract/cached.rs
```diff
@@ -139,19 +142,20 @@ mod tests {
...
 
-            async fn from_request(_req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
+            async fn from_request(_req: &mut RequestParts<B, S>) -> Result<Self, Self::Rejection> {
                 COUNTER.fetch_add(1, Ordering::SeqCst);
                 Ok(Self(Instant::now()))
             }
         }
 
-        let mut req = RequestParts::new(Request::new(()));
+        let mut req = RequestParts::with_state(Request::new(()), ());
```
jplatte, line 158:
```suggestion
        let mut req = RequestParts::new(Request::new(()));
```

axum-extra/src/routing/spa.rs
```diff
@@ -147,7 +147,7 @@ impl<B, T, F> SpaRouter<B, T, F> {
     }
 }
 
-impl<B, F, T> From<SpaRouter<B, T, F>> for Router<B>
+impl<B, F, T> From<SpaRouter<B, T, F>> for Router<(), B>
```
jplatte, line 150: Is there a specific reason `SpaRouter` doesn't become generic as well?

axum-macros/tests/from_request/fail/generic_without_via.stderr
```diff
@@ -12,15 +12,24 @@ warning: unused import: `Extension`
...
 
-error[E0277]: the trait bound `fn(Extractor<()>) -> impl Future<Output = ()> {foo}: Handler<_, _>` is not satisfied
+error[E0599]: no function or associated item named `new` found for struct `Router<Body>` in the current scope
+  --> tests/from_request/fail/generic_without_via.rs:10:21
+   |
+10 |     Router::<Body>::new().route("/", get(foo));
+   |                     ^^^ function or associated item not found in `Router<Body>`
+   |
+   = note: the function or associated item was found for
+           - `Router<(), B>`
```
jplatte, line 22: This seems unintentional, looks like the test should be updated to use `Router::<_, Body>`.

axum-macros/tests/from_request/fail/generic_without_via_rejection.stderr
```diff
@@ -12,15 +12,24 @@ warning: unused import: `Extension`
...
 
-error[E0277]: the trait bound `fn(Extractor<()>) -> impl Future<Output = ()> {foo}: Handler<_, _>` is not satisfied
+error[E0599]: no function or associated item named `new` found for struct `Router<Body>` in the current scope
+  --> tests/from_request/fail/generic_without_via_rejection.rs:11:21
+   |
+11 |     Router::<Body>::new().route("/", get(foo));
+   |                     ^^^ function or associated item not found in `Router<Body>`
+   |
+   = note: the function or associated item was found for
+           - `Router<(), B>`
```
jplatte, line 22: Same here.

axum-macros/tests/from_request/fail/generic_without_via_rejection_derive.stderr
```diff
@@ -12,15 +12,24 @@ warning: unused import: `Extension`
...
 
-error[E0277]: the trait bound `fn(Extractor<()>) -> impl Future<Output = ()> {foo}: Handler<_, _>` is not satisfied
+error[E0599]: no function or associated item named `new` found for struct `Router<Body>` in the current scope
+  --> tests/from_request/fail/generic_without_via_rejection_derive.rs:11:21
+   |
+11 |     Router::<Body>::new().route("/", get(foo));
+   |                     ^^^ function or associated item not found in `Router<Body>`
+   |
+   = note: the function or associated item was found for
+           - `Router<(), B>`
```
jplatte, line 22: Here too.

axum-macros/tests/from_request/pass/named_via.rs
```diff
@@ -25,7 +25,7 @@ struct Extractor {
 
 fn assert_from_request()
 where
-    Extractor: FromRequest<Body, Rejection = ExtractorRejection>,
+    Extractor: FromRequest<Body, (), Rejection = ExtractorRejection>,
```
jplatte, line 28: Just noticed this: I think the order of `B` and `S` in `FromRequest` and `Router` should match.

axum/CHANGELOG.md
```diff
@@ -35,6 +35,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  `Router::with_state` and gives compile errors for missing states, whereas
+  `Extension` would result in runtime errors ([#1155])
+- **breaking:** The following types or traits have a new `S` type param
+  (`()` by default) which represents the state ([#1155]):
+  - `FromRequest`
+  - `RequestParts`
+  - `Router`
+  - `MethodRouter`
+  - `Handler`
+  - `Resource`
```
jplatte, line 48: This is axum-extra though, right?

axum-extra/src/routing/mod.rs
```diff
@@ -244,41 +241,35 @@ where
...
-        H: Handler<T, B>,
+        H: Handler<T, S, B>,
         T: FirstElementIs<P> + 'static,
         P: TypedPath,
     {
         self.route(P::PATH, axum::routing::trace(handler))
     }
 
-    fn route_with_tsr<T>(mut self, path: &str, service: T) -> Self
+    fn route_with_tsr(mut self, path: &str, method_router: MethodRouter<S, B>) -> Self
```
jplatte, line 251: We need `route_service_with_tsr` now, right?

axum-extra/src/routing/mod.rs
```diff
@@ -159,22 +158,20 @@ pub trait RouterExt<B>: sealed::Sealed {
     ///     .route_with_tsr("/bar/", get(|| async {}));
     /// # let _: Router = app;
     /// ```
-    fn route_with_tsr<T>(self, path: &str, service: T) -> Self
+    fn route_with_tsr(self, path: &str, method_router: MethodRouter<S, B>) -> Self
```
jplatte, line 161: Also here (although I don't know off the top of my head why this exists twice in the first place).

axum/src/extract/state.rs
```diff
@@ -0,0 +1,206 @@
...
+/// Note that you don't need to use the `State` extractor since you can access the state directly
+/// from [`RequestParts`].
+#[derive(Debug, Default, Clone, Copy)]
+pub struct State<S>(pub S);
+
+#[async_trait]
+impl<B, OuterState, InnerState> FromRequest<B, OuterState> for State<InnerState>
+where
+    B: Send,
+    OuterState: Clone + Into<InnerState> + Send,
```
jplatte, line 183: I wonder whether cloning the outer state to then access a single field could prove to be a problematic pessimization. The only alternative I can think of is using a custom trait instead of `Into<InnerState>`, something that has `Clone` as a supertrait and a `&self` conversion method, and blanket `impl<T: Clone> ThatTrait<T> for T` like `From` (but in this case mapping `&T -> T`.

axum/src/handler/mod.rs
```diff
@@ -260,58 +210,116 @@ all_the_tuples!(impl_handler);
...
     }
 }
 
-impl<S, T> Layered<S, T> {
-    pub(crate) fn new(svc: S) -> Self {
-        Self {
-            svc,
-            _input: PhantomData,
-        }
+/// Extension trait for [`Handler`]s who doesn't have state.
```
jplatte, line 280:
```suggestion
/// Extension trait for [`Handler`]s that don't have state.
```

jplatte (commented):
Screenshot_2022-08-17_105123

Will continue review some other time 😮‍💨

axum-extra/src/routing/mod.rs
davidpdrsn: Yes!

axum-extra/src/routing/mod.rs
davidpdrsn: Its because its defined on an extension trait (`RouterExt`)

axum-macros/tests/from_request/fail/generic_without_via.stderr
davidpdrsn: oh yeah good catch!

axum-extra/CHANGELOG.md
```diff
@@ -25,9 +25,13 @@ and this project adheres to [Semantic Versioning].
 - **added:** `WithRejection` extractor for customizing other extractors' rejections ([#1262])
 - **added:** Add sync constructors to `CookieJar`, `PrivateCookieJar`, and
   `SignedCookieJar` so they're easier to use in custom middleware
+- **breaking:** `Resource` has a new `S` type param which represents the state ([#1155])
+- **breaking:** `RouterExt::route_with_tsr` now only accepts `MethodRouter`s ([#1155])
+- **added:** `RouterExt::route_service_with_tsr` for routing to any `Service` ([#1155])
```
jplatte, line 30: What happened here? Why do these entries come after some of the link definitions?

axum-extra/src/routing/spa.rs
davidpdrsn: Yeah because it doesn't need to access the state and the user cannot add handlers directly to `SpaRouter`. I added a test in b6a2f39ca3dca just to make sure it actually works with routers that have state.

axum-macros/tests/from_request/pass/named_via.rs
davidpdrsn: Good point! I agree. The order on `FromRequest` and `RequestParts` was a hold over from when we used to have `S = ()`. So that had to be last since we don't have a default for `B`.

I don't think we should have a default type param for `FromRequest` and `RequestParts` because I'd worry library authors accidentally doing

```rust
impl<B> FromRequest<B> for Foo {}
```

without realizing that their extractor now only works for `Router<()>`.

I'll fix it (and change all the files again 🥲)

axum-extra/CHANGELOG.md
davidpdrsn: Not sure what you mean 🤔 

image

jplatte: You can click on the filename here to see the diff I was seeing:
Screenshot_2022-08-17_143057

But looks like it's not an actual problem then 👍🏼

davidpdrsn: Oh yeah I see it when I click the filename. No idea what github is doing.

axum-macros/tests/from_request/pass/named_via.rs
davidpdrsn: Done!

axum/src/extract/state.rs
davidpdrsn: Yeah thats a good point. We could do something like this

```rust
pub trait FromRef<T>: Clone {
    fn from_ref(value: &T) -> Self;
}

impl<T> FromRef<T> for T
where
    T: Clone,
{
    fn from_ref(value: &T) -> Self {
        value.clone()
    }
}
```

jplatte: Slight nitpick: With a general name like `FromRef`, maybe the trait itself shouldn't have `Clone` as a supertrait, i.e. maybe it should only be a bound on the blanket impl.

axum-extra/src/routing/mod.rs
```diff
@@ -244,41 +253,56 @@ where
...
-        self = self.route(path, service);
+        self = self.route_service(path, service);
 
         let redirect = Redirect::permanent(path);
 
         if let Some(path_without_trailing_slash) = path.strip_suffix('/') {
             self.route(
                 path_without_trailing_slash,
-                (move || ready(redirect.clone())).into_service(),
+                any(move || ready(redirect.clone())),
```
jplatte, line 295: Should this be `route_service` + `into_service` so there's no unnecessary `MethodRouter` involved?

axum-extra/src/routing/mod.rs
```diff
@@ -244,41 +253,56 @@ where
...
             )
         } else {
-            self.route(
-                &format!("{}/", path),
-                (move || ready(redirect.clone())).into_service(),
-            )
+            self.route(&format!("{}/", path), any(move || ready(redirect.clone())))
```
jplatte, line 298: Same pattern here.

axum/src/routing/mod.rs
```diff
@@ -131,39 +156,53 @@ where
...
+    {
+        if path.is_empty() {
+            panic!("Paths must start with a `/`. Use \"/\" for root routes");
+        } else if !path.starts_with('/') {
+            panic!("Paths must start with a `/`");
+        }
+
+        let service = match try_downcast::<Router<S, B>, _>(service) {
+            Ok(_) => {
+                panic!("Invalid route: `Router::route_service` cannot be used with `Router`s. Use `Router::nest` instead")
```
jplatte, line 197: This is an arbitrary restriction, right? I.e. there is no *technical* reason to forbid this? Just want to know I'm understanding.

axum/src/routing/tests/mod.rs
```diff
@@ -641,6 +641,54 @@ async fn limited_body_with_streaming_body() {
     assert_eq!(res.status(), StatusCode::PAYLOAD_TOO_LARGE);
 }
 
+#[tokio::test]
+async fn extracting_state() {
```
jplatte, line 645: May be very nitpicky, but this reads much better to me:

```suggestion
async fn extract_state() {
```

axum/src/routing/tests/mod.rs
```diff
@@ -641,6 +641,54 @@ async fn limited_body_with_streaming_body() {
...
+
+    let app = Router::with_state(state).route("/", get(handler));
+    let client = TestClient::new(app);
+
+    let res = client.get("/").send().await;
+    assert_eq!(res.status(), StatusCode::OK);
+}
+
+#[tokio::test]
+async fn explicitly_setting_state() {
```
jplatte, line 681:
```suggestion
async fn explicitly_set_state() {
```

examples/testing/src/main.rs
```diff
@@ -34,7 +34,7 @@ async fn main() {
 /// Having a function that produces our app makes it easy to call it from tests
 /// without having to create an HTTP server.
 #[allow(dead_code)]
-fn app() -> Router {
+fn app() -> Router<()> {
```
jplatte, line 37: This change isn't actually needed, is it?

axum/src/extract/state.rs
davidpdrsn: I made a separate PR for this https://github.com/tokio-rs/axum/pull/1268

axum/src/routing/mod.rs
davidpdrsn: Yeah its arbitrary. It doesn't really make sense to do and we've had some people ask why it didn't work. Generally people want `nest`.

examples/testing/src/main.rs
davidpdrsn: Nope!

jplatte (approved):

axum/src/lib.rs
```diff
@@ -168,13 +168,48 @@
...
+//! ) {
+//!     // ...
+//! }
+//! # async {
+//! # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
+//! # };
+//! ```
+//!
+//! You should prefer using [`State`] if possible since it's more type safe. The downside is that
+//! its less dynamic than request extensions.
```
pickfire, line 205: I was trying a similar experiment to have an `AnyMap` within `State` to allow using dynamic `Extension` that is not type-safe. But the issue with that is that the impl crashes due to `impl FromRef<T> for State` crashes with those specific implementations of `FromRef`.

I believe this is also more performant since we no longer have to go through a `HashMap` of types (`AnyMap`) and can access the fields directly. Not sure if we want to write it down but I will try to do a benchmark and see what's the difference.

davidpdrsn: We already have a benchmark of extension vs state in the repo.

axum-extra/src/extract/cookie/signed.rs
```diff
@@ -63,22 +61,36 @@ use std::{convert::Infallible, fmt, marker::PhantomData};
...
+///     // that holds the key used to sign cookies
+///     key: Key,
+/// }
+///
+/// // this impl tells `SignedCookieJar` how to access the key from our state
+/// impl FromRef<AppState> for Key {
+///     fn from_ref(state: &AppState) -> Self {
+///         state.key.clone()
+///     }
+/// }
```
pickfire, line 76: Is it a good idea to recommend implementing `FromRef` for a cloned version that requires `.clone()` instead of a borrowed version? The `FromRef` implementations from the docs seemed to always have `.clone()`.

davidpdrsn: How else would you do it?

pickfire: Maybe recommend the borrowed version like `impl FromRef<AppState> for &Key` in cases where clone is heavy and user does not need the owned version? Like show an example.

davidpdrsn: Yeah exactly. I don't think it'll actually work.

wangluyi1982:
looking forward to the 0.6 release. this is wonderful!

PR #1163, opened 2022-07-14
added local-addr code

sugihjamin:
## Motivation

To provide example code to obtain server's IP:port used in request Uri.

On a server with multiple IPs, it is sometimes useful to know which IP a client request came in from.

## Solution

Obtain server's IP:port first from the request's pseudo-header authority field, in case of HTTP/2, then from the header's host field, in case of HTTP/1.

Example code is built off the "hello-world" example code.

### DIFF
--- a/examples/local-addr/Cargo.toml
+++ b/examples/local-addr/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "local-addr"
+version = "0.1.0"
+edition = "2021"
+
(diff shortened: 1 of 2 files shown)


davidpdrsn:
This is what the [`Host`](https://docs.rs/axum/latest/axum/extract/struct.Host.html) extractor does so don't feel we need an example for it.

sugihjamin:
Right. I missed the `Host` extractor.

PR #1165, opened 2022-07-15
Accept `S::Response: IntoResponse` everywhere

davidpdrsn:
This is a continuation of https://github.com/tokio-rs/axum/pull/1152 by doing the same for all methods that accept some `S: Service`. I think this is a win as it makes things more flexible for advanced users while still allowing `Response`s to be returned.

### DIFF
--- a/axum-extra/src/routing/spa.rs
+++ b/axum-extra/src/routing/spa.rs
@@ -1,7 +1,7 @@
 use axum::{
     body::{Body, HttpBody},
     error_handling::HandleError,
-    response::Response,
+    response::IntoResponse,
     routing::{get_service, Route},
     Router,
 };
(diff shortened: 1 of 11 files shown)


axum-extra/src/routing/spa.rs
```diff
@@ -150,8 +150,8 @@ impl<B, T, F> SpaRouter<B, T, F> {
 impl<B, F, T> From<SpaRouter<B, T, F>> for Router<B>
 where
     F: Clone + Send + 'static,
-    HandleError<Route<B, io::Error>, F, T>:
-        Service<Request<B>, Response = Response, Error = Infallible>,
+    HandleError<Route<B, io::Error>, F, T>: Service<Request<B>, Error = Infallible>,
+    <HandleError<Route<B, io::Error>, F, T> as Service<Request<B>>>::Response: IntoResponse + Send,
```
jplatte, line 154: It's interesting that in a few cases like this one, `Send` is also required while in others it isn't. Did you investigate why that is? Should we maybe have it everywhere for consistency and allow a little more leeway with implementations?

jplatte (approved):

PR #1166, opened 2022-07-15
Redirect HTTP to HTTPS in `tls-rustls` example

sugihjamin:
## Motivation

Redirect http request for "http://<SERVER>:<HTTP_PORT>/" to "https://<SERVER>:<HTTPS_PORT>/"

## Solution

Run an axum server listening on HTTP_PORT and a rustls axum_server listening on HTTPS_PORT. Http server returns an axum::response::Redirect with original uri reconstructed to point to the https server.


davidpdrsn:
Nice! I think this makes sense to have an example for. Maybe we can just add it to `tls-rustls` rather than making a whole new one that is 90% identical.

examples/http-redirect/src/main.rs
```diff
@@ -0,0 +1,88 @@
...
+        .await
+        .unwrap();
+}
+
+async fn handler() -> Html<&'static str> {
+    Html("<h1>Hello secured world!</h1>")
+}
+
+async fn redirect() {
+    let app = Router::new().fallback(fallback.into_service());
```
davidpdrsn, line 67: You don't need to create a whole `Router`. `fallback.into_make_service()` also works thanks to [`Handler::into_make_service`](https://docs.rs/axum/latest/axum/handler/trait.Handler.html#method.into_make_service)

sugihjamin:
Yes, it works merged into `tls-rustls.

I wasn't sure whether you'd want to "complicate" `tls-rustls` with a second issue. It's nice that each example currently illustrates a single issue.

Let me read up on `Handler::into_make_service` and learn how to use it.

davidpdrsn:
> I wasn't sure whether you'd want to "complicate" tls-rustls with a second issue. It's nice that each example currently illustrates a single issue.

But copying an existing example and changing like 5 lines isn't worth it 😜 We should add it to `tls-rustls`

sugihjamin:
I've merged the bits to do with redirection into `tls-rustls` (and removed the `http-redirect` directory).

examples/tls-rustls/src/main.rs
```diff
@@ -42,3 +54,25 @@ async fn main() {
...
+    tracing::debug!("308: Permanent Redirect");
+
+    // Can remove call to `.replace()` if using default
+    // ports for http (80) and https (443).
+    Redirect::permanent(&*format!(
+        "https://{}{}",
+        &host.replace("7878", "3000"),
+        &uri
+    ))
+}
```
davidpdrsn, line 78: What do you think about doing something like this instead:

```suggestion
#[derive(Clone, Copy)]
struct Ports {
    http: u16,
    https: u16,
}

async fn redirect_http_to_https(ports: Ports) {
    fn make_https(host: String, uri: Uri, ports: Ports) -> Result<Uri, BoxError> {
        let mut parts = uri.into_parts();

        parts.scheme = Some(axum::http::uri::Scheme::HTTPS);

        if parts.path_and_query.is_none() {
            parts.path_and_query = Some("/".parse().unwrap());
        }

        let https_host = host.replace(&ports.http.to_string(), &ports.https.to_string());
        parts.authority = Some(https_host.parse()?);

        Ok(Uri::from_parts(parts)?)
    }

    let redirect = move |Host(host): Host, uri: Uri| async move {
        match make_https(host, uri, ports) {
            Ok(uri) => Ok(Redirect::permanent(&uri.to_string())),
            Err(error) => {
                tracing::warn!(%error, "failed to convert URI to HTTPS");
                Err(StatusCode::BAD_REQUEST)
            }
        }
    };

    let addr = SocketAddr::from(([127, 0, 0, 1], ports.http));
    tracing::debug!("http redirect listening on {}", addr);

    axum::Server::bind(&addr)
        .serve(redirect.into_make_service())
        .await
        .unwrap()
}
```

So
- Make a `Ports` struct instead of duplicating the ports
- Using `Uri::into_parts` and handling errors, rather than working with strings

davidpdrsn:
cc @programatik29 because I believe you worked on this example in the past.

sugihjamin:
Very nice!  The new version is built from first principles, more "generic" in terms of port numbers used, and with proper error handling.

What do you think of letting `ports` be of type `Option<Ports>` instead? The internal function `make_https()` can then default to the default ports if `None` is passed to `redirect_http_to_https(ports: Option<Ports>)`:
```
-        let https_host = host.replace(&ports.http.to_string(), &ports.https.to_string());
+       let https_host = if let Some(ports) = ports { host.replace(&ports.http.to_string(), &ports.https.to_string()) } else { host };
```

I also like that the new version retains the use of the `Host` extractor. (Running grep in the example directory shows that this will be the first use of the `Host` extractor in the examples.)

davidpdrsn:
> What do you think of letting `ports` be of type `Option<Ports>` instead? The internal function `make_https()` can then default to the default ports if `None` is passed to `redirect_http_to_https(ports: Option<Ports>)`:

But why would `Ports` be `None`? When you're creating the server I suppose you always know the ports.

sugihjamin:
So that one can be lazy in the "production" case and not have to specify the default ports (80 and 443)?

One additional change to use `Option<Ports>` though:
```
-    let addr = SocketAddr::from(([127, 0, 0, 1], ports.http));
+    let addr = SocketAddr::from(([127, 0, 0, 1], if let Some(port) = ports { ports.http } else { 80 }));
```

Either way is fine with me.  I just thought we could take advantage of the default cases. Thanks.

sugihjamin:
I take back the suggestion to use `Option` since one always has to specify the port in constructing a `SocketAddr` anyway.

examples/tls-rustls/src/main.rs
```diff
@@ -42,3 +64,38 @@ async fn main() {
...
+    };
+
+    let addr = SocketAddr::from(([127, 0, 0, 1], ports.http));
+    tracing::debug!("http redirect listening on {}", addr);
+
+    axum::Server::bind(&addr)
+        .serve(redirect.into_make_service())
+        .await
+        .unwrap()
+}
```
programatik29, line 101: I suggest depending on `url` crate and:

```rust
async fn redirect_http_to_https(ports: Ports) {
    let addr = SocketAddr::from(([127, 0, 0, 1], ports.http));
    tracing::debug!("http redirect listening on {}", addr);

    let app = redirect.layer(Extension(ports));

    axum::Server::bind(&addr)
        .await
        .unwrap()
}

async fn redirect(
    Host(host): Host,
    uri: Uri,
    Extension(ports): Extension<Ports>,
) -> Result<Redirect, StatusCode> {
    let path_and_query = uri
        .path_and_query()
        .map(ToString::to_string)
        .unwrap_or("".to_owned());

    let mut url = Url::parse(&format!("http://{}{}", host, path_and_query))
        .map_err(|_| StatusCode::BAD_REQUEST)?;

    url.set_scheme("https").unwrap();
    let _ = url.set_port(Some(ports.https));

    Ok(Redirect::permanent(&url.to_string()))
}
```

davidpdrsn: What advantage does that give? The code doesn't seem less complicated than the version using `http::Uri` 🤔

sugihjamin:
I'm not sure if I the "check (examples)" issue was that the last statement in `redirect_http_to_https()` was missing a ';'?

examples/tls-rustls/src/main.rs
sugihjamin: The version using `url` crate has a more functional approach, which may be preferred by some developers.  As a teaching example, though, I like that the version using `http::Uri` show how to work with the parts of `Uri` explicitly. I also prefer not to have to rely on an additional crate and middleware layer and extension extractor.

programatik29: I think this is simpler and less error prone. Things are more likely to go wrong when using `http::Uri`.

davidpdrsn: I don't think pulling in an extra dependency for an example is worth it.

examples/tls-rustls/src/main.rs
sugihjamin: Sorry I didn't see the "Commit suggestion" button.  Next time I'll know to use it.

davidpdrsn (approved):

davidpdrsn:
@sugihjamin thanks for adding this!

PR #1168, opened 2022-07-15
Add note on default responses if extraction fails

### LINKED ISSUE #1167: Add documentation on default rejection behavior if extraction fails
This is supposed to make the exposure of error messages more clear.

Mentioned in #1116.

SzudemJ:
This is useful in case one does not want to expose internal error
messages through the interface.

Fixes #1167

### DIFF
--- a/axum/src/lib.rs
+++ b/axum/src/lib.rs
@@ -112,6 +112,10 @@
 //! async fn json(Json(payload): Json<serde_json::Value>) {}
 //! ```
 //!
+//! If an extraction fails, axum [`rejects`](crate::extract::rejection) requests
+//! and responds with an appropriate status code and the respective error message.
+//! This behavior can be adapted by customizing extractor responses.
+//!
 //! See [`extract`](crate::extract) for more details on extractors.
 //!
 //! # Responses


davidpdrsn:
Maybe it's more appropriate to mention [here](https://github.com/tokio-rs/axum/blob/main/axum/src/docs/extract.md#customizing-extractor-responses). wdyt?

SzudemJ:
Yes, I thought about it. Default behavior is sort-of mentioned [here](https://github.com/tokio-rs/axum/blob/main/axum/src/docs/extract.md#customizing-extractor-responses) but I though one might skip reading this section if not interested in customizing responses.

Actually I digged a bit deeper into how default rejections are implemented and I saw, that library internals are only exposed [when extracting query parameters](https://docs.rs/axum/latest/src/axum/extract/rejection.rs.html#125-133).

```rust
impl std::fmt::Display for FailedToDeserializeQueryString {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(
            f,
            "Failed to deserialize query string. Expected something of type `{}`. Error: {}",
            self.type_name, self.error,
        )
    }
}
```

For comparison, the default json response does not expose such details, even if a field is missing. It simply names the field in the response body:

Example of Json response:
`Failed to deserialize the JSON body into the target type: missing field 'has_accepted_terms' at line 4 column 1`

Example Query response:
`Failed to deserialize query string. Expected something of type 'registration_service::routes::registrations::RequestParams'. Error: missing field 'email'`


**Just to avoid misunderstandings:**
You really think it is necessary to include `self.type_name` in the error message?

SzudemJ:
Closing as this has been resolved by #1171

PR #1171, opened 2022-07-18
Avoid exposure of type names by QueryRejection

SzudemJ:
## Motivation

This is neccessary, to avoid public exposure of implementation details
by the default error message if the extractor fails to deserialize
the query parameters into the target type.

## Solution

Removing the `type_name` field from `FailedToDeserializeQueryString`.

### DIFF
--- a/axum/src/extract/rejection.rs
+++ b/axum/src/extract/rejection.rs
@@ -100,7 +100,6 @@ define_rejection! {
 #[derive(Debug)]
 pub struct FailedToDeserializeQueryString {
     error: Error,
-    type_name: &'static str,
 }
 
 impl FailedToDeserializeQueryString {
@@ -111,7 +110,6 @@ impl FailedToDeserializeQueryString {
     {
         FailedToDeserializeQueryString {
             error: Error::new(error),
-            type_name: std::any::type_name::<T>(),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


axum/src/extract/rejection.rs
```diff
@@ -111,7 +110,6 @@ impl FailedToDeserializeQueryString {
     {
         FailedToDeserializeQueryString {
             error: Error::new(error),
-            type_name: std::any::type_name::<T>(),
```
davidpdrsn, line 114: I think we should remove the `T` type param as well since it's not used anymore.

davidpdrsn (changes requested):

jplatte (approved):

axum/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+- **fixed:** `QueryRejection` response exposes type names through public interface ([#1171])
```
davidpdrsn, line 10: This almost sounds like the fix was to expose internal types. Maybe this is more clear?

```suggestion
- **fixed:** Don't expose internal type names in `QueryRejection` responses ([#1171])
```

SzudemJ: huh.. I guess you are right.

davidpdrsn (approved):
Thanks!

PR #1173, opened 2022-07-19
Update ECOSYSTEM.md

lingdu1234:
add axum_admin

### DIFF
--- a/ECOSYSTEM.md
+++ b/ECOSYSTEM.md
@@ -38,7 +38,7 @@ If your project isn't listed here and you would like it to be, please feel free 
 - [Pinging.net](https://www.pinging.net) ([repository](https://github.com/benhansenslc/pinging)): A new way to check and monitor your internet connection.
 - [wastebin](https://github.com/matze/wastebin): A minimalist pastebin service.
 - [sandbox_axum_observability](https://github.com/davidB/sandbox_axum_observability) A Sandbox/showcase project to experiment axum and observability (tracing, opentelemetry, jaeger, grafana tempo,...)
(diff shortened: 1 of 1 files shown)


ECOSYSTEM.md
```diff
@@ -38,6 +38,7 @@ If your project isn't listed here and you would like it to be, please feel free
 - [Pinging.net](https://www.pinging.net) ([repository](https://github.com/benhansenslc/pinging)): A new way to check and monitor your internet connection.
 - [wastebin](https://github.com/matze/wastebin): A minimalist pastebin service.
 - [sandbox_axum_observability](https://github.com/davidB/sandbox_axum_observability) A Sandbox/showcase project to experiment axum and observability (tracing, opentelemetry, jaeger, grafana tempo,...)
+- [axum_admin](https://github.com/lingdu1234/axum_admin)   An admin panel built with **axum**, Sea-orm and Vue 3.
```
jplatte, line 41:
```suggestion
- [axum_admin](https://github.com/lingdu1234/axum_admin): An admin panel built with **axum**, Sea-orm and Vue 3.
```

matze:
Stupid question and not wanting to sound in anyway discriminating/exclusionary but shouldn't we restrict ecosystem to code and examples that reach the widest possible audience? I don't have a single clue what exactly that repo/app does because I am unable to read Chinese (?) which is used both in docs, comments and commit messages.

jplatte:
I hadn't actually looked at the link. Personally I feel like there should be an indication that the contents are in a different language, but it shouldn't be kicked out of the list.

davidpdrsn:
I didn't actually check the link either, will do that from now 😅 I agree it's probably good to point out things that aren't in English but don't think such items should be removed from the list.

@lingdu1234 would you be able to provide and English translation?

Issue #1174, opened 2022-07-19
Add an example for multiple listen addresses


jplatte:
We should have an example for multiple listen addresses. Most of the code was already written by David, you can copy it from https://github.com/tokio-rs/axum/discussions/1063#discussioncomment-3051746. Somebody just needs to create an example from that and open a PR for it, and I'm always low on time.

There should probably also be a small comment about IPv6 catch-all listeners applying to IPv4 as well in some OS configurations (see https://serverfault.com/q/21657).

DesmondWillowbrook:
I'd like to work on this.

PR #1179, opened 2022-07-20
Add example for multiple listen addrs

DesmondWillowbrook:
Fixes: #1174


DesmondWillowbrook:
I tried to find an [authoritative source](https://docs.microsoft.com/en-us/windows/win32/winsock/dual-stack-sockets) for the Windows socket thing instead of citing the SO question directly.

davidpdrsn (approved):
Thanks!

PR #1184, opened 2022-07-22
Fix: Broken commonly used middleware links in middleware module

### LINKED ISSUE #1180: "Commonly used middleware" links are broken in middleware module docs
Title says it all.

anuvratsingh:
Fixes  #1180 
 - Fixes broken commonly used middleware links.
 - Make some link name shorter, e.g. `axum::middleware::from_fn` ⇾ `middleware::from_fn`
 
 Issue
  - Heading which starts with backtick(\`), e.g. (## \`middleware::from_fn\`) shows as `__` when opening docs(`cargo doc --open`) on both nightly and stable, probably my machine(x86_64-unknown-linux-gnu) specific issue.

### DIFF
--- a/axum/src/docs/middleware.md
+++ b/axum/src/docs/middleware.md
@@ -61,14 +61,12 @@ let app = Router::new()
 
 Some commonly used middleware are:
 
-- [`TraceLayer`](tower_http::trace::TraceLayer) for high level tracing/logging.
+- [`TraceLayer`] for high level tracing/logging.
 - [`CorsLayer`](tower_http::cors::CorsLayer) for handling CORS.
-- [`CompressionLayer`](tower_http::compression::CompressionLayer) for automatic compression of
-  responses.
-- [`SetRequestIdLayer`](tower_http::request_id::SetRequestIdLayer) and
-  [`PropagateRequestIdLayer`](tower_http::request_id::PropagateRequestIdLayer) set and propagate request
+- [`CompressionLayer`](tower_http::compression::CompressionLayer) for automatic compression of responses.
+- [`RequestIdLayer`] set and propagate request
   ids.
-- [`TimeoutLayer`](tower::timeout::TimeoutLayer) for timeouts. Note this
+- [`TimeoutLayer`] for timeouts. Note this
   requires using [`HandleErrorLayer`](crate::error_handling::HandleErrorLayer)
   to convert timeouts to responses.
 
@@ -170,22 +168,22 @@ mentally which is one of the reasons `ServiceBuilder` is recommended.
 axum offers many ways of writing middleware, at different levels of abstraction
 and with different pros and cons.
 
-## `middleware::from_fn`
+## `axum::middleware::from_fn`
 
-Use [`middleware::from_fn`] to write your middleware when:
+Use [`axum::middleware::from_fn`] to write your middleware when:
 
 - You're not comfortable with implementing your own futures and would rather use
   the familiar `async`/`await` syntax.
 - You don't intend to publish your middleware as a crate for others to use.
   Middleware written like this are only compatible with axum.
 
-## `middleware::from_extractor`
+## `axum::middleware::from_extractor`
 
-Use [`middleware::from_extractor`] to write your middleware when:
+Use [`axum::middleware::from_extractor`] to write your middleware when:
 
 - You have a type that you sometimes want to use as an extractor and sometimes
   as a middleware. If you only need your type as a middleware prefer
-  [`middleware::from_fn`].
+  [`axum::middleware::from_fn`].
 
 ## tower's combinators
 
@@ -210,7 +208,7 @@ by implementing [`tower::Service`]:
 Use [`tower::Service`] with `Pin<Box<dyn Future>>` to write your middleware when:
 
 - Your middleware needs to be configurable for example via builder methods on
-  your [`tower::Layer`] such as [`TraceLayer`](tower_http::trace::TraceLayer).
+  your [`tower::Layer`] such as [`TraceLayer`].
 - You do intend to publish your middleware as a crate for others to use.
 - You're not comfortable with implementing your own futures.
 
@@ -275,7 +273,7 @@ Use [`tower::Service`] with manual futures to write your middleware when:
 
 - You want your middleware to have the lowest possible overhead.
 - Your middleware needs to be configurable for example via builder methods on
-  your [`tower::Layer`] such as [`TraceLayer`](tower_http::trace::TraceLayer).
+  your [`tower::Layer`] such as [`TraceLayer`].
 - You do intend to publish your middleware as a crate for others to use, perhaps
   as part of tower-http.
 - You're comfortable with implementing your own futures, or want to learn how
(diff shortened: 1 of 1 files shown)


axum/src/docs/middleware.md
```diff
@@ -61,12 +61,12 @@ let app = Router::new()
 
 Some commonly used middleware are:
 
-- [`TraceLayer`](tower_http::trace) for high level tracing/logging.
-- [`CorsLayer`](tower_http::cors) for handling CORS.
-- [`CompressionLayer`](tower_http::compression) for automatic compression of
+- [`TraceLayer`](tower_http::trace::TraceLayer) for high level tracing/logging.
```
jplatte, line 64: I don't see why intra-doc links would work for the type, but not the module. In general I wouldn't really have expected this to work at all, they're called **intra**-doc links afterall. @davidpdrsn are you sure this used to work?

anuvratsingh: - \- \[\`TraceLayer\`\](tower_http::trace) works fine locally, but somehow on [docs.rs](https://docs.rs/axum/latest/axum/middleware/) it's broken.
- Also, earlier I thought the issue was with linking the whole module rather than the specific Layer.

jplatte: I see. Can you try this:

* Make sure you have an up-to-date nightly toolchain
* Create `.cargo/config.toml` with the following content:
  ```toml
  [doc.extern-map.registries]
  crates-io = "https://docs.rs/"
  
  [unstable]
  rustdoc-map = true
  ```
* Run `RUSTDOCFLAGS='--cfg docsrs' cargo +nightly doc --no-deps --all-features`

That should get you very close to how docs.rs generates documentation, maybe you can reproduce the issue then.

anuvratsingh: Following the steps specified, both branches `upstream/main` and `origin/brokendocslink` works fine. Former linking the whole module, while latter the specific layer.

```
rustc 1.64.0-nightly (f8588549c 2022-07-18)
binary: rustc
commit-hash: f8588549c3c3d45c32b404210cada01e2a45def3
commit-date: 2022-07-18
host: x86_64-unknown-linux-gnu
release: 1.64.0-nightly
LLVM version: 14.0.6
```

jplatte: Interesting. I wonder whether this was just a nightly regression then that got fixed since we last published a release on docs.rs, or some way in which docs.rs does things differently to that setup.

anuvratsingh: Should I try it on the released doc's nightly version? But I don't know how to get that.

davidpdrsn: Just tried shipping a new version and thus rebuilding the docs. [But this issue is still there.](https://github.com/tokio-rs/axum/issues/1180#issuecomment-1194327015). So seems we do need this after all. Not clue why it works locally but nto on docs.rs

davidpdrsn: These links shouldn't go to the `Layer` but the module, so `tower_http::trace` instead of `tower_http::trace::TraceLayer`. The modules contains the docs an examples.

Also seems we need to use absolute links rather than intra docs links. Not sure why their broken on docs.rs 🤷

axum/src/docs/middleware.md
```diff
@@ -61,12 +61,12 @@ let app = Router::new()
...
+- [`CorsLayer`](tower_http::cors::CorsLayer) for handling CORS.
+- [`CompressionLayer`](tower_http::compression::CompressionLayer) for automatic compression of
   responses.
-- [`RequestIdLayer`](tower_http::request_id) and
-  [`PropagateRequestIdLayer`](tower_http::request_id) set and propagate request
+- [`SetRequestIdLayer`](tower_http::request_id::SetRequestIdLayer) and
+  [`PropagateRequestIdLayer`](tower_http::request_id::PropagateRequestIdLayer) set and propagate request
   ids.
 - [`TimeoutLayer`](tower::timeout::TimeoutLayer) for timeouts. Note this
```
davidpdrsn, line 71: Wanna change this to link to https://docs.rs/tower-http/latest/tower_http/timeout/index.html instead?

axum/src/docs/middleware.md
```diff
@@ -170,18 +170,18 @@ mentally which is one of the reasons `ServiceBuilder` is recommended.
 axum offers many ways of writing middleware, at different levels of abstraction
 and with different pros and cons.
 
-## `axum::middleware::from_fn`
+## `middleware::from_fn`
```
davidpdrsn, line 173: I think we should keep the `axum::` prefix here since this section contains references to tower and users might not know if `middleware::from_fn` exists in axum or tower.

davidpdrsn (changes requested):

axum/src/docs/middleware.md
anuvratsingh: Does absolute links here means direct link to `docs.rs`?

davidpdrsn: Yes.

axum/src/docs/middleware.md
anuvratsingh: Reasoning behind this was, most of the links referencing to other crates already had crate name in it like `tower::Layer/Service` rather than just `Layer/Serive`. Making `axum` kind of redundant and `axum::middleware::*` really long for a heading.

davidpdrsn: I still think we should just keep it like it was 🤷

axum/src/docs/middleware.md
anuvratsingh: So I should add another commonly used middleware, and it would look like this:
- [`TraceLayer`](docs,rs link) for high level tracing/logging.
- [`CorsLayer`](docs,rs link) for handling CORS.
- [`CompressionLayer`](docs,rs link) for automatic compression of responses.
- [`RequsetIdLayer`](docs,rs link) set and propagate requests IDs.
- [`tower::TimeoutLayer`](docs,rs link) and [`tower_http::TimeoutLayer`](docs,rs link) for timeouts.

davidpdrsn: No need to mention `tower::TimeoutLayer`. The one in tower-http is what people should use.

And the links should to be to the modules, not the layers themselves. For example https://docs.rs/tower-http/latest/tower_http/trace/

anuvratsingh:
@davidpdrsn 
Looking at other links like `tower` and `tower-service` which either uses:
- Just link external crate's module like this [\`tower::ServiceBuilder\`] .
- Use reference like [\`ServiceBuilder::map_response\`] then provide link at the bottom like this 
[\`ServiceBuilder::map_response\`]: tower::ServiceBuilder::map_response.

As `tower`, `tower-service`, `tower-layer` etc already uses this, it should work for `tower-http` too.

Can I give this a try?

davidpdrsn:
Sure! If intra docs links can work that's certainly nicer.

jplatte:
I just tried, and can't reproduce the `docs.rs` problem locally at all, which makes it impossible (for me) to verify the fix. I also upgraded my Nightly toolchain from 2022-07-19 to 2022-07-27, but that didn't change anything.

Issue #1186, opened 2022-07-23
axum::extract::multipart::Field should implement Clone


Haider-Ali-Dev:
### Version
0.5.5

├── axum v0.5.5
│   ├── axum-core v0.2.4
│   ├── axum-core v0.2.4 (*)

### Platform

Linux asus 5.18.12-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 15 Jul 2022 15:33:02 +0000 x86_64 GNU/Linux

### Description

`axum::extract::multipart::Field` should implement `Clone` its really hard to if a `field.name()` equals to what we are looking for because it can't be cloned, all getter methods are `self` not `&self`. Implementing `Clone` will make it easy or else its really hard.
```rust
fn create_post(mut form: Multipart) -> Result<Json<String>, ApiError> {
  while let Some(field) = form.next_field().await.unwrap() {
      // &self - no problem
      match field.name() {
         // Did call field.text() which is &self
         "foo" => {..}
         ...
         _ => ()
      }
      // Checking if the content_type is image or video and then processes it.
      // Error here because field was deleted by text() method
      match field.content_type() { ... }
   }
}
```

Haider-Ali-Dev:
Just made a pull request to fix it

davidpdrsn:
`Field` cannot be `Clone` because it contains a `&mut Multipart`. That is required to uphold an invariant.

Think of `Field` as a "pointer into a the request body stream" meaning it doesn't make sense to clone since you cannot consume a stream twice. That is why the `Field::text` takes `self` and is `async`.

Seems to me that putting a `continue` in your loop should fix your problem.

PR #1187, opened 2022-07-23
Field can now be cloned

Haider-Ali-Dev:
#1186 
## Motivation
`axum::extract::multipart::Field` couldn't be cloned and it was hard to get the value of field because getter methods were on `self` not `&self`

## Solution
Just derived `Clone`

### DIFF
--- a/axum/src/extract/multipart.rs
+++ b/axum/src/extract/multipart.rs
@@ -87,7 +87,7 @@ impl Multipart {
 }
 
 /// A single field in a multipart stream.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Field<'a> {
     inner: multer::Field<'static>,
     // multer requires there to only be one live `multer::Field` at any point. This enforces that


davidpdrsn (changes requested):
I don't think that's correct. How do you wanna clone the `&'a mut Multipart`?

davidpdrsn:
Also, please at least run `cargo check` before submitting a PR. Not doing so is disrespectful to the maintainers time.

Haider-Ali-Dev:
> Also, please at least run `cargo check` before submitting a PR. Not doing so is disrespectful to the maintainers time.

I’m sorry, I’ll fix it. I ran `cargo test` Everything passed so I thought it’s working

davidpdrsn:
You need to run `cargo test --all-features` since multipart is behind a non-default feature.

See my comment [here](https://github.com/tokio-rs/axum/issues/1186). Not sure it can work and not sure it's the right solution to your problem.

Haider-Ali-Dev:
@davidpdrsn Yea there can’t be two mutable references but I’m trying to add some interior mutability maybe that’ll help, I’m trying different methods and making sure it doesn’t slow Axum

davidpdrsn:
I don't think we want to add any interior mutability here. Also not sure how that would help. We don't have any methods that take `&mut self`. 

Try the other solution I mentioned in the issues. I don't think we need to change anything in axum.

PR #1189, opened 2022-07-23
Add an example for head request

### LINKED ISSUE #1175: Add an example for a GET-or-HEAD handler that special-cases HEAD
Registering a `GET` route also registers a `HEAD` route that calls the same handler. axum will strip the response body if this handler is called with `HEAD`, but that is potentially much slower than never generating a response body in the first place. We should have an example with a handler that uses `http::Method` as an extractor and checks its value to decide whether to build a response with or without a body.

lexcao:
## Motivation
As #84 add get-head request, adding an example for a GET-or-HEAD handler that special-cases HEAD.
Closes #1175 

## Solution
Add an `handle-head-request` example.

### DIFF
--- a/examples/handle-head-request/src/main.rs
+++ b/examples/handle-head-request/src/main.rs
@@ -1,0 +1,82 @@
+//! Run with
+//!
+//! ```not_rust
+//! cd examples && cargo run -p example-handle-head-request
+//! ```
+
+use axum::response::{IntoResponse, Response};
+use axum::{http, routing::get, Router};
+use std::net::SocketAddr;
+
+fn app() -> Router {
+    Router::new().route("/get-head", get(get_head_handler))
+}
+
+#[tokio::main]
+async fn main() {
+    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
+
+    axum::Server::bind(&addr)
+        .serve(app().into_make_service())
+        .await
+        .unwrap();
+}
+
+// GET routes will also be called for HEAD requests but will have the response body removed.
+// You can handle the HEAD method explicitly by extracting `http::Method` from the request.
+async fn get_head_handler(method: http::Method) -> Response {
+    // it usually only makes sense to special-case HEAD
+    // if computing the body has some relevant cost
+    if method == http::Method::HEAD {
+        return ([("x-some-header", "header from HEAD")]).into_response();
+    }
+
+    // then do some computing task in GET
+    do_some_computing_task();
+
+    ([("x-some-header", "header from GET")], "body from GET").into_response()
+}
+
+fn do_some_computing_task() {
+    // TODO
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use axum::body::Body;
+    use axum::http::{Request, StatusCode};
+    use tower::ServiceExt;
+
+    #[tokio::test]
+    async fn test_get() {
+        let app = app();
+
+        let response = app
+            .oneshot(Request::get("/get-head").body(Body::empty()).unwrap())
+            .await
+            .unwrap();
+
+        assert_eq!(response.status(), StatusCode::OK);
+        assert_eq!(response.headers()["x-some-header"], "header from GET");
+
+        let body = hyper::body::to_bytes(response.into_body()).await.unwrap();
+        assert_eq!(&body[..], b"body from GET");
+    }
+
+    #[tokio::test]
+    async fn test_implicit_head() {
+        let app = app();
+
+        let response = app
+            .oneshot(Request::head("/get-head").body(Body::empty()).unwrap())
+            .await
+            .unwrap();
+
+        assert_eq!(response.status(), StatusCode::OK);
+        assert_eq!(response.headers()["x-some-header"], "header from HEAD");
+
+        let body = hyper::body::to_bytes(response.into_body()).await.unwrap();
+        assert!(body.is_empty());
+    }
+}
--- a/examples/handle-head-request/Cargo.toml
+++ b/examples/handle-head-request/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "example-handle-head-request"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+axum = { path = "../../axum" }
+tokio = { version = "1.0", features = ["full"] }
+
+[dev-dependencies]
+hyper = { version = "0.14", features = ["full"] }
+tower = { version = "0.4", features = ["util"] }


examples/handle-head-request/src/main.rs
```diff
@@ -0,0 +1,99 @@
...
+    if method == http::Method::HEAD {
+        return ([("x-some-header", "header from HEAD")], "");
+    }
+
+    return ([("x-some-header", "header from GET")], "body from GET");
+}
+
+async fn explicit_handler() -> impl IntoResponse {
+    http::StatusCode::OK
+}
```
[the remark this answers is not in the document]
lexcao: Hi, this shows the ability of the explicit head handler, which use in 
```
route("/explicit", head(explicit_handler))
```

If it does not make sense, I could remove them.

lexcao:
Hi, @davidpdrsn 
Thanks for your suggestion. It's helpful for me.

examples/handle-head-request/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+
+// get routes will also be called for HEAD requests but will have the response body removed
+// can handle implicit head method request by extract `http::Method` from request
+async fn get_head_handler(method: http::Method) -> Response {
+    if method == http::Method::HEAD {
+        return ([("x-some-header", "header from HEAD")]).into_response();
+    }
+
+    ([("x-some-header", "header from GET")], "body from GET").into_response()
```
jplatte, line 32: Can we have something a little more elaborate here? Or at least a comment above this line that you would usually only special-case `HEAD` if computing the body has some relevant compute cost?

examples/handle-head-request/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+    use tower::ServiceExt;
+
+    #[tokio::test]
+    async fn test_get() {
+        let app = app();
+
+        let response = app
+            .oneshot(
+                Request::builder()
+                    .uri("/get-head")
```
jplatte, line 49:
```suggestion
                Request::get("/get-head")
```

examples/handle-head-request/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    #[tokio::test]
+    async fn test_implicit_head() {
+        let app = app();
+
+        let response = app
+            .oneshot(
+                Request::builder()
+                    .uri("/get-head")
```
jplatte, line 70:
```suggestion
                Request::head("/get-head")
```

examples/handle-head-request/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+                    .method(http::Method::HEAD)
```
jplatte, line 71:
```suggestion
                Request::head("/get-head")
```

lexcao:
Hi, @jplatte 
Thanks for your suggestion.
I have added a case.

examples/handle-head-request/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+        .await
+        .unwrap();
+}
+
+// get routes will also be called for HEAD requests but will have the response body removed
+// can handle implicit head method request by extract `http::Method` from request
+async fn get_head_handler(method: http::Method) -> Response {
+    let something = Some(Thing);
+
+    // it would usually handle only special-case HEAD
```
jplatte, line 30:
```suggestion
    // it usually makes sense to special-case HEAD
```

examples/handle-head-request/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+    // if computing the body has some relevant compute cost
```
jplatte, line 31: Double 'compute' sounds a bit weird

```suggestion
    // if computing the body has some relevant cost
```

examples/handle-head-request/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+    if method == http::Method::HEAD {
+        // it could do lightweight check in HEAD
+        return if something.is_some() {
+            ([("x-some-header", "header from HEAD")]).into_response()
+        } else {
+            (http::StatusCode::NOT_FOUND).into_response()
+        };
```
jplatte, line 38: I don't really understand what this is about.

examples/handle-head-request/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    // then do some computing task in GET
+    if let Some(thing) = something {
+        thing.do_some_computing_task();
+    }
```
jplatte, line 44: Maybe just `Thing.do_some_computing_task();`?

lexcao:
Hi, 
I have fixed some wording. 
Thank you!

examples/handle-head-request/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+    axum::Server::bind(&addr)
+        .serve(app().into_make_service())
+        .await
+        .unwrap();
+}
+
+// get routes will also be called for HEAD requests but will have the response body removed
+// can handle implicit head method request by extract `http::Method` from request
+async fn get_head_handler(method: http::Method) -> Response {
+    // it usually makes sense to handle only special-case HEAD
```
jplatte, line 28: Wording can still be improved a bit, I think I suggested something similar before?

```suggestion
    // it usually only makes sense to special-case HEAD
```

examples/handle-head-request/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
+
+    axum::Server::bind(&addr)
+        .serve(app().into_make_service())
+        .await
+        .unwrap();
+}
+
+// get routes will also be called for HEAD requests but will have the response body removed
+// can handle implicit head method request by extract `http::Method` from request
```
jplatte, line 26:
```suggestion
// GET routes will also be called for HEAD requests but will have the response body. removed
// You can handle the HEAD method explicitly by extracting `http::Method` from the request.
```

jplatte (approved):
Other than some more wording suggestions, this looks very good now 👍🏼

examples/handle-head-request/src/main.rs
lexcao: Yes, thank you.
I just did not notice that.

davidpdrsn (approved):
Great! Thanks :D

davidpdrsn:
@lexcao it seems CI is failing because the dependencies aren't sorted.

lexcao:
Hi, @davidpdrsn I just fixed the sorting.
Thank you!

PR #1192, opened 2022-07-25
Improve build times by generating less IR

davidpdrsn:
I used [`cargo-llvm-lines`](https://crates.io/crates/cargo-llvm-lines) to find functions that generates lots of IR and tweaked them (mostly by making inner functions) to generate less. The biggest offender was `MethodRouter::merge`.

I tested on a small example with a couple of routes and some middleware and this shaved ~1 second of the clean debug build time (12 seconds to 11 seconds) and a 3% reduction in IR.

### DIFF
--- a/axum/src/routing/method_routing.rs
+++ b/axum/src/routing/method_routing.rs
@@ -899,29 +899,26 @@ impl<ReqBody, E> MethodRouter<ReqBody, E> {
         S::Response: IntoResponse + 'static,
         S::Future: Send + 'static,
     {
-        macro_rules! set_service {
-            (
-                $filter:ident,
-                $svc:ident,
-                $allow_header:ident,
-                [
-                    $(
-                        ($out:ident, $variant:ident, [$($method:literal),+])
-                    ),+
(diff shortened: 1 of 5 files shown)


axum/src/routing/method_routing.rs
```diff
@@ -939,22 +936,89 @@ impl<ReqBody, E> MethodRouter<ReqBody, E> {
             mut allow_header,
             _request_body: _,
         } = self;
-        let svc = Some(Route::new(svc));
-        set_service!(
+
+        let svc = Route::new(svc);
+
+        set_service(
```
jplatte, line 878: These were somewhat more readable before. Generic closure would be a great solution, but we don't have those yet of course :/

jplatte (approved):

axum/src/routing/method_routing.rs
davidpdrsn: Yeah I agree its a bit unfortunate but this method was in the top 3 so think its worth it.

Issue #1177, opened 2022-07-20
JSON performance in TechEmpower benchmarks


nuvacore:
The latest TechEmpower benchmarks have been released. Axum is doing extremely well and is in the top ten. Very well done!

https://www.techempower.com/benchmarks/#section=data-r21&test=composite 

What surprised me is that the JSON benchmark shows especially the JSON benchmark being way down, almost 50% of the top performer. I thought that there was little overhead in Rust for JSON. "may-minihttp", which is also Rust, is almost double of Axum's JSON performance.

What's the reason for this? Is this something that can be improved?

jplatte:
axum's JSON support uses serde_json. I can't think of anything that would make it particularily fast, or slow, compared to others like actix. Do you know what exactly is being benchmarked there for the JSON column?

jplatte:
Sooo I actually had an idea, looked into it, here's a PR that should improve performance: #1178
Don't have time to look into benchmarking right now, though at worst this should have the same performance as before.

ibraheemdev:
Most other rust frameworks use simd-json, and some even re-use a byte buffer per thread, which is probably the difference.

davidpdrsn:
Which frameworks? [simd-json's dependents](https://crates.io/crates/simd-json/reverse_dependencies) don't show much.

actix-web uses serde-json but reads into a buffer [with some reserved capacity](https://docs.rs/actix-web/4.1.0/src/actix_web/types/json.rs.html#140-159). We could do that as well.

jplatte:
Maybe before we tweak our JSON (de)serialization code more, we should start some benchmarking? I'm curious whether my change helped / how much :)

davidpdrsn:
That is a good idea 😅

davidpdrsn:
Generally I don't put much care into micro benchmarks but small tweaks that don't impact the user experience are fine imo.

ibraheemdev:
@davidpdrsn You have to look at the benchmark code not the frameworks default json responder 🙃. Looking again though, hyper just uses serde-json and performs significantly better than axum. It also does much better on plaintext, and is only worse on fortunes but is using an out of date postgres dependency.... which suggests to me the difference is elsewhere.

davidpdrsn:
> You have to look at the benchmark code not the frameworks default json responder

🤦

> Looking again though, hyper just uses serde-json and performs significantly better than axum

Uh that sounds very weird 🤔

[...]

PR #1196, opened 2022-07-26
Create BytesMut for Json with initial capacity

ryanfowler:
This PR initializes the `BytesMut` for serializing `Json` with a small initial capacity to reduce allocations as the inner `Vec<u8>` grows.

A [recent PR](https://github.com/tokio-rs/axum/pull/1178) removed an unnecessary copy by introducing `BytesMut` and switching away from `serde_json::to_vec`, but I noticed a small regression that I believe is due to `serde_json::to_vec` creating a `Vec<u8>` with an initial capacity ([code](https://github.com/serde-rs/json/blob/v1.0.82/src/ser.rs#L2189)). I figured matching the initial capacity of "128 bytes" made sense. In a simple benchmark like below (mostly pulled from [this comment](https://github.com/tokio-rs/axum/issues/1177#issuecomment-1194007420)), I'm seeing a modest ~1% speed bump.

```rust
use axum::{routing::get, Json, Router};

#[tokio::main]
async fn main() {
    let app = Router::new().route("/json", get(json));

    axum::Server::bind(&"0.0.0.0:8000".parse().unwrap())
        .await
        .unwrap();
}

async fn json() -> Json<Message> {
    let message = Message {
        name: "http-benchmark",
        version: "v0.1.0",
        message: "Hello, World!",
    };
    Json(message)
}

#[derive(serde::Serialize)]
pub struct Message {
    pub name: &'static str,
    pub version: &'static str,
    pub message: &'static str,
}
```

### DIFF
--- a/axum/src/json.rs
+++ b/axum/src/json.rs
@@ -186,7 +186,7 @@ where
     T: Serialize,
 {
     fn into_response(self) -> Response {
-        let mut buf = BytesMut::new().writer();
+        let mut buf = BytesMut::with_capacity(128).writer();
         match serde_json::to_writer(&mut buf, &self.0) {
             Ok(()) => (
                 [(
--- a/axum/src/json.rs
+++ b/axum/src/json.rs
@@ -187,7 +187,7 @@ where
 {
     fn into_response(self) -> Response {
         // Use a small initial capacity of 128 bytes like serde_json::to_vec
-        // https://github.com/serde-rs/json/blob/v1.0.82/src/ser.rs#L2189
+        // https://docs.rs/serde_json/1.0.82/src/serde_json/ser.rs.html#2189
         let mut buf = BytesMut::with_capacity(128).writer();
         match serde_json::to_writer(&mut buf, &self.0) {
             Ok(()) => (
--- a/axum/src/json.rs
+++ b/axum/src/json.rs
@@ -186,6 +186,8 @@ where
     T: Serialize,
 {
     fn into_response(self) -> Response {
+        // Use a small initial capacity of 128 bytes like serde_json::to_vec
+        // https://github.com/serde-rs/json/blob/v1.0.82/src/ser.rs#L2189
         let mut buf = BytesMut::with_capacity(128).writer();
         match serde_json::to_writer(&mut buf, &self.0) {
             Ok(()) => (


axum/src/json.rs
```diff
@@ -186,7 +186,7 @@ where
     T: Serialize,
 {
     fn into_response(self) -> Response {
-        let mut buf = BytesMut::new().writer();
+        let mut buf = BytesMut::with_capacity(128).writer();
```
davidpdrsn, line 189: Maybe add a comment about why `128` specifically. Can just link to the code in `serde_json`.

davidpdrsn (commented):
Makes sense!

[actix-web uses an initial capacity of 8192 bytes](https://docs.rs/actix-web/4.1.0/src/actix_web/types/json.rs.html#371). I'm wondering if we should benchmark a slightly bigger initial capacity as well 🤔

axum/src/json.rs
jplatte: Could also use a docs.rs link: https://docs.rs/serde_json/1.0.82/src/serde_json/ser.rs.html#2189
(could be longer-lived, but who knows)

ryanfowler:
> [actix-web uses an initial capacity of 8192 bytes](https://docs.rs/actix-web/4.1.0/src/actix_web/types/json.rs.html#371). I'm wondering if we should benchmark a slightly bigger initial capacity as well 🤔

Oh, interesting! I suppose it completely depends on the size of JSON being serialized. I would guess that JSON responses in real-world servers are usually >128 bytes...

axum/src/json.rs
ryanfowler: Done!

jplatte (approved):
Intuitively, 128 seems like a good initial capacity to me but 8k seems a little much.

davidpdrsn (approved):
Yep its cool with me as well. We can always tweak later.

ryanfowler:
Hmm it appears that the required checks for the `main` branch may need to be updated for [this](https://github.com/tokio-rs/axum/commit/3bcc8fb07d0abfb8548c701b7028a9f63b1cab9b) naming change made to CI

davidpdrsn:
Yeah I'll do that separately.

Thanks!

PR #1204, opened 2022-07-28
Make `IntoMakeService::new()` public

### LINKED ISSUE #1203: Make `IntoMakeService::new()` public
### Motivation

I would like to wrap a user supplied app, like a `Router`, in a global layer. For that purpose I stumbled upon ["Routing to services/middleware and backpressure"](https://docs.rs/axum/0.5.13/axum/middleware/index.html#routing-to-servicesmiddleware-and-backpressure), talking about [`ServiceBuilder`](https://docs.rs/tower/0.4.13/tower/struct.ServiceBuilder.html).

The issue is that to use it for example with `axum-server`, I need to turn it into a service factory, exactly like [`Router::into_make_service()`](https://docs.rs/axum/0.5.13/axum/routing/struct.Router.html#method.into_make_service), which isn't available for `ServiceBuilder`.

### Proposal

~~Add `ServiceBuilder::into_make_service()`, wrapping `ServiceBuilder` in [`IntoMakeService`](https://docs.rs/axum/0.5.13/axum/routing/struct.IntoMakeService.html).~~

I just realized that `ServiceBuilder` doesn't live in `axum`, so we can't actually add methods. So I changed my proposal to make `IntoMakeService::new()` public instead.

### Alternatives

I'm actually not sure at all if this the appropriate solution to begin with, maybe there is a better alternative?

~~I have also tried to roll my own `IntoMakeService`, because `IntoMakeService::new()` is private, which worked as expected and requires very little code.~~

Maybe a trait adding a `into_make_service()` method to appropriate types would be a better solution.

daxpedda:
See #1203 for more details, copied here for convenience.

>### Motivation
>
>I would like to wrap a user supplied app, like a `Router`, in a global layer. For that purpose I stumbled upon ["Routing to services/middleware and backpressure"](https://docs.rs/axum/0.5.13/axum/middleware/index.html#routing-to-servicesmiddleware-and-backpressure), talking about [`ServiceBuilder`](https://docs.rs/tower/0.4.13/tower/struct.ServiceBuilder.html).
>
>The issue is that to use it for example with `axum-server`, I need to turn it into a service factory, exactly like [`Router::into_make_service()`](https://docs.rs/axum/0.5.13/axum/routing/struct.Router.html#method.into_make_service), which isn't available for `ServiceBuilder`.
>
>### Proposal
>
>~~Add `ServiceBuilder::into_make_service()`, wrapping `ServiceBuilder` in [`IntoMakeService`](https://docs.rs/axum/0.5.13/axum/routing/struct.IntoMakeService.html).~~
>
>I just realized that `ServiceBuilder` doesn't live in `axum`, so we can't actually add methods. So I changed my proposal to make `IntoMakeService::new()` public instead.

Fixes #1203.

### DIFF
--- a/axum/src/routing/into_make_service.rs
+++ b/axum/src/routing/into_make_service.rs
@@ -14,7 +14,8 @@ pub struct IntoMakeService<S> {
 }
 
 impl<S> IntoMakeService<S> {
-    pub(crate) fn new(svc: S) -> Self {
+    /// Create a new `IntoMakeService`.
+    pub fn new(svc: S) -> Self {
         Self { svc }
     }
 }


davidpdrsn (changes requested):
Sorry but I don't think we should do this. See my comment [here](https://github.com/tokio-rs/axum/issues/1203).

PR #1207, opened 2022-07-28
Use v2 of the Cargo feature resolver and make CI work with it

jplatte:
Let's see whether this succeeds. Locally it seems to work. Split out from #1205.
The "Fix up Cargo.lock" step is only required as long as Cargo selects crc32fast 1.1.0, which I think has an invalid minimum version on its semver dependency (https://github.com/srijs/rust-crc32fast/issues/27).


jplatte:
Okay, looks like this can't be decoupled from #1205, closing.

PR #1209, opened 2022-07-29
JSON streaming support

abdolence:
## Motivation

Frequently there's a need for a stream of JSON objects to be sent over without allocating huge buffer server side. Reading them as a batch from some source such as database, file, etc and sending to the client side.

## Solution

This implementation gives a new type of body as `JsonStreamBody` that works with a standard `futures::Stream` of objects:

```rust
#[derive(Debug, Clone, Serialize)]
struct MyTestStructure {
    some_test_field: String
}

async fn test_json_stream() -> impl IntoResponse {
      let my_stream: BoxStream<'static, MyTestStructure> = ...
     JsonStreamBody::new(my_stream)
}
```

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -24,6 +24,7 @@ json-lines = ["serde_json", "serde", "tokio-util/io", "tokio-stream/io-util"]
 query = ["serde", "serde_html_form"]
 spa = ["tower-http/fs"]
 typed-routing = ["axum-macros", "serde", "percent-encoding"]
+json-stream-body = ["serde", "serde_json", "tokio-stream"]
 
 [dependencies]
 axum = { path = "../axum", version = "0.5", default-features = false }
--- a/axum-extra/src/body/json_stream_body.rs
+++ b/axum-extra/src/body/json_stream_body.rs
@@ -1,0 +1,192 @@
+use axum::body::HttpBody;
+use axum::response::{IntoResponse, Response};
+use futures_util::stream::BoxStream;
+use futures_util::StreamExt;
+use http::HeaderMap;
+use serde::Serialize;
+use std::fmt::Formatter;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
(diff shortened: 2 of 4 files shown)


davidpdrsn:
This is pretty much what [`JsonLines`](https://docs.rs/axum-extra/latest/axum_extra/json_lines/struct.JsonLines.html) in axum-extra does except it streams the json values separated by new lines and not as a single array which seems to be what you do.

Since we have `JsonLines` I'm not sure we need as well. Wdyt?

abdolence:
I didn't know that `JsonLines` supported Stream, probably didn't read docs well.
However, it seems it supports only JSON per line format which is not possible to read as common json when needed.

I'm thinking maybe I could convert this solution to a more general approach to be able to serialize a different kind of JSON Streaming (separator or not) or even maybe some like CSV-like protocols?

davidpdrsn:
> I didn't know that JsonLines supported Stream

The first line of the docs is "A stream of newline delimited JSON." 😅

> However, it seems it supports only JSON per line format which is not possible to read as common json when needed.

I think newline delimited JSON is fairly common for large payloads. It also allows the client to save memory since they don't to buffer everything.

> I'm thinking maybe I could convert this solution to a more general approach to be able to serialize a different kind of JSON Streaming (separator or not) or even maybe some like CSV-like protocols?

That sounds like it would fit well in a stand alone crate 😊

I have never seen an api that used csv instead of json so not sure how useful that is.

abdolence:
> I have never seen an api that used csv instead of json so not sure how useful that is.

CSV is still the most popular format in ML world :) especially datasets. However I see your point, so I'll keep it separately from this crate.

davidpdrsn:
Hehe I'm not surprised. Personally I think CSV is a really bad format and I don't understand why people use it. But I've done any ML 🤷

If you do end up writing a crate please add it to ECOSYSTEM.md 😀

abdolence:
Just FYI: I ended up with this for now:
https://github.com/abdolence/axum-streams-rs

PR #1212, opened 2022-07-30
Add svelte-axum-project into ECOSYSTEM.md

jbertovic:
Enjoy Axum - great work!

Adding link to a starter template in combining a frontend like Svelte with an Axum backend.  Shows how to use authentication and other middleware.  Not sure if it should be in one of the other categories instead like "Project showcase" or "Tutorials".


davidpdrsn (approved):
Thanks!

PR #1214, opened 2022-08-01
feat(axum-extra): Add tokio-rs/prost Protocol Buffer extractor and response

diceride:
Allows users to send and receive [tokio-rs/prost](https://github.com/tokio-rs/prost) Protocol Buffers.<br>This is an optional feature and can be enabled using the `prost` feature flag.

Example - Decoding a Protocol Buffer

```toml
[dependencies]
axum = { version = "...", features = ["prost"] }
```

```rs
use axum::{..., extract::Prost};

#[derive(prost::Message)]
struct Input {
  #[prost(string, tag="1")]
  foo: prost::alloc::string::String,
}

async fn create_user(Prost(input): Prost<Input>) {
  // input is `Input`
}
```

The default Content-Type header is `application/octet-stream`, a MIME type that is used for unknown binary files.

A good summary: [What is the correct Protobuf content type?](https://stackoverflow.com/questions/30505408/what-is-the-correct-protobuf-content-type)

https://stackoverflow.com/questions/30505408/what-is-the-correct-protobuf-content-type

> Content types should map to encoding schemes, and there are multiple encoding schemes for Protocol Buffers. The wire format is the most common and important one, but there's also [text format](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.text_format), and potentially others.
>
> Protocol Buffers are just a way of describing schema, and are not tightly coupled with any particular way of encoding data in that schema.

To anticipate a discussion about the “Content-Type ”, I took an *agnostic* approach. 

The default Content-Type header is `application/octet-stream` and error messages are phrased in general terms. The underlying [tokio-rs/prost](https://github.com/tokio-rs/prost) Protocol Buffers implementation is never exposed to the client.

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -28,6 +28,7 @@ json-lines = [
     "tokio-util?/io",
     "tokio-stream?/io-util"
 ]
+protobuf = ["dep:prost"]
 query = ["dep:serde", "dep:serde_html_form"]
 spa = ["tower-http/fs"]
 typed-routing = ["dep:axum-macros", "dep:serde", "dep:percent-encoding"]
@@ -49,6 +50,7 @@ tower-service = "0.3"
 axum-macros = { path = "../axum-macros", version = "0.2.2", optional = true }
 cookie = { package = "cookie", version = "0.16", features = ["percent-encode"], optional = true }
 percent-encoding = { version = "2.1", optional = true }
+prost = { version = "0.11", optional = true }
 serde = { version = "1.0", optional = true }
 serde_html_form = { version = "0.1", optional = true }
 serde_json = { version = "1.0.71", optional = true }
--- a/axum-extra/src/lib.rs
+++ b/axum-extra/src/lib.rs
@@ -78,6 +78,9 @@ pub mod routing;
 #[cfg(feature = "json-lines")]
 pub mod json_lines;
 
+#[cfg(feature = "protobuf")]
+pub mod protobuf;
+
 /// Combines two extractors or responses into a single type.
 #[derive(Debug, Copy, Clone)]
 pub enum Either<L, R> {
--- a/axum-extra/src/protobuf.rs
+++ b/axum-extra/src/protobuf.rs
@@ -1,0 +1,319 @@
+//! Protocol Buffer extractor and response.
+
+use axum::{
+    async_trait,
+    body::{Bytes, HttpBody},
+    extract::{rejection::BytesRejection, FromRequest, RequestParts},
+    response::{IntoResponse, Response},
+    BoxError,
+};
+use bytes::BytesMut;
+use http::StatusCode;
+use prost::Message;
+use std::ops::{Deref, DerefMut};
+
+/// A Protocol Buffer message extractor and response.
+///
+/// This can be used both as an extractor and as a response.
+///
+/// # As extractor
+///
+/// When used as an extractor, it can decode request bodies into some type that
+/// implements [`prost::Message`]. The request will be rejected (and a [`ProtoBufRejection`] will
+/// be returned) if:
+///
+/// - The body couldn't be decoded into the target Protocol Buffer message type.
+/// - Buffering the request body fails.
+///
+/// See [`ProtoBufRejection`] for more details.
+///
+/// The extractor does not expect a `Content-Type` header to be present in the request.
+///
+/// # Extractor example
+///
+/// ```rust,no_run
+/// use axum::{routing::post, Router};
+/// use axum_extra::protobuf::ProtoBuf;
+///
+/// #[derive(prost::Message)]
+/// struct CreateUser {
+///     #[prost(string, tag="1")]
+///     email: String,
+///     #[prost(string, tag="2")]
+///     password: String,
+/// }
+///
+/// async fn create_user(ProtoBuf(payload): ProtoBuf<CreateUser>) {
+///     // payload is `CreateUser`
+/// }
+///
+/// let app = Router::new().route("/users", post(create_user));
+/// # async {
+/// # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
+/// # };
+/// ```
+///
+/// # As response
+///
+/// When used as a response, it can encode any type that implements [`prost::Message`] to
+/// a newly allocated buffer.
+///
+/// If no `Content-Type` header is set, the `Content-Type: application/octet-stream` header
+/// will be used automatically.
+///
+/// # Response example
+///
+/// ```
+/// use axum::{
+///     extract::Path,
+///     routing::get,
+///     Router,
+/// };
+/// use axum_extra::protobuf::ProtoBuf;
+///
+/// #[derive(prost::Message)]
+/// struct User {
+///     #[prost(string, tag="1")]
+///     username: String,
+/// }
+///
+/// async fn get_user(Path(user_id) : Path<String>) -> ProtoBuf<User> {
+///     let user = find_user(user_id).await;
+///     ProtoBuf(user)
+/// }
+///
+/// async fn find_user(user_id: String) -> User {
+///     // ...
+///     # unimplemented!()
+/// }
+///
+/// let app = Router::new().route("/users/:id", get(get_user));
+/// # async {
+/// # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
+/// # };
+/// ```
+#[derive(Debug, Clone, Copy, Default)]
+#[cfg_attr(docsrs, doc(cfg(feature = "protobuf")))]
+pub struct ProtoBuf<T>(pub T);
+
+#[async_trait]
+impl<T, B> FromRequest<B> for ProtoBuf<T>
+where
+    T: Message + Default,
+    B: HttpBody + Send,
+    B::Data: Send,
+    B::Error: Into<BoxError>,
+{
+    type Rejection = ProtoBufRejection;
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
+        let mut bytes = Bytes::from_request(req).await?;
+
+        match T::decode(&mut bytes) {
+            Ok(value) => Ok(ProtoBuf(value)),
+            Err(err) => Err(ProtoBufDecodeError::from_err(err).into()),
+        }
+    }
+}
+
+impl<T> Deref for ProtoBuf<T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl<T> DerefMut for ProtoBuf<T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+impl<T> From<T> for ProtoBuf<T> {
+    fn from(inner: T) -> Self {
+        Self(inner)
+    }
+}
+
+impl<T> IntoResponse for ProtoBuf<T>
+where
+    T: Message + Default,
+{
+    fn into_response(self) -> Response {
+        let mut buf = BytesMut::with_capacity(128);
+        match &self.0.encode(&mut buf) {
+            Ok(()) => buf.into_response(),
+            Err(err) => (StatusCode::INTERNAL_SERVER_ERROR, err.to_string()).into_response(),
+        }
+    }
+}
+
+/// Rejection type for [`ProtoBuf`].
+///
+/// This rejection is used if the request body couldn't be decoded into the target type.
+#[derive(Debug)]
+pub struct ProtoBufDecodeError(pub(crate) axum::Error);
+
+impl ProtoBufDecodeError {
+    pub(crate) fn from_err<E>(err: E) -> Self
+    where
+        E: Into<axum::BoxError>,
+    {
+        Self(axum::Error::new(err))
+    }
+}
+
+impl std::fmt::Display for ProtoBufDecodeError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Failed to decode the body: {:?}", self.0)
+    }
+}
+
+impl std::error::Error for ProtoBufDecodeError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        Some(&self.0)
+    }
+}
+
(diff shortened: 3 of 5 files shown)


diceride:
Yes, I thought about the same thing. And I put it into axum on purpose. 

[tokio-rs/prost](https://github.com/tokio-rs/prost) is not some third party dependency. It belongs to the Tokio universe. It is a premier club member. A project that is actively endorsed and supported by the same Tokio community, that provides [tokio-rs/axum](https://github.com/tokio-rs/axum). In my opinion, Prost deserves a chance to stand next to `Json` and `Form`, because of this very special *Tokio* connection.

diceride:
Yes, I thought about the same thing. And I put it into axum on purpose.

[tokio-rs/prost](https://github.com/tokio-rs/prost) is not some third party dependency. It belongs to the Tokio universe. It is a premier club member. A project that is actively endorsed and supported by the same Tokio community, that provides [tokio-rs/axum](https://github.com/tokio-rs/axum). In my opinion, Prost deserves a chance to stand next to Json and Form, because of this very special Tokio connection. Let me know what you think.

diceride:
Yes, I thought about the same thing. And I put it into axum on purpose.

[tokio-rs/prost](https://github.com/tokio-rs/prost) is not some third party dependency. It belongs to the Tokio universe. It is a premier club member. A project that is actively endorsed and supported by the same Tokio community, that provides [tokio-rs/axum](https://github.com/tokio-rs/axum). In my opinion, Prost deserves a chance to stand next to Json and Form, because of this very special _Tokio_ connection. Let me know what you think.

davidpdrsn:
I agree it makes sense in axum-extra. Been thinking that I'd like to expand it with other input/output formats.

> tokio-rs/prost is not some third party dependency. It belongs to the Tokio universe. It is a premier club member. A project that is actively endorsed and supported by the same Tokio community, that provides tokio-rs/axum. In my opinion, Prost deserves a chance to stand next to Json and Form, because of this very special Tokio connection. Let me know what you think

I disagree. I think what matters most is how common a format is and my gut feeling is that protobuf (without gRPC) is not common enough for warrant a spot in axum but it fits well in axum-extra imo.

davidpdrsn:
Re: content type is was thinking that since there doesn't seem to any consensus maybe we shouldn't require a specific one for the extractor version. For example actix-web's protobuf responder uses `application/protobuf`. So it would annoying to use with axum.

I haven't thought too much about the response content type but it's probably fine. Maybe we could compare with what other frameworks do?

I also think we should call the `ProtoBuf` since that's what the format is called, same reason that it's `Json` and not `SerdeJson` 😅

diceride:
* Moved the feature from axum -> axum-extra.
* Renamed `Prost` -> `ProtoBuf`.
* Added a new function `bytes()` to axum/test_client::TestResponse.

davidpdrsn:
@diceride what do you think about the content type stuff I wrote [here](https://github.com/tokio-rs/axum/pull/1214#issuecomment-1201626863)?

diceride:
@davidpdrsn Yes, I would like to avoid introducing something that might be seen as "opinionated" (like `application/protobuf` or `application/x-protobuffers` etc.). 
I mean, we could go with something like this

```rs
pub struct ProtoBuf<T>{
  message: pub T,
  content_type: Option<String>,
};
```

or even simpler

```rs
pub struct ProtoBuf<T>(pub T, Option<String>);
```

But this breaks with the general style of extractors and responses and it seems kinda over-engineered. Otherwise, I would stick with Google and keep it simple for now.

https://github.com/google/protorpc/blob/95849c9a3e414b9ba2d3da91fd850156348fe558/protorpc/protobuf.py#L50

davidpdrsn:
Why do you want the extractor to contain the content type?

I was suggesting the FromRequest implementation should decode the body, regardless of its content type.

diceride:
Ah now I see, that's a good idea. axum will set the Content-Type to `application/octet-stream` for `Vec<u8>` automatically.

diceride:
* Removed the Content-Type verification from the extractor.
* When used as a response, no Content-Type is set by default.

axum-extra/Cargo.toml
```diff
@@ -42,6 +42,7 @@ tower-service = "0.3"
 axum-macros = { path = "../axum-macros", version = "0.2.2", optional = true }
 cookie-lib = { package = "cookie", version = "0.16", features = ["percent-encode"], optional = true }
 percent-encoding = { version = "2.1", optional = true }
+prost = { version = "0.11.0", optional = true }
```
davidpdrsn, line 45: Thinking we should call the feature `protobuf` instead of `prost`.

axum-extra/src/protobuf.rs
```diff
@@ -0,0 +1,303 @@
...
+use prost::Message;
+use std::ops::{Deref, DerefMut};
+
+/// Protocol Buffer extractor and response.
+///
+/// When used as an extractor, it can decode request bodies into some type that
+/// implements [`prost::Message`]. The request will be rejected (and a [`ProtoBufRejection`] will
+/// be returned) if:
+///
+/// - The body couldn't be decoded into the target Protocol Buffer.
```
davidpdrsn, line 20:
```suggestion
/// - The body couldn't be decoded into the target Protocol Buffer message type.
```

axum-extra/src/protobuf.rs
```diff
@@ -0,0 +1,303 @@
...
+/// ```rust,no_run
+/// use axum::{routing::post, Router};
+/// use axum_extra::protobuf::ProtoBuf;
+///
+/// #[derive(prost::Message)]
+/// struct CreateUser {
+///     #[prost(string, tag="1")]
+///     email: prost::alloc::string::String,
+///     #[prost(string, tag="2")]
+///     password: prost::alloc::string::String,
```
davidpdrsn, line 36: I don't suppose the full qualification is necessary.

```suggestion
///     email: String,
///     #[prost(string, tag="2")]
///     password: String,
```

axum-extra/src/protobuf.rs
```diff
@@ -0,0 +1,303 @@
...
+};
+use http::StatusCode;
+use prost::Message;
+use std::ops::{Deref, DerefMut};
+
+/// Protocol Buffer extractor and response.
+///
+/// When used as an extractor, it can decode request bodies into some type that
+/// implements [`prost::Message`]. The request will be rejected (and a [`ProtoBufRejection`] will
+/// be returned) if:
```
davidpdrsn, line 18: I think we should put this into the extractor section below.

axum-extra/src/protobuf.rs
```diff
@@ -0,0 +1,303 @@
...
+///
+/// - The body couldn't be decoded into the target Protocol Buffer.
+/// - Buffering the request body fails.
+///
+/// See [`ProtoBufRejection`] for more details.
+///
+/// # Extractor example
```
davidpdrsn, line 25: For consistency with [`JsonLines`](https://docs.rs/axum-extra/latest/axum_extra/json_lines/struct.JsonLines.html)

```suggestion
/// # As extractor
```

axum-extra/src/protobuf.rs
```diff
@@ -0,0 +1,303 @@
...
+/// let app = Router::new().route("/users", post(create_user));
+/// # async {
+/// # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
+/// # };
+/// ```
+///
+/// When used as a response, it can encode any type that implements [`prost::Message`] to
+/// a newly allocated buffer, and will automatically set `Content-Type: application/octet-stream` header.
+///
+/// # Response example
```
davidpdrsn, line 52:
```suggestion
/// # As response
```

axum-extra/src/protobuf.rs
```diff
@@ -0,0 +1,303 @@
...
+/// }
+///
+/// let app = Router::new().route("/users", post(create_user));
+/// # async {
+/// # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
+/// # };
+/// ```
+///
+/// When used as a response, it can encode any type that implements [`prost::Message`] to
+/// a newly allocated buffer, and will automatically set `Content-Type: application/octet-stream` header.
```
davidpdrsn, line 50: Like the extractor: I think we should put this paragraph into the response section below.

axum-extra/src/protobuf.rs
davidpdrsn: Probably also worth documenting that the extractor doesn't care about the `content-type`.

axum-extra/src/protobuf.rs
```diff
@@ -0,0 +1,303 @@
...
+///     extract::Path,
+///     routing::get,
+///     Router,
+/// };
+/// use axum_extra::protobuf::ProtoBuf;
+///
+/// #[derive(prost::Message)]
+/// struct User {
+///     #[prost(string, tag="1")]
+///     username: prost::alloc::string::String,
```
davidpdrsn, line 65:
```suggestion
///     username: String,
```

axum-extra/src/protobuf.rs
```diff
@@ -0,0 +1,303 @@
...
+        Self(inner)
+    }
+}
+
+impl<T> IntoResponse for ProtoBuf<T>
+where
+    T: Message + Default,
+{
+    fn into_response(self) -> Response {
+        self.0.encode_to_vec().into_response()
```
davidpdrsn, line 132: I think using `Message::encode` with a `BytesMut` will be faster since `Vec` bodies will be converted to a `Bytes`. Should probably also set an initial capacity on the buffer [like we do for `Json`](https://github.com/tokio-rs/axum/blob/main/axum/src/json.rs#L191).

axum-extra/src/protobuf.rs
```diff
@@ -0,0 +1,303 @@
...
+
+impl std::error::Error for ProtoBufDecodeError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        Some(&self.0)
+    }
+}
+
+impl IntoResponse for ProtoBufDecodeError {
+    fn into_response(self) -> Response {
+        (StatusCode::UNPROCESSABLE_ENTITY).into_response()
```
davidpdrsn, line 165:
```suggestion
        StatusCode::UNPROCESSABLE_ENTITY.into_response()
```

davidpdrsn (changes requested):
Great work! I mostly had minor comments and suggestions.

axum-extra/Cargo.toml
jplatte: @davidpdrsn By the way, I think we should start using namespaced features so we don't have a bunch of extra features like `axum-macros` and `cookie-libs`. Would you mind us bumping the MSRV for that?

davidpdrsn: Sure!

axum-extra/src/protobuf.rs
diceride: Oh, I did not know that. Bueno 🤌.

axum-extra/src/protobuf.rs
diceride: I updated the documentation and adapted the documentation style from the [JsonLines](https://docs.rs/axum-extra/latest/axum_extra/json_lines/struct.JsonLines.html) extractor.

axum-extra/Cargo.toml
diceride: We should not rename the feature from `prost` to `protobuf`. This will confuse users.

Actually, there is a competing Protocol Buffer implementation for Rust provided by the [`grpc`](https://crates.io/crates/grpc) crate which is called [`grpc-protobuf`](https://crates.io/crates/grpc) (please notice, that the grpc crate is not an official implementation maintained by the gRPC group).

I once introduced support for both Protocol Buffer implementations in the [`opentelemetry-otlp`](https://docs.rs/crate/opentelemetry-otlp) crate. Users can choose their gRPC backend by using the feature flags`tonic` and`prost` or `grpc-sys` (for `grpc-protobuf`).

davidpdrsn: I mean there are other serialization frameworks than serde, but the feature is still called `json`.

I say we rename it.

axum-extra/src/protobuf.rs
```diff
@@ -0,0 +1,330 @@
...
+    fn into_response(self) -> Response {
+        let mut buf = BytesMut::with_capacity(128);
+        match &self.0.encode(&mut buf) {
+            Ok(()) => buf.freeze().into_response(),
+            Err(err) => (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                [(
+                    header::CONTENT_TYPE,
+                    HeaderValue::from_static(mime::TEXT_PLAIN_UTF_8.as_ref()),
+                )],
```
davidpdrsn, line 155: `IntoResponse` for `String` already sets a text/plain content-type. So don't need to do that here.

axum-extra/src/protobuf.rs
```diff
@@ -0,0 +1,330 @@
...
+}
+
+impl<T> IntoResponse for ProtoBuf<T>
+where
+    T: Message + Default,
+{
+    fn into_response(self) -> Response {
+        let mut buf = BytesMut::with_capacity(128);
+        match &self.0.encode(&mut buf) {
+            Ok(()) => buf.freeze().into_response(),
```
davidpdrsn, line 149: This should work too.

```suggestion
            Ok(()) => buf.into_response(),
```

diceride:
I applied the changes as requested. The feature can be enabled by using the `protobuf` feature flag.

axum-extra/Cargo.toml
```diff
@@ -28,6 +28,7 @@ json-lines = [
     "tokio-util?/io",
     "tokio-stream?/io-util"
 ]
+protobuf = ["prost"]
```
davidpdrsn, line 31:
```suggestion
protobuf = ["dep:prost"]
```

axum-extra/src/lib.rs
```diff
@@ -16,6 +16,7 @@
 //! `erased-json` | Enables the `ErasedJson` response | No
 //! `form` | Enables the `Form` extractor | No
 //! `json-lines` | Enables the `json-lines` extractor and response | No
+//! `prost` | Enables the `ProtoBuf` Protocol Buffer extractor and response | No
```
davidpdrsn, line 19:
```suggestion
//! `prost` | Enables the `ProtoBuf` extractor and response | No
```

davidpdrsn (approved):
Thanks for working on this!

PR #1218, opened 2022-08-02
Add exhaustive flag.

BratSinot:
## Motivation

Motivation is simple -- I want to handle all errors and don't ignore new one.
In my opinion, if someone want `non_exhaustive` he / she should use `if-then-else` chain. `match` is supposed to covering all options.

## Solution

By adding ability to suppress `non_exhaustive` for pattern matching.

### DIFF
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -24,6 +24,7 @@ original-uri = []
 query = ["serde_urlencoded"]
 tower-log = ["tower/log"]
 ws = ["tokio-tungstenite", "sha-1", "base64"]
+exhaustive = []
 
 # Required for intra-doc links to resolve correctly
 __private_docs = ["tower/full", "tower-http/full"]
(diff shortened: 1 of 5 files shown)


jplatte (commented):
This would have to be an unstable feature IMO, similar to some things in tokio that opt you into outside-semver breakage.

BratSinot:
> This would have to be an unstable feature IMO, similar to some things in tokio that opt you into outside-semver breakage.

For this reason, it disabled by default. In other case, I can't be sure what everything work fine, because things can be added silently even in minor version update.

davidpdrsn (changes requested):
I'm afraid we can't do this. Having a feature that changes the API probably isn't a good idea.

davidpdrsn:
A library that depends on axum can changes this for other crates that depend on that library.

I don't wanna do what tokio does for unstable features just for something like this.

jplatte:
`#[non_exhaustive]` exists so it's not a breaking change to add further variants to an enum.

PR #1229, opened 2022-08-08
Improve workspace caching on CI

maxcountryman:
This improves workspace caching performance by using the workspace parameter of `rust-cache@v2`.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,7 +26,10 @@ jobs:
         override: true
         profile: minimal
         components: clippy, rustfmt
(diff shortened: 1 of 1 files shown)


davidpdrsn:
How does it improve it? Where can we read about how v2 is different?

maxcountryman:
You might start with the action README.

For an example of this working in practice my crate, axum-sessions, saw builds go from a few minutes to just about a minute (examples were never cached previously; the same behavior is evident in axum's suite based on action logs).

davidpdrsn (approved):

davidpdrsn:
Thanks!

PR #1231, opened 2022-08-09
docs: remove double arc clone

MarkTuddenham:
## Motivation
Using closure captures for shared state example creates a local variable for the cloned arc, only to not use it.

## Solution
Remove the unnecessary variable.

### DIFF
--- a/axum/src/lib.rs
+++ b/axum/src/lib.rs
@@ -232,17 +232,11 @@
 //! let app = Router::new()
 //!     .route(
 //!         "/users",
-//!         post({
-//!             let shared_state = Arc::clone(&shared_state);
-//!             move |body| create_user(body, Arc::clone(&shared_state))
-//!         }),
+//!         post(move |body| create_user(body, Arc::clone(&shared_state))),
 //!     )
 //!     .route(
 //!         "/users/:id",
-//!         get({
-//!             let shared_state = Arc::clone(&shared_state);
-//!             move |path| get_user(path, Arc::clone(&shared_state))
-//!         }),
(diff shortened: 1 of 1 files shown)


davidpdrsn:
Are you sure? I get this error:

```
---- src/lib.rs - (line 223) stdout ----
error[E0382]: use of moved value: `shared_state`
  --> src/lib.rs:246:13
   |
17 | let shared_state = Arc::new(State { /* ... */ });
   |     ------------ move occurs because `shared_state` has type `Arc<State>`, which does not implement the `Copy` trait
...
22 |         post(move |body| create_user(body, Arc::clone(&shared_state))),
   |              ----------- value moved into closure here ------------ variable moved due to use in closure
...
26 |         get(move |path| get_user(path, Arc::clone(&shared_state))),
   |             ^^^^^^^^^^^ value used here after move ------------ use occurs due to use in closure

error: aborting due to previous error

For more information about this error, try `rustc --explain E0382`.
Couldn't compile the test.

failures:
    src/lib.rs - (line 223)
```

jplatte (commented):
This change isn't correct. I just checked locally and as I was expecting, the tests break when removing the double clone.

MarkTuddenham:
Serves me right for not running the tests, the second clone is still unnecessary though, so it should be:
```rust
     .route(
         "/users",
         post({
             let shared_state = Arc::clone(&shared_state);
             move |body| create_user(body, shared_state)
         }),
     )
```

davidpdrsn:
Feel free to submit a new PR, if the tests pass ;)

PR #1232, opened 2022-08-09
docs: use local variable in closure

MarkTuddenham:
## Motivation

The "Using closure captures for shared state" example creates a local variable for the cloned arc and then clones that clone.

The current example works since the clone in the closure is actually cloning the overshadowing variable, which has been moved into the closure.

## Solution

Since the overshadowing variable is moved into the closure, just use that in the function call.


davidpdrsn (approved):
Thanks!

PR #1233, opened 2022-08-09
Fix accidental breaking change to `FailedToDeserializeQueryString`

### LINKED ISSUE #1202: Ship new version of axum and axum-extra
We accidentally shipped https://github.com/tokio-rs/axum/pull/1171 in a minor version which included a breaking change in axum but didn't merge a new version of axum-extra so axum-extra broke.

I've yanked axum 0.5.14 so no one can update to it.

I'll ship a new version of axum and axum-extra (that depends on the version of axum) in a week from today.

davidpdrsn:
Fixes #1202

I'll update the changelogs, with notes about why 0.5.14 was yanked, in
the release PR.

### DIFF
--- a/axum-extra/src/extract/form.rs
+++ b/axum-extra/src/extract/form.rs
@@ -67,7 +67,7 @@ where
         if req.method() == Method::GET {
             let query = req.uri().query().unwrap_or_default();
             let value = serde_html_form::from_str(query)
-                .map_err(FailedToDeserializeQueryString::__private_new)?;
+                .map_err(FailedToDeserializeQueryString::__private_new::<(), _>)?;
             Ok(Form(value))
         } else {
             if !has_content_type(req, &mime::APPLICATION_WWW_FORM_URLENCODED) {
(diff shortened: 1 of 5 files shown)


jplatte (approved):

axum-extra/src/extract/form.rs
```diff
@@ -67,7 +67,7 @@ where
         if req.method() == Method::GET {
             let query = req.uri().query().unwrap_or_default();
             let value = serde_html_form::from_str(query)
-                .map_err(FailedToDeserializeQueryString::__private_new)?;
+                .map_err(FailedToDeserializeQueryString::__private_new::<(), _>)?;
```
jplatte, line 70: Actually, why is this `()` instead of T?

jplatte (changes requested):

axum-extra/src/extract/form.rs
jplatte: This means that assuming this is released, and the new axum-extra is used with an older version of axum, the message will still include "for type _", but `_` will be `()`.

I guess we can release the new axum 0.5.x release, then update axum-extra's dependency so it has a matching minimum version, and release that 🤔

jplatte (approved):

axum-extra/src/extract/form.rs
davidpdrsn: > I guess we can release the new axum 0.5.x release, then update axum-extra's dependency so it has a matching minimum version, and release that

Yeah that was my plan.

PR #1234, opened 2022-08-09
Replace async-graphql with link to async-graphql's own examples

davidpdrsn:
async-graphql has shipped a breaking version and rather than updating our examples I think we should just link to their examples.


.github/workflows/CI.yml
```diff
@@ -230,6 +230,9 @@ jobs:
     - name: Install cargo-sort
       run: |
         cargo install cargo-sort
+    - name: Remove non-crate folder
+      run: rm -rf async-graphql || true
+      working-directory: ${{ matrix.pwd }}
```
jplatte, line 235: Why is this needed?

.github/workflows/CI.yml
```diff
@@ -230,6 +230,9 @@ jobs:
...
     - name: Check dependecy tables
```
jplatte, line 236: Just noticed:

```suggestion
    - name: Check dependency tables
```

jplatte (approved):

examples/Cargo.toml
```diff
@@ -1,4 +1,4 @@
 [workspace]
 members = ["*"]
-exclude = ["target", "async-graphql"]
+exclude = ["async-graphql", "target"]
```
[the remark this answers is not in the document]
davidpdrsn: Yeah it really should have 😜 I might send them a PR

PR #1236, opened 2022-08-09
axum-extra: Version 0.3.7

davidpdrsn:


jplatte (approved):
I'm wondering whether this release really makes sense. I guess it could slightly improve compile times / decrease code size by always invoking the unnecessarily-generic rejection constructor with the same type parameter.

davidpdrsn:
Yeah I guess. I'd rather be on the safe side and it shouldn't hurt anyways.

PR #1239, opened 2022-08-09
Use new Cargo features to avoid implicit features for optional dependencies

jplatte:


axum-extra/Cargo.toml
```diff
@@ -14,16 +14,23 @@ version = "0.3.5"
...
 
-async-read-body = ["tokio-util/io"]
-cookie = ["cookie-lib"]
-cookie-private = ["cookie", "cookie-lib/private"]
-cookie-signed = ["cookie", "cookie-lib/signed"]
-erased-json = ["serde_json", "serde"]
-form = ["serde", "serde_html_form"]
-json-lines = ["serde_json", "serde", "tokio-util/io", "tokio-stream/io-util"]
-query = ["serde", "serde_html_form"]
+async-read-body = ["dep:tokio-util", "tokio-util?/io"]
```
davidpdrsn, line 17: What does the `?` do?

davidpdrsn (approved):
Great!

.github/workflows/CI.yml
```diff
@@ -2,7 +2,7 @@ name: CI
 
 env:
   CARGO_TERM_COLOR: always
-  MSRV: 1.56
+  MSRV: 1.60
```
davidpdrsn, line 5: Eh does github trim trailing zeros... sigh

image

Maybe this works

```suggestion
  MSRV: 1.60.0
```

axum-extra/Cargo.toml
jplatte: It enables that feature on the dependency only if that dependency is activated by something else. It's also necessary to refer to features of optional dependencies that are activated using `dep:` syntax (otherwise it would be unnecessary here).

davidpdrsn: Ah I see!

.github/workflows/CI.yml
jplatte: I think we just need to quote it so it's not interpreted as a number.

Issue #1238, opened 2022-08-09
Server sent events without data are ignored


Fxxxlei:
### Version

0.5.13-0.5.15

### Platform

all

### Description

Server-sent events without a data field are ignored and not dispatched in the browser.
This is documented in the specs (https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation step 2), but I couldn't find that info anywhere else.

Since axum::response::sse::Event doesn't prevent Events without a data field from being sent, I think this behaviour should be documented and/or prevented in axum.

davidpdrsn:
I just tested this and they not totally ignored. If you send `Event::default()` axum sends a `\n` in the body. I guess its just the browser that ignores that event.

I'm honestly not sure what the most correct thing to do is but I guess we could just ignore events if they're empty and not send anything in the body. What do you think?

Fxxxlei:
The specs say the browser must ignore the event, if it doesn't have a data field. It doesn't prohibit sending events that are ignored this way.

Question is if we want to allow sending events we know won't be dispatched in a compliant client.
On the other hand, I have a use cases for non-dispatched events (to check if the client is still connected).

I'd say the current behaviour is as intended but should be documented. I was sending named events without data and was wondering why they were not dispatched in the client.

davidpdrsn:
Sure. I'm fine with just documenting the current behavior. Do you wanna submit a PR?

davidpdrsn:
I've made a PR for this https://github.com/tokio-rs/axum/pull/1242

PR #1242, opened 2022-08-10
Document browsers ignoring SSE events without data fields

davidpdrsn:
From https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation:

> If the data buffer is an empty string, set the data buffer and the event type buffer to the empty string and return.

Fixes https://github.com/tokio-rs/axum/issues/1238

### DIFF
--- a/axum/src/response/sse.rs
+++ b/axum/src/response/sse.rs
@@ -204,8 +204,6 @@ impl Event {
     ///
     /// This corresponds to [`MessageEvent`'s data field].
     ///
-    /// Note that events with an empty data field will be ignored by the browser.
-    ///
     /// # Panics
     ///
     /// Panics if `data` or `json_data` have already been called.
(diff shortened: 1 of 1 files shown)


jplatte (approved):

axum/src/response/sse.rs
```diff
@@ -202,11 +204,13 @@ impl Event {
     ///
     /// This corresponds to [`MessageEvent`'s data field].
     ///
-    /// [`MessageEvent`'s data field]: https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent/data
+    /// Note that events with an empty data field will be ignored by the browser.
```
jplatte, line 207: Actually, is it possible to send an empty data field using `json_data`? Maybe this remark only makes sense on `data`.

Issue #1199, opened 2022-07-27
Add benchmarks


davidpdrsn:
The recent discussions around the [TechEmpower benchmarks](https://github.com/tokio-rs/axum/issues/1177) showed that, while axum is plenty fast, having some benchmarks in the repo would be useful:

- The purpose of the benchmarks should be to compare axum to itself such that one can easily see if some change improves performance.
- I don't they should do git shenanigans to automatically compare to a previous commit. Just print numbers for the current commit.
- They should not be designed to compare axum to other frameworks. There are other benchmarks that do that.
- They should be easy to run. Ideally just `cargo bench` if the output is good enough. Some bash script using `rewrk` is also fine with me.
- They should exercise things most users will use like routing and JSON input/output. Non axum specifics, like databases, aren't relevant imo.
- It's important that the benchmarks don't use unrealistic code. They should use the APIs we expect users to use.
- I don't necessarily think we should run the benchmarks on CI but should at least type check them to prevent bit rot.

I don't intend to chase micro benchmarks but having some convenient scripts to run does have value imo.

PR #1243, opened 2022-08-10
Add initial benchmarks

davidpdrsn:
This adds a basic setup for writing benchmarks and adds a couple of benchmarks.

The output looks like this

```
❯ cargo bench -p axum -- send-json
   Compiling axum v0.5.10
    Finished bench [optimized] target(s) in 0.68s
     Running unittests src/lib.rs (target/release/deps/axum-732f709dd139ce93)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 203 filtered out; finished in 0.00s

     Running benches/benches.rs (target/release/deps/benches-ddcccec176823e7f)
Running "send-json" benchmark
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:58714 for 10 second(s)
    Latencies:
      0.06ms   0.03ms   0.02ms   4.06ms
    Requests:
      Total: 1689726 Req/Sec: 168960.39
    Transfer:
      Total: 228.83 MB Transfer Rate: 22.88 MB/Sec
```

It basically runs a default hyper server and shells out to [rewrk](https://github.com/lnx-search/rewrk) which does all the heavy lifting.

I went with rewrk rather than the default cargo bench tooling (or criterion) because I'm mainly interested in seeing requests per second and not how long a single request takes. I would also like to avoid configuring and running our own client. This seems fine to be me but let me know what you think.

Fixes https://github.com/tokio-rs/axum/issues/1199

### DIFF
--- a/axum/benches/benches.rs
+++ b/axum/benches/benches.rs
@@ -1,0 +1,186 @@
+use axum::{
+    routing::{get, post},
+    Json, Router, Server,
+};
+use hyper::server::conn::AddrIncoming;
+use serde::{Deserialize, Serialize};
+use std::{
+    io::BufRead,
+    process::{Command, Stdio},
+};
+
+fn main() {
+    ensure_rewrk_is_installed();
+
+    benchmark("minimal").run(Router::new);
+
+    benchmark("basic").run(|| Router::new().route("/", get(|| async { "Hello, World!" })));
+
+    benchmark("routing").path("/foo/bar/baz").run(|| {
+        let mut app = Router::new();
+        for a in 0..10 {
+            for b in 0..10 {
+                for c in 0..10 {
+                    app = app.route(&format!("/foo-{}/bar-{}/baz-{}", a, b, c), get(|| async {}));
+                }
+            }
+        }
+        app.route("/foo/bar/baz", get(|| async {}))
+    });
+
+    benchmark("receive-json")
+        .method("post")
+        .headers(&[("content-type", "application/json")])
+        .body(r#"{"n": 123, "s": "hi there", "b": false}"#)
+        .run(|| Router::new().route("/", post(|_: Json<Payload>| async {})));
+
+    benchmark("send-json").run(|| {
+        Router::new().route(
+            "/",
+            get(|| async {
+                Json(Payload {
+                    n: 123,
+                    s: "hi there".to_owned(),
+                    b: false,
+                })
+            }),
+        )
+    });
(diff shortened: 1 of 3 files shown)


axum/benches/benches.rs
```diff
@@ -0,0 +1,186 @@
...
+    fn run<F>(self, f: F)
+    where
+        F: FnOnce() -> Router,
+    {
+        // support only running some benchmarks with
+        // ```
+        // cargo bench -- routing send-json
+        // ```
+        let args = std::env::args().collect::<Vec<_>>();
+        let names = &args[1..args.len() - 1];
```
jplatte, line 100: So wasteful to collect the commandline args once per call of `run` 😭
At least we can do slightly less work w/o complicating the code:

```suggestion
        let names = std::env::args().skip(1).collect::<Vec<_>>();
```

axum/benches/benches.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        let mut cmd = Command::new("rewrk");
+        cmd.stdout(Stdio::piped());
+
+        cmd.arg("--host");
+        if let Some(path) = self.path {
+            cmd.arg(format!("http://{}{}", addr, path));
+        } else {
+            cmd.arg(format!("http://{}", addr));
+        }
```
jplatte, line 135: Could slightly simplify to `format!("http://{}{}, addr, path.unwrap_or(""))`.

axum/benches/benches.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        for (key, value) in self.headers.into_iter().flatten() {
+            cmd.arg("--header");
+            cmd.arg(format!("{}: {}", key, value));
+        }
+
+        if let Some(body) = self.body {
+            cmd.arg("--body");
+            cmd.arg(body);
```
jplatte, line 153:
```suggestion
            cmd.args(&["--body", body]);
```

jplatte:
Overall looks very decent :)

programatik29:
I observed rewrk doesn't exit with error code on error.
Errors can be detected by checking stderr.

davidpdrsn:
> I observed rewrk doesn't exit with error code on error.
> 
> Errors can be detected by checking stderr.

We should fix that. Checking stderr is tedious 😅

jplatte (approved):
The bulk of this looks good.

PR #1248, opened 2022-08-11
Annotate panicking functions with `#[track_caller]`

davidpdrsn:
This annotates panicking functions with `#[track_caller]` so the error message points to where the user added the invalid router, rather than somewhere internally in axum.

Before:

```
thread 'routing::tests::foobar' panicked at 'Invalid route "/:b": insertion failed due to conflict with previously registered route: /:a', axum/src/routing/mod.rs:174:13
```

After

```
thread 'routing::tests::foobar' panicked at 'Invalid route "/:b": insertion failed due to conflict with previously registered route: /:a', axum/src/routing/tests/mod.rs:667:10
```

This should make it even easier to hunt down which route actually causes the conflict.


jplatte:
Your "Before" and "After" are equal 😆

davidpdrsn:
Whoops! Fixed 😅

jplatte (approved):

PR #1250, opened 2022-08-12
Port cookie extractors to use state to extract keys

davidpdrsn:
This PR builds on https://github.com/tokio-rs/axum/pull/1155 by porting `SignedCookieJar` and `PrivateCookieJar` to extract the keys via state rather than loosely typed extensions.

### DIFF
--- a/axum-extra/src/extract/cookie/mod.rs
+++ b/axum-extra/src/extract/cookie/mod.rs
@@ -208,7 +208,7 @@ fn set_cookies(jar: cookie::CookieJar, headers: &mut HeaderMap) {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use axum::{body::Body, http::Request, routing::get, Extension, Router};
+    use axum::{body::Body, http::Request, routing::get, Router};
     use tower::ServiceExt;
 
     macro_rules! cookie_test {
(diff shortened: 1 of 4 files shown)


axum-extra/src/extract/cookie/mod.rs
```diff
@@ -280,6 +283,24 @@ mod tests {
...
+    struct AppState {
+        key: Key,
+        custom_key: CustomKey,
+    }
+
+    impl From<AppState> for Key {
+        fn from(state: AppState) -> Key {
+            state.key
+        }
+    }
```
jplatte, line 296: So, what does the compiler error look like when this impl is missing?

davidpdrsn: Its pretty good actually

```
error[E0277]: the trait bound `cookie::Key: From<AppState>` is not satisfied
   --> axum-extra/src/extract/cookie/mod.rs:236:36
    |
236 |                     .route("/set", get(set_cookie))
    |                                    ^^^ the trait `From<AppState>` is not implemented for `cookie::Key`
...
281 |     cookie_test!(signed_cookies, SignedCookieJar);
    |     --------------------------------------------- in this macro invocation
    |
    = help: the trait `From<CustomKey>` is implemented for `cookie::Key`
    = note: required because of the requirements on the impl of `Into<cookie::Key>` for `AppState`
note: required because of the requirements on the impl of `axum::extract::FromRequest<hyper::Body, AppState>` for `signed::SignedCookieJar`
```

jplatte (approved):

PR #1251, opened 2022-08-12
Avoid unhelpful compiler suggestion

jplatte:


davidpdrsn (approved):

jplatte:
I think it's actually kind of wrong to say we don't support generic functions, it's just that we can't *fully* check them. Also we could check input types that don't mention any of the generic parameters, but that's non-trivial and might not be worth the time needed to implement it.

I've had an idea for a while that would make generic handlers pretty much fully checkable, that is:

```rust
#[debug_handler(with(T = String, T = u64; U = i8, U = i16; N = 0))
fn handler<T, U, const N: usize>(extract_t: Path<T>, extract_u: Json<U>, extract_n: Foo<N>)
where
    T: Deserialize,
    U: Deserialize,
{
    // ...
}
```

could make `debug_handler` add `FromRequest` assertions for `Path<String>`, `Path<u64>`, `Json<i8>`, `Json<i16>` and `Foo<0>` while also adding `Send` assertions for the futures returned by

* `handler::<String, i8, 0>`
* `handler::<String, i16, 0>`
* `handler::<u64, i8, 0>`
* `handler::<u64, i16, 0>`

PR #1254, opened 2022-08-12
Improve performance in low-level-rustls example

kkocdko:
Avoid await `IntoMakeSeviceFuture` in main thread.


kkocdko:
Await `make_service` will take a little time, it should not be placed in main thread. Inspired by [hyper's code](https://github.com/hyperium/hyper/blob/v0.14.20/src/server/server.rs#L176).

And I wrapped `Http` into Arc, to avoid unnecessary memory allocation.

These make about 4% performance improvement in my machine.

davidpdrsn (approved):
Thanks!

Issue #1116, opened 2022-06-24
Improve the custom rejection experience


jplatte:
Currently, rejections from extractors in axum and axum-extra turn into responses with plain-text bodies. In some cases this is not a problem but many people are using axum to implement API servers, and at least for rejections that are caused by invalid API usage (rather than server implementation errors), you would most often want a machine-readable error instead.

This means that many extractors such as `Json` or `Query` are rarely suitable for production – instead you have to write your own¹. We have examples for this, but it's really unfortunate you can't just do the obvious thing of using the provided ones. There have been many requests for being able to customize the rejections somehow, but we have not found a design that's clearly better than requiring people to create their own extractors.

These are the ideas we had so far, and why they were rejected:

* Add a defaulted generic const parameter to the extractors, something like
  ```rust
  struct Json<T, const F: fn(serde_json::Error) -> Response = default_error_to_response>(T);
  ```
  * __Rejected because__ it isn't possible on stable (and nightly right now, actually, though it will likely become possible at some point), and makes the type rather unwieldy in the documentation
* Add a defaulted generic type parameter to the extractors, something like
  ```rust
  struct Json<T, R = PlainTextJsonRejection>(T);
  ```
  where `R` has to implement both `From<serde_json::Error>` and `IntoResponse`
  * __Rejected because__ it isn't possible with Rust right now, generic type parameters on a struct need to be mentioned in one of the fields, but adding a `PhantomData<R>` breaks usage of `Json(value)` as a pattern and constructor
* Add some config mechanism, the most basic form of which would be
  ```rust
  my_router.layer(Extension(JsonConfig {
      serialize_error: |e: serde_json::Error| /* turn e into axum::response::Response */,
      ..Default::default()
  }))
  ```
  * __Rejected because__ it is non-local and has an unnecessary runtime cost (this really seems like something that should be configured at compile-time, not at runtime)

¹ these can sometimes delegate to the provided ones, but not always: for example if you want to include line / column as machine-readable fields of a json rejection, you have to write JSON extractor "from scratch" – though this is hardly more complex than delegating to `axum::Json`

SzudemJ:
Maybe this is a good place to mention, that I have to implement custom rejections for Json and Query.

I am doing this, because the default rejections return error messages that contain internal specifics such as module paths on an public rest-api, e.g.:

`Failed to deserialize query string. Expected something of type 'registration_service::routes::registrations::RequestParams'. Error: missing field 'email'`.

[...]

PR #1256, opened 2022-08-12
Support using a different rejection for `#[derive(FromRequest)]`

davidpdrsn:
This implements the suggestion I made [here](https://github.com/tokio-rs/axum/issues/1116#issuecomment-1195392516). I.e. one can change the rejection type in `#[derive(FromRequest)]` like so:

```rust
#[derive(FromRequest)]
#[from_request(via(axum::Json), rejection(MyRejection))]
struct MyJson<T>(T);

struct MyRejection(Response);

impl From<JsonRejection> for MyRejection {
    fn from(rejection: JsonRejection) -> Self {
        let response = (
            axum::Json(json!({ "error": rejection.to_string() })),
        ).into_response();

        MyRejection(response)
    }
}

impl IntoResponse for MyRejection {
        self.0
    }
}

#[derive(Deserialize)]
struct Payload {}

async fn handler(MyJson(payload): MyJson<Payload>) {}
```

I like this quite a bit as it takes most of the boilerplate out of wrapping extractors and importantly doesn't require users write any complex trait bounds (like the ones `Json` requires).

It also supports a couple of more use cases which are shown in the docs.

### DIFF
--- a/axum-macros/src/from_request/attr.rs
+++ b/axum-macros/src/from_request/attr.rs
@@ -11,7 +11,11 @@ pub(crate) struct FromRequestFieldAttr {
 }
 
 pub(crate) enum FromRequestContainerAttr {
-    Via(syn::Path),
+    Via {
+        path: syn::Path,
+        rejection: Option<syn::Path>,
+    },
+    Rejection(syn::Path),
     RejectionDerive(kw::rejection_derive, RejectionDeriveOptOuts),
     None,
 }
@@ -19,6 +23,7 @@ pub(crate) enum FromRequestContainerAttr {
 pub(crate) mod kw {
     syn::custom_keyword!(via);
     syn::custom_keyword!(rejection_derive);
+    syn::custom_keyword!(rejection);
     syn::custom_keyword!(Display);
     syn::custom_keyword!(Debug);
     syn::custom_keyword!(Error);
@@ -51,6 +56,7 @@ pub(crate) fn parse_container_attrs(
 
     let mut out_via = None;
     let mut out_rejection_derive = None;
+    let mut out_rejection = None;
 
     // we track the index of the attribute to know which comes last
     // used to give more accurate error messages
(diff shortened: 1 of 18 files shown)


axum-macros/src/from_request/attr.rs
```diff
@@ -73,11 +79,18 @@ pub(crate) fn parse_container_attrs(
...
+                } else {
+                    out_rejection = Some((idx, rejection, path));
+                }
+            }
         }
     }
 
-    match (out_via, out_rejection_derive) {
-        (Some((via_idx, via, _)), Some((rejection_derive_idx, rejection_derive, _))) => {
+    match (out_via, out_rejection_derive, out_rejection) {
```
davidpdrsn, line 92: This is getting out of hand 🙃 should probably be structured differently to be easier to extend.

axum-macros/tests/from_request/pass/override_rejection_with_via_on_struct.rs
```diff
@@ -0,0 +1,41 @@
...
+#[derive(Deserialize)]
+struct Payload {}
+
+async fn handler(_: MyJson<Payload>) {}
+
+async fn handler_result(_: Result<MyJson<Payload>, MyJsonRejection>) {}
+
+#[derive(FromRequest)]
+#[from_request(
+    // TODO(david): `rejection(...)` without `via(...)`
```
jplatte, line 23: Don't forget your `TODO`s 🙂

jplatte (approved):

axum-macros/src/from_request/attr.rs
jplatte: I'd say it's still manageable. But with the next attribute it will be bad ^^

axum-macros/tests/from_request/pass/override_rejection_with_via_on_struct.rs
davidpdrsn: Oh yeah! I did actually fix this but forgot to remove the todo.

davidpdrsn, line 23:
```suggestion
```

PR #1262, opened 2022-08-15
axum-extra: WithRejection

Altair-Bueno:
### Motivation

See https://github.com/tokio-rs/axum/issues/1116#issuecomment-1215034454

### Solution

- [x] Implement `WithRequest`
- [x] Add `with_request` feature
- [x] Include tests
- [x] Add documentation
- [x] Include an example

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -32,7 +32,6 @@ protobuf = ["dep:prost"]
 query = ["dep:serde", "dep:serde_html_form"]
 spa = ["tower-http/fs"]
 typed-routing = ["dep:axum-macros", "dep:serde", "dep:percent-encoding"]
-with_rejection = []
 
 [dependencies]
 axum = { path = "../axum", version = "0.5", default-features = false }
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -32,6 +32,7 @@ protobuf = ["dep:prost"]
 query = ["dep:serde", "dep:serde_html_form"]
 spa = ["tower-http/fs"]
 typed-routing = ["dep:axum-macros", "dep:serde", "dep:percent-encoding"]
+with_rejection = []
 
 [dependencies]
 axum = { path = "../axum", version = "0.5", default-features = false }
--- a/axum-extra/src/extract/with_rejection.rs
+++ b/axum-extra/src/extract/with_rejection.rs
@@ -5,16 +5,16 @@ use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut};
 
 /// Extractor for customizing extractor rejections
-/// 
-/// `WithRejection` wraps another extractor and gives you the result. If the 
-/// extraction fails, the `Rejection` is transformed into `R` and returned as a 
+///
+/// `WithRejection` wraps another extractor and gives you the result. If the
+/// extraction fails, the `Rejection` is transformed into `R` and returned as a
 /// response
-/// 
+///
 /// `E` is expected to implement [`FromRequest`]
-/// 
+///
 /// `R` is expected to implement [`IntoResponse`] and [`From<E::Rejection>`]
-/// 
-/// 
+///
+///
 /// # Example
 ///
 /// ```rust
@@ -48,9 +48,9 @@ use std::ops::{Deref, DerefMut};
 ///     WithRejection(Json(Person), _): WithRejection<Json<Person>, MyRejection>
 /// ) { /* ... */ }
 /// ```
-/// 
+///
 /// For a full example see the [customize-extractor-error] example
-/// 
+///
 /// [customize-extractor-error]: https://github.com/tokio-rs/axum/blob/main/examples/customize-extractor-error/src/main.rs
 /// [`FromRequest`]: axum::extract::FromRequest
 /// [`IntoResponse`]: axum::response::IntoResponse
--- a/axum-extra/src/extract/with_rejection.rs
+++ b/axum-extra/src/extract/with_rejection.rs
@@ -47,7 +47,7 @@ use std::ops::{Deref, DerefMut};
 ///     // client using the `IntoResponse` impl
 ///     WithRejection(Json(Person), _): WithRejection<Json<Person>, MyRejection>
 /// ) { /* ... */ }
-# let _: Router = Router::new().route("/", get(handler));
+/// # let _: axum::Router = axum::Router::new().route("/", axum::routing::get(handler));
 /// ```
 ///
 /// For a full example see the [customize-extractor-error] example
--- a/axum-extra/src/extract/with_rejection.rs
+++ b/axum-extra/src/extract/with_rejection.rs
@@ -4,10 +4,62 @@ use axum::response::IntoResponse;
 use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut};
 
+/// Extractor for customizing extractor rejections
+/// 
+/// `WithRejection` wraps another extractor and gives you the result. If the 
+/// extraction fails, the `Rejection` is transformed into `R` and returned as a 
+/// response
+/// 
+/// `E` is expected to implement [`FromRequest`]
+/// 
+/// `R` is expected to implement [`IntoResponse`] and [`From<E::Rejection>`]
+/// 
+/// 
+/// # Example
+///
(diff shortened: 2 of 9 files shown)


jplatte:
Please remove the `@` in front of my username in your commit message, it's generating extra notifications for me that I don't want.

axum-extra/Cargo.toml
```diff
@@ -32,6 +32,7 @@ protobuf = ["dep:prost"]
 query = ["dep:serde", "dep:serde_html_form"]
 spa = ["tower-http/fs"]
 typed-routing = ["dep:axum-macros", "dep:serde", "dep:percent-encoding"]
+with_rejection = []
```
jplatte, line 35: I don't think this needs to be behind a feature flag, since it doesn't pull in any additional dependencies.

axum-extra/src/extract/with_rejection.rs
```diff
@@ -0,0 +1,45 @@
...
+    E: FromRequest<B>,
+    R: From<E::Rejection> + IntoResponse,
+{
+    type Rejection = R;
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
+        match req.extract::<E>().await {
+            Ok(extractor) => Ok(WithRejection(extractor, PhantomData)),
+            Err(err) => Err(err.into()),
+        }
```
jplatte, line 43: If we use `From` instead of `Into`, this can be

```suggestion
        let extractor = req.extract::<E>().await?;
        Ok(WithRejection(extractor, PhantomData))
```

examples/customize-extractor-error/src/main.rs
```diff
@@ -5,15 +5,13 @@
 //! ```
 
 use axum::{
-    async_trait,
```
davidpdrsn, line 8: I don't think we should change this example. By doing that we are sorta saying "this is the best/recommended way to solve this" which I'm not sure is right.

However I do think we should link to `WithRejection` so people know that it's an alternative.

axum-extra/src/extract/with_rejection.rs
```diff
@@ -0,0 +1,138 @@
...
+/// ```
+///
+/// For a full example see the [customize-extractor-error] example
+///
+/// [customize-extractor-error]: https://github.com/tokio-rs/axum/blob/main/examples/customize-extractor-error/src/main.rs
+/// [`FromRequest`]: axum::extract::FromRequest
+/// [`IntoResponse`]: axum::response::IntoResponse
+/// [`From<E::Rejection>`]: std::convert::From
+#[derive(Debug, Clone, Copy, Default)]
+pub struct WithRejection<E, R>(pub E, pub PhantomData<R>);
```
davidpdrsn, line 59: Would `MapRejection` be a more accurate name?

axum-extra/src/extract/with_rejection.rs
```diff
@@ -0,0 +1,138 @@
...
+/// ) { /* ... */ }
+/// ```
+///
+/// For a full example see the [customize-extractor-error] example
+///
+/// [customize-extractor-error]: https://github.com/tokio-rs/axum/blob/main/examples/customize-extractor-error/src/main.rs
+/// [`FromRequest`]: axum::extract::FromRequest
+/// [`IntoResponse`]: axum::response::IntoResponse
+/// [`From<E::Rejection>`]: std::convert::From
+#[derive(Debug, Clone, Copy, Default)]
```
davidpdrsn, line 58: These derives will also require `R: Clone`, etc. Don't think we want that so have to write manual impls.

axum-extra/src/extract/with_rejection.rs
```diff
@@ -0,0 +1,138 @@
...
+/// use axum::extract::rejection::JsonRejection;
+/// use axum::response::{Response, IntoResponse};
+/// use axum::Json;
+/// use axum_extra::extract::WithRejection;
+/// use serde::Deserialize;
+///
+/// struct MyRejection { /* ... */ }
+///
+/// impl From<JsonRejection> for MyRejection {
+///     fn from(_:JsonRejection) -> MyRejection {
```
davidpdrsn, line 30:
```suggestion
///     fn from(rejection: JsonRejection) -> MyRejection {
```

axum-extra/src/extract/with_rejection.rs
```diff
@@ -0,0 +1,138 @@
...
+/// }
+/// #[derive(Debug, Deserialize)]
+/// struct Person { /* ... */ }
+///
+/// async fn handler(
+///     // If the `Json` extractor ever fails, `MyRejection` will be sent to the
+///     // client using the `IntoResponse` impl
+///     WithRejection(Json(Person), _): WithRejection<Json<Person>, MyRejection>
+/// ) { /* ... */ }
+/// ```
```
davidpdrsn, line 50: We usually add this to ensure `handler` actually implements `Handler`. Just to make sure the docs don't go out of sync.

```suggestion
# let _: Router = Router::new().route("/", get(handler));
/// ```
```

axum-extra/src/extract/with_rejection.rs
Altair-Bueno: `PhantomData<T>` always implements `Clone` for `T:?Sized`. See https://doc.rust-lang.org/std/marker/struct.PhantomData.html#impl-Clone

examples/customize-extractor-error/src/main.rs
Altair-Bueno: Maybe we should keep both of them? `customize-extractor-error` for `WithRejection` and `custom-extractor` for the old example? `WithRejection` is perfect for just replacing the rejection message, but the old example was better if you wanted to access the request + async

axum-extra/src/extract/with_rejection.rs
Altair-Bueno: I would say that `MapRejection` implies the existence of other functional combinators such as `FilterRejection` or `FoldRejection`

This raises the question: Should we consider adding this suite of combinators too?

davidpdrsn: What would they do?

axum-extra/src/extract/with_rejection.rs
davidpdrsn: Yes, but the derive macro still adds `R: Clone` which we don't need.

examples/customize-extractor-error/src/main.rs
davidpdrsn: Imo linking to it is fine since the docs contains an example.

axum-extra/src/extract/with_rejection.rs
davidpdrsn: > I would say that MapRejection implies the existence of other functional combinators

But just because "filter" doesn't make sense for some type doesn't mean you cannot add "map", if that does make sense.

examples/customize-extractor-error/src/main.rs
davidpdrsn: And to be clear I'm not saying we necessarily should rename it. I'm was just wondering if `MapRejection` is clearer.

jplatte: `MapRejection` was my first idea but I then proposed `WithRejection` because `.map_` (and correponding `Map` types) to me imply some sort of closure-accepting API.

davidpdrsn: Oh sorry I commented in the wrong place :P

axum-extra/src/extract/with_rejection.rs
davidpdrsn: Alright lets just keep the name `WithRejection`

davidpdrsn (commented):
I think it looks good!

Altair-Bueno:
> Wanna update the changelog as well?

Sure, omw

examples/customize-extractor-error/src/main.rs
Altair-Bueno: > What would they do?

Tbh `FilterRejection` doesn't sound interesting to me, but some kind of `EitherRejection` does (dispatching different variants to different rejections, for example). Nonetheless, I'll do some prototyping and see if any of these are actually useful

davidpdrsn (approved):
Thanks!

PR #1263, opened 2022-08-16
Add `Either{2..8}` to axum-extra

davidpdrsn:
This adds `Either{2..8}` types which combines extractors and responses into a single type. They serve a similar purpose as https://github.com/tokio-rs/axum/pull/1170 but are a bit more low level.

### DIFF
--- a/axum-extra/src/either.rs
+++ b/axum-extra/src/either.rs
@@ -53,7 +53,7 @@
 //! ```
 //!
 //! The general recommendation is to use [`IntoResponse::into_response`] to return different response
-//! types, but if you need to preserve the extract type then `Either*` works as well.
+//! types, but if you need to preserve the exact type then `Either*` works as well.
 //!
 //! [`BytesRejection`]: axum::extract::rejection::BytesRejection
 //! [`IntoResponse::into_response`]: https://docs.rs/axum/0.5/axum/response/index.html#returning-different-response-types
--- a/axum-extra/src/either.rs
+++ b/axum-extra/src/either.rs
@@ -16,10 +16,10 @@
 //!     }
 //! }
 //! #
-//! # let _: axum::Router = axum::Router::new().route("/", axum::routing::get(handler));
+//! # let _ = axum::routing::get(handler);
 //! ```
 //!
-//! Note that if all the inner extractors rejects the request then the rejection from the last
+//! Note that if all the inner extractors reject the request, the rejection from the last
 //! extractor will be returned. For the example above that would be [`BytesRejection`].
 //!
 //! # As a response
@@ -56,7 +56,7 @@
 //! types, but if you need to preserve the extract type then `Either*` works as well.
 //!
 //! [`BytesRejection`]: axum::extract::rejection::BytesRejection
-//! [`IntoResponse::into_response`]: https://docs.rs/axum/latest/axum/response/index.html#returning-different-response-types
+//! [`IntoResponse::into_response`]: https://docs.rs/axum/0.5/axum/response/index.html#returning-different-response-types
 
 use axum::{
     async_trait,
--- a/axum-extra/src/either.rs
+++ b/axum-extra/src/either.rs
@@ -68,7 +68,7 @@ use axum::{
 ///
 /// See the [module docs](self) for examples.
 #[derive(Debug, Clone)]
-pub enum Either2<E1, E2> {
+pub enum Either<E1, E2> {
     #[allow(missing_docs)]
     E1(E1),
     #[allow(missing_docs)]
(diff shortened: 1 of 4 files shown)


axum-extra/src/either.rs
```diff
@@ -0,0 +1,233 @@
...
+//!     body: Either3<Json<serde_json::Value>, String, Bytes>,
+//! ) {
+//!     match body {
+//!         Either3::E1(json) => { /* ... */ }
+//!         Either3::E2(string) => { /* ... */ }
+//!         Either3::E3(bytes) => { /* ... */ }
+//!     }
+//! }
+//! #
+//! # let _: axum::Router = axum::Router::new().route("/", axum::routing::get(handler));
```
jplatte, line 19: This should be enough to validate that the function is a handler, assuming that's what this line tests for.

```suggestion
//! # let _ = axum::routing::get(handler);
```

axum-extra/src/either.rs
```diff
@@ -0,0 +1,233 @@
...
+//! ```
+//!
+//! Note that if all the inner extractors rejects the request then the rejection from the last
```
jplatte, line 22:
```suggestion
//! Note that if all the inner extractors reject the request, the rejection from the last
```

axum-extra/src/either.rs
```diff
@@ -0,0 +1,233 @@
...
+//! fn something_else() -> bool {
+//!     // ...
+//!     # false
+//! }
+//! #
+//! # let _: axum::Router = axum::Router::new().route("/", axum::routing::get(handler));
+//! ```
+//!
+//! The general recommendation is to use [`IntoResponse::into_response`] to return different response
+//! types, but if you need to preserve the extract type then `Either*` works as well.
```
jplatte, line 56:
```suggestion
//! types, but if you need to preserve the exact type then `Either*` works as well.
```

axum-extra/src/either.rs
```diff
@@ -0,0 +1,233 @@
...
+//!
+//! [`BytesRejection`]: axum::extract::rejection::BytesRejection
+//! [`IntoResponse::into_response`]: https://docs.rs/axum/latest/axum/response/index.html#returning-different-response-types
```
jplatte, line 59:
```suggestion
//! [`IntoResponse::into_response`]: https://docs.rs/axum/0.5/axum/response/index.html#returning-different-response-types
```

jplatte: … to ensure this link doesn't break in the future. Could also put in a full version, since changing headers in docs could conceivably happen in a future 0.5.x release.

axum-extra/src/either.rs
```diff
@@ -0,0 +1,233 @@
...
+    async_trait,
+    extract::{FromRequest, RequestParts},
+    response::{IntoResponse, Response},
+};
+
+/// Combines two extractors or responses into a single type.
+///
+/// See the [module docs](self) for examples.
+#[derive(Debug, Clone)]
+pub enum Either2<E1, E2> {
```
jplatte, line 71: This type is usually just called `Either`. But I don't care too much whether we have the `2` or not.

axum-extra/src/either.rs
```diff
@@ -0,0 +1,233 @@
...
+    #[allow(missing_docs)]
+    E1(E1),
+    #[allow(missing_docs)]
+    E2(E2),
```
jplatte, line 75: Maybe this should be `A`, `B` and so on? I've seen that a few times, but haven't seen numbered variants like this.

axum-extra/src/either.rs
```diff
@@ -0,0 +1,233 @@
...
+    E7(E7),
+    #[allow(missing_docs)]
+    E8(E8),
+}
+
+macro_rules! impl_traits_for_either {
+    (
+        $either:ident =>
+        [$($ident:ident),* $(,)?],
+        $last:ident $(,)?
```
jplatte, line 190: I think this can probably be simplified, but I can play around with that after this PR is merged.

axum-extra/src/either.rs
davidpdrsn: I was thinking numbers might be easier. Doesn't require people know the alphabet. Off the top of my head I dunno what the 6th number is.

jplatte: Hm, but do you need to know the `n`th variant at any time? I would expect virtually all usage of this to be an exhaustive match handling every possible case. I'm fine leaving it as-is though if you prefer.

jplatte (approved):

axum-extra/src/either.rs
davidpdrsn: Hm perhaps. Think I'll just leave it as is 🤷

PR #1264, opened 2022-08-16
Add sync constructors to `CookieJar`, `SignedCookieJar`, and `PrivateCookieJar`

davidpdrsn:
This makes them easier to use in middleware which I think is a use case we should support.

### DIFF
--- a/axum-extra/src/extract/cookie/mod.rs
+++ b/axum-extra/src/extract/cookie/mod.rs
@@ -82,7 +82,7 @@ pub use cookie::Key;
 ///     .route("/me", get(me));
 /// # let app: Router<axum::body::Body> = app;
 /// ```
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct CookieJar {
     jar: cookie::CookieJar,
 }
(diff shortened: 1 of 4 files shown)


axum-extra/src/extract/cookie/mod.rs
```diff
@@ -95,26 +95,41 @@ where
...
+        let mut jar = cookie::CookieJar::new();
+        for cookie in cookies_from_request(headers) {
+            jar.add_original(cookie);
+        }
+        Self { jar }
+    }
+}
+
 impl CookieJar {
+    /// Create a new empty `CookieJar` from a map of request headers.
```
jplatte, line 122: I prefer the new function with a parameter (as documented here, but not as implemented) over the `From` impl because that is harder to discover. The `Default` implementation probably also makes sense for some cases, maybe its existence should be called out in the c'tor description.

jplatte: I disagree, I think `new` should cover the most common use case, which I think is what's now `from_headers`. But I don't insist of having it set up that way, this should be just as easy to use.

axum-extra/src/extract/cookie/signed.rs
```diff
@@ -101,20 +102,52 @@ where
 
     async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
         let key = Extension::<K>::from_request(req).await?.0.into();
```
jplatte, line 104:
```suggestion
        let key = req.extract::<Extension<K>>().await?.0.into();
```

axum-extra/src/extract/cookie/private.rs
```diff
@@ -83,20 +84,52 @@ where
 
     async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
         let key = Extension::<K>::from_request(req).await?.0.into();
```
jplatte, line 86: Same here.

jplatte (approved):
Small nitpick, we could use `RequestParts::extract`. Otherwise LGTM.

axum-extra/src/extract/cookie/mod.rs
davidpdrsn: But this way the constructors are consistent across the three kinds of jars:
- Empty jar: `CookieJar::new()`, `SignedCookieJar::new(key)`, `PrivateCookieJar::new(key)`
- With headers: `CookieJar::from_headers(headers)`, `SignedCookieJar::from_headers(headers, key)`, `PrivateCookieJar::from_headers(headers, key)`

Private and signed jars cannot implement `Default` because they need a `Key`.

jplatte: Right, I guess that is a good reason to do it this way 👍

PR #1268, opened 2022-08-17
Add `FromRef` trait

davidpdrsn:
Makes it possible to extract substates without cloning the outer state.

Part of https://github.com/tokio-rs/axum/pull/1155#pullrequestreview-1075733928

### DIFF
--- a/axum-core/src/extract/from_ref.rs
+++ b/axum-core/src/extract/from_ref.rs
@@ -1,0 +1,23 @@
+/// Used to do reference-to-value conversions thus not consuming the input value.
+///
+/// This is mainly used with [`State`] to extract "substates" from a reference to main application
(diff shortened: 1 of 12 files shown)


jplatte (commented):
With this, most of the `S: Clone` bounds can be removed, right? (e.g. on all `Router` impls except `impl Clone`)

davidpdrsn:
> With this, most of the `S: Clone` bounds can be removed, right? (e.g. on all `Router` impls except `impl Clone`)

Nope. We still need it for places like https://github.com/tokio-rs/axum/blob/router-state-extractor--from-ref/axum/src/routing/mod.rs#L487

jplatte (approved):

PR #1271, opened 2022-08-17
Support changing state type in `#[debug_handler]`

davidpdrsn:
Previously `#[debug_handler]` assumed your state was `()`. This makes it possible to change that, similarly to [changing the request body type](https://docs.rs/axum-macros/latest/axum_macros/attr.debug_handler.html#changing-request-body-type).

```rust
#[debug_handler(state = AppState)]
async fn handler(_: State<AppState>) {}

#[derive(Clone)]
struct AppState;
```


axum-macros/src/debug_handler.rs
```diff
@@ -39,29 +39,55 @@ pub(crate) fn expand(attr: Attrs, item_fn: ItemFn) -> TokenStream {
...
 
 impl Parse for Attrs {
     fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
         let mut body_ty = None;
+        let mut state_ty = None;
 
         while !input.is_empty() {
-            let ident = input.parse::<syn::Ident>()?;
-            if ident == "body" {
+            let lh = input.lookahead1();
```
davidpdrsn, line 58: One day we should refactor the argument parsing to be done the same way for each macro, but today is not that day 😅

axum-macros/src/lib.rs
```diff
@@ -513,12 +513,29 @@ pub fn derive_from_request(item: TokenStream) -> TokenStream {
...
+///
+/// By default `#[debug_handler]` assumes your state type is `()`. Thus if you're using
+/// [`axum::extract::State`] you must change the state type to match:
+///
+/// ```
+/// use axum::extract::State;
+/// # use axum_macros::debug_handler;
+///
+/// #[debug_handler(state = AppState)]
+/// async fn handler(State(staste): State<AppState>) {}
```
jplatte, line 526: Why not parse the argument / only have this state attribute as a fallback for indirect state type requirements (through other extractors / nested uses of `State`)?

davidpdrsn: Sure!

axum-macros/src/debug_handler.rs
```diff
@@ -371,6 +408,68 @@ fn self_receiver(item_fn: &ItemFn) -> Option<TokenStream> {
...
+        .sig
+        .inputs
+        .iter()
+        .filter_map(|input| match input {
+            FnArg::Receiver(_) => None,
+            FnArg::Typed(pat_type) => Some(pat_type),
+        })
+        .map(|pat_type| &pat_type.ty)
+        .filter_map(|ty| {
+            if let Type::Path(path) = &**ty {
```
jplatte, line 435: All the `if let`s in the function would probably be shorter if written as `match`. Also all the `filter_map` (and one `map`) could be collapsed into a single `filter_map`, but not sure whether that makes things simpler or not.

axum-macros/src/debug_handler.rs
```diff
@@ -371,6 +408,68 @@ fn self_receiver(item_fn: &ItemFn) -> Option<TokenStream> {
...
+            if let syn::GenericArgument::Type(ty) = generic_arg {
+                Some(ty)
+            } else {
+                None
+            }
+        })
+        .collect::<HashSet<_>>();
+
+    if state_inputs.len() == 1 {
+        state_inputs.iter().next().map(|&ty| ty.clone())
```
jplatte, line 467:
```suggestion
        state_inputs.into_iter().next().cloned()
```

jplatte (approved):
Looks good! Below are just nitpicks.

PR #1272, opened 2022-08-17
Only allow last extractor to mutate the request

davidpdrsn:
Fixing https://github.com/tokio-rs/axum/issues/1115 requires touching pretty much every file (see https://github.com/tokio-rs/axum/pull/1121). So rather than implementing everything at once I think we should split the work into smaller PRs and then merge them into this branch. That way we don't have to review the whole thing at once.

Fixes https://github.com/tokio-rs/axum/issues/1115

## TODO

Things to fix before we can merge this

- [x] docs for axum-extra
- [x] docs for axum-macros
- [x] On `FromRequestParts` document having to implement `FromRequest` if you need the body. To help with https://github.com/tokio-rs/axum/pull/1283/files#r950838603
- [x] changelog
- [x] document how to write extractors that wrap other extractors (needs to implement both traits manually). See https://github.com/tokio-rs/axum/pull/1288/files#r950879490
- [x] Can we remove any rejections for Path, like the one for the extension missing? No probably not. Someone might run `Path` from anywhere without running the request through `Router` and that shouldn't panic.
- [x] `FromRequest` and `FromRequestParts` for tuples (see `tuple.rs`)
- [x] Update `Handler::or` in `axum-extra`. Note this depends on the tuple impls
- [x] Update `#[derive(FromRequest)]`
- [x] Do we want `#[derive(FromRequestParts)]`? If so, that can be done after merging this. File issue if we want it.
- [x] Decide on final names for `Mut` and `Once`.


davidpdrsn:
@jplatte I took at stab at the tuple impls but couldn't quite make it work, even with adding another possible value of `M`. Do you wanna give it a shot?

jplatte:
I'll do that, yeah. Won't be in front of a capable computer until Sunday though.

jplatte:
Arg, can't stop thinking through this 😅

Have you tried sth. like

- `impl FromRequestParts for (impl FromRequestParts, impl FromRequestParts)`
- `impl FromRequest<Tup> for (impl FromRequestParts, impl FromRequest<Once>)`
- `impl FromRequest<Tup2> for (inpl FromRequestParts, impl FromRequest<Tup>)` (last tuple element nesting, could also do Tup3..TupN but likely not even this one would ever be used)

jplatte:
Open question from #1288: What should the final names for `Mut` and `Once` be (they will appear in the docs, even if they don't have doc pages of their own)?

* Keep `Mut` and `Once`
* `ViaParts` and something?
  * `ViaRequest`
  * Something like `Regular` / `Default` / `Direct`

davidpdrsn:
Re `Mut`/`Once` naming I like

- `Mut` becomes `ViaParts`
- `Once` becomes `ViaRequest`

`impl<S, B, T> FromRequest<S, B, ViaParts> for T` does kinda make sense since we're extracting a `T` "via the request parts".

`trait FromRequest<S, B, M = ViaRequest>` is repetitive but at least it matches the other.

davidpdrsn:
I guess this is ready for review now 😊 All the todos are done and CI is green.

jplatte (approved):
I already reviewed everything individually, is more review needed?

PR #1275, opened 2022-08-18
Change `FromRequest` and add `FromRequestParts` trait

davidpdrsn:
This is to get the ball rolling on https://github.com/tokio-rs/axum/pull/1272

This totally doesn't compile, and that isn't the intention, but all other changes will follow from this. `cargo check -p axum-core` does work though.

This PR doesn't add any docs. I'll work on that next.


davidpdrsn:
And here are my answers to the unresolved questions:

> Is this the way we wanna go, as supposed to having different `FromRequest` traits? That doesn't seem to be possible however given the discussion in https://github.com/tokio-rs/axum/issues/1115.

From the fact that I'm opening this PR, I think so :) It does work and given we don't have any other possible solutions I think its okay.

> Are the names `Mut` and `Once` good enough?

I don't think the names are _fine_ and don't have any better suggestions. It is a little weird to have a `&mut RequestParts<Once>`. Shouldn't you only be allowed to have an owned `RequestParts<Once>`?

> Should the type parameter be called `M`?

In the previous PR I called it `R` :shrug: Not sure which is better or if there is a more obvious choice. I don't have a good reason for `R` or `M`.

> `RequestParts<Mut, _, _>` is not allowed to remove extensions. This is a bit annoying if you have a middleware that inserts an extension which you wanna remove (to avoid a clone) in the handler. This is safe because the extension is only removed once. Do we wanna allow `RequestParts<Mut, _, _>` to call `remove_extension`?

I think we should keep it like as it is in this PR. If someone comes along with a really good use case for needing `remove_extension` from a `RequestParts<Mut>` we can always move the method. That isn't a breaking change.

jplatte:
> > Is this the way we wanna go, as supposed to having different `FromRequest` traits? That doesn't seem to be possible however given the discussion in #1115.
> 
> From the fact that I'm opening this PR, I think so :) It does work and _given we don't have any other possible solutions_ I think its okay.

Challenge accepted! I've just had one more idea, it is really ugly but so is the `Mut` / `Once` thing IMO.

davidpdrsn:
> Challenge accepted! I've just had one more idea, it is really ugly but so is the Mut / Once thing IMO.

Great! Excited to see it :)

jplatte:
Wow, it actually looks like it works. The only unfortunate thing about this is that you can't apply `#[doc(hidden)]` to generic parameters 😶‍🌫️

axum-core/src/extract/mod.rs
```diff
@@ -65,256 +72,91 @@ pub use self::from_ref::FromRef;
...
-    pub(crate) state: Arc<S>,
-    method: Method,
-    uri: Uri,
-    version: Version,
-    headers: HeaderMap,
-    extensions: Extensions,
-    body: Option<B>,
-}
+#[async_trait]
+pub trait FromRequestOnce<S, B, M = Once>: Sized {
```
davidpdrsn, line 85: Oh so if you add `M` then it just works? Cool! :D

axum-core/src/extract/mod.rs
```diff
@@ -65,256 +72,91 @@ pub use self::from_ref::FromRef;
...
 #[async_trait]
-pub trait FromRequest<S, B>: Sized {
+pub trait FromRequestMut<S, B>: Sized {
     /// If the extractor fails it'll use this "rejection" type. A rejection is
     /// a kind of error that can be converted into a response.
     type Rejection: IntoResponse;
 
     /// Perform the extraction.
-    async fn from_request(req: &mut RequestParts<S, B>) -> Result<Self, Self::Rejection>;
+    async fn from_request_mut(req: &mut Request<B>, state: &S) -> Result<Self, Self::Rejection>;
```
davidpdrsn, line 81: Really liking just passing the request here but its still possible to consume the body even with an `&mut Request<B>`:

```rust
#[async_trait]
impl<S, T, B> FromRequestMut<S, B> for Foo
where
    T: FromRequestMut<S, B>,
    B: http_body::Body + Send + Unpin,
    S: Send + Sync,
{
    type Rejection = Infallible;

    async fn from_request_mut(
        req: &mut Request<B>,
        state: &S,
    ) -> Result<Option<T>, Self::Rejection> {
        let pull_something_from_the_body = req.data().await;

        ...
    }
}
```

I think thats pretty unfortunate. I suppose we could do

```rust
async fn from_request_mut(parts: &mut http::request::Parts, state: &S) -> Result<Self, Self::Rejection>

// or maybe `from_request_parts`?
```

ie not even give you the body at all. That still allows you nuke the extensions, but maybe thats fine?

jplatte: Well, `Bytes` / `String` and friends all only implementing `FromRequestOnce` should already heavily discourage trying to extract the body in `FromRequestMut`, but I wouldn't mind passing `&mut http::request::Parts` either.

axum-core/src/extract/mod.rs
jplatte: Yeah, the trick is that a `FromRequestOnce<S, B, Once>` impl can't possibly overlap with a `FromRequestOnce<S, B, Mut>` impl. It could also be `FromRequestOnce<S, B, const REALLY_ONCE: bool = true>` with the blanket impl being for `FromRequestOnce<S, B, false>`, but then `Handler` also needs that const-generic param.

axum-core/src/extract/mod.rs
davidpdrsn: Yeah it would definitely be shooting yourself in the foot

> but I wouldn't mind passing `&mut http::request::Parts` either

I think that would be interesting to explore. Could even call the traits `FromRequestParts` and `FromRequest` 🤔

jplatte: I like that, it mirrors `IntoResponse` and `IntoResponseParts`! I think being able to nuke extensions is fine by the way.

davidpdrsn: Oh yeah thats true! Its all coming together 😃

I'll change it :)

axum-core/src/extract/mod.rs
```diff
@@ -18,6 +17,14 @@ mod tuple;
 
 pub use self::from_ref::FromRef;
 
+mod __private {
+    #[derive(Debug, Clone, Copy)]
+    pub enum Mut {}
+
+    #[derive(Debug, Clone, Copy)]
+    pub enum Once {}
```
davidpdrsn, line 25: I figure these can be private since users should never use them.

axum-core/src/extract/mod.rs
```diff
@@ -65,256 +72,95 @@ pub use self::from_ref::FromRef;
...
-    ///
-    /// You generally shouldn't need to construct this type yourself, unless
-    /// using extractors outside of axum for example to implement a
-    /// [`tower::Service`].
-    ///
-    /// [`tower::Service`]: https://docs.rs/tower/lastest/tower/trait.Service.html
-    pub fn new(req: Request<B>) -> Self {
-        Self::with_state((), req)
-    }
+    /// TODO
```
davidpdrsn, line 90: I'll update the docs later I think.

davidpdrsn:
@jplatte ready for another review

jplatte (approved):

PR #1276, opened 2022-08-18
examples: `error-handling` example

Altair-Bueno:
## Motivation

See https://github.com/tokio-rs/axum/issues/1116#issuecomment-1219430083

## Solution

- [x] Create a new `error-handling` example crate
  - [x] Include `WithRejection` version (https://github.com/tokio-rs/axum/issues/1116#issuecomment-1218276084)
  - [x] Include `#[derive(FromRequest)]` version (https://github.com/tokio-rs/axum/issues/1116#issuecomment-1195392516)
  - [x] Include manual version (https://github.com/tokio-rs/axum/tree/main/examples/customize-extractor-error)
- [x] Add comments


examples/error-handling/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "error-handling"
```
davidpdrsn, line 2: I still think we should add this to the `customize-extractor-error` crate. Don't see why we need another crate.

We also cannot call it "error handling" because that means something in axum. See https://docs.rs/axum/latest/axum/error_handling/index.html

davidpdrsn: I see now that my comment about not touching existing examples was confusing. I meant not touching other examples than `customize-extractor-error`. Sorry about that.

examples/error-handling/src/with_rejection.rs
```diff
@@ -0,0 +1,36 @@
...
+            "message": self.to_string(),
+            "timestamp": Utc::now(),
+            "origin": "with_rejection"
+        });
+        let code = match self {
+            ApiError::JsonExtractorRejection(x) => match x {
+                JsonRejection::JsonDataError(_) | JsonRejection::MissingJsonContentType(_) => {
+                    StatusCode::BAD_REQUEST
+                }
+                _ => StatusCode::INTERNAL_SERVER_ERROR,
```
davidpdrsn, line 31: I'm slightly worried that people will just copy/paste these examples so I think its important to use the right status codes.

- `JsonDataError` should be `StatusCode::UNPROCESSABLE_ENTITY`
- `JsonSyntaxError` should be `StatusCdoe::BAD_REQUEST`
- `MissingJsonContentType` should be `StatusCode::UNSUPPORTED_MEDIA_TYPE`

Same goes for the other mods.

examples/error-handling/src/main.rs
```diff
@@ -0,0 +1,31 @@
...
+            std::env::var("RUST_LOG").unwrap_or_else(|_| "error_handling=debug".into()),
+        ))
+        .with(tracing_subscriber::fmt::layer())
+        .init();
+
+    // Build our application with some routes
+    let app = Router::new()
+        .route("/withRejection", post(with_rejection::handler))
+        .route("/customExtractor", post(custom_extractor::handler))
+        .route("/deriveFromRequest", post(derive_from_request::handler));
```
davidpdrsn, line 22: nit: We normally use `snake-case` for paths.

examples/error-handling/src/with_rejection.rs
```diff
@@ -0,0 +1,36 @@
...
+pub enum ApiError {
+    #[error(transparent)]
+    JsonExtractorRejection(#[from] JsonRejection),
+}
+
+impl IntoResponse for ApiError {
+    fn into_response(self) -> axum::response::Response {
+        let payload = json!({
+            "message": self.to_string(),
+            "timestamp": Utc::now(),
```
davidpdrsn, line 23: I don't think depending on chrono just to add a timestamp is worth it. Hyper already adds a `Date` header anyway.

Altair-Bueno:
Some notes:

- The timestamp field originates from my example, i should probably remove it
- Because `derive(FromRequest)` also uses `From<E::Rejection>` we can also use `thiserror` crate to transform rejections the same way as `WithRejection`. I updated the `derive_from_request` mod with this change
- Comments are currently missing. Im wondering if i should add a README with a more in depth explanation about the differences between the different approaches to rejections
- `custom_extractor` includes an extra field to highlight the ability of custom extractors on accessing the request. 
- I think I've found a bug with `MatchedPath`, probably due to some breaking change on master. 


Replace https://github.com/Altair-Bueno/axum/blob/e39433ea5f816b132bf3a4db6c9210dd0a798faf/examples/error-handling/src/custom_extractor.rs#L37 with this throws this error

```rust
let path = req.extract::<MatchedPath>().await.map(|x|x.as_str().to_owned()).ok();
```


```text
   Compiling error-handling v0.1.0 (/Users/compux72/Developer/axum/examples/error-handling)
error: future cannot be sent between threads safely
  --> error-handling/src/custom_extractor.rs:32:90
   |
32 |       async fn from_request(req: &mut RequestParts<S, B>) -> Result<Self, Self::Rejection> {
   |  __________________________________________________________________________________________^
33 | |         match axum::Json::<T>::from_request(req).await {
34 | |             Ok(value) => Ok(Self(value.0)),
35 | |             Err(rejection) => {
...  |
59 | |         }
60 | |     }
   | |_____^ future created by async block is not `Send`
   |
note: future is not `Send` as this value is used across an await
  --> error-handling/src/custom_extractor.rs:41:56
   |
33 |         match axum::Json::<T>::from_request(req).await {
   |               ---------------------------------------- has type `Result<axum::Json<T>, JsonRejection>` which is not `Send`
...
41 |                 let path = req.extract::<MatchedPath>().await.map(|x|x.as_str().to_owned()).ok();
   |                                                        ^^^^^^ await occurs here, with `axum::Json::<T>::from_request(req).await` maybe used later
...
60 |     }
   |     - `axum::Json::<T>::from_request(req).await` is later dropped here
   = note: required for the cast to the object type `dyn Future<Output = Result<custom_extractor::Json<T>, (StatusCode, axum::Json<serde_json::Value>)>> + Send`
help: consider further restricting this bound
   |
25 |     T: DeserializeOwned + std::marker::Send,
   |                         +++++++++++++++++++

error: could not compile `error-handling` due to previous error
```

Haven't checked yet if this is a known issue. If not, let me know and I open an issue (with pr) tomorrow

davidpdrsn:
> The timestamp field originates from my example, i should probably remove it

I agree. We should remove it.

> Because derive(FromRequest) also uses From<E::Rejection> we can also use thiserror crate to transform rejections the same way as WithRejection. I updated the derive_from_request mod with this change

I think we should just use `thiserror` in one example. Otherwise people might assume thats the only way to do it.

> I think I've found a bug with MatchedPath, probably due to some breaking change on master.

Uhm not sure whats going on there but don't think that should be broken on `main`. Don't think we've changed anything there 🤔

examples/error-handling/src/with_rejection.rs
Altair-Bueno: I copy & pasted from https://github.com/tokio-rs/axum/blob/568394a28e62904f8db4872da0d10e90d831ff42/examples/customize-extractor-error/src/main.rs#L76 so I'm afraid we should probably double check all examples too. My bad for not reading the match statement

davidpdrsn: Ah okay yeah then that example was wrong to begin with 😜

Altair-Bueno:
> I think we should just use thiserror in one example. Otherwise people might assume thats the only way to do it.

Alright, i'll rollback those changes

> Uhm not sure whats going on there but don't think that should be broken on main. Don't think we've changed anything there 🤔

Np i'll look into that tomorrow with a smaller sample

examples/customize-extractor-error/Cargo.toml
```diff
@@ -2,12 +2,16 @@
 name = "example-customize-extractor-error"
 version = "0.1.0"
 edition = "2021"
-publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-axum = { path = "../../axum" }
+axum = { path = "../../axum", features=["matched-path"] }
```
davidpdrsn, line 9: This feature is enabled by default.

```suggestion
axum = { path = "../../axum" }
```

examples/customize-extractor-error/Cargo.toml
```diff
@@ -2,12 +2,16 @@
 name = "example-customize-extractor-error"
 version = "0.1.0"
 edition = "2021"
-publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
davidpdrsn, line 6: We'll wanna keep this so we don't accidentally publish this crate.

```suggestion
publish = false
```

Altair-Bueno:
> Np i'll look into that tomorrow with a smaller sample

It was just a missing trait bound. `T` must be `Send` to use `RequestParts::extract` on the request. Looks like `Json` doesn't require this because it `awaits` before calling `serde_json::from_slice`

- Keep as it is, using the original trait bounds imposed by `Json`
- Add `T:Send` and replace `extension().get::<MatchedPath>()` with `extract::<MatchedPath>()`

examples/customize-extractor-error/src/with_rejection.rs
```diff
@@ -0,0 +1,57 @@
...
+    // The second constructor argument is not meaningful and can be safely ignored
+    WithRejection(Json(value), _): WithRejection<Json<Value>, ApiError>,
+) -> impl IntoResponse {
+    Json(dbg!(value))
+}
+
+// We derive `thiserror::Error`
+#[derive(Debug, Error)]
+pub enum ApiError {
+    // The `#[ƒrom]` attribute generates `From<JsonRejection> for ApiError`
```
davidpdrsn, line 35: Thats a strange `f` 😜

```suggestion
    // The `#[from]` attribute generates `From<JsonRejection> for ApiError`
```

davidpdrsn (commented):
One final nitpick otherwise I think it looks good!

examples/customize-extractor-error/src/custom_extractor.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    async fn from_request(req: &mut RequestParts<S, B>) -> Result<Self, Self::Rejection> {
+        match axum::Json::<T>::from_request(req).await {
+            Ok(value) => Ok(Self(value.0)),
+            // convert the error from `axum::Json` into whatever we want
+            Err(rejection) => {
+                let path = req
+                    .extensions()
+                    .get::<MatchedPath>()
+                    .map(|x| x.as_str().to_owned());
```
davidpdrsn, line 45: Should probably use `req.extract::<MatchedPath>().await` as its more future proof. There is no guarantee that `MatchedPath` will remain an extension. Thats an implementation detail.

PR #1277, opened 2022-08-18
Port `Handler` and most extractors

davidpdrsn:
This ports `Handler` and most extractors to the new setup. `cargo t --all-features -p axum --lib` now passes 🎉 

The only things I didn't port was `Handler::or` in axum-extra. I'll look at that later.

Biggest changes:
- `Handler` got a new `M` type param.
- `FromRequestParts` doesn't actually need to be generic over the body. So its `B` type param is gone.

Overall I'm really liking this new setup! Its very straight forward to port things. The only thing you have to be aware of is when writing an extractor that wraps another (like `WithRejection` or `Cached`) you have to write both `FromRequest` and `FromRequestParts` impls. That is because you need to call one or the other methods on the wrapped extractor which requires different bounds. I'll make a note that we have to document that.

### DIFF
--- a/axum/src/extract/multipart.rs
+++ b/axum/src/extract/multipart.rs
@@ -2,12 +2,13 @@
 //!
 //! See [`Multipart`] for more details.
 
-use super::{rejection::*, BodyStream, FromRequest, RequestParts};
+use super::{BodyStream, FromRequest};
 use crate::body::{Bytes, HttpBody};
 use crate::BoxError;
 use async_trait::async_trait;
 use futures_util::stream::Stream;
 use http::header::{HeaderMap, CONTENT_TYPE};
+use http::Request;
 use std::{
     fmt,
     pin::Pin,
@@ -58,10 +59,12 @@ where
 {
     type Rejection = MultipartRejection;
 
-    async fn from_request(req: &mut RequestParts<S, B>) -> Result<Self, Self::Rejection> {
-        let stream = BodyStream::from_request(req).await?;
-        let headers = req.headers();
-        let boundary = parse_boundary(headers).ok_or(InvalidBoundary)?;
+    async fn from_request(req: Request<B>, state: &S) -> Result<Self, Self::Rejection> {
+        let boundary = parse_boundary(req.headers()).ok_or(InvalidBoundary)?;
+        let stream = match BodyStream::from_request(req, state).await {
+            Ok(stream) => stream,
+            Err(err) => match err {},
+        };
         let multipart = multer::Multipart::new(stream, boundary);
         Ok(Self { inner: multipart })
     }
@@ -224,7 +227,6 @@ composite_rejection! {
     ///
     /// Contains one variant for each way the [`Multipart`] extractor can fail.
     pub enum MultipartRejection {
-        BodyAlreadyExtracted,
         InvalidBoundary,
     }
 }
--- a/axum-core/src/extract/mod.rs
+++ b/axum-core/src/extract/mod.rs
@@ -17,10 +17,15 @@ mod tuple;
 
 pub use self::from_ref::FromRef;
 
-mod __private {
+#[doc(hidden)]
+pub mod private {
+    /// Private APIs
+
+    #[doc(hidden)]
     #[derive(Debug, Clone, Copy)]
     pub enum Mut {}
 
+    #[doc(hidden)]
     #[derive(Debug, Clone, Copy)]
     pub enum Once {}
 }
@@ -72,7 +77,7 @@ mod __private {
 /// [`http::Request<B>`]: http::Request
 /// [`axum::extract`]: https://docs.rs/axum/latest/axum/extract/index.html
 #[async_trait]
-pub trait FromRequestParts<S, B>: Sized {
+pub trait FromRequestParts<S>: Sized {
     /// If the extractor fails it'll use this "rejection" type. A rejection is
     /// a kind of error that can be converted into a response.
     type Rejection: IntoResponse;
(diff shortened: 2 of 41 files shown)


davidpdrsn:
Something isn't quite right with tuple extractors because this doesn't work

```rust
async fn handler(_: (HeaderMap,)) {}
let _: Router = Router::new().route("/", get(handler));
```

I'm trying to figure out whats wrong. Probably something in https://github.com/tokio-rs/axum/blob/from-request-mut-once--round-1/axum-core/src/extract/tuple.rs

axum/src/extract/multipart.rs
```diff
@@ -224,7 +227,6 @@ composite_rejection! {
     ///
     /// Contains one variant for each way the [`Multipart`] extractor can fail.
     pub enum MultipartRejection {
-        BodyAlreadyExtracted,
         InvalidBoundary,
     }
```
jplatte, line 231: Do you want this to remain an enum for forwards compatibility?

davidpdrsn:
@jplatte So yeah I don't think we can implement `FromRequest{Parts}` for tuples, i.e.
`(T1,)`, `(T1, T2)`, etc.

To make something like `(HeaderMap,)` work we have to write this implementation:

```rust
impl<S, B, T1> FromRequest<S, B, /* Once */> for (T1,)
where
    T1: FromRequest<S, B, /* Once */> + Send,
{
    type Rejection = Response;

    async fn from_request(req: Request<B>, state: &S) -> Result<Self, Self::Rejection> {
        ...
    }
}
```

However since `HeaderMap` implements `FromRequest` via the blanket impl

```rust
impl<S, B, T> FromRequest<S, B, private::Mut> for T
where
    T: FromRequestParts<S>,
{
    type Rejection = <Self as FromRequestParts<S>>::Rejection;

    async fn from_request(req: Request<B>, state: &S) -> Result<Self, Self::Rejection> {
        ...
    }
}
```

It doesn't implement `FromRequest<S, B, Once>`, like we required in the `(T1,)`
impl.

If we instead require `T1: FromRequest<S, B, Mut>` then things like `Json<_>`
wouldn't work because they implement `FromRequest<S, B, Once>`.

We also cannot use a generic `M` param as that gives a conflicting impl error:

```rust
impl<S, B, M, T1> FromRequest<S, B, M> for (T1,)
where
    T1: FromRequest<S, B, M> + Send,
{
    type Rejection = Response;

    async fn from_request(req: Request<B>, state: &S) -> Result<Self, Self::Rejection> {
        ...
    }
}
```

That conflicts with

```rust
impl<S, B, T> FromRequest<S, B, private::Mut> for T
where
    T: FromRequestParts<S>,
{
    type Rejection = <Self as FromRequestParts<S>>::Rejection;

    async fn from_request(req: Request<B>, state: &S) -> Result<Self, Self::Rejection> {
        ...
    }
}
```

The error is

```
error[E0119]: conflicting implementations of trait `extract::FromRequest<_, _, extract::private::Mut>` for type `(_,)`
   --> axum-core/src/extract/mod.rs:100:1
    |
100 | / impl<S, B, T> FromRequest<S, B, private::Mut> for T
101 | | where
102 | |     B: Send + 'static,
103 | |     S: Send + Sync,
...   |
111 | |     }
112 | | }
    | |_^ conflicting implementation for `(_,)`
    |
   ::: axum-core/src/extract/tuple.rs:34:1
    |
34  | / impl<S, B, M, T1> FromRequest<S, B, M> for (T1,)
35  | | where
36  | |     T1: FromRequest<S, B, M> + Send,
37  | |     B: Send + 'static,
...   |
44  | |     }
45  | | }
    | |_- first implementation here
    |
    = note: downstream crates may implement trait `extract::FromRequestParts<_>` for type `(_,)`
```

I'm not sure how to resolve this :thinking: Kinda feels like we need to drop the
tuple impls :/

davidpdrsn:
and `cargo t --all-features -p axum --lib` still passes even if I remove the tuple impls, so not sure we really need them.

jplatte (commented):
Looks pretty good! Why did you add the `M` parameter to handlers though, rather than just having it on the impls and including it in the `T` tuple like I did in the test? Then the routing wouldn't need an extra parameter.

jplatte:
I think we might be able to make tuples work with another dummy type to work around the overlapping impls.

axum/src/extract/multipart.rs
davidpdrsn: Nah I'd rather just remove it rather than not using it.

davidpdrsn:
> Why did you add the M parameter to handlers though, rather than just having it on the impls and including it in the T tuple like I did in the test? Then the routing wouldn't need an extra parameter.

That was probably just being silly. I'll try that!

> I think we might be able to make tuples work with another dummy type to work around the overlapping impls. (not another parameter, just another distinct value for M)

I'll give it a shot :)

axum/src/extract/multipart.rs
jplatte: I'm not sure you understood my question 😅

I wanted to ask whether you want to keep `MultipartRejection` even if it only has one variant now, or remove it in favor of using the inner type directly.

davidpdrsn:
> Why did you add the M parameter to handlers though, rather than just having it on the impls and including it in the T tuple like I did in the test? Then the routing wouldn't need an extra parameter.

Done! So much cleaner 😅

I've commented out the tuple impls. Think we should do that in its own PR as this is big enough as is.

axum/src/extract/multipart.rs
davidpdrsn: Ah sorry. I thought you meant `BodyAlreadyExtracted`. Kinda wanna keep the enum especially since multipart interacts with an external dependency. So like if multer adds a new error type we might need a new variant here.

jplatte (approved):

PR #1283, opened 2022-08-21
Port `#[derive(TypedPath)]` and `#[debug_handler]`

davidpdrsn:
I'll do `#[derive(FromRequest)]` separately since it probably requires more work. We might even want `#[derive(FromRequestParts]` as well.


axum-macros/src/debug_handler.rs
```diff
@@ -159,6 +159,12 @@ fn check_inputs_impls_from_request(
     body_ty: &Type,
     state_ty: Type,
 ) -> TokenStream {
+    let takes_self = item_fn
+        .sig
+        .inputs
+        .first()
+        .map_or(false, |arg| matches!(arg, FnArg::Receiver(_)));
```
jplatte, line 166: Should also check for the other `FnArg` variant where the `self` keyword is followed by a colon and type like any other parameter.

jplatte (approved):
One nitpick, otherwise LGTM!

axum-macros/src/debug_handler.rs
davidpdrsn: Ah yes good catch!

PR #1285, opened 2022-08-21
Port docs for axum and axum-core

davidpdrsn:
`cargo t --all-features -p axum -p axum-core` now passes 🎉

### DIFF
--- a/axum/src/docs/extract.md
+++ b/axum/src/docs/extract.md
@@ -219,7 +219,7 @@ async fn handler(
 ```
 
 axum enforces this by requiring the last extractor implements [`FromRequest`]
-and all others implements [`FromRequestParts`].
+and all others implement [`FromRequestParts`].
 
 # Optional extractors
 
--- a/axum/src/docs/extract.md
+++ b/axum/src/docs/extract.md
@@ -5,12 +5,12 @@ Types and traits for extracting data from requests.
 - [Intro](#intro)
 - [Common extractors](#common-extractors)
 - [Applying multiple extractors](#applying-multiple-extractors)
-- [Be careful when extracting `Request`](#be-careful-when-extracting-request)
+- [The order of extractors](#the-order-of-extractors)
 - [Optional extractors](#optional-extractors)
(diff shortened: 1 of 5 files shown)


axum/src/docs/extract.md
```diff
@@ -604,27 +653,20 @@ async fn auth_middleware<B>(
...
-    // running extractors requires a `RequestParts`
-    let mut request_parts = RequestParts::new(request);
+    // running extractors requires a `axum::http::request::Parts`
+    let (mut parts, body) = request.into_parts();
 
-    // `TypedHeader<Authorization<Bearer>>` extracts the auth token but
-    // `RequestParts::extract` works with anything that implements `FromRequest`
-    let auth = request_parts.extract::<TypedHeader<Authorization<Bearer>>>()
+    // `TypedHeader<Authorization<Bearer>>` extracts the auth token
+    let auth = TypedHeader::<Authorization<Bearer>>::from_request_parts(&mut parts, &())
```
jplatte, line 660: I think we should provide `RequestExt { fn extract(self) -> _; fn extract_parts(&mut self) -> _; }` and `PartsExt { fn extract(&mut self) -> _; }` or sth. like that, such that people can use method call syntax, and also don't need to do the whole `into_parts` + `from_parts` dance, to extract headers and such from a request.

axum/src/lib.rs
```diff
@@ -376,6 +377,7 @@
 //! [tower-guides]: https://github.com/tower-rs/tower/tree/master/guides
 //! [`Uuid`]: https://docs.rs/uuid/latest/uuid/
 //! [`FromRequest`]: crate::extract::FromRequest
+//! [`FromRequetsParts`]: crate::extract::FromRequetsParts
```
jplatte, line 380:
```suggestion
//! [`FromRequestParts`]: crate::extract::FromRequestParts
```

axum/src/docs/extract.md
davidpdrsn: Yep I agree. My plan was to add those when the big PR is merged.

jplatte (approved):

PR #1286, opened 2022-08-21
Implement FromRequest[Parts] for tuples

jplatte:
This means that `FromRequest` is implemented for `(Method,)`, `(Bytes,)` as well as `((Method,),)` but not `((Bytes,),)`. I could add more impls such that that is allowed, but only up to a fixed level of tuple nesting.


davidpdrsn (approved):
Great!

PR #1287, opened 2022-08-21
Port axum-extra

davidpdrsn:
This ports the code and docs in axum-extra. `cargo t --all-features -p axum-extra` now passes. Getting closer 😊

### DIFF
--- a/axum-core/Cargo.toml
+++ b/axum-core/Cargo.toml
@@ -9,7 +9,7 @@ license = "MIT"
 name = "axum-core"
 readme = "README.md"
 repository = "https://github.com/tokio-rs/axum"
-version = "0.2.6" # remember to also bump the version that axum and axum-extra depends on
+version = "0.2.6" # remember to also bump the version that axum and axum-extra depend on
 
 [dependencies]
 async-trait = "0.1"
--- a/axum-core/Cargo.toml
+++ b/axum-core/Cargo.toml
@@ -9,7 +9,7 @@ license = "MIT"
 name = "axum-core"
 readme = "README.md"
 repository = "https://github.com/tokio-rs/axum"
-version = "0.2.6" # remember to also bump the version that axum depends on
+version = "0.2.6" # remember to also bump the version that axum and axum-extra depends on
 
 [dependencies]
 async-trait = "0.1"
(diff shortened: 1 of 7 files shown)


axum-extra/src/either.rs
```diff
@@ -184,17 +219,22 @@ pub enum Either8<E1, E2, E3, E4, E5, E6, E7, E8> {
...
+
 macro_rules! impl_traits_for_either {
     (
         $either:ident =>
         [$($ident:ident),* $(,)?],
         $last:ident $(,)?
     ) => {
         #[async_trait]
-        impl<S, B, $($ident),*, $last> FromRequest<S, B> for $either<$($ident),*, $last>
+        impl<S, B, $($ident),*, $last> FromRequest<S, B, private::EitherOnce> for $either<$($ident),*, $last>
```
davidpdrsn, line 234: This required the same trick as the tuples.

axum-core/Cargo.toml
```diff
@@ -9,7 +9,7 @@ license = "MIT"
 name = "axum-core"
 readme = "README.md"
 repository = "https://github.com/tokio-rs/axum"
-version = "0.2.6" # remember to also bump the version that axum depends on
+version = "0.2.6" # remember to also bump the version that axum and axum-extra depends on
```
jplatte, line 12:
```suggestion
version = "0.2.6" # remember to also bump the version that axum and axum-extra depend on
```

axum-extra/src/either.rs
```diff
@@ -184,17 +219,22 @@ pub enum Either8<E1, E2, E3, E4, E5, E6, E7, E8> {
...
         where
             $($ident: FromRequestParts<S>),*,
-            $last: FromRequest<S, B>,
+            $last: FromRequest<S, B, axum_core::extract::private::Once>,
```
jplatte, line 237: I wonder how useful this impl is at all, with just the last of the alternatives being allowed to take the request. Maybe we should only provide `FromRequestParts`? We would also need another impl with `TupleOnce` instead `Once` if we wanted to have a request-consuming tuples as `Either`s last type parameter.

davidpdrsn: Yeah I agree. I've removed it.

jplatte (approved):

PR #1288, opened 2022-08-21
New FromRequest[Parts] trait cleanup

jplatte:
Who would have thought that it's so easy? 😅 (the tuple stuff)

### DIFF
--- a/axum-core/src/extract/tuple.rs
+++ b/axum-core/src/extract/tuple.rs
@@ -16,11 +16,6 @@ where
     }
 }
 
-mod private {
-    #[derive(Debug, Clone, Copy)]
-    pub enum TupleOnce {}
-}
-
 macro_rules! impl_from_request {
     (
         [$($ty:ident),*], $last:ident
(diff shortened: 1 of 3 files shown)


axum-core/src/extract/tuple.rs
```diff
@@ -53,10 +48,10 @@ macro_rules! impl_from_request {
         // implementation of `FromRequest<S, B, Mut>` for `T: FromRequestParts<S>`.
         #[async_trait]
         #[allow(non_snake_case, unused_mut, unused_variables)]
-        impl<S, B, $($ty,)* $last> FromRequest<S, B, private::TupleOnce> for ($($ty,)* $last,)
+        impl<S, B, $($ty,)* $last> FromRequest<S, B> for ($($ty,)* $last,)
         where
             $( $ty: FromRequestParts<S> + Send, )*
-            $last: FromRequest<S, B, super::private::Once> + Send,
+            $last: FromRequest<S, B> + Send,
```
jplatte, line 54: If we want to write some docs about how to write generic extractors that wrap other extractors (that might or might not consume the request), this seems like pretty straight-forward advice: Implement both `FromRequestParts` and `FromRequest`, without any generic-ness over `M` in the latter case.

jplatte:
One more thought: Maybe we should rename `Mut` to `ViaParts` now that the traits are no longer `FromRequestOnce`, `FromRequestMut`? Not sure about the `Once` name.

davidpdrsn (approved):
Haha awesome!

axum-core/src/extract/tuple.rs
davidpdrsn: Yes! Its on the todo list [here](https://github.com/tokio-rs/axum/pull/1272)

davidpdrsn:
> One more thought: Maybe we should rename Mut to ViaParts now that the traits are no longer FromRequestOnce, FromRequestMut? Not sure about the Once name.

@jplatte `ViaParts` and `ViaRequest` maybe?

PR #1293, opened 2022-08-22
Add `FromRequest` changes to changelogs

davidpdrsn:


axum/CHANGELOG.md
```diff
@@ -237,18 +237,48 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  axum enforces this by only allowing the _last_ extractor to consume the
+  request.
+
+  For example:
+
+  ```rust
+  use axum::{Json, http::HeaderMap};
+
+  // this wont compile on 0.6 because both `Json` and `String` needs to consume
+  // the request body. You must use either `Json` or `String` but not both.
```
jplatte, line 253:
```suggestion
  // This wont compile on 0.6 because both `Json` and `String` need to consume
  // the request body. You can use either `Json` or `String`, but not both.
```

axum/CHANGELOG.md
```diff
@@ -237,18 +237,48 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  async fn handler_1(
+      json: Json<serde_json::Value>,
+      string: String,
+  ) {}
+
+  // this wont either since `Json` is not the last extractor
```
jplatte, line 259:
```suggestion
  // This won't work either since `Json` is not the last extractor.
```

axum/CHANGELOG.md
```diff
@@ -237,18 +237,48 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  async fn handler_2(
+      json: Json<serde_json::Value>,
+      headers: HeaderMap,
+  ) {}
+
+  // this works!
```
jplatte, line 265:
```suggestion
  // This works!
```

axum/CHANGELOG.md
```diff
@@ -266,22 +296,53 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
       }
   }
   ```
 
+- **breaking:** `RequestParts` has been removed as part of the `FromRequest`
+  rework ([#1272])
+- **breaking:** `BodyAlreadyExtracted` has been removed ([#1272])
+- **breaking:** The following types or traits have a new `S` type param
+  (`()` by default) which represents the state ([#1155]):
+  - `FromRequest`
```
jplatte, line 341: This isn't right, `FromRequest` doesn't have the `()` default. `Handler` does I guess, but should it actually?

davidpdrsn: > `Handler` does I guess, but should it actually?

Hm good question 🤔 The main reason for the defaults in the first place was to make `impl Handler` easier to use. But I don't think thats possible anymore since the `M` types are private. I don't mind that actually, `State` and closure captures is easier to use anyway.

So I've removed the default for `Handler` since we probably don't need it.

jplatte (approved):

Issue #1298, opened 2022-08-22
`#[debug_handler]` doesn't generate good error when non-last extractor doesn't implement `FromRequestParts`


davidpdrsn:
The error message for

```rust
use axum::{http::Method, routing::get, Router};
use axum_macros::debug_handler;

#[tokio::main]
async fn main() {
    let _: Router = Router::new().route("/", get(handler));
}

#[debug_handler]
async fn handler(_: String, _: Method) {}
```

Is still the bad one from rust

```
    Checking example-hello-world v0.1.0 (/Users/davidpdrsn/dev/major/axum/examples/hello-world)
error[E0277]: the trait bound `fn(String, Method) -> impl Future<Output = ()> {handler}: Handler<_, _, _>` is not satisfied
   --> hello-world/src/main.rs:12:50
    |
12  |     let _: Router = Router::new().route("/", get(handler));
    |                                              --- ^^^^^^^ the trait `Handler<_, _, _>` is not implemented for `fn(String, Method) -> impl Future<Output = ()> {handler}`
    |                                              |
    |                                              required by a bound introduced by this call
    |
    = help: the trait `Handler<T, S, B>` is implemented for `Layered<L, H, T, S, B>`
note: required by a bound in `axum::routing::get`
   --> /Users/davidpdrsn/dev/major/axum/axum/src/routing/method_routing.rs:400:1
    |
400 | top_level_handler_fn!(get, GET);
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `axum::routing::get`
    = note: this error originates in the macro `top_level_handler_fn` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I thought I handled this in https://github.com/tokio-rs/axum/pull/1283 but apparently not.

The code that gets generated is

```rust
async fn handler(_: String, _: Method) {}
#[allow(warnings)]
fn __axum_macros_check_handler_0_from_request_check<M>()
where
    String: ::axum::extract::FromRequest<(), axum::body::Body, M> + Send,
{
}
#[allow(warnings)]
fn __axum_macros_check_handler_0_from_request_call_check() {
    __axum_macros_check_handler_0_from_request_check();
}
#[allow(warnings)]
fn __axum_macros_check_handler_1_from_request_check<M>()
where
    Method: ::axum::extract::FromRequest<(), axum::body::Body, M> + Send,
{
}
#[allow(warnings)]
[...]

PR #1299, opened 2022-08-22
Generate correct bound for non-last extractor in `#[debug_handler]`

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1298

While working on this I also discovered another annoying possible bug. If you implement both `FromRequest` and `FromRequestParts` that wont work:

```rust
fn main() {
    let _: Router = Router::new().route("/", get(handler));
}

async fn handler(_: A, _: A) {}

struct A {}

#[async_trait]
impl<S, B> FromRequest<S, B> for A
where
    B: Send + 'static,
    S: Send + Sync,
{
    type Rejection = ();
    async fn from_request(req: Request<B>, state: &S) -> Result<Self, Self::Rejection> {
        unimplemented!()
    }
}

#[async_trait]
impl<S> FromRequestParts<S> for A
where
    S: Send + Sync,
{
    type Rejection = ();
    async fn from_request_parts(_parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
        unimplemented!()
    }
}
```

This fails with

```
error[E0283]: type annotations needed
   --> hello-world/src/main.rs:16:46
    |
16  |     let _: Router = Router::new().route("/", get(handler));
    |                                              ^^^ cannot infer type for type parameter `M`
    |
note: multiple `impl`s satisfying `A: FromRequest<(), Body, _>` found
   --> hello-world/src/main.rs:24:1
    |
24  | / impl<S, B> FromRequest<S, B> for A
25  | | where
26  | |     B: Send + 'static,
27  | |     S: Send + Sync,
...   |
32  | |     }
33  | | }
    | |_^
    = note: and another `impl` found in the `axum_core` crate:
            - impl<S, B, T> FromRequest<S, B, axum_core::extract::private::ViaParts> for T
              where B: 'static, B: Send, S: Send, S: Sync, T: FromRequestParts<S>;
    = note: required because of the requirements on the impl of `Handler<(_, A, A), ()>` for `fn(A, A) -> impl Future<Output = ()> {handler}`
note: required by a bound in `axum::routing::get`
   --> /Users/davidpdrsn/dev/major/axum/axum/src/routing/method_routing.rs:400:1
    |
400 | top_level_handler_fn!(get, GET);
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `axum::routing::get
```

There is no `#[debug_handler]` used here so its not the macros fault. However it doesn't improve the error either.

I'm not sure if this is something that can be fixed but might be worth exploring and possibly documenting.

---

I also added support for setting `AXUM_TEST_ONLY` to only run a single UI test. I often need to iterate on just one test and waiting for everything else to run is both slow and confusing since trybuild produces so much output.

Example usage:

```
AXUM_TEST_ONLY=tests/debug_handler/fail/doesnt_implement_from_request_parts.rs ca t --all-features -p axum-macros --lib debug_handler
```

### DIFF
--- a/axum-macros/src/with_position.rs
+++ b/axum-macros/src/with_position.rs
@@ -1,0 +1,116 @@
+// this is copied from itertools under the following license
+//
+// Copyright (c) 2015
+//
+// Permission is hereby granted, free of charge, to any
+// person obtaining a copy of this software and associated
+// documentation files (the "Software"), to deal in the
+// Software without restriction, including without
+// limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of
+// the Software, and to permit persons to whom the Software
+// is furnished to do so, subject to the following
+// conditions:
+//
+// The above copyright notice and this permission notice
+// shall be included in all copies or substantial portions
+// of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF
+// ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+// TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+// PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT
+// SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
+// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+use std::iter::{Fuse, FusedIterator, Peekable};
+
+pub(crate) struct WithPosition<I>
+where
+    I: Iterator,
+{
+    handled_first: bool,
+    peekable: Peekable<Fuse<I>>,
+}
+
+impl<I> WithPosition<I>
+where
+    I: Iterator,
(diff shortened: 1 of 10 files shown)


jplatte:
> While working on this I also discovered another annoying possible bug. If you implement both `FromRequest` and `FromRequestParts` that wont work:
>
> I'm not sure if this is something that can be fixed but might be worth exploring and possibly documenting.

This is impossible to fix, the whole `Handler` thing only works when `Handler` is implemented exactly once for a given type. It really is an abuse of rustc's type inference which in some situations selects the single applicable impl when it probably should never have allowed that for forwards compatibility (i.e. because it's generally considered a non-breaking change to add a trait implementation). You could already implement `Handler<Foo, ..>` and `Handler<Bar, ..>` for a single type in any previous version of axum, and get the same error.

So yeah, let's document that this is only allowed for extractors generically wrapping another extractor that may or may not consume the request 😅

axum-macros/src/with_position.rs
```diff
@@ -0,0 +1,116 @@
+// this is copied from itertools under the following license
+//
+// Copyright (c) 2015
```
jplatte, line 3: Copyright nobody? 😄

jplatte (approved):
Regarding a "only implements one of `FromRequest` and `FromRequestParts`" check, you could have a look at what [`static_assertions` does](https://docs.rs/static_assertions/latest/static_assertions/macro.assert_impl_one.html).

axum-macros/src/with_position.rs
davidpdrsn: Yep 🤷 https://github.com/rust-itertools/itertools/blob/master/LICENSE-MIT

PR #1300, opened 2022-08-22
Document limitation with implementing both `FromRequest` and `FromRequestParts`

davidpdrsn:
Documents the shortcoming discussed in https://github.com/tokio-rs/axum/pull/1299

### DIFF
--- a/axum/src/docs/extract.md
+++ b/axum/src/docs/extract.md
@@ -499,6 +499,65 @@ let app = Router::new().route("/foo", get(handler));
 # };
 ```
 
+## Cannot implement both `FromRequest` and `FromRequestParts`
+
+Note that you cannot implement both `FromRequest` and `FromRequestParts` for the
+same type, unless it is wrapping another extractor:
+
+```rust,compile_fail
+use axum::{
+    Router,
+    routing::get,
+    extract::{FromRequest, FromRequestParts},
+    http::{Request, request::Parts},
+    async_trait,
+};
+use std::convert::Infallible;
+
+// Some extractor that doesn't wrap another extractor
+struct MyExtractor;
+
+// `MyExtractor` implements both `FromRequest`
+#[async_trait]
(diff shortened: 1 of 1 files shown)


axum/src/docs/extract.md
```diff
@@ -499,6 +499,67 @@ let app = Router::new().route("/foo", get(handler));
 # };
 ```
 
+## Cannot implement both `FromRequest` and `FromRequestParts`
+
+Note that you cannot implement both `FromRequest` and `FromRequestParts` for the
+same type, unless it is wrapping another extractor:
```
jplatte, line 505: We should probably note again here the blanket impl, plus make it clear that there's not going to be a compile error from the two direct impls.

```suggestion
Note that you will make your extractor unusable by implementing both `FromRequest` and `FromRequestParts` directly for the
same type, unless it is wrapping another extractor:
```

axum/src/docs/extract.md
```diff
@@ -499,6 +499,67 @@ let app = Router::new().route("/foo", get(handler));
 # };
 ```
 
+## Cannot implement both `FromRequest` and `FromRequestParts`
+
+Note that you will make your extractor unusable by implementing both `FromRequest` and `FromRequestParts` directly for the
+same type, unless it is wrapping another extractor:
```
jplatte, line 505: Now you need to re-wrap this 😛 

Also how about mentioning the blanket impl of `FromRequest` here? I guess it could be confusing to people how they're allowed to implement both at all then?

davidpdrsn: > Also how about mentioning the blanket impl of FromRequest here? I guess it could be confusing to people how they're allowed to implement both at all then?

Hm how would you phrase that?

jplatte: I'm not sure. Maybe not that important? The whole handler thing is probably just "magic" to most users anyways 😅

jplatte (approved):

Issue #1297, opened 2022-08-22
Add helper methods to run easily run extractors


davidpdrsn:
In axum 0.5 you could call `request_parts.extract::<T>()` which is a bit nicer than `T::from_request(request_parts)` but https://github.com/tokio-rs/axum/pull/1272 removed the `RequestParts` type and thus also this method.

For 0.6 we should think about adding back something similar.

Haven't thought much about it but perhaps something like

```rust
pub trait RequestExt {
    async fn extract<T>(self) -> Result<(Self, T), T::Rejection>
    where
        T: FromRequest,
    {
        ...
    }
}

pub trait PartsExt {
    async fn extract<T>(&mut self) -> Result<T, T::Rejection>
    where
        T: FromRequestParts,
    {
        ...
    }
}
```

Should also think about how to pass state and whether we need separate methods for when the state is and isn't `()`.

jplatte:
We should also have `extract_parts` in `RequestExt`, so users don't need to do `into_parts` + `from_parts` themselves. Although calling that multiple times consecutively might be slower than doing one `into_parts` + `from_parts` and using `.extract` on the `Parts`.

PR #1301, opened 2022-08-22
Add `RequestExt` and `RequestPartsExt`

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1297

### DIFF
--- a/axum/src/ext_traits/request.rs
+++ b/axum/src/ext_traits/request.rs
@@ -1,0 +1,191 @@
+use async_trait::async_trait;
+use axum_core::extract::{FromRequest, FromRequestParts};
+use http::Request;
+
+mod sealed {
+    pub trait Sealed<B> {}
+    impl<B> Sealed<B> for http::Request<B> {}
+}
+
+/// Extension trait that adds additional methods to [`Request`].
+#[async_trait]
+pub trait RequestExt<B>: sealed::Sealed<B> + Sized {
+    /// Apply an extractor to this `Request`.
+    ///
+    /// This is just a convenience for `E::from_request(req, &())`.
+    ///
+    /// Note this consumes the request. Use [`RequestExt::extract_parts`] if you're not extracting
+    /// the body and don't want to consume the request.
+    async fn extract<E, M>(self) -> Result<E, E::Rejection>
+    where
+        E: FromRequest<(), B, M>;
+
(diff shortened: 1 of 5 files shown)


axum/src/ext_traits/request.rs
```diff
@@ -0,0 +1,200 @@
...
+        E: FromRequestParts<S> + 'static,
+        S: Send + Sync,
+    {
+        let mut req = Request::new(());
+        *req.version_mut() = self.version();
+        *req.method_mut() = self.method().clone();
+        *req.uri_mut() = self.uri().clone();
+        *req.headers_mut() = std::mem::take(self.headers_mut());
+        *req.extensions_mut() = std::mem::take(self.extensions_mut());
+        let (mut parts, _) = req.into_parts();
```
jplatte, line 94: Hm, this is kinda unfortunate. As an alternative, we could require `B: Default` and use `mem::take`, wdyt about that? Pretty much everything in `http-body` including the boxed body types seems to implement `Default` if the inner buffer does, the only exception being `Limited` but I'm sure there it can be added too.

davidpdrsn: Yeah thats probably fine. I'll change it.

davidpdrsn: Hm actually its kinda weird having to write this on your extractor. Feels kinda arbitrary from the user perspective 🤔

```diff
@@ -180,7 +171,7 @@ mod tests {
     impl<S, B> FromRequest<S, B> for WorksForCustomExtractor
     where
         S: Send + Sync,
-        B: Send + 'static,
+        B: Send + Default + 'static,
         String: FromRef<S> + FromRequest<(), B>,
     {
         type Rejection = <String as FromRequest<(), B>>::Rejection;
```

jplatte: Right, I didn't consider that you'd want to use this from functions generic over `B`. Maybe `http` can be updated to provide `.parts(&self) -> &Parts` and `.parts_mut(&mut self) -> &mut Parts` on `Request` and `Response`, the current implementation of the types would trivially allow this.

davidpdrsn: I remember there was some discussion around that a while ago https://github.com/hyperium/http/pull/511.

Issue #1120, opened 2022-06-25
Middleware cannot change the request in a way that impacts routing


davidpdrsn:
As part of removing trailing slash redirects (https://github.com/tokio-rs/axum/issues/1118) I've added a `NormalizePath` middleware to tower-http, however that requires changing the path before routing. That means `Router::layer` cannot be used, since those middleware only run after routing.

Instead you have to do something like

```rust
NormalizePath::trim_trailing_slash(
    Router::new().route("/foo", get(handler))
);
```

which isn't very obvious. It also means you cannot use `Router::into_make_service`, since your service is now `NormalizePath<Router>` and not `Router`. Instead you have to use `tower::make::Shared` or something similar.

We should think of a more ergonomic solution to this. Maybe a third `*_layer` method?

jplatte:
Maybe as an easy first step, `NormalizePath` itself could just have an `into_make_service` method? And have its docs call out that it should usually be (very close to) the outermost layer / has to run before routing.

davidpdrsn:
What if axum provided a trait like

```rust
pub trait ServiceExt<R>: Service<R> + Sized {
    fn into_make_service(self) -> IntoMakeService<Self>;

    // fn into_make_service_with_connect_info(self) -> ...
}
```

`Router::into_make_service` should still take precedence since that method exists directly on `Router`. But if you've added a `Layer` around a `Router` you can use `ServiceExt::into_make_service`.

We could technically remove `Router::into_make_service` but I'm not a fan of users having to always import that trait, although I suppose it is more consistent.

jplatte:
Sure, why not. Agree about keeping the inherent method.

PR #1302, opened 2022-08-22
Support turning any `Service` into a `MakeService`

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1120

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -364,7 +364,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   message points to where the user added the invalid route, rather than
   somewhere internally in axum ([#1248])
 - **added:** Add `ServiceExt` with methods for turning any `Service` into a
-  `MakeService` similarly to `Router::into_make_service`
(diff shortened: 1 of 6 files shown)


jplatte (commented):
Looks good, I just wonder: Is axum the right crate for this functionality? (as opposed to `hyper` / `tower-util` / `tower-http`)

davidpdrsn:
tower already has `Shared` which is identical to `IntoMakeService`. We could add a helper to `tower::ServiceExt`. But I kinda still wanna keep it in axum 😕 feels a bit more discoverable plus its one fewer dependency users have to add.

I don't think `IntoMakeServiceWithConnectInfo` could live in tower-http however. It uses the [`Connected`](https://docs.rs/axum/latest/axum/extract/connect_info/trait.Connected.html) trait which has an impl for [`AddrStream`](https://docs.rs/axum/latest/axum/extract/connect_info/trait.Connected.html#impl-Connected%3C%26AddrStream%3E-for-SocketAddr) from hyper. Don't think tower-http should have a public dependency on hyper.

jplatte (approved):

PR #1306, opened 2022-08-23
Show path in panic message when merging overlapping `MethodRouter`s

davidpdrsn:
When you do

```rust
Router::new()
    .route("/foo/bar", get(handler))
    .route("/foo/bar", get(handler));
```

or

```rust
Router::new().route("/foo/bar", get(handler)).merge(
    Router::new().route("/foo/bar", get(handler))
);
```

The panic message will now contain the path (`/foo/bar`) so the error is hopefully easier to track down. Example:

```
Overlapping method route. Handler for `GET /foo/bar` already exists
```

It doesn't work for

```rust
Router::new().route("/foo/bar", get(handler).get(handler));
```

because the `get(_).get(_)` panics before `Router::route` is called so we don't know the path at that point.

### DIFF
--- a/axum/src/routing/method_routing.rs
+++ b/axum/src/routing/method_routing.rs
@@ -877,16 +877,30 @@ where
         self
     }
 
-    #[doc = include_str!("../docs/method_routing/merge.md")]
     #[track_caller]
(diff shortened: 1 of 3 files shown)


jplatte (approved):
Nice!

PR #1308, opened 2022-08-23
Use `.fallback()` instead of (maybe old?) `.fallback_service()`

frederikhors:

### DIFF
--- a/examples/static-file-server/src/main.rs
+++ b/examples/static-file-server/src/main.rs
@@ -34,7 +34,7 @@ async fn main() {
     // as the fallback to a `Router`
     let app: _ = Router::new()
         .route("/foo", get(|| async { "Hi from /foo" }))
-        .fallback_service(get_service(ServeDir::new(".")).handle_error(handle_error))
+        .fallback(get_service(ServeDir::new(".")).handle_error(handle_error))
         .layer(TraceLayer::new_for_http());
 
     let addr = SocketAddr::from(([127, 0, 0, 1], 3000));


jplatte:
Have you tested this locally? It shouldn't actually work. `fallback` was changed recently to only accept handlers, not arbitrary services. (this change is going to be released in axum 0.6)

davidpdrsn:
Did you run `cargo check` for this? `fallback` is only for handler functions, which this isn't. Its a `Service`. This was recently changed on `main` but not published to crates.io yet.

jplatte:
Yeah, as CI shows this doesn't compile.

frederikhors:
> Have you tested this locally? It shouldn't actually work. `fallback` was changed recently to only accept handlers, not arbitrary services. (this change is going to be released in axum 0.6)

I'm actually using this code with no problems.

```rust
app = app
        .fallback(get_service(ServeDir::new("./static")).handle_error(handle_error))
        .layer(TraceLayer::new_for_http());
```

```toml
axum = "0.5.15"
hyper = { version = "0.14.20", features = ["client"] }
tokio = { version = "1.20.1", features = ["full"] }
tower-http = { version = "0.3.4", features = ["fs", "trace", "cors"] }
```

frederikhors:
I cannot find any `fallback_service()` method on my `app`.

jplatte:
Yes, in axum 0.5 there was no `.fallback_service`, we recently added it and adjusted `.fallback`s signature.

davidpdrsn:
You gotta check it on `main` since there might be differences, which there are in this case. See [this note](https://github.com/tokio-rs/axum#breaking-changes) in the README.

PR #1310, opened 2022-08-23
Bump version of all crates

davidpdrsn:
Lets get the release train going 🚀

I think we should look at https://github.com/tokio-rs/axum/pull/1054 during this week but I don't think it should block shipping the first rc.


davidpdrsn:
So I think I'm gonna yolo this. Would like to get it out before its too late here.

PR #1315, opened 2022-08-24
FromRequestParts does not have a 'B' type

yanns:
## Motivation

The release notes for 0.6-rc.1 contains the following code:
```
// implement `FromRequestParts` if you don't need to consume the request body
#[async_trait]
impl<S> FromRequestParts<S> for MyExtractor
where
    S: Send + Sync,
    B: Send + 'static,
{
    type Rejection = StatusCode;

        // ...
    }
}
```

`B: Send + 'static` cannot compile as `B` is not known for this implementation.

## Solution

Remove `B: Send + 'static`.

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -313,7 +313,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   impl<S> FromRequestParts<S> for MyExtractor
   where
       S: Send + Sync,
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
Thanks!

yanns:
@davidpdrsn You will have to edit https://github.com/tokio-rs/axum/releases/tag/axum-v0.6.0-rc.1 manually I guess.

PR #1317, opened 2022-08-24
add rust-embed-spa example

frederikhors:
## Motivation

I think we need an example like this.


davidpdrsn:
I think this example belongs in rust-embed's repo. It's quite specific to that crate and they already have [an axum example](https://github.com/pyrossh/rust-embed/blob/master/examples/axum.rs) that you can add to.

frederikhors:
Yeah. I know I opened https://github.com/pyrossh/rust-embed/pull/184 too.

But I think it doesn't hurt to host this example on this repo as well.

I do not want to insist, it is a wonderful project, you are wonderful people and if you decide that it is not correct to add it I Gladly ACCEPT the decision.

davidpdrsn:
I don't think having the example in both places is worth much. I'll close this since I think rust-embed is the right place for this to live.

PR #1320, opened 2022-08-24
Add example for parsing body based on `Content-Type`

davidpdrsn:
Gotten a few questions from people trying to do

```rust
async fn handler(
    form: Option<Form<BTreeMap<String, String>>>,
    multipart: Option<Multipart>,
) {
```

basically parse the body based on the content-type.

I think writing a custom extractor that inspects the content-type and delegates the corresponding extractor is a good way to solve it. So this adds an example for that.

### DIFF
--- a/examples/parse-body-based-on-content-type/src/main.rs
+++ b/examples/parse-body-based-on-content-type/src/main.rs
@@ -1,16 +1,9 @@
 //! Provides a RESTful web server managing some Todos.
 //!
-//! API will be:
-//!
-//! - `GET /todos`: return a JSON list of Todos.
-//! - `POST /todos`: create a new Todo.
-//! - `PUT /todos/:id`: update a specific Todo.
-//! - `DELETE /todos/:id`: delete a specific Todo.
-//!
 //! Run with
 //!
 //! ```not_rust
-//! cd examples && cargo run -p example-todos
+//! cd examples && cargo run -p example-parse-body-based-on-content-type
 //! ```
 
 use axum::{
--- a/examples/parse-body-based-on-content-type/src/main.rs
+++ b/examples/parse-body-based-on-content-type/src/main.rs
@@ -57,14 +57,14 @@ enum JsonOrForm<T, K = T> {
 }
 
 #[async_trait]
-impl<S, B, T, K> FromRequest<S, B> for JsonOrForm<T, K>
+impl<S, B, T, U> FromRequest<S, B> for JsonOrForm<T, U>
 where
     B: Send + 'static,
     S: Send + Sync,
     Json<T>: FromRequest<(), B>,
-    Form<K>: FromRequest<(), B>,
+    Form<U>: FromRequest<(), B>,
     T: 'static,
-    K: 'static,
+    U: 'static,
 {
     type Rejection = Response;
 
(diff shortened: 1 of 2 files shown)


examples/parse-body-based-on-content-type/src/main.rs
```diff
@@ -0,0 +1,98 @@
...
+//! - `GET /todos`: return a JSON list of Todos.
+//! - `POST /todos`: create a new Todo.
+//! - `PUT /todos/:id`: update a specific Todo.
+//! - `DELETE /todos/:id`: delete a specific Todo.
+//!
+//! Run with
+//!
+//! ```not_rust
+//! cd examples && cargo run -p example-todos
+//! ```
```
jplatte, line 14: Copy-paste error

examples/parse-body-based-on-content-type/src/main.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+async fn handler(payload: JsonOrForm<Payload>) -> Response {
+    match payload {
+        JsonOrForm::Json(payload) => Json(payload).into_response(),
+        JsonOrForm::Form(payload) => Form(payload).into_response(),
+    }
+}
+
+enum JsonOrForm<T, K = T> {
```
jplatte, line 61: Why is the second type parameter named `K`?

examples/parse-body-based-on-content-type/src/main.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        if content_type.starts_with("application/json") {
+            let Json(payload) = req.extract().await.map_err(IntoResponse::into_response)?;
+            return Ok(Self::Json(payload));
+        }
+
+        if content_type.starts_with("application/x-www-form-urlencoded") {
+            let Form(payload) = req.extract().await.map_err(IntoResponse::into_response)?;
+            return Ok(Self::Form(payload));
+        }
```
jplatte, line 94:
```suggestion
        let content_type_header = req.headers().get(CONTENT_TYPE);
        let content_type = content_type_header.and_then(|value| value.to_str().ok());

        if let Some(content_type) = content_type {
            if content_type.starts_with("application/json") {
                let Json(payload) = req.extract().await.map_err(IntoResponse::into_response)?;
                return Ok(Self::Json(payload));
            }

            if content_type.starts_with("application/x-www-form-urlencoded") {
                let Form(payload) = req.extract().await.map_err(IntoResponse::into_response)?;
                return Ok(Self::Form(payload));
            }
        }
```

examples/parse-body-based-on-content-type/src/main.rs
davidpdrsn: Might be over engineered but the idea was to be able to extract different types for json and form requests.

jplatte: Yeah I got that, the question was just why `T` and `K`, rather than e.g. `T` and `U` or `A` and `B` which are a much more common combination for generic parameters without specific meaning.

davidpdrsn: Ah okay. No particular reason. I'll change it to A and B

davidpdrsn: Actually I'll use T and U because B is taken by the request body.

jplatte (approved):

examples/parse-body-based-on-content-type/src/main.rs
davidpdrsn: I'll just leave it like this :P

Issue #1321, opened 2022-08-24
`Router::with_state_arc` might be misleading


denschub:
Note that this probably is neither a Bug Report or a Feature Request, and just a me-problem. But I got confused by it and @davidpdrsn asked me to file a bug on Reddit. :)

### Version

0.6.0-rc.1

### Description

I have a small app that has a non-Clone'able state, so I've been using an `Arc` in an `Extension` for now. I just poked around with the new State in `0.6.0-rc.1`, and found `axum::routing::Router::with_state_arc`.

To me, this function felt like an invitation to pass an Arc to something Unclonable into it:

```rs
struct UnclonableExample(usize);

#[tokio::main]
async fn main() {
    let state = UnclonableExample(42);
    let state_arc = Arc::new(state);
    let router = Router::with_state_arc(state_arc).route("/", get(handler));
    // ...
}

async fn handler(State(state): State<Arc<UnclonableExample>>) -> impl IntoResponse {
    Html("<h1>foo</h1>")
}
```

But that doesn't work. This specific example fails with

<details>

```
error[E0308]: mismatched types
   --> src/main.rs:16:63
    |
16  |     let router = Router::with_state_arc(state_arc).route("/", get(handler));
    |                                                    -----      ^^^^^^^^^^^^ expected struct `UnclonableExample`, found struct `Arc`
    |                                                    |
    |                                                    arguments to this function are incorrect
    |
    = note: expected struct `MethodRouter<UnclonableExample, _>`
               found struct `MethodRouter<Arc<UnclonableExample>, _>`
note: associated function defined here
   --> /Users/denschub/.cargo/registry/src/github.com-1ecc6299db9ec823/axum-0.6.0-rc.1/src/routing/mod.rs:154:12
    |
154 |     pub fn route(mut self, path: &str, method_router: MethodRouter<S, B>) -> Self {
    |            ^^^^^

For more information about this error, try `rustc --explain E0308`.
```

</details>

and changing the function signature to `State<UnclonableExample>` of course also doesn't work, with a rather cryptic compiler error:

<details>

```
error[E0277]: the trait bound `fn(State<UnclonableExample>) -> impl Future<Output = impl IntoResponse> {handler}: Handler<_, _, _>` is not satisfied
   --> src/main.rs:16:67
    |
16  |     let router = Router::with_state_arc(state_arc).route("/", get(handler));
[...]

PR #1324, opened 2022-08-25
[doc] Added notes about extractor precedence

chrisglass:
Added a note in the structs docs  pointing to the relevant part of the documentation regarding extractor precedence.

## Motivation

During discussion of a slightly unrelated problem (https://github.com/tokio-rs/axum/issues/1321), it was discovered that my example was ordering the extractors wrong, causing a compile-time problem.

## Solution

This PR should hopefully make the solution for the above problem (reordering extractors) a little bit more discoverable.

### DIFF
--- a/axum/src/extract/content_length_limit.rs
+++ b/axum/src/extract/content_length_limit.rs
@@ -6,8 +6,16 @@ use std::ops::Deref;
 
 /// Extractor that will reject requests with a body larger than some size.
 ///
+///
 /// `GET`, `HEAD`, and `OPTIONS` requests are rejected if they have a `Content-Length` header,
 /// otherwise they're accepted without the body being checked.
+///
+/// Note: `ContentLengthLimit` can wrap types that extract the body (for example, Form or Json)
+/// if that is the case, the inner type will consume the request's body, which means the
+/// ContentLengthLimit must come *last* if the handler uses several extractors. See
+/// ["the order of extractors"][order-of-extractors]
+///
+/// [order-of-extractors]: crate::extract#the-order-of-extractors
 ///
 /// # Example
 ///
--- a/axum/src/extract/content_length_limit.rs
+++ b/axum/src/extract/content_length_limit.rs
@@ -6,7 +6,6 @@ use std::ops::Deref;
 
 /// Extractor that will reject requests with a body larger than some size.
 ///
-///
 /// `GET`, `HEAD`, and `OPTIONS` requests are rejected if they have a `Content-Length` header,
 /// otherwise they're accepted without the body being checked.
 ///
--- a/axum/src/extract/content_length_limit.rs
+++ b/axum/src/extract/content_length_limit.rs
@@ -15,6 +15,8 @@ use std::ops::Deref;
 /// ["the order of extractors"][order-of-extractors]
 ///
 /// [order-of-extractors]: crate::extract#the-order-of-extractors
+/// [`Form`]: crate::form::Form
+/// [`Json`]: crate::json::Json
 ///
 /// # Example
 ///
--- a/axum/src/extract/content_length_limit.rs
+++ b/axum/src/extract/content_length_limit.rs
@@ -9,9 +9,9 @@ use std::ops::Deref;
 /// `GET`, `HEAD`, and `OPTIONS` requests are rejected if they have a `Content-Length` header,
 /// otherwise they're accepted without the body being checked.
 ///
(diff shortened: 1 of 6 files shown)


axum/src/form.rs
```diff
@@ -16,6 +16,11 @@ use std::ops::Deref;
 /// If used as an extractor `Form` will deserialize `application/x-www-form-urlencoded` request
 /// bodies into some target type via [`serde::Deserialize`].
 ///
+/// Since parsing form data requires consuming the request body, the `Form` extractor must be
+/// *last* if there are multiple extractors in a handler. See ["the order of extractors"][order-of-extractors]
+///
+/// [order-of-extractors]: extract/index.html#the-order-of-extractors
```
davidpdrsn, line 22: Should use an intra docs link

```suggestion
/// [order-of-extractors]: crate::extract#the-order-of-extractors
```

axum/src/json.rs
```diff
@@ -25,6 +25,11 @@ use std::ops::{Deref, DerefMut};
 /// type.
 /// - Buffering the request body fails.
 ///
+/// Since parsing JSON requires consuming the request body, the `Json` extractor must be
+/// *last* if there are multiple extractors in a handler. See ["the order of extractors"][order-of-extractors]
+///
+/// [order-of-extractors]: extract/index.html#the-order-of-extractors
```
davidpdrsn, line 31:
```suggestion
/// [order-of-extractors]: crate::extract#the-order-of-extractors
```

davidpdrsn (changes requested):
I think this is worth having!

I think we should add it to the other types we have that only implement `FromRequest:
- `BodyStream`
- `RawBody`
- `Multipart`
- `ContentLengthLimit`, note this one is conditional, only if what its wrapping implements `FromRequest`. The docs should mention that.

axum/src/form.rs
davidpdrsn: Oh you're right. We should change that on `State`. Should be an intra docs link as well.

chrisglass: Done.

axum/src/json.rs
chrisglass: Done.

chrisglass:
All comments are addressed (I think!)

axum/src/extract/content_length_limit.rs
```diff
@@ -6,9 +6,17 @@ use std::ops::Deref;
 
 /// Extractor that will reject requests with a body larger than some size.
 ///
+///
```
davidpdrsn, line 9: nit: do we need this blank line?

By the way you can click "apply suggestion" or "commit" on the suggestion in GitHub. You don't need to manually write the changes.

```suggestion
```

chrisglass: Yes, I know :) 
Since I had to change the content in my editor anyway, I just did it there.

chrisglass:
Well, CI failed :) will fix that later today.

axum/src/extract/content_length_limit.rs
```diff
@@ -9,6 +9,13 @@ use std::ops::Deref;
 /// `GET`, `HEAD`, and `OPTIONS` requests are rejected if they have a `Content-Length` header,
 /// otherwise they're accepted without the body being checked.
 ///
+/// Note: `ContentLengthLimit` can wrap types that extract the body (for example, Form or Json)
```
davidpdrsn, line 12:
```suggestion
/// Note: `ContentLengthLimit` can wrap types that extract the body (for example, [`Form`] or [`Json`])
```

axum/src/extract/content_length_limit.rs
```diff
@@ -9,6 +9,13 @@ use std::ops::Deref;
...
+/// if that is the case, the inner type will consume the request's body, which means the
+/// ContentLengthLimit must come *last* if the handler uses several extractors. See
```
davidpdrsn, line 14:
```suggestion
/// `ContentLengthLimit` must come *last* if the handler uses several extractors. See
```

axum/src/extract/multipart.rs
```diff
@@ -17,6 +17,12 @@ use std::{
 
 /// Extractor that parses `multipart/form-data` requests (commonly used with file uploads).
 ///
+/// Since extracting multipart form data from the request requires consuming it, the
```
davidpdrsn, line 20:
```suggestion
/// Since extracting multipart form data from the request requires consuming the body, the
```

davidpdrsn (approved):
Thanks!

PR #1327, opened 2022-08-25
Panic when attempting to add a route_layer to an empty router

jplatte:
Maybe we should also disallow / heavily discourage routers that only consist of a fallback and panic in the `layer` methods as well?


davidpdrsn (approved):
Makes sense! I've definitely heard some people who were tripped by this.

I think `Router::new().layer(...)` should be allowed, even though its a very roundabout way of handling all requests and adding a middleware to that.

PR #1332, opened 2022-08-26
Fix broken docs links

liigo:
(maybe this is an issue of docs.rs, which does not redirect to latest rc version's link)


davidpdrsn (approved):
Thanks!

PR #1334, opened 2022-08-26
Fix doc

randomairborne:
## Motivation
The doc comment wasn't updated to three state methods, rather then two.

## Solution
Change two to three. Though if one of the others is going to be removed, that works too


davidpdrsn (approved):
Thanks!

Issue #1284, opened 2022-08-21
Refactor macro argument parsing code


davidpdrsn:
The argument parsing code for `#[derive(TypedPath)]`, `#[derive(FromRequest)]`, and `#[debug_handler]` is pretty different. It would be nice to refactor it so it's consistent, while making sure its easy to add new arguments, validations etc.

`#[derive(FromRequest)]` has the most complex arguments so probably worth it to start with that and then bring the structure to the other macros.

I'm not sure using a dedicated argument parsing crate (like [darling](https://crates.io/crates/darling)) is worth it but could be convinced otherwise.

Issue #1314, opened 2022-08-24
`#[derive(FromRequest[Parts])]` doesn't work with `State`


davidpdrsn:
```rust
#[derive(FromRequest)]
struct Foo {
    state: State<String>,
}
```

fails with

```
error[E0277]: the trait bound `String: FromRef<S>` is not satisfied
 --> src/main.rs:6:12
  |
6 |     state: State<String>,
  |            ^^^^^^^^^^^^^ the trait `FromRef<S>` is not implemented for `String`
  |
  = note: required because of the requirements on the impl of `FromRequestParts<S>` for `State<String>`
help: consider extending the `where` clause, but there might be an alternative better way to express this requirement
  |
4 | #[derive(FromRequest, String: FromRef<S>)]
  |                     ++++++++++++++++++++
```

We can probably do something similar to what [`#[debug_handler]` does](https://docs.rs/axum-macros/0.3.0-rc.1/axum_macros/attr.debug_handler.html#changing-state-type) where it tries to infer the state, otherwise requires you to specify it.

PR #1339, opened 2022-08-29
Refactor macro argument parsing code

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1284

This alignes how argument parsing is done in the four macros. It also makes it easier to add more arguments in the future (required for https://github.com/tokio-rs/axum/issues/1314)

### DIFF
--- a/axum-macros/src/debug_handler.rs
+++ b/axum-macros/src/debug_handler.rs
@@ -5,11 +5,7 @@ use crate::{
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote, quote_spanned};
 use std::collections::HashSet;
-use syn::{
-    parse::{Parse, ParseStream},
-    spanned::Spanned,
-    FnArg, ItemFn, Token, Type,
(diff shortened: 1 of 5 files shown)


jplatte (commented):
Are you sure all these abstractions, especially the macros, are pulling their weight?

genusistimelord:
I think the Macro's do indeed help reduce the overall coding people will need to do when dealing with importing multiple Extensions, a State and other FromRequestPart calls. I can honestly say it was a lot of code reduced due to these macros.  Downside is Macro's are harder to maintain and update for new stuff right away. 

Probably just need to make more People aware that they Exist.

jplatte:
I was talking about the new internal `macro_rules!` macros, not whether the proc-macros are pulling their weight 😄

davidpdrsn:
I have actually worked on a very similar thing at embark where I didn't make any new macros. I'll take another pass at this at some point.

PR #1346, opened 2022-09-03
Limit size of request bodies in `Bytes` extractor

davidpdrsn:
## Motivation

We recently received a report of a vulnerability in axum caused by `Bytes::from_request` calling [`hyper::body::into_bytes`](https://docs.rs/hyper/latest/hyper/body/fn.to_bytes.html) directly without setting a limit. This meant if someone sent an infinite request body the server would attempt to buffer the whole thing in memory and eventually OOM.

This also applies to extractors that calls `Bytes::from_request` internally like `String`, `Json`, and `Form`.

## Solution

I think the right move is to set a default limit on how much `Bytes::from_request` will consume. I don't think just documenting it is enough since users might not notice.

This PR fixes it by wrapping the body in [`http_body::Limited`](https://docs.rs/http-body/latest/http_body/struct.Limited.html) before calling `into_bytes`. This can be disabled by adding the new `DefaultBodyLimit::disable()` middleware. That sets a private extension which `Bytes::from_request` looks for.

Once this is merged I'll backport this to 0.5

### DIFF
--- a/axum-core/src/extract/request_parts.rs
+++ b/axum-core/src/extract/request_parts.rs
@@ -1,4 +1,6 @@
-use super::{rejection::*, FromRequest, FromRequestParts};
+use super::{
+    default_body_limit::DefaultBodyLimitDisabled, rejection::*, FromRequest, FromRequestParts,
+};
 use crate::BoxError;
 use async_trait::async_trait;
 use bytes::Bytes;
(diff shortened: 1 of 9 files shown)


axum-core/src/extract/request_parts.rs
```diff
@@ -82,11 +84,20 @@ where
...
 
     async fn from_request(req: Request<B>, _: &S) -> Result<Self, Self::Rejection> {
-        let body = req.into_body();
-
-        let bytes = crate::body::to_bytes(body)
-            .await
-            .map_err(FailedToBufferBody::from_err)?;
+        // update docs in `axum-core/src/extract/default_body_limit.rs` and
+        // `axum/src/docs/extract.md` if this changes
+        const DEFAULT_LIMIT: usize = 2_097_152; // 2 mb
```
davidpdrsn, line 89: I picked 2mb because thats what actix-web uses for their `Json` extractor.

jplatte:
If we do this, why not let the user configure the limit using `DefaultBodyLimit`, rather than requiring a separate middleware to be used when one wants a different limit than 2MB?

davidpdrsn:
> If we do this, why not let the user configure the limit using `DefaultBodyLimit`, rather than requiring a separate middleware to be used when one wants a different limit than 2MB?

Yeah we could do that. I did consider it but kinda wanted to this as minimal as possible and then add more stuff as we go.

yanns:
For those who are already using `RequestBodyLimitLayer`, we should disable the new layer right?

davidpdrsn:
@yanns yes. See https://docs.rs/axum/latest/axum/extract/struct.DefaultBodyLimit.html

davidpdrsn:
Yes. That is also in the docs 😅

> Note that if an extractor consumes the body directly with `Body::data`, or similar, the default limit is not applied.

DanielJoyce:
What version of axum-core is this in and has it been pushed?

davidpdrsn:
@DanielJoyce yes. See the changelog.

PR #1352, opened 2022-09-06
expose FromRequest and FromRequestParts macros in axum

tamasfe:
I would expect to just `#[derive(FromRequest)]` if the `macros` feature is enabled, this tiny PR enables that.

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -13,6 +13,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   without any routes will now result in a panic. Previously, this just did
   nothing. [#1327]
 
+## Extractors
+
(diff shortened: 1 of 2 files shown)


davidpdrsn (commented):
Sure!

How does the docs look for this? Since you're exporting a proc-macro from a non-root module. Does that work?

Also, wanna add a note to the changelog as well?

tamasfe:
The docs seem to be inlined so it's the same as in the `axum-macros` crate, the `axum` links work as well since they point to `docs.rs`.

<details>
  <summary>screenshot</summary>

image 
</details>

I've added an entry to the changelog.

jplatte (approved):

davidpdrsn (approved):

RReverser:
FWIW the changelog seems wrong - this somehow moved from "unreleased" to "0.6.0-rc.2" section

image

but 0.6.0-rc.2 doesn't have the derives yet:

image

davidpdrsn:
@RReverser thanks for noticing. I pushed a fix in a57bd9a118b34.

PR #1363, opened 2022-09-10
Ship rc.2

davidpdrsn:
To get https://github.com/tokio-rs/axum/pull/1346 out.

@jplatte do you if there is a way to avoid bumping the version of axum-macros? It doesn't have any changes but currently have to bump it so it depends on the new version of axum.

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -9,7 +9,7 @@ license = "MIT"
 name = "axum-extra"
 readme = "README.md"
 repository = "https://github.com/tokio-rs/axum"
-version = "0.4.0-rc.1"
+version = "0.4.0-rc.2"
 
 [features]
 default = []
@@ -34,7 +34,7 @@ spa = ["tower-http/fs"]
 typed-routing = ["dep:axum-macros", "dep:serde", "dep:percent-encoding"]
 
 [dependencies]
-axum = { path = "../axum", version = "0.6.0-rc.1", default-features = false }
+axum = { path = "../axum", version = "0.6.0-rc.2", default-features = false }
 bytes = "1.1.0"
 futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
 http = "0.2"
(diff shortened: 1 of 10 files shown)


jplatte:
I don't think you have to do anything special. The dev-deps of axum-macros don't really matter. Everything should work if you just don't re-release it.

This time you should use exact version constraints for everything though 😉

davidpdrsn:
That makes sense!

I suppose we still need a new release of axum-extra since it depends on axum as a regular dependency and not a dev-dep.

axum/Cargo.toml
```diff
@@ -30,7 +30,7 @@ __private_docs = ["tower/full", "tower-http/full"]
 
 [dependencies]
 async-trait = "0.1.43"
-axum-core = { path = "../axum-core", version = "0.3.0-rc.1" }
+axum-core = { path = "../axum-core", version = "0.3.0-rc.2" }
```
jplatte, line 33:
```suggestion
axum-core = { path = "../axum-core", version = "=0.3.0-rc.2" }
```

jplatte:
> I suppose we still need a new release of axum-extra since it depends on axum as a regular dependency and not a dev-dep.

Well, since it doesn't have an exact version constraint, the existing `-rc.1` will support depending on `-rc.2` of axum, actually.

davidpdrsn:
Does that mean we should do

```diff
-axum = { path = "../axum", version = "0.6.0-rc.2", default-features = false }
+axum = { path = "../axum", version = "=0.6.0-rc.2", default-features = false }
```

for axum-extra?

axum-extra/Cargo.toml
```diff
@@ -34,7 +34,7 @@ spa = ["tower-http/fs"]
 typed-routing = ["dep:axum-macros", "dep:serde", "dep:percent-encoding"]
 
 [dependencies]
-axum = { path = "../axum", version = "0.6.0-rc.1", default-features = false }
+axum = { path = "../axum", version = "0.6.0-rc.2", default-features = false }
```
jplatte, line 37: I think we should do one of two things here: Either add an exact version dependency so future rc's or the full axum 0.6.0 don't get used as a dependency of axum-extra 0.4.0-rc.2; or not release another rc of axum-extra at all (it doesn't seem to be required, right?).

davidpdrsn: I assumed it was required but if axum-extra-rc.1 can use axum-rc.2 then yeah we don't need a release of axum-extra. I assumed rc versions were considered incompatible.

jplatte: Yeah they really should be, but aren't. That's why adding the `=` to the axum-core dependency is important.

davidpdrsn: Alright I've reverted the new version of axum-extra.

jplatte (approved):

davidpdrsn:
I will publish this tomorrow.

Issue #1326, opened 2022-08-25
Nesting/merging routers that use state could be more ergonomic


davidpdrsn:
If you wanna nest routers that use state you currently have to pass the state to each router:

```rust
let state = AppState {};
let api = Router::with_state(state.clone());
let app = Router::with_state(state).nest("/api", api);
```

This is a little inconvenient but its also a source of confusion.

Which state will requests to `/api/*` use? `api`'s state or `app`'s state? Of course in this example it doesn't matter such since we clone the state. But in general its still confusing.

With the current implementation it'll use `api`'s state but it would be great if this wasn't even a question, i.e. the state was only set once.

For 0.6.0 I think we should explore some alternative APIs. @jplatte suggested this which I quite like

```rust
impl<S> Router<S> {
    fn new() -> Self; // this is in the generic impl now

    // need to supply state to turn router into a service
    fn into_service(state: S) -> RouterService<S>;
}

let app = Router::new()
    // nest is updated to take a Router with the same state param
    .nest("/api", Router::new().route("/", get(handler))
    // nest_service is added for arbitrary services, consistent with route and stuff
    .nest_service(
        "/foo",
        // this router uses different state, so it's nested with nest_service
        Router::new()
            .route("/", get(other_handler))
            .into_service(other_state),
    )
    .into_make_service(main_state);
```

All this applies to `Router::merge` as well.

Altair-Bueno:
It's similar to my original proposal, but instead of manually adding `Layer`s, the service is built using a `Router` method. My original `State<Fallback<Router<S>>>` would be equivalent to `RouterService` but without the `Fallback` component

The fallback was meant to transform a `Service<RequestParts, Response = R, Error = RouterError>` into `Service<RequestParts, Response = R, Error = Infalible>` by applying some `Fn`

By allowing the Router to return `RouterError`, we ensure that Fallbacks can be captured on parent routers. We just need to provide a `From<Infalible> for RouterError` impl in order to nest routers

```rs
pub enum RouterError {
	NotFound,
}

let nested = Router::new().route("/foo", get(foo))
// nested.with_state(()) creates a
// `Service<RequestParts<()>, Response = R, Error = RouterError>`

let nested_with_fallback = Fallback::new(|x|"Some message").layer(router)
// nested_with_fallback.with_state(()) creates a 
// `Service<RequestParts<()>, Response = R, Error = Infalible>`

let parent = Router::new()
	// /baz/foo/err will be caught and transformed into a response
[...]

PR #1364, opened 2022-09-10
Small routing module refactoring

jplatte:
Small things I found while working on #1326.

### DIFF
--- a/axum/src/routing/method_routing.rs
+++ b/axum/src/routing/method_routing.rs
@@ -917,8 +917,7 @@ where
                         panic!("Overlapping method route. Cannot merge two method routes that both define `{name}`")
                     }
                 }
-                (Some(svc), None) => Some(svc),
-                (None, Some(svc)) => Some(svc),
+                (Some(svc), None) | (None, Some(svc)) => Some(svc),
                 (None, None) => None,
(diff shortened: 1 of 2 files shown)


jplatte:
Sooo.. The test fails because previously, we had different fallback merging strategies for the `Router` and the `MethodRouter`, specifically when both merged instances have the default, in `Router`s case the "other" fallback was preferred while in `MethodRouter`s case the `self` fallback was preferred. Was that intentional?

davidpdrsn (approved):

davidpdrsn:
Oh wait sorry. Didn't notice the test failure. I'll look more into this tomorrow.

davidpdrsn:
> the `MethodRouter`, specifically when both merged instances have the default, in `Router`s case the "other" fallback was preferred while in `MethodRouter`s case the `self` fallback was preferred. Was that intentional?

Oh yeah that's not intentional. I think we should pick the "other" fallback, so what `Router` does.

As far as I remember it matters most when the default fallback has middleware.

jplatte:
Since this is is approved and it turned out I didn't need to adjust any tests with the strategy you preferred, I'll merge.

PR #1365, opened 2022-09-10
Allow Routers to inherit state and fallback route

### LINKED ISSUE #1326: Nesting/merging routers that use state could be more ergonomic

jplatte:
Nowhere near complete, I just wanted to post this so people know I've started work on this, and so we can potentially collaborate on it.

Depends on #1364.
Closes #1326.

### DIFF
--- a/axum/src/routing/method_routing.rs
+++ b/axum/src/routing/method_routing.rs
@@ -923,21 +923,6 @@ where
             }
         }
 
-        #[track_caller]
-        fn merge_fallback<B, E>(
-            fallback: Fallback<B, E>,
-            fallback_other: Fallback<B, E>,
-        ) -> Fallback<B, E> {
-            match (fallback, fallback_other) {
-                (pick @ Fallback::Default(_), Fallback::Default(_)) => pick,
-                (Fallback::Default(_), pick @ Fallback::Custom(_)) => pick,
-                (pick @ Fallback::Custom(_), Fallback::Default(_)) => pick,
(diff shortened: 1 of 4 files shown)


jplatte:
I realized that a lot of things I did here are orthogonal to the key change that might make it all work. I'll close this to preserve it as a snapshot, and open a new PR with only the important bits.

PR #1371, opened 2022-09-12
Show the errored path on `JsonDataError`

dahlia:
## Motivation

Previously, it was difficult to find out the path in the deep JSON tree at which a deserialization error occurred.

## Solution

This patch makes an error message to contain the errored path.  In order to find out the path, I added [serde_path_to_error](https://crates.io/crates/serde_path_to_error), a new optional dependency.


axum/src/json.rs
```diff
@@ -296,4 +297,37 @@ mod tests {
...
+        let res = client
+            .post("/")
+            .body("{\"a\": 1, \"b\": [{\"x\": 2}]}")
+            .header("content-type", "application/json")
+            .send()
+            .await;
+
+        assert_eq!(res.status(), StatusCode::UNPROCESSABLE_ENTITY);
+        let body_text = res.text().await;
+        assert!(body_text.contains("b[0]"), "body: {:?}", body_text);
```
davidpdrsn, line 331: Wanna change this to match on the whole text, instead of using `contains`? Then its easier to see what the final result actually is.

davidpdrsn (commented):
Thanks! I think it looks good.

Worth nothing that this changes the inner error of a `JsonRejection` from `serde_json::Error` to `serde_path_to_error::Error<serde_json::Error>`. Thats not considered a breaking change but I think we should mention it in the changelog regardless.

jplatte:
I wonder whether this has significant performance implications. Maybe it should only be part of axum-extra (for now) if that is the case?

davidpdrsn:
Good point. I can the benchmark:

```
main
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:49942 for 10 second(s)
    Latencies:
      0.07ms   0.02ms   0.02ms   2.63ms
    Requests:
      Total: 1490865 Req/Sec: 149077.94
    Transfer:
      Total: 106.63 MB Transfer Rate: 10.66 MB/Sec
```

`json-error-path` branch
```
  Benchmarking 10 connections @ http://0.0.0.0:49918 for 10 second(s)
    Latencies:
      0.07ms   0.02ms   0.02ms   0.67ms
    Requests:
      Total: 1517780 Req/Sec: 151771.58
    Transfer:
      Total: 108.56 MB Transfer Rate: 10.86 MB/Sec
```

This branch is faster, which is odd so might just be noise and no loss of performance.

jplatte:
That's the `receive-json` bench? Seems good if there's no (measureable) slowdown.

davidpdrsn:
> That's the `receive-json` bench?

Yes

axum/src/json.rs
dahlia: Amended!

axum/CHANGELOG.md
```diff
@@ -7,7 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- **changed**: The inner error of a `JsonRejection` is now
+  `serde_path_to_error::Error<serde_json::Error>`.  Previously it was
+  `serde_json::Error`. ([#1371])
+- **changed**: `JsonRejection` now displays the path at which a deserialization
+  error occurred too. ([#1371])
```
davidpdrsn, line 14:
```suggestion
- **added**: `JsonRejection` now displays the path at which a deserialization
  error occurred too ([#1371])
```

You also need to add a link for `[#1371]` like the other changes below.

axum/CHANGELOG.md
```diff
@@ -7,7 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- **changed**: The inner error of a `JsonRejection` is now
+  `serde_path_to_error::Error<serde_json::Error>`.  Previously it was
+  `serde_json::Error`. ([#1371])
```
davidpdrsn, line 12:
```suggestion
  `serde_json::Error` ([#1371])
```

davidpdrsn (approved):
Thanks!

dahlia:
@davidpdrsn Fixed the changelogs!

davidpdrsn (approved):
Thanks!

PR #1379, opened 2022-09-16
Relax Multipart FromRequest implementation bounds

jplatte:
Without this, it was impossible to use [tower_http::limit](https://docs.rs/tower-http/latest/tower_http/limit/index.html) with `Multipart` because `http_body::Limited` doesn't implement `Default`.

The first commit is not a breaking change and can thus be backported to 0.5.

### DIFF
--- a/axum/src/extract/multipart.rs
+++ b/axum/src/extract/multipart.rs
@@ -249,7 +249,7 @@ define_rejection! {
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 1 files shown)


davidpdrsn (commented):
Nice! Wanna add a test?

jplatte:
Not now, but if you want to wait I'll look into that soon-ish.

davidpdrsn (approved):

PR #1380, opened 2022-09-16
Use regular non-exhaustive debug representation instead of custom one

jplatte:


davidpdrsn (approved):
Cool! Didn't know about this one.

PR #1381, opened 2022-09-16
Split RouterService off of Router

jplatte:
I guess this needs some docs. Can you add those @davidpdrsn? 🙏🏼 

Also I wonder whether the `WithState` type for `Method` router should also be renamed to `MethodRouterService` 🤔

### DIFF
--- a/axum-extra/src/extract/cookie/mod.rs
+++ b/axum-extra/src/extract/cookie/mod.rs
@@ -258,7 +258,8 @@ mod tests {
                 let app = Router::<_, Body>::with_state(state)
                     .route("/set", get(set_cookie))
                     .route("/get", get(get_cookie))
-                    .route("/remove", get(remove_cookie));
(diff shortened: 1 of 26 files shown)


jplatte:
Should we do another intermediary branch to merge the state inheritance stuff into? This PR probably doesn't make sense on its own, only if combined with a split of `nest` / `nest_service`.

davidpdrsn (approved):
I'm fine with merging this as is with the docs still pending. We can just make an issue about adding the docs later.

davidpdrsn:
> Also I wonder whether the WithState type for Method router should also be renamed to MethodRouterService 🤔

We could. I named it `WithState` because [its already in the `method_router` module](https://docs.rs/axum/0.6.0-rc.2/axum/routing/method_routing/struct.WithState.html). I don't feel strongly either way.

PR #1382, opened 2022-09-17
Add tokio feature & make tokio optional for WASM support

logankeenan:
## Motivation

This PR is a continuation of @fisherdarling's [PR](https://github.com/tokio-rs/axum/pull/1054) since they are very busy and don't have time to see it through at the moment.

It allows Axum to be compiled to WASM, therefore, executable in a browser, Cloudflare Worker, etc...

## Solution

This PR adds a tokio feature that enables tokio as well as hyper/tcp and hyper/server. The tokio feature also enables SSE and extract::connect_info.

This is a breaking change since this breaks usages of default_features = false.


examples/simple-router-wasm/src/main.rs
```diff
@@ -0,0 +1,51 @@
+//! Run with
+//!
+//! ```not_rust
+//! cd examples && cargo run -p example-simple-router-wasm
```
EdorianDark, line 4: But this doesn't compile to Wasm. What about adding `--target wasm32-unknown-unknown` and running it in a Wasm runtime?

examples/simple-router-wasm/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+//! ```
+//!
+//! This example shows what using axum in a wasm context might look like. This example should
+//! always compile with `--target wasm32-unknown-unknown`.
+//!
+//! [`mio`](https://docs.rs/mio/latest/mio/index.html), tokio's IO layer, does not support the
+//! `wasm32-unknown-unknown` target which is why this crate requires `default-features = false`
+//! for axum.
+//!
+//! Most "serverless" runtimes use Chrome's V8 engine for executing Javascript, and in our
```
EdorianDark, line 14: There are also other Wasm runtimes using [Wasmtime](https://bytecodealliance.org/articles/wasmtime-1-0-fast-safe-and-production-ready). I would prefer adding more examples or leaving the example out entirely.

axum/Cargo.toml
```diff
@@ -61,6 +61,7 @@ serde_path_to_error = { version = "0.1.8", optional = true }
 serde_urlencoded = { version = "0.7", optional = true }
 sha-1 = { version = "0.10", optional = true }
 tokio-tungstenite = { version = "0.17.2", optional = true }
+tokio_cr = { package = "tokio", version = "1", features = ["time"], optional = true }
```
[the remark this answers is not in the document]
fisherdarling: @logankeenan I think when I originally made this PR `dep:` syntax wasn't available for axum's MSRV. That's why there's a `tokio_cr` hack.

examples/simple-router-wasm/src/main.rs
fisherdarling: The goal here is to have an example that runs on the user's local machine. We could add a wasm runtime, but then we'd need to have to determine a runtime, add a runtime specific interface, get the user to install the runtime, and then support that runtime in CI. That feels like much more of a maintenance burden. I think it's better to provide some guidance and let the user figure it out.

examples/simple-router-wasm/src/main.rs
fisherdarling: How about we just strike the first sentence and write:

>Most serverless runtimes expect an exported function that takes in a single request and returns a single response, much like axum's `Handler`.

examples/simple-router-wasm/src/main.rs
davidpdrsn: I agree and good that we call this out in the example.

examples/simple-router-wasm/src/main.rs
davidpdrsn: I agree, I don't think we should callout any specific runtimes.

davidpdrsn:
@logankeenan fyi I've started committing to this a bit, so plz don't force push 🙏

davidpdrsn (approved):
Thank you @fisherdarling and @logankeenan for working on this. I did some final touch ups and this is good to merge 🎉

PR #1391, opened 2022-09-18
Support `State` with `#[derive(FromRequest[Parts])]`

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1314

This makes it possible to extract things via `State` in `#[derive(FromRequet)]`:

```rust
#[derive(FromRequet)]
#[from_request(state(AppState))]
struct Foo {
    state: State<AppState>,
}
```

The state can also be inferred in a lot of cases so you only need to write:

```rust
#[derive(FromRequet)]
struct Foo {
    // since we're using `State<AppState>` we know the state has to be
    // `AppState`
    state: State<AppState>,
}
```

Same for

```rust
#[derive(FromRequet)]
struct Foo {
    #[from_request(via(State))]
    state: AppState,
}
```

And

```rust
#[derive(FromRequet)]
#[from_request(via(State))]
struct AppState {}
```

I think I've covered all the edge cases but there are (unsurprisingly) a few.

## Todo

- [x] Test that these extractors can be combined with others in the same handler, ie that we don't get errors because of `M`

### DIFF
--- a/axum-macros/src/from_request.rs
+++ b/axum-macros/src/from_request.rs
@@ -4,9 +4,11 @@ use crate::{
     from_request::attr::FromRequestFieldAttrs,
 };
 use proc_macro2::{Span, TokenStream};
-use quote::{quote, quote_spanned};
-use std::fmt;
-use syn::{punctuated::Punctuated, spanned::Spanned, Ident, Token};
+use quote::{quote, quote_spanned, ToTokens};
+use std::{collections::HashSet, fmt, iter};
+use syn::{
+    parse_quote, punctuated::Punctuated, spanned::Spanned, Fields, Ident, Path, Token, Type,
+};
 
 mod attr;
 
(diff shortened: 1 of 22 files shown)


genusistimelord:
I will give this a Test since I use a ton of different extractors in my full state struct.

davidpdrsn:
@genusistimelord Thanks!

axum-macros/src/from_request.rs
```diff
@@ -78,14 +162,28 @@ pub(crate) fn expand(item: syn::Item, tr: Trait) -> syn::Result<TokenStream> {
...
+                rejection,
+                state,
+            } = parse_attrs("from_request", &attrs)?;
+
+            let state = match state {
+                Some((_, state)) => State::Custom(state),
+                None => (|| {
+                    let via = via.as_ref().map(|(_, via)| via)?;
+                    state_from_via(&ident, via).map(State::Custom)
+                })()
```
jplatte, line 176:
```suggestion
                None => via
                    .as_ref()
                    .and_then(|(_, via)| state_from_via(&ident, via))
                    .map(State::Custom)
```

axum-macros/src/lib.rs
```diff
@@ -613,6 +615,50 @@ where
...
+where
+    I: Iterator<Item = &'a Type>,
+{
+    let state_inputs = types
+        .filter_map(|ty| {
+            if let Type::Path(path) = ty {
+                Some(&path.path)
+            } else {
+                None
+            }
```
jplatte, line 628: Will be a bit shorter if converted to `match`.

axum-macros/src/lib.rs
```diff
@@ -613,6 +615,50 @@ where
...
+        })
+        .filter_map(|generic_arg| {
+            if let syn::GenericArgument::Type(ty) = generic_arg {
+                Some(ty)
+            } else {
+                None
+            }
```
jplatte, line 651: Same here.

jplatte:
Will have to come back to this, but left a few first comments.

axum-macros/tests/from_request/pass/state_infer.rs
```diff
@@ -0,0 +1,18 @@
+use axum_macros::FromRequest;
+use axum::extract::State;
+
+#[derive(FromRequest)]
+struct Extractor {
+    inner_state: State<AppState>,
```
jplatte, line 6: You made the derive take the first `State<_>` field even if there are multiple, right? Would be good to also have a test for that so it doesn't accidentally regress.

jplatte (commented):
Just one more comment, have now reviewed everything.

axum-macros/tests/from_request/pass/state_infer.rs
davidpdrsn: You mean something like

```rust
#[derive(FromRequest)]
struct Extractor {
    inner_state: State<AppState>,
    also_inner_state: State<OtherState>,
}
```

There it'll require you to specify the state because there are two different options.

I tried to make that work with generating bounds like `AppState: FromRef<S>` and `OtherState: FromRef<S>` but ran into issues with that that broke other things I think 🤔

jplatte: Maybe have a failing UI test for that case then? Or is there already one?

davidpdrsn: Sure. I've added one.

axum-macros/tests/from_request/fail/state_infer_multiple_different_types.stderr
```diff
@@ -0,0 +1,23 @@
...
+ --> tests/from_request/fail/state_infer_multiple_different_types.rs:7:18
+  |
+7 |     other_state: State<OtherState>,
+  |                  ^^^^^ the trait `FromRef<S>` is not implemented for `OtherState`
+  |
+  = note: required because of the requirements on the impl of `FromRequestParts<S>` for `State<OtherState>`
+help: consider extending the `where` clause, but there might be an alternative better way to express this requirement
+  |
+4 | #[derive(FromRequest, OtherState: FromRef<S>)]
+  |                     ++++++++++++++++++++++++
```
jplatte, line 23: I think there's room for improvement on this error message, but the important bit is that we have a test for it not working. I can work on the error message after this is merged.

jplatte (approved):

Issue #1370, opened 2022-09-12
Tracking issue for allowing configurable request body size limit


kellytk:
Re: https://github.com/tokio-rs/axum/pull/1346#issuecomment-1240700262, https://github.com/tokio-rs/axum/pull/1346#issuecomment-1240773691

_Some_ limit is a welcome addition!  IMHO an ability to configure the limit, keeping the value consistent with [other similar limits](https://github.com/caddyserver/caddy/issues/4558) I utilize, would be ideal.

PR #1397, opened 2022-09-19
Add `DefaultBodyLimit::max` to change the body size limit

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1370

We now have a few ways to set body limits:
- The new `DefaultBodyLimit::max`
- `ContentLengthLimit`
- `tower_http::limit::RequestBodyLimit` and `http_body::Limited`

I don't we should remove anything tower-http and http-body but I feel we should consider removing `ContentLengthLimit`. It can essentially be replaced with

```rust
Router::new().route(
    "/",
    post(handler.layer(DefaultBodyLimit::max(new_limit))),
);
```

### DIFF
--- a/axum-core/CHANGELOG.md
+++ b/axum-core/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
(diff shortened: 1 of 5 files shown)


jplatte (approved):
I agree, this seems strictly better than the extractor.

PR #1400, opened 2022-09-22
Remove `ContentLengthLimit`

### LINKED ISSUE #1399: Remove `ContentLengthLimit` and make `DefaultBodyLimit` check `Content-Length`
See https://github.com/tokio-rs/axum/pull/1397

nylonicious:
Closes #1399

Before I continue I wanted to ask if the implementation should live in the `Bytes` extractor, and if so should we also check the method like `ContentLengthLimit` impl used to? If the impl should live in in the Bytes extractor what with rejections, should I just extend it with the required fields or maybe add one variant to `BytesRejection` named something like `BodyLimitRejection` and store more detailed rejection inside that variant?


davidpdrsn:
> Before I continue I wanted to ask if the implementation should live in the Bytes extractor, and if so should we also check the method like ContentLengthLimit impl used to?

Yes I think that would be good to add as well. I think that should be its own PR though.

> If the impl should live in in the Bytes extractor what with rejections, should I just extend it with the required fields or maybe add one variant to BytesRejection named something like BodyLimitRejection and store more detailed rejection inside that variant?

Hm I didn't really consider that. I suppose yes, we would have to add the rejection variants from

axum-core/src/extract/request_parts.rs
```diff
@@ -94,6 +94,16 @@ where
...
             Some(DefaultBodyLimitKind::Limit(limit)) => {
+                let content_length: usize = req
+                    .headers()
+                    .get(http::header::CONTENT_LENGTH)
+                    .and_then(|val| val.to_str().ok()?.parse().ok())
+                    .unwrap();
+
+                if content_length > limit {
+                    todo!()
+                }
```
davidpdrsn, line 105: Sorry but I don't think we should add this in this PR. I'd rather have that split up. I think if this PR just removes `ContentLengthLimit` then thats best.

axum/src/extract/multipart.rs
```diff
@@ -49,7 +49,7 @@ use std::{
 /// ```
 ///
 /// For security reasons it's recommended to combine this with
-/// [`ContentLengthLimit`](super::ContentLengthLimit) to limit the size of the request payload.
+/// [`DefaultBodyLimit`](super::DefaultBodyLimit) to limit the size of the request payload.
```
davidpdrsn, line 52: Since `DefaultBodyLimit` doesn't (yet) apply to `Multipart` I think we need to mention https://docs.rs/tower-http/latest/tower_http/limit/index.html instead.

davidpdrsn (changes requested):

nylonicious:
I couldn't find any other nice way of adding `RequestBodyLimitLayer` to key value store.

davidpdrsn (commented):
Looks good!

Wanna update the changelog as well? Then we can merge it 😊

axum/CHANGELOG.md
```diff
@@ -18,12 +18,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **added:** Add `middleware::from_extractor_with_state` and
   `middleware::from_extractor_with_state_arc` ([#1396])
 - **added:** Add `DefaultBodyLimit::max` for changing the default body limit ([#1397])
+- **breaking:** `ContentLengthLimit` has been removed ([#1400])
```
davidpdrsn, line 21:
```suggestion
- **breaking:** `ContentLengthLimit` has been removed. `Use DefaultBodyLimit` instead ([#1400])
```

davidpdrsn (approved):

PR #1413, opened 2022-09-25
Refactor `all_the_tuples!` macros

davidpdrsn:
Changes `all_the_tuples` to use the array/last special casing by default.

### DIFF
--- a/axum-core/src/macros.rs
+++ b/axum-core/src/macros.rs
@@ -102,6 +102,36 @@ macro_rules! composite_rejection {
 
 macro_rules! all_the_tuples {
     ($name:ident) => {
+        $name!([], T1);
+        $name!([T1], T2);
+        $name!([T1, T2], T3);
+        $name!([T1, T2, T3], T4);
+        $name!([T1, T2, T3, T4], T5);
+        $name!([T1, T2, T3, T4, T5], T6);
+        $name!([T1, T2, T3, T4, T5, T6], T7);
+        $name!([T1, T2, T3, T4, T5, T6, T7], T8);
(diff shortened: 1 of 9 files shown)


axum-core/src/macros.rs
```diff
@@ -101,6 +101,36 @@ macro_rules! composite_rejection {
...
+        $name!([T1, T2, T3, T4, T5, T6, T7], T8);
+        $name!([T1, T2, T3, T4, T5, T6, T7, T8], T9);
+        $name!([T1, T2, T3, T4, T5, T6, T7, T8, T9], T10);
+        $name!([T1, T2, T3, T4, T5, T6, T7, T8, T9, T10], T11);
+        $name!([T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11], T12);
+        $name!([T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12], T13);
+        $name!(
+            [T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13],
+            T14
+        );
```
jplatte, line 121: Noo, it starts breaking into multiple lines. Does rustfmt actually enforce this? Is `#[rustfmt::skip]` on the `macro_rules!` in order?

jplatte (approved):

axum-core/src/macros.rs
davidpdrsn: Oh yeah that is rustfmt. I didn't think rustfmt attempted to format macros. But I guess it does in some cases.

jplatte: Yeah, I think it stops as soon as you start adding repetitions, but for simple cases like this it does (and I'm generally glad that's the case) 🙂

Issue #1394, opened 2022-09-19
`middleware::map_request[_with_state[_arc]]` and `map_response[_with_state[_arc]]`


jplatte:
### Motivation

`axum`s `from_fn[_with_state[_arc]]` can be confusing to newcomers who haven't read about / fully understood tower's "onion principle" of layers yet. Its generality also makes for a little more boilerplate than necessary when one wants to write a middleware that only concerns itself with either the request or response.

### Proposal

Add `map_request[_with_state[_arc]]` and `map_response[_with_state[_arc]]` to the `middleware` module. In the case of `map_response*` it is notable that async functions passed to these functions will never have to be generic.

### Alternatives

Do nothing.

davidpdrsn:
I think that makes sense to support. I'm a bit worried about duplicating stuff that is in tower but I think these two are probably fine.

jplatte:
Well, ours are going to be more ergonomic by supporting extractors.

jplatte:
I've been discussing this a bit with @genusistimelord, one question that came up is what the function signature should look like, in particular for `map_request*`. For some use cases, you may want to return early inside the closure, so a `Result<Request<B>, impl IntoResponse>` return type could make sense. But then people who have an infallible closure have to write `Ok::<_, Infallible>(req)` because of the generic `E` type on the `Result`.

The easy way out of this is providing both `map_request*` and `try_map_request*`. For `map_response`, none of this is needed because that can just return any type that implements `IntoResponse`.

davidpdrsn:
Maybe https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html?

jplatte:
Can you elaborate? That's pretty much just a different way of spelling the same thing (w/ some different rules for `?`) and still leaves the question of whether there should be separate `try_` variants or not, no?

genusistimelord:
I think a map_request* and try_map_request* make the most sense here as we can dictate that map_request never fails and try Can fail returning a Result<Request<B>, IntoResponse> or a ControlFlow<IntoResponse> and make req passable via &mut req?

davidpdrsn:
I was thinking something like https://github.com/tokio-rs/axum/pull/1408

PR #1414, opened 2022-09-25
Add `map_response` and friends

davidpdrsn:
Part of https://github.com/tokio-rs/axum/issues/1394

### DIFF
--- a/axum/src/middleware/map_response.rs
+++ b/axum/src/middleware/map_response.rs
@@ -1,0 +1,341 @@
+use crate::response::{IntoResponse, Response};
+use axum_core::extract::FromRequestParts;
+use futures_util::future::BoxFuture;
+use http::Request;
+use std::{
+    any::type_name,
+    convert::Infallible,
+    fmt,
+    future::Future,
+    marker::PhantomData,
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+};
+use tower_layer::Layer;
+use tower_service::Service;
+
+/// Create a middleware from an async function that transforms a response.
+///
+/// This differs from [`tower::util::MapResponse`] in that it allows you to easily run axum-specific
+/// extractors.
+///
+/// # Example
+///
+/// ```
+/// use axum::{
+///     Router,
+///     routing::get,
+///     middleware::map_response,
+///     response::Response,
+/// };
+///
+/// async fn set_header<B>(mut response: Response<B>) -> Response<B> {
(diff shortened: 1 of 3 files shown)


axum/src/middleware/map_response.rs
```diff
@@ -0,0 +1,341 @@
...
+
+macro_rules! impl_service {
+    (
+        $($ty:ident),*
+    ) => {
+        #[allow(non_snake_case, unused_mut)]
+        impl<F, Fut, S, I, B, ResBody, $($ty,)*> Service<Request<B>> for MapResponse<F, S, I, ($($ty,)*)>
+        where
+            F: FnMut($($ty,)* Response<ResBody>) -> Fut + Clone + Send + 'static,
+            $( $ty: FromRequestParts<S> + Send, )*
```
davidpdrsn, line 214: Note that we require `FromRequestParts` and not `FromRequest`. Not sure there would be a way to recover the request afterwards so we can pass it to the inner service 🤔

jplatte: Yeah, this is exactly right.

axum/src/middleware/map_response.rs
```diff
@@ -0,0 +1,341 @@
...
+///     Router,
+///     routing::get,
+///     middleware::map_response,
+///     extract::Path,
+///     response::Response,
+/// };
+/// use std::collections::HashMap;
+///
+/// async fn log_path_params<B>(
+///     Path(path_params): Path<HashMap<String, String>>,
```
jplatte, line 57: I think we should call out that the request extractors run before any inner services / the handler and so isn't going to necessarily receive the same data from extractors as a handler would. It might not be obvious for people who only expect this layer to run after the handler (which of course it can't, since the handler consumes the request).

axum/src/middleware/map_response.rs
```diff
@@ -0,0 +1,341 @@
...
+///     response: Response<B>,
+/// ) -> Response<B> {
```
jplatte, line 59: I think we should also call out that users can return any `impl IntoResponse`, not just `Response<B>`.

jplatte (approved):

jplatte:
Oh, you had enabled auto-merge. Well, docs can still be improved separately :)

davidpdrsn:
> Oh, you had enabled auto-merge. Well, docs can still be improved separately :)

🚀 😅

I'll do a follow-up

PR #1417, opened 2022-09-25
Rewrite how state is passed from Router to MethodRouter

jplatte:
I think we should also move `WithState` into its own module, but that can be a follow-up PR.


davidpdrsn (approved):
Nice!

PR #1421, opened 2022-09-26
Add From impls for extract::ws::Message

Rapptz:
## Motivation

I wanted to do something like `ws.send("hello".into()).await?` but the `From` impls in tungstenite were not exposed.

## Solution

This merely reimplements those `From` impls from tungstenite.

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -545,6 +545,30 @@ impl Message {
     }
 }
 
+impl From<String> for Message {
+    fn from(string: String) -> Self {
+        Message::Text(string)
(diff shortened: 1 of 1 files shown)


jplatte (approved):

jplatte:
Can you add a changelog entry?

davidpdrsn (commented):
Thanks! Wanna add a note the changelog as well?

Rapptz:
I've added the changelog entry, unsure if the wording needs tweaking!

davidpdrsn (approved):
LGTM!

PR #1426, opened 2022-09-27
rmv claim that state can't be passed to middleware

chuckberrypi:
axum::middleware::from_fn_with_state now allows middleware to access application state.

## Motivation
The docs currently state that the State extractor cannot be passed to middleware. axum::middleware::from_fn_with_state allows state to be passed to middleware, so the statement is either incorrect or at the least confusing. 

## Solution
Removed the statement.


davidpdrsn (approved):
Thanks!

PR #1432, opened 2022-09-29
Improve the error message for state type inference failure in FromRequest(Parts) derive macro

jplatte:
Happy to do similar things for other macro errors, if somebody writes a test case with a suboptimal error message 🙂


davidpdrsn (approved):

jplatte:
There doesn't seem to be a test covering the state type inference failure for `#[debug_handler]`, do you want me to add that?

davidpdrsn:
Sure!

PR #1436, opened 2022-09-30
Improve compile errors for unimplemented traits

weiznich:
These two commits try to improve error messages generated by rustc if some trait bound is not fulfilled. 

The first commit simply improves the `#[debug_handler]` helper proc-macro attribute to set spans ins certain situation. This results in the corresponding error messages always pointing to a concrete handler argument instead of the attribute itself (see the updated compile tests).

The second commit introduces a rust nightly only `nightly-error-messages` feature flag + uses unstable rust attributes to improve some other error messages as well. It currently mostly points the users to use the `#[debug_handler]` attribute. [In specific situation it additionally adds more information.](https://github.com/weiznich/axum/commit/a151aac96a8569c59df2de2c0ae3d645ab1c6430). This is mostly meant as demonstration of what's possible. I currently try to push for stabilization of the underlying `#[rustc_on_unimplemented]` attribute.


axum-macros/src/debug_handler.rs
```diff
@@ -231,11 +231,11 @@ fn check_inputs_impls_from_request(
             };
 
             let from_request_bound = if must_impl_from_request_parts {
-                quote! {
+                quote_spanned! { span =>
```
jplatte, line 234: In syn's docs, this is always written without any extra spaces, even with the docs pointing out that it should look like that to stand out. I think we should follow that.

```suggestion
                quote_spanned! {span=>
```

axum-macros/src/debug_handler.rs
```diff
@@ -231,11 +231,11 @@ fn check_inputs_impls_from_request(
...
                     #ty: ::axum::extract::FromRequestParts<#state_ty> + Send
                 }
             } else {
-                quote! {
+                quote_spanned! { span =>
```
jplatte, line 238:
```suggestion
                quote_spanned! {span=>
```

axum/src/handler/mod.rs
```diff
@@ -99,6 +99,12 @@ pub(crate) use self::{
 /// {}
 /// ```
 #[doc = include_str!("../docs/debugging_handler_type_errors.md")]
+#[cfg_attr(
+    feature = "nightly-error-messages",
+    rustc_on_unimplemented(
+        note = "Consider using `#[axum_macros::debug_handler]` to improve the error message"
```
jplatte, line 105: Nice! 😄 

Can you maybe add one or two tests for the `on_unimplemented` errors?

weiznich: The question is where and how such tests can be added? Running them would always require a nightly compiler + trybuild/compiletests. I'm not that familiar with the axum project setup to answer this question, but if someone points me into the right direction I'm happy to add them there.

jplatte: Although if you're doing this anyways, it's probably more sensible to always compile the test itself, and ignore it on other compiler versions using `#[rustversion::attr(not(nightly), ignore)]`. (instead of having the test do nothing)

jplatte: @weiznich We can also merge this as-is and add those tests in a separate PR, if you prefer?

weiznich: That would be a good idea. I'm currently quite busy with other stuff, so I cannot address this comment now (although I've planed to do that next week or the week after).

jplatte: Okay. Looks like CI needs to be updated though (we're using `--all-features` which is now trying to activate the nightly-only stuff on stable / beta).

weiznich: I'm also fine with making the required changes later this month.

weiznich: I've pushed a new version that enables this by default on nightly compilers and removes the explicit feature flag. Also this is now integrated in the `axum-macros` test suite output, although it's not clear for me if that's run on CI yet?

axum-macros/rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly-2022-10-20
```
davidpdrsn, line 1: Is there any difference to having this file here vs in the root of the workspace?

axum-core/src/extract/mod.rs
```diff
@@ -39,6 +39,17 @@ mod private {
...
+    nightly_error_messages,
+    rustc_on_unimplemented(
+        on(
+            _Self = "axum::http::Request<axum::body::Body>",
+            message = "Request must always be the last argument to a handler function",
+            label = "Move this argument to the end of the argument list",
+        ),
+        note = "Function argument is no valid axum extractor. \nSee `https://docs.rs/axum/latest/axum/extract/index.html` for details",
+    )
+)]
```
davidpdrsn, line 52: I think this could be tweaked a bit and also need one for `FromRequest`. But we can play around with that once this is merged.

davidpdrsn (approved):
Sorry for taking so long to review this. I think this looks really good! Getting on stable would be awesome ✨

axum-macros/rust-toolchain
weiznich: I choose to follow the same strategy that I use for diesel. Only the compile tests should use a nightly compiler, anything else should use the default tool chain. By putting the rust-toolchain file into this sub-directory it's only applied for running tests from inside this directory.

davidpdrsn (changes requested):
Wait a second. Doesn't seem like the nightly tests are run on CI. Need to fix that before we can merge this. I'll look at that later.

I'm also getting an odd error from running this locally:

```
$ cargo +nightly-2022-10-20 check --tests --all-features
...
error: unreachable `pub` item
   --> axum/src/extract/mod.rs:56:9
    |
56  | pub use self::matched_path::MatchedPath;
    | ---     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | |
    | help: consider restricting its visibility: `pub(crate)`
    |
    = help: or consider exporting it for use by other crates
note: the lint level is defined here
   --> axum/src/lib.rs:429:9
    |
429 | #![deny(unreachable_pub, private_in_public)]
    |         ^^^^^^^^^^^^^^^
```

jplatte:
Maybe try a newer Nightly and bump the toolchain file if things work?

davidpdrsn (approved):
@weiznich thanks a lot of working on this!

PR #1444, opened 2022-10-03
Start second server on different port for metrics endpoint

vbrandl:
## Motivation

The `prometheus-metrics` example makes the `/metrics` endpoint publicly available.

## Solution

As discussed in #1443, this change starts a second server on a different port for the `/metrics` endpoint.

### DIFF
--- a/examples/prometheus-metrics/src/main.rs
+++ b/examples/prometheus-metrics/src/main.rs
@@ -23,6 +23,48 @@ use std::{
 };
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
+fn metrics_app() -> Router {
+    let recorder_handle = setup_metrics_recorder();
+    Router::new()
+        .route("/metrics", get(move || ready(recorder_handle.render())))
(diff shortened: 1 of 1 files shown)


examples/prometheus-metrics/src/main.rs
```diff
@@ -23,35 +23,62 @@ use std::{
...
-                .unwrap_or_else(|_| "example_todos=debug,tower_http=debug".into()),
-        ))
-        .with(tracing_subscriber::fmt::layer())
-        .init();
-
+fn metrics_app() -> Router {
     let recorder_handle = setup_metrics_recorder();
+    Router::new()
+        .route("/metrics", get(move || ready(recorder_handle.render())))
+        .route_layer(middleware::from_fn(track_metrics))
```
davidpdrsn, line 30: Probably don't need to track metrics on the metrics server :P

```suggestion
```

vbrandl: You are absolutely right. Fixed in 2151bb6

davidpdrsn (approved):

PR #1460, opened 2022-10-09
Don't internally `Arc` the state

davidpdrsn:
The fact that the state is `Arc`'ed internally in `Router` has caused confusion for several people. I've been thinking about and think we should remove the `Arc` and give the users the choice to `Arc` things or not.

### DIFF
--- a/axum-extra/src/routing/mod.rs
+++ b/axum-extra/src/routing/mod.rs
@@ -178,7 +178,7 @@ pub trait RouterExt<S, B>: sealed::Sealed {
 impl<S, B> RouterExt<S, B> for Router<S, B>
 where
     B: axum::body::HttpBody + Send + 'static,
-    S: Send + Sync + 'static,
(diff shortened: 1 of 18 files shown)


axum-extra/src/routing/mod.rs
```diff
@@ -178,7 +178,7 @@ pub trait RouterExt<S, B>: sealed::Sealed {
 impl<S, B> RouterExt<S, B> for Router<S, B>
 where
     B: axum::body::HttpBody + Send + 'static,
-    S: Send + Sync + 'static,
+    S: Clone + Send + Sync + 'static,
```
jplatte, line 181: I'm pretty sure `Sync` is now no longer needed, and one of `Send` or `'static` might not be either. We can relax these in a separate PR though, if you prefer.

jplatte (approved):

Issue #1441, opened 2022-10-02
Extracting `MatchedPath` in middleware will not always give the final matched path


dcormier:
### Version

axum v0.6.0-rc.2

### Platform

Darwin 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:23 PDT 2021; root:xnu-8019.41.5~1/RELEASE_X86_64 x86_64

### Description

When extracting `MatchedPath` in middleware, it may not have the correct value if the request end up being handled by a nested `Router`.

A simplified example:

```rust
use std::{
    convert::Infallible,
    fmt::Debug,
    mem,
    task::{Context, Poll},
};

use axum::{
    body::Body, extract::MatchedPath, http::Request, response::Response, routing::get, Router,
};
use futures::future::BoxFuture;
use tower::{Layer, Service, ServiceExt};

#[derive(Debug, Clone)]
pub struct LogService<S>
where
    S: Debug + Clone,
{
    inner: S,
}

impl<S, B> Service<Request<B>> for LogService<S>
where
    S: Service<Request<B>, Response = Response, Error = Infallible>
        + Debug
        + Clone
        + Send
        + 'static,
    S::Future: Send,
    B: Send + 'static,
{
    type Response = Response;
    type Error = Infallible;
    type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;

    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
        self.inner.poll_ready(cx)
    }

    fn call(&mut self, req: Request<B>) -> Self::Future {
        let clone = self.inner.clone();
        let mut inner = mem::replace(&mut self.inner, clone);

        Box::pin(async move {
            let matched_path: MatchedPath = req
                .extensions()
                .get()
                .cloned()
                .expect("Request is missing MatchedPath");

            let resp = inner.call(req).await;

            println!(
                "Request to {} complete: {}",
                matched_path.as_str(),
                resp.as_ref().unwrap().status()
            );

            resp
        })
    }
}

pub struct LogLayer;

impl<S> Layer<S> for LogLayer
where
    S: Debug + Clone,
{
    type Service = LogService<S>;

    fn layer(&self, inner: S) -> Self::Service {
        LogService { inner }
    }
}

#[tokio::test]
async fn bug() {
    let mut router = Router::new()
        .route(
            "/root",
            get(|matched_path: MatchedPath| async move {
[...]

PR #1462, opened 2022-10-09
Don't allow extracting `MatchedPath` in middleware for nested routes

davidpdrsn:
This changes `MatchedPath` so it cannot be extracted in middleware of nested routes. This means users wont see the internal wildcard route (`/nested/*__private__axum_nest_tail_param`) that actually matched. This is an implementation detail and shouldn't be exposed.

I didn't make `MatchedPath` accessible in response extensions like discussed in https://github.com/tokio-rs/axum/issues/1441 because thats not a breaking change and shouldn't block shipping 0.6.0

I also reworked the tests so there are multiple smaller tests instead of one big one. Should be easier to read.

Part of https://github.com/tokio-rs/axum/issues/1441


axum/src/extract/matched_path.rs
```diff
@@ -49,10 +53,47 @@ use std::sync::Arc;
...
 ///     );
-/// # async {
-/// # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
-/// # };
+/// # let _: Router = app;
 /// ```
+///
+/// # Matched path in nested routers
+///
+/// Because of how [nesting] works `MatchedPath` isn't accessible in middleware on nested routes:
```
jplatte, line 61:
```suggestion
/// Because of how [nesting] works, `MatchedPath` isn't accessible in middleware on nested routes:
```

axum/src/extract/matched_path.rs
```diff
@@ -49,10 +53,47 @@ use std::sync::Arc;
...
+///     // if `/foo/bar` is called this will be `Err(_)` since that matches
+///     // a nested route
+///     let matched_path: Result<MatchedPath, MatchedPathRejection> =
+///         request.extract_parts::<MatchedPath>().await;
+///
+///     request
+/// }
+///
+/// // `MatchedPath` is always accessible on handlers regardless
+/// // if its for a nested route or not
```
jplatte, line 84:
```suggestion
/// // of whether it's for a nested route or not
```

jplatte: Also, what if you do `.nest("/foo", handler.into_service())`? Doesn't count as nesting a handler? 😄

axum/src/extract/matched_path.rs
```diff
@@ -82,130 +123,181 @@ where
...
+    let matched_path = if let Some(matched_path) = route_id_to_path.get(&id) {
+        matched_path
+    } else {
+        #[cfg(debug_assertions)]
+        panic!("should always have a matched path for a route id");
+    };
+
+    let matched_path = append_nested_matched_path(matched_path, extensions);
+
+    if matched_path.contains(NEST_TAIL_PARAM_CAPTURE) {
```
jplatte, line 143:
```suggestion
    if matched_path.ends_with(NEST_TAIL_PARAM_CAPTURE) {
```

axum/src/extract/matched_path.rs
```diff
@@ -82,130 +123,181 @@ where
...
+        extensions.insert(MatchedNestedPath(matched_path));
+        extensions.remove::<MatchedPath>();
```
jplatte, line 145: Why would there be a `MatchedPath` in the extensions in this branch?

axum/src/extract/matched_path.rs
```diff
@@ -82,130 +123,181 @@ where
...
+    if let Some(previous) = extensions
+        .get::<MatchedPath>()
+        .map(|matched_path| matched_path.as_str())
+        .or_else(|| Some(&extensions.get::<MatchedNestedPath>()?.0))
+    {
+        let previous = if let Some(previous) = previous.strip_suffix(NEST_TAIL_PARAM_CAPTURE) {
+            previous
+        } else {
+            previous
+        };
```
jplatte, line 163:
```suggestion
        let previous = previous
            .strip_suffix(NEST_TAIL_PARAM_CAPTURE)
            .unwrap_or(previous);
```

dcormier:
It would be nice if the error returned by `impl FromRequestParts for MatchedPath` made it clear that it failed because there's a rested router.

dcormier:
Yes, I agree. I don't think they would be able to handle it differently. I was thinking purely for discoverability of the fact that it won't work with nested routers.

axum/src/extract/matched_path.rs
davidpdrsn: Oh right good point. Such handlers should be able to access the matched path either. I've added a test case for it and clarified the docs a bit.

axum/src/extract/matched_path.rs
davidpdrsn: There shouldn't be. It was mostly a "better safe than sorry". I've changed it to a `debug_assert!`.

axum/src/extract/matched_path.rs
```diff
@@ -82,130 +122,198 @@ where
...
+    } else {
+        #[cfg(debug_assertions)]
+        panic!("should always have a matched path for a route id");
+    };
+
+    let matched_path = append_nested_matched_path(matched_path, extensions);
+
+    if matched_path.contains(NEST_TAIL_PARAM_CAPTURE) {
+        extensions.insert(MatchedNestedPath(matched_path));
+        debug_assert!(extensions.remove::<MatchedPath>().is_none());
```
jplatte, line 144:
```suggestion
        debug_assert_eq!(extensions.get::<MatchedPath>(), None);
```

axum/src/extract/matched_path.rs
jplatte: Marked as resolved, but there's still contains. Or is GitHub fooling me?

axum/src/extract/matched_path.rs
```diff
@@ -82,130 +122,198 @@ where
...
+        .map(|matched_path| matched_path.as_str())
+        .or_else(|| Some(&extensions.get::<MatchedNestedPath>()?.0))
+    {
+        let previous = if let Some(previous) = previous.strip_suffix(NEST_TAIL_PARAM_CAPTURE) {
+            previous
+        } else {
+            previous
+        };
+
+        let matched_path = format!("{}{}", previous, matched_path);
```
jplatte, line 164: Can use implicit format captures.

jplatte (approved):
Outside of some small code style things, looks good!

PR #1466, opened 2022-10-10
Re-integrate examples into the root workspace

jplatte:
## Motivation

Currently, when working on axum (with VSCode), you don't get in-editor errors when working on examples. This was caused by #978 moving examples into their own workspace to make development of axum easier, but I've found the opposite to be true.

## Solution

Re-integrate the examples into the workspace, this time also setting `default-members` so that they don't get built / checked by a regular `cargo check` / `cargo build` (only with `--workspace`). I hope that retains most of the positives of #978 while fixing its downsides.


davidpdrsn:
For me rust-analyzer still checks all the examples which I'd like to avoid. I tried a bunch of ways to set `overrideCommand` (both `rust-analyzer.cargo.buildScripts.overrideCommand` and `rust-analyzer.checkOnSave.overrideCommand`) but couldn't get it to stop doing that. Unless there is a way to fix that I don't think we should do this.

jplatte:
This seems to work for me on the latest release:

```json
"rust-analyzer.checkOnSave.overrideCommand": [
    "cargo",
    "check", // or clippy
    "--message-format=json",
    "--all-targets"
]
```

davidpdrsn:
I was able to get it working with that `overrideCommand`. So let's give this a try and see how it works! I'll deal with the conflicts.

davidpdrsn (approved):

PR #1469, opened 2022-10-10
Relax bounds for `FromExtractor`

### LINKED ISSUE #1467: `from_extractor` not compatible with `RequestBodyLimitLayer`
I tried adding the `RequestBodyLimitLayer` to my application. This works for everything, except the routes that contain a layer built with `from_extractor`.

My `axum` version is `0.5.16`, `tower-http` is `0.3.4` and `rustc` is `1.63.0`.

## A minimized example

This does not compile, but it does as soon as I comment out one of the two `layer` lines.

```rust
use async_trait::async_trait;
use axum::{
    extract::{FromRequest, RequestParts},
    middleware::from_extractor,
    response::Html,
    routing::get,
    Router,
};
use std::{convert::Infallible, net::SocketAddr};

#[tokio::main]
async fn main() {
    let app = Router::new()
        .layer(from_extractor::<MyExtractor>())
        .layer(RequestBodyLimitLayer::new(1))
        .route("/", get(handler));

    axum::Server::bind(&addr)
        .await
        .unwrap();
}

async fn handler() -> Html<&'static str> {
    Html("<h1>Hello, World!</h1>")
}

struct MyExtractor;

#[async_trait]
impl<B> FromRequest<B> for MyExtractor
where
    B: Send,
{
    type Rejection = Infallible;

    async fn from_request(req: &mut RequestParts<B>) -> std::result::Result<Self, Self::Rejection> {
        todo!()
    }
}
```
## The Error Message

```
error[E0277]: the trait bound `http_body::limited::Limited<_>: Default` is not satisfied
   --> src/main.rs:15:16
    |
15  |         .layer(from_extractor::<MyExtractor>())
    |          ----- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Default` is not implemented for `http_body::limited::Limited<_>`
    |          |
    |          required by a bound introduced by this call
    |
    = help: the trait `tower_service::Service<Request<ReqBody>>` is implemented for `ExtractorMiddleware<S, E>`
    = note: required for `ExtractorMiddleware<Route<http_body::limited::Limited<_>>, MyExtractor>` to implement `tower_service::Service<Request<http_body::limited::Limited<_>>>`
note: required by a bound in `Router::<B>::layer`
   --> /home/hak/.cargo/registry/src/github.com-1ecc6299db9ec823/axum-0.5.16/src/routing/mod.rs:305:50
    |
305 |         L::Service: Service<Request<NewReqBody>, Response = Response<NewResBody>, Error = Infallible>
    |                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Router::<B>::layer`

For more information about this error, try `rustc --explain E0277`.
```

davidpdrsn:
We didn't actually use the bounds. I guess they were left over from a previous version.

Fixes #1467

### DIFF
--- a/axum/src/middleware/from_extractor.rs
+++ b/axum/src/middleware/from_extractor.rs
@@ -201,7 +201,7 @@ where
 impl<T, E, B, S> Service<Request<B>> for FromExtractor<T, E, S>
 where
     E: FromRequestParts<S> + 'static,
-    B: Default + Send + 'static,
+    B: Send + 'static,
     T: Service<Request<B>> + Clone,
     T::Response: IntoResponse,
     S: Clone + Send + Sync + 'static,
@@ -266,7 +266,6 @@ where
     E: FromRequestParts<S>,
     T: Service<Request<B>>,
     T::Response: IntoResponse,
-    B: Default,
 {
     type Output = Result<Response, T::Error>;
 
@@ -305,9 +304,10 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{handler::Handler, routing::get, test_helpers::*, Router};
+    use crate::{async_trait, handler::Handler, routing::get, test_helpers::*, Router};
     use axum_core::extract::FromRef;
     use http::{header, request::Parts, StatusCode};
+    use tower_http::limit::RequestBodyLimitLayer;
 
     #[tokio::test]
     async fn test_from_extractor() {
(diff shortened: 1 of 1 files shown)


jplatte (approved):

kamulos:
Thank you very much for the quick fix :heart: 

Would it be possible to get a `0.5` release containing this commit at some point in the coming days? If it takes more time, no worries... I'll integrate it in my application in another way

davidpdrsn:
The bounds are actually being used on 0.5 so would require more work to backport, so since 0.6 is quite far along I'd rather no backport it. Once https://github.com/tokio-rs/axum/pull/1462 is merged I'm gonna do another 0.6 rc with this change and then you can use that, assuming that you're able to use 0.6 in the first place. Otherwise your best bet is probably not to use `from_extractor` but write your own middleware.

kamulos:
All right, sounds good. The 0.6 features look really nice :+1:

PR #1477, opened 2022-10-13
Enforce Sync for inner services and handlers

jplatte:
**Motivation:** Makes `Router` and `RouterService` `Sync`.

The `box_clone_sync_service` is copied from tower and adjusted to include `Sync`, none of the docs since it is private. It could probably be upstreamed into tower if we decide that we want this.

### DIFF
--- a/axum-extra/src/handler/mod.rs
+++ b/axum-extra/src/handler/mod.rs
@@ -166,7 +166,7 @@ pub struct IntoHandler<H, T, S, B> {
 
 impl<H, T, S, B> Handler<T, S, B> for IntoHandler<H, T, S, B>
 where
-    H: HandlerCallWithExtractors<T, S, B> + Clone + Send + 'static,
+    H: HandlerCallWithExtractors<T, S, B> + Clone + Send + Sync + 'static,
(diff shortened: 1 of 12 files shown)


davidpdrsn:
I'm a little afraid of this since we used to require `Sync` in the past but reverted it because it broke other stuff (https://github.com/tokio-rs/axum/pull/273). At least requires a bunch of testing with real projects that uses axum to see the impact.

jplatte:
Yeah I agree this needs extra testing. After the coming `rc.3`, I can port a few things to that first, to then see whether this commit works equally well.

davidpdrsn:
I'll also test it with our stuff at Embark.

davidpdrsn:
Putting this on the 0.7 milestone. I'm reluctant to increase the scope of 0.6 now.

jplatte:
Rebased. Hope you find some time to test it, I'm probably not available for that at least for a few weeks.

davidpdrsn:
I've added this to the todo list for 0.7 but probably easier to redo it there rather than trying to rebase it given all the changes. I'll close this for now but still think its something we should consider for 0.7.

PR #1487, opened 2022-10-19
Add RawForm extractor

### LINKED ISSUE #1483: Add a RawForm extractor
### Motivation

In my route, at the moment of extraction, I don't yet know which struct I want to decode the parameters into, however, I do want axum to take care of extracting raw data from either the query string in GET method, or from the request body in other methods, together with checking for the `application/x-www-form-urlencoded` mime type.

### Proposal

Similarly to the existing `RawBody` and `RawQuery`, add a `RawForm` extractor that just extract the params as a `String`, but doesn't attempt to deserialize it with serde.

### Alternatives

One alternative could be to specialize `Form<String>` somehow to deserialize into raw query string. Considering that it's powered by a 3rd-party library, which doesn't allow it today, I guess this would be a more involved process, plus it might be a less obvious solutions to users as it's not consistent with the earlier mentioned existing `RawBody` / `RawQuery` extractors.

nylonicious:
Closes #1483

### DIFF
--- a/axum/src/extract/raw_form.rs
+++ b/axum/src/extract/raw_form.rs
@@ -10,8 +10,10 @@ use super::{
 
 use crate::{body::HttpBody, BoxError};
 
-/// Extractor that extracts the query bytes from the GET request or body in other methods
-/// with expecting `Content-Type` header value to be `application/x-www-form-urlencoded`.
+/// Extractor that extracts raw form requests.
+///
+/// For `GET` requests it will extract the raw query. For other methods it extracts the raw
+/// `application/x-www-form-urlencoded` encoded request body.
 ///
 /// # Example
 ///
--- a/axum/src/extract/raw_form.rs
+++ b/axum/src/extract/raw_form.rs
@@ -1,0 +1,50 @@
+use async_trait::async_trait;
+use axum_core::extract::FromRequest;
+use http::{Method, Request};
+
+use super::{
+    has_content_type,
+    rejection::{InvalidFormContentType, RawFormRejection},
+};
+
+use crate::{body::HttpBody, BoxError};
+
+/// Extractor that extracts the raw form string, without parsing it.
+///
+/// # Example
+/// ```rust,no_run
+/// use axum::{
+///     extract::RawForm,
+///     routing::get,
+///     Router
+/// };
+///
+/// async fn handler(RawForm(form): RawForm) {}
+///
+/// let router = Router::new().route("/", get(handler));
+/// ```
+#[derive(Debug)]
+pub struct RawForm(pub String);
+
+#[async_trait]
+impl<S, B> FromRequest<S, B> for RawForm
+where
+    B: HttpBody + Send + 'static,
+    B::Data: Send,
+    B::Error: Into<BoxError>,
+    S: Send + Sync,
+{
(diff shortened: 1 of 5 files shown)


axum/src/extract/raw_form.rs
```diff
@@ -0,0 +1,54 @@
...
+///
+/// async fn handler(RawForm(form): RawForm) {}
+///
+/// let app = Router::new().route("/", get(handler));
+/// # async {
+/// # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
+/// # };
+/// ```
+#[derive(Debug)]
+pub struct RawForm(pub Option<String>);
```
davidpdrsn, line 31: I think using an `Option` here might give the impression that it'll `None` instead of returning a rejection if something fails.

I'm wondering if we should just use an empty string in the case there is no query and remove the Option. What do you think?

axum/src/extract/raw_form.rs
```diff
@@ -0,0 +1,54 @@
...
+use http::{Method, Request};
+
+use super::{
+    has_content_type,
+    rejection::{InvalidFormContentType, RawFormRejection},
+};
+
+use crate::{body::HttpBody, BoxError};
+
+/// Extractor that extracts the raw form string, without parsing it.
```
davidpdrsn, line 12: Wanna mention that this looks at the query for GET and body for other methods, and that it requires the right content-type?

axum/src/extract/raw_form.rs
```diff
@@ -0,0 +1,54 @@
...
+            Ok(Self(req.uri().query().map(String::from)))
+        } else {
+            if !has_content_type(req.headers(), &mime::APPLICATION_WWW_FORM_URLENCODED) {
+                return Err(InvalidFormContentType.into());
+            }
+
+            Ok(Self(Some(String::from_request(req, state).await?)))
+        }
+    }
+}
```
davidpdrsn, line 54: Wanna add some tests?

axum/src/extract/raw_form.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    async fn from_request(req: Request<B>, state: &S) -> Result<Self, Self::Rejection> {
+        if req.method() == Method::GET {
+            Ok(Self(req.uri().query().map(String::from)))
+        } else {
+            if !has_content_type(req.headers(), &mime::APPLICATION_WWW_FORM_URLENCODED) {
+                return Err(InvalidFormContentType.into());
+            }
+
+            Ok(Self(Some(String::from_request(req, state).await?)))
```
davidpdrsn, line 51: Would using `Bytes` to allow non-utf8 be better?

jplatte:
Should we implement `Form` in terms of `RawForm` maybe (@davidpdrsn)?

davidpdrsn:
> Should we implement `Form` in terms of `RawForm` maybe (@davidpdrsn)?

Yes thats a good idea! I don't necessarily think it needs to happen in this PR though.

nylonicious:
CI failure seems unrelated.

jplatte:
Yeah, CI failure is unrelated, fix in #1488.

davidpdrsn (approved):
Thanks!

PR #1492, opened 2022-10-20
Add example for `Router::inherit_state`

davidpdrsn:


axum/src/routing/mod.rs
```diff
@@ -147,6 +147,31 @@ where
...
+    ///
+    /// // A router that will be merged into the `app` router.
+    /// let merged_router = Router::inherit_state()
+    ///     .route("/baz", get(|state: State<AppState>| async {}));
+    ///
+    /// let app = Router::with_state(AppState {})
+    ///     .route("/", get(|state: State<AppState>| async {}))
+    ///     .nest("/foo", nested_router)
+    ///     .merge(merged_router);
+    /// # let _: Router<AppState> = app;
```
jplatte, line 173: Maybe for clarity add a note at the end like "app now contains handlers for `/`, `/foo/bar` and `/baz`"?

jplatte (approved):

axum/src/routing/mod.rs
```diff
@@ -147,6 +147,33 @@ where
...
+    /// // A router that will be merged into the `app` router.
+    /// let merged_router = Router::inherit_state()
+    ///     .route("/baz", get(|state: State<AppState>| async {}));
+    ///
+    /// let app = Router::with_state(AppState {})
+    ///     .route("/", get(|state: State<AppState>| async {}))
+    ///     .nest("/foo", nested_router)
+    ///     .merge(merged_router);
+    ///
+    /// `app` now has routes for `/`, `/foo/bar`, and `/baz` that all use the same state.
```
jplatte, line 174:
```suggestion
    /// // `app` now has routes for `/`, `/foo/bar`, and `/baz` that all use the same state.
```

PR #1501, opened 2022-10-24
Add example of static-file-server using serve dir with handler as service

SET001:
This PR simply adds another case in static-file-server example showing how it is possible to use a handler function with ServeDir

### DIFF
--- a/examples/static-file-server/src/main.rs
+++ b/examples/static-file-server/src/main.rs
@@ -89,12 +89,12 @@ fn using_serve_dir_only_from_root_via_fallback() -> Router {
 }
 
 fn using_serve_dir_with_handler_as_service() -> Router {
-    // you can convert handler function to service
-    async fn handler_404() -> impl IntoResponse {
+    async fn handle_404() -> (StatusCode, &'static str) {
         (StatusCode::NOT_FOUND, "Not found")
     }
-    
-    let service = handler_404
+
+    // you can convert handler function to service
+    let service = handle_404
         .into_service()
         .map_err(|_| io::Error::from(io::ErrorKind::Other));
 
--- a/examples/static-file-server/src/main.rs
+++ b/examples/static-file-server/src/main.rs
@@ -103,7 +103,7 @@ fn using_serve_dir_with_handler_as_service() -> Router {
 
     Router::new()
         .route("/foo", get(|| async { "Hi from /foo" }))
-        .fallback_service(serve_dir)    
+        .fallback_service(serve_dir)
 }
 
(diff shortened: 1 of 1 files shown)


examples/static-file-server/src/main.rs
```diff
@@ -86,6 +88,24 @@ fn using_serve_dir_only_from_root_via_fallback() -> Router {
...
 
+fn using_serve_dir_with_handler_as_service() -> Router {
+    // you can convert handler function to service
+    async fn handler_404() -> impl IntoResponse {
+        (StatusCode::NOT_FOUND, "Not found")
+    }
+    
+    let service = handler_404
+        .into_service()
+        .map_err(|_| io::Error::from(io::ErrorKind::Other));
```
davidpdrsn, line 99: Does this work?

```suggestion
        .map_err(|err| match err {}));
```

examples/static-file-server/src/main.rs
```diff
@@ -86,6 +88,24 @@ fn using_serve_dir_only_from_root_via_fallback() -> Router {
         .fallback_service(serve_dir)
 }
 
+fn using_serve_dir_with_handler_as_service() -> Router {
+    // you can convert handler function to service
+    async fn handler_404() -> impl IntoResponse {
+        (StatusCode::NOT_FOUND, "Not found")
+    }
+    
+    let service = handler_404
```
davidpdrsn, line 97:
```suggestion
    async fn handle_404() -> (StatusCode, &'static str) {
        (StatusCode::NOT_FOUND, "Not found")
    }

    // you can convert handler function to service
    let service = handle_404
```

examples/static-file-server/src/main.rs
SET001: No. With this it throws error:
```
error[E0277]: the trait bound `std::io::Error: From<()>` is not satisfied
   --> static-file-server/src/main.rs:103:21
    |
103 |     let serve_dir = get_service(serve_dir).handle_error(handle_error);
    |                     ^^^^^^^^^^^ the trait `From<()>` is not implemented for `std::io::Error`
```

davidpdrsn: Ah right. This works:

```rust
        .map_err(|err| -> std::io::Error { match err {} });
```

I think we should use that because it makes it clear that we never actually create an `io::Error`.

SET001: yeah, this one works. Commited

davidpdrsn (approved):
Thanks!

Issue #1515, opened 2022-11-09
app building fine with rc.2 fails to build with rc.3, in matched_path.rs


olivier-fs:
### Version

```
axum = { version = "0.6.0-rc.3", features = ["http1", "json", "tokio"] }
tokio = { version = "1.21.2", default-features = false, features = ["macros", "rt-multi-thread"] }
tower-http = { version = "0.3.4", features = ["compression-gzip", "fs", "set-header", "timeout", "trace"] }
```

```shell
$ cargo tree | grep axum
├── axum v0.6.0-rc.3
│   ├── axum-core v0.3.0-rc.3
```
```shell
$ cargo --version
cargo 1.65.0 (4bc8f24d3 2022-10-20)
```

### Platform

Windows 10

### Description

An axum app that was building fine with rc.2 doesn't build with rc.3.
I fixed some breaking but trivial fixes (e.g. nest() => nest_service()) but still :
```shell
$ cargo run --release
   Compiling proc-macro2 v1.0.43
   ...
   ...
   Compiling axum v0.6.0-rc.3
error[E0308]: `if` and `else` have incompatible types
   --> C:\Users\olivier\.cargo\registry\src\github.com-1ecc6299db9ec823\axum-0.6.0-rc.3\src\extract\matched_path.rs:135:12
    |
133 |       let matched_path = if let Some(matched_path) = route_id_to_path.get(&id) {
    |                          ----------------------------------------------------- `if` and `else` have incompatible types
134 |           matched_path
    |           ------------ expected because of this
135 |       } else {
    |  ____________^
136 | |         #[cfg(debug_assertions)]
137 | |         panic!("should always have a matched path for a route id");
138 | |     };
    | |_____^ expected `&Arc<str>`, found `()`
    |
    = note: expected reference `&Arc<str>`
               found unit type `()`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `axum` due to previous error
warning: build failed, waiting for other jobs to finish...
```
The odd part is :
- the same code builds fine in debug mode i.e. without --release
- another axum with the "same" setup i.e. route() with get() handlers, nest_service(), and fallback(), builds OK

rustc, that is nearly always giving good details about the problem and the way to fix it, doesn't in my case, and doesn't even give some filename/line featuring the "culprit" code...

Building with nightly (cargo 1.67.0-nightly (9286a1beb 2022-11-04)) doesn't change anything (as expected, it is a simple if let statement, not the new let chain).

[...]

PR #1517, opened 2022-11-09
Fix compile error in release mode

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1515

### DIFF
--- a/axum/src/extract/matched_path.rs
+++ b/axum/src/extract/matched_path.rs
@@ -125,6 +125,7 @@ where
 #[derive(Clone, Debug)]
 struct MatchedNestedPath(Arc<str>);
 
+#[allow(unreachable_code)]
 pub(crate) fn set_matched_path_for_request(
     id: RouteId,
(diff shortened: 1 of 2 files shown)


axum/src/extract/matched_path.rs
```diff
@@ -125,6 +125,7 @@ where
 #[derive(Clone, Debug)]
 struct MatchedNestedPath(Arc<str>);
 
+#[allow(unreachable_code)]
```
jplatte, line 128:
```suggestion
```

axum/src/extract/matched_path.rs
```diff
@@ -135,6 +136,7 @@ pub(crate) fn set_matched_path_for_request(
     } else {
         #[cfg(debug_assertions)]
         panic!("should always have a matched path for a route id");
+        return;
```
jplatte, line 139:
```suggestion
        #[cfg(not(debug_assertions))]
        return;
```

Issue #1416, opened 2022-09-25
Figure out fallback inheritance


davidpdrsn:
I'll elaborate on this later. This issue exists to be part of the 0.6 milestone so people can see that there is more work to be done.

PR #1521, opened 2022-11-11
Add fallback inheritance for nested routers

davidpdrsn:
That was actually easier than I had feared 😅

Fixes https://github.com/tokio-rs/axum/issues/1416

### DIFF
--- a/axum/src/routing/service.rs
+++ b/axum/src/routing/service.rs
@@ -1,4 +1,6 @@
-use super::{future::RouteFuture, url_params, Endpoint, Node, Route, RouteId, Router};
+use super::{
+    future::RouteFuture, url_params, Endpoint, FallbackRoute, Node, Route, RouteId, Router,
+};
 use crate::{
     body::{Body, HttpBody},
     response::Response,
@@ -11,6 +13,7 @@ use std::{
     sync::Arc,
     task::{Context, Poll},
 };
+use sync_wrapper::SyncWrapper;
 use tower::Service;
 
 /// A [`Router`] converted into a [`Service`].
@@ -18,7 +21,7 @@ use tower::Service;
 pub struct RouterService<B = Body> {
     routes: HashMap<RouteId, Route<B>>,
     node: Arc<Node>,
-    fallback: Route<B>,
+    fallback: FallbackRoute<B>,
 }
 
 impl<B> RouterService<B>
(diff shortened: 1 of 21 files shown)


axum/src/routing/service.rs
```diff
@@ -11,14 +13,15 @@ use std::{
...
+use sync_wrapper::SyncWrapper;
 use tower::Service;
 
 /// A [`Router`] converted into a [`Service`].
 #[derive(Debug)]
 pub struct RouterService<B = Body> {
     routes: HashMap<RouteId, Route<B>>,
     node: Arc<Node>,
-    fallback: Route<B>,
+    fallback: FallbackRoute<B>,
```
jplatte, line 24: I'm not a big fan of this still being a thing at the service level (and using request extensions to make things work). Can we not copy the fallback handler to nested routers that don't have a custom one when building the `RouterService`?

jplatte: Well we could store nested routers (that want to inherit state) differently from nested services, same as we already do for method routers vs. services, no?

davidpdrsn: Almost 😅 we get into trouble when we have to apply middleware to such nested routers. When applying layers to such routers we have to call `Router::layer`, since they don't implement `Service` so we cannot apply the layer around the whole thing [like we do today](https://github.com/tokio-rs/axum/blob/main/axum/src/routing/mod.rs#L426). But doing that subtly changes the behavior:

```rust
Router::new()
    .nest(
        "/foo",
        Router::new().route("/bar", ...),
    )
    .layer(log_url)
```

Here if `/foo/bar` is called `log_url` will see the URL with the `/foo` prefix, as it should because its applied around the whole router.

We would instead end up doing something equivalent to calling `Router::layer` on the inner router:

```rust
Router::new()
    .nest(
        "/foo",
        Router::new().route("/bar", ...).layer(log_url),
    )
```

But now `log_url` will see the URL with `/foo` removed, i.e. `/bar`.

We also cannot store the layer and apply it later, since it changes the request body and introduces new generic parameters.

davidpdrsn: I tried but failed 😞 Ran into some issues that I dunno how to resolve. See https://github.com/tokio-rs/axum/pull/1531

davidpdrsn: I'm leaning towards merging this as is. We can always refactor things internally.

This together with https://github.com/tokio-rs/axum/pull/1532 _might_ be the final release candidate 🤞

jplatte (approved):

davidpdrsn:
For the record we agreed to get this in now as is and refactor the internals later. See https://github.com/tokio-rs/axum/pull/1531.

PR #1523, opened 2022-11-11
Add sample demonstrating static file served behind authentication

ledoyen:
This change adds a sample, demonstrating how to use `ServeDir` (and `ServeFile`) inside a method handler.

The purpose is to be able to leverage other injection mechanisms (`FromRequest`, `FromRequestParts`, `Extension`, etc.).

My use case was to gate static files behind authentication.

## Motivation
As a rust newbie I invested quite some time to figure out how to do this, so I figure it could help someone else.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,6 @@ members = [
     "axum",
     "axum-core",
     "axum-extra",
-    "axum-macros",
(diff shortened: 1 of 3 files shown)


ledoyen:
ok

PR #1524, opened 2022-11-12
Update link to rust on nails

parK-dev:
## Motivation

I was reading through the documentation on axum and found that Rust On Nails has changed its URL.

## Solution

Simply changed the URL in the doc.


davidpdrsn (approved):
Thanks!

PR #1531, opened 2022-11-15
Fallback inheritance, attempt 2

davidpdrsn:
This doesn't actually work. See https://github.com/tokio-rs/axum/blob/fallback-inheritance-attempt-2/axum/src/boxed.rs#L170-L177


jplatte:
After looking at it for a bit, I'm pretty sure this can be made to work if `Map` stores the fallback alongside the layer and applies it later. I think that applying later thing "just" needs an extra interface somewhere. I also think this would be a little easier to design if at least at first boxed handlers and boxed routers were separate. I can try to find some time for a third attempt at this problem, if you think more internal complexity is acceptable.

davidpdrsn:
That'd be great! I don't mind the additional complexity. Having used the current boxed handler stuff in anger now I feel I understand it better 😊

How do you feel about merging https://github.com/tokio-rs/axum/pull/1521 as is (using extensions) and then changing the internals later? Both approaches would have the same public api I suppose.

jplatte:
Yeah, let's do that.

davidpdrsn:
Closing in favor of https://github.com/tokio-rs/axum/pull/1521

Issue #1516, opened 2022-11-09
Can't turn a `Router` that wants to inherit state into a service


gd87429:
Updating from `0.6.0-rc.2` to `0.6.0-rc.3`:

I have a bunch of mods & sub-mods that look like this:

```rust
pub fn get_routes(shared_state: Arc<ServerState>) -> Router<Arc<ServerState>> {
	Router::with_state(shared_state.clone())
		.nest("/sdfg", get_sdfg_routes(shared_state))
}
```

However when trying to remove passing in clones of state and using `inherit_state()`, everything compiles fine but during startup I get: `Can't turn a 'Router' that wants to inherit state into a service`

Which makes me wonder:

1. https://docs.rs/axum/0.6.0-rc.3/axum/handler/trait.Handler.html#method.into_service says that handlers that need state should use `with_state`. Doesn't that defeat the purpose of not passing in state? How do I turn it into something like this and pass in state once at the very top `merge`:

	```rust
	pub fn get_routes() -> Router {
		Router::inherit_state()
			.nest("/sdfg", get_sdfg_routes())
	}
	```

2. Can't it throw an error during compilation instead?

davidpdrsn:
You should only set the state once, at the top level router:

```rust
use axum::Router;

#[tokio::main]
async fn main() {
    let app = Router::with_state(AppState {}).nest("/api", api_router());
}

#[derive(Clone)]
struct AppState {}

fn api_router() -> Router<AppState> {
    // don't pass state here
    Router::inherit_state()
}
```

> Can't it throw an error during compilation instead?

It can but that complicates the types quite a lot. We considered it and currently don't think its worth the additional complexity.

gd87429:
Thank you for the quick response. I seem to be doing the same, but continue having the exact same error. Is something like this supposed to work?

```rust
use axum::Router;
use std::sync::Arc;

#[derive(Clone)]
struct AppState {}
let state = Arc::new(AppState {});

#[tokio::main]
async fn main() {
    let app = Router::with_state(state).merge(routes());
}

fn routes() -> Router<Arc<AppState>> {
    Router::inherit_state().route("/v1", v1_routes());
}

fn v1_routes() -> Router<Arc<AppState>> {
    Router::inherit_state().route("/get", get());
}
```

jplatte:
[...]

PR #1532, opened 2022-11-15
Type safe state inheritance

davidpdrsn:
Lately I've been wanting to explore a more type safe way to passing state to routers, that don't run things like https://github.com/tokio-rs/axum/issues/1516. It think its a little unfortunate that axum 0.6's headline feature is type safe state, only for `inherit_state` to not actually be type safe.

This changes the API such that `Router::with_state` requires the state which is then passed to everything when building the final `RouterService`:

```rust
let service: RouterService = Router::new()
    .route(...)
    .with_state(state);
```

The upside of this approach is that we don't need additional type parameters or runtime stuff to track whether the state has been supplied or not.

Nesting (and merging) is only possible with a `Router` which doesn't have the state so it becomes more type safe. You can still use `nest_service` but that requires turning the `Router` into a `RouterService` by supplying the state.

@jplatte what do you think?

Fixes https://github.com/tokio-rs/axum/issues/1516

## TODO

- [x] Update docs
- [x] Changelog


jplatte:
> You can still use `nest_service` but that requires turning the `Router` into a `RouterService` by supplying the state.

I think the reason I didn't do it this way is that you then can't have fallback inheritance for nested routers that have a different state type.

davidpdrsn:
> I think the reason I didn't do it this way is that you then can't have fallback inheritance for nested routers that have a different state type.

Yeah I ran into that here. I changed `nest` and `merge` to require the same state type. I think thats an okay compromise. I think I'd rather have type safety.

jplatte:
Well, I still somewhat strongly prefer the previous compromise because while you could be surprised about that too, the surprise with nesting a `RouterService` and expecting fallback inheritance to work same as with nesting a `Router` can happen while the service is already in prod. That's not the case with the panic from attempting to turn an inheriting `Router` into a service, whether directly or indirectly.

axum/CHANGELOG.md
```diff
@@ -30,6 +31,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 [#1529]: https://github.com/tokio-rs/axum/pull/1529
 [#1532]: https://github.com/tokio-rs/axum/pull/1532
+=======
+- **fixed:** Nested routers will now inherit fallbacks from outer routers ([#1521])
+- **added:** Add `accept_unmasked_frames` setting in WebSocketUpgrade ([#1529])
+
+[#1521]: https://github.com/tokio-rs/axum/pull/1521
+>>>>>>> main
```
jplatte, line 39: You left in some conflict markers here ;)

jplatte:
We discussed this on Discord (privately, sorry) and because of how state inheritance is implemented now, this can't actually cause the problem that I was worried about.

I guess it makes it impossible to merge two routers with different state though? Maybe that's fine. It's at least not something that you notice when the service is already running 🙂

jplatte (approved):

jplatte:
The changelog for this is very weird, I only noticed now. There is a double `:` and it talks about inherit_**fallback** and that you should use `with_state` (rather than just `new`).

davidpdrsn:
@jplatte oh good catch! Fixed in 2e3000f1a302b

zvkemp:
This change seems to have broken `into_make_service()` for routers not using the default state; was that decision was intentional? As `IntoMakeService::new()` is crate-private, it's unclear how we are now supposed to serve routers with a custom state.

davidpdrsn:
@zvkemp it didn't. You gotta call `Router::with_state` to provide the state. See the examples in the docs or in the repo.

zvkemp:
Yes, we are calling `with_state`.

This method (as well as `into_make_service_with_connect_info`) is only implemented on `Router<(), B>`, not `Router<S, B>` as they were previously.

https://docs.rs/axum/0.6.0/axum/routing/struct.Router.html#method.into_make_service

zvkemp:
Oh actually I think I understand what is happening — I didn't realize that the signature of `with_state` doesn't specify that the state argument will become the router's state, and we had a type hint specifying our router would be built as a `Router<MyState, _>` rather than the required `Router<(), _>` for it to be a Service. Removing that type hint fixes the issue.

jplatte:
It sounds like you probably have a type annotation for the router somewhere that makes it `Router<AppState>` after `with_state`. Conceptually the `S` type parameter no longer represents the state type used by already-added routes now in the final 0.6.0, it represents the type that is used for further routes being added, and that still has to be supplied.

Anyways, if you still can't figure it out based on that, please open a discussion with a reproducible example instead of posting to a merged PR.

PR #1535, opened 2022-11-17
Add example of how to test WebSockets

davidpdrsn:
We occasionally get questions about how to test websockets so figured it was worth writing an example for.

### DIFF
--- a/examples/testing-websockets/src/main.rs
+++ b/examples/testing-websockets/src/main.rs
@@ -1,0 +1,149 @@
+//! Run with
+//!
+//! ```not_rust
+//! cargo test -p example-testing-websockets
+//! ```
+
+use axum::{
+    extract::{
+        ws::{Message, WebSocket},
+        WebSocketUpgrade,
+    },
+    response::Response,
+    routing::get,
+    Router,
+};
+use futures::{Sink, SinkExt, Stream, StreamExt};
+use std::net::SocketAddr;
+
+#[tokio::main]
+async fn main() {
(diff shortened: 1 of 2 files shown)


examples/testing-websockets/src/main.rs
```diff
@@ -0,0 +1,149 @@
...
+        // `Stream`
+        let (socket_write, mut test_rx) = futures::channel::mpsc::channel(1024);
+        let (mut test_tx, socket_read) = futures::channel::mpsc::channel(1024);
+
+        tokio::spawn(unit_testable_handle_socket(socket_write, socket_read));
+
+        test_tx
+            .send(Ok(Message::Text("foo".to_owned())))
+            .await
+            .unwrap();
```
programatik29, line 140: Can be improved slightly like:

```rust
let (socket_write, mut test_rx) = futures::channel::mpsc::channel(1024);

let msgs = vec![Message::Text("foo".to_owned())];
let msg_stream = stream::iter(msgs.into_iter().map(Ok));

tokio::spawn(unit_testable_handle_socket(socket_write, msg_stream));
```

davidpdrsn: I kinda wanna keep it simple as I think thats easier for people to adapt to their needs 🤔

programatik29: Both patterns look useful. This one is more convenient for multiple messages sent at once.

programatik29: This pattern can be added separately or we can keep it simple and omit it. Either looks fine.

PR #1536, opened 2022-11-17
Fix `Handler::with_state` not working if request body was changed via layer

davidpdrsn:
Previously

```rust
handler.layer(RequestBodyLimitLayer::new(...)).with_state(...)
```

didn't work because we required the same request body all the way through.

### DIFF
--- a/axum/src/handler/mod.rs
+++ b/axum/src/handler/mod.rs
@@ -139,9 +139,10 @@ pub trait Handler<T, S, B = Body>: Clone + Send + Sized + 'static {
     /// # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
     /// # };
     /// ```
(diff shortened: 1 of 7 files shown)


axum/src/handler/mod.rs
```diff
@@ -220,13 +221,13 @@ all_the_tuples!(impl_handler);
 /// A [`Service`] created from a [`Handler`] by applying a Tower middleware.
 ///
 /// Created with [`Handler::layer`]. See that method for more details.
-pub struct Layered<L, H, T, S, B> {
+pub struct Layered<L, H, T, S, B, B2> {
```
davidpdrsn, line 224: How many type params are too many 🙃 Couldn't find a way to make it work with adding `B2`.

jplatte (approved):
I haven't fully grokked the problem, but it's all type system stuff so considering you have a test that failed before and works now, seems fine.

PR #1538, opened 2022-11-17
Improve error when nesting twice at root

davidpdrsn:
Heard from someone in Discord today who tried to do

```rust
Router::new()
    .nest("/", foo)
    .nest("/", bar)
```

In that case the error should probably point towards `Router::merge`

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -69,6 +69,7 @@ pub struct Router<S = (), B = Body> {
     routes: HashMap<RouteId, Endpoint<S, B>>,
     node: Arc<Node>,
     fallback: Fallback<S, B>,
+    service_nested_at_root: bool,
 }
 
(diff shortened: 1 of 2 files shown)


axum/src/routing/mod.rs
```diff
@@ -300,6 +303,14 @@ where
             path = "/";
         }
 
+        if path == "/" {
+            if self.service_nested_at_root {
+                panic!("Cannot nest more than one service at `/`. Consider using `Router::merge` instead.");
```
jplatte, line 308: Maybe add "if applicable"? Since this is only possible if the nested service is another router. Alternatively, we could have this check in `nest` instead of `nest_service`.

davidpdrsn:
Think I'll close this. If we move to nest flattering routes this wont be as big a problem anyway.

PR #1539, opened 2022-11-17
Add `on_failed_upgrade` callback to `WebSocketUpgrade`

davidpdrsn:
Previously if upgrading a connection to a WebSocket connection failed the background task would panic. There was no way to customize that so users that might wanna report the error was out of luck.

Panicking also wasn't great because users might abort on panics which would bring down the server.

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -134,18 +134,30 @@ use tokio_tungstenite::{
 /// rejected.
 ///
 /// See the [module docs](self) for an example.
-#[derive(Debug)]
 #[cfg_attr(docsrs, doc(cfg(feature = "ws")))]
-pub struct WebSocketUpgrade {
+pub struct WebSocketUpgrade<F = DefaultOnFailedUpdgrade> {
     config: WebSocketConfig,
     /// The chosen protocol sent in the `Sec-WebSocket-Protocol` header of the response.
     protocol: Option<HeaderValue>,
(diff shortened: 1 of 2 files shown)


axum/src/extract/ws.rs
```diff
@@ -134,18 +134,30 @@ use tokio_tungstenite::{
...
 }
 
-impl WebSocketUpgrade {
+impl<F> std::fmt::Debug for WebSocketUpgrade<F> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("WebSocketUpgrade")
+            .field("config", &self.config)
+            .field("protocol", &self.protocol)
+            .field("sec_websocket_key", &self.sec_websocket_key)
+            .field("on_upgrade", &self.on_upgrade)
```
jplatte, line 154:
```suggestion
```

https://docs.rs/hyper/latest/src/hyper/upgrade.rs.html#225-229

axum/src/extract/ws.rs
```diff
@@ -134,18 +134,30 @@ use tokio_tungstenite::{
...
+            .field("sec_websocket_protocol", &self.sec_websocket_protocol)
+            .finish()
```
jplatte, line 156:
```suggestion
            .finish_non_exhaustive()
```

jplatte (approved):

jplatte (approved):

PR #1540, opened 2022-11-18
Make sure fallbacks are still inherited for opaque services

davidpdrsn:
Adds tests to https://github.com/tokio-rs/axum/pull/1532 that ensures fallbacks are still inherited for `RouterService` and opaque services that contain a `RouterService`.


jplatte (approved):
♥️

Issue #1313, opened 2022-08-24
Document how state works with `merge` and `nest`


davidpdrsn:
Gotten a few questions about how states work with `merge` and `nest`. Things like whether the state needs to be passed to each router and whether the state of the inner routers is dropped. This is something we should document.

genusistimelord:
Yeah I noticed this when using AxumFlash and passing the config as the State. I had to change my Routing functions to take in the State to set them individually. 

```
Router::with_state(flash_config.clone())
        .route("/", get(root))
       .merge(main_routes(flash_config.clone()))
```

which results in 
```
pub fn main_routes(config: axum_flash::Config) -> Router<axum_flash::Config> {
    Router::with_state(config)
        .route("/main", get(main_page))
}
```

davidpdrsn:
https://github.com/tokio-rs/axum/pull/1368 is making things more flexible. When that is in I'll write some docs for it.

davidpdrsn:
I think we're in a good place in terms for docs. We can expand things when people ask questions but for now I'll close this.

davidpdrsn:
Good idea. Do you wanna make a pr?

PR #1543, opened 2022-11-19
Improve `State` and `Router` docs

jimmycuadra:
Per [my comment in #1313](https://github.com/tokio-rs/axum/issues/1313#issuecomment-1314955380), this improves discoverability of how `State` works with nested/merged `Router`s.

I made a few other minor documentation fixes while I was reading:

* Corrected the docs for `Router::nest` to say that it nests another `Router` rather than a `Service`. There is a separate function (`Router::nest_service`) for the latter.
* Changed one of the headings in nest.md to use more idiomatic English.
* Changed awkward phrasing under the "Sharing state with handlers" heading of the root documentation page.
* Removed a trailing period from one of three list items for consistency.

### DIFF
--- a/axum/src/docs/routing/merge.md
+++ b/axum/src/docs/routing/merge.md
@@ -43,9 +43,6 @@ When combining [`Router`]s with this function, each [`Router`] must have the
 same type of state. See ["Combining stateful routers"][combining-stateful-routers]
 for details.
 
-If you want to compose axum services with different types of state, use
-[`Router::nest_service`].
-
 # Panics
 
 - If two routers that each have a [fallback](Router::fallback) are merged. This
--- a/axum/src/docs/routing/merge.md
+++ b/axum/src/docs/routing/merge.md
@@ -37,7 +37,18 @@ let app = Router::new()
 # };
 ```
 
-## Panics
+# Merging routers with state
+
+When combining [`Router`]s with this function, each [`Router`] must have the
+same type of state. See ["Combining stateful routers"][combining-stateful-routers]
+for details.
+
+If you want to compose axum services with different types of state, use
+[`Router::nest_service`].
+
+# Panics
 
 - If two routers that each have a [fallback](Router::fallback) are merged. This
   is because `Router` only allows a single fallback.
+
+[combining-stateful-routers]: crate::extract::State#combining-stateful-routers
--- a/axum/src/extract/state.rs
+++ b/axum/src/extract/state.rs
@@ -46,13 +46,76 @@ use std::{
 /// # let _: axum::routing::RouterService = app;
 /// ```
 ///
-/// [`Router`]s that are combined with [`Router::nest`] or [`Router::merge`]
-/// generally require the same type of state. See [`Router::nest`] for how to
-/// combine routers with different types of state.
-/// 
+/// ## Combining stateful routers
+///
+/// Multiple [`Router`]s can be combined with [`Router::nest`] or [`Router::merge`]
+/// When combining [`Router`]s with one of these methods, the [`Router`]s must have
+/// the same state type. Generally, this can be inferred automatically:
+///
+/// ```
+/// use axum::{Router, routing::get, extract::State};
+///
+/// #[derive(Clone)]
+/// struct AppState {}
+///
+/// let state = AppState {};
+///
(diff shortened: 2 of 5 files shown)


axum/src/extract/state.rs
```diff
@@ -46,6 +46,15 @@ use std::{
...
 /// ```
 ///
+/// [`Router`][router]s that are combined with [`Router::nest`][router-nest]
+/// or [`Router::merge`][router-merge] generally require the same type of state.
+/// See [`Router::nest`][router-nest] for how to combine routers with different
+/// types of state.
+///
+/// [router]: crate::Router
+/// [router-nest]: crate::Router::nest
+/// [router-merge]: crate::Router::merge
```
davidpdrsn, line 56:
```suggestion
/// [`Router`]s that are combined with [`Router::nest`] or [`Router::merge`]
/// generally require the same type of state. See [`Router::nest`] for how to
/// combine routers with different types of state.
/// 
/// [`Router`]: crate::Router
/// [`Router::nest`]: crate::Router::nest
/// [`Router::merge`]: crate::Router::merge
```

davidpdrsn (approved):
Thanks!

davidpdrsn:
> This doesn't really make sense, because you can't construct a Router with a non-unit state type to pass to Router::nest. 

Sure you can. `Router::new` doesn't _always_ give a router whose state is `()`. The state type param only defaults to `()`. So unless specified it becomes `()`.

Just do:

```rust
fn api_v1() -> Router<AppState> {
    Router::new()
        .route("/teams", get(get_teams))
        .route("/teams/:slug", get(get_team))
}
```

davidpdrsn:
Nesting routers with different state is possible by calling `with_state` to get a `RouterService` then using `nest_service`. Like shown [here](https://docs.rs/axum/0.6.0-rc.5/axum/routing/struct.Router.html#nesting-a-router-with-a-different-state-type)

davidpdrsn:
> Ah, I see. So the state type of a router is inferred if `with_state` is called on it, but it can also be specified with an explicit annotation, even if the router doesn't actually (or yet) contain state of that type.

Yes exactly.

> I'm going to revise my changes tomorrow now that I understand it better. I suspect I won't be the last person to be confused by this without better docs. Thanks for the example!

Thanks!

axum/src/routing/mod.rs
```diff
@@ -212,6 +212,41 @@ where
...
+    ///
+    /// let app = Router::new()
+    ///     .route("/", get(outer_handler))
+    ///     .nest_service("/foo", inner_router)
+    ///     .with_state(OuterState {});
+    /// # let _: axum::routing::RouterService = app;
+    /// ```
+    ///
+    /// Note that the inner router will still inherit the fallback from the outer
+    /// router.
```
davidpdrsn, line 249: I think this section should go on the docs for `nest`. When people try to use `nest` with different states I think they're most likely to read the docs for `nest` to see how to do that.

axum/src/docs/routing/merge.md
```diff
@@ -37,7 +37,18 @@ let app = Router::new()
...
 
-## Panics
+# Merging routers with state
+
+When combining [`Router`]s with this function, each [`Router`] must have the
+same type of state. See ["Combining stateful routers"][combining-stateful-routers]
+for details.
+
+If you want to compose axum services with different types of state, use
+[`Router::nest_service`].
```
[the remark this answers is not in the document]
davidpdrsn: That isn't possible. `merge` needs a `Router` because it literally copied the routes from one to the other and random services don't have routes.

https://github.com/tokio-rs/axum/pull/1552 is gonna make it work though.

axum/src/routing/mod.rs
davidpdrsn: I disagree. The docs should be where users need them and that is when using `nest`.

davidpdrsn: I still disagree but this isn't a hill I'm willing to die on. So you win 😊

davidpdrsn (approved):

axum/src/routing/mod.rs
davidpdrsn: The whole thing except the first sentence.

PR #1550, opened 2022-11-19
Add `RouterService::{layer, route_layer}`

davidpdrsn:
Figure we might as well have these.


jplatte (approved):
This might lock us into the `RouterService` not erasing further details about the routes, but I guess that's fine at 0.6 since 0.7 or 1.0 is going to be a thing at some point anyways.

PR #1555, opened 2022-11-20
edition 2021 implies resolver v2

tshepang:


jplatte:
No, it doesn't for virtual manifests:

> If you are using a [virtual workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-manifest), you will still need to explicitly set the [`resolver` field](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) in the [workspace] definition if you want to opt-in to the new resolver.

Source: https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html

PR #1560, opened 2022-11-21
docs: add how to use Arc<AppState> with the cookies examples

woile:
## Motivation

Discussed [here](https://github.com/tokio-rs/axum/discussions/1556#discussioncomment-4192874)

## Solution

Add information about how to do it. I didn't want to be vague with the solution, so I included David's whole sample. But if you prefer otherwise I'll update 👍🏻

### DIFF
--- a/axum-extra/src/extract/cookie/private.rs
+++ b/axum-extra/src/extract/cookie/private.rs
@@ -69,6 +69,38 @@ use std::{convert::Infallible, fmt, marker::PhantomData};
 ///     .route("/get", get(get_secret));
 /// # let app: Router<_> = app;
 /// ```
+///
+/// If you have been using `Arc<AppState>` you cannot implement `FromRef<Arc<AppState>> for Key`.
+/// You can use a new type instead:
+///
+/// ```rust
+/// # use axum::extract::FromRef;
+/// # use axum_extra::extract::cookie::{PrivateCookieJar, Cookie, Key};
+/// use std::sync::Arc;
+/// use std::ops::Deref;
+///
+/// #[derive(Clone)]
+/// struct AppState(Arc<InnerState>);
+///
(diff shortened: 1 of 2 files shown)


axum-extra/src/extract/cookie/private.rs
```diff
@@ -69,6 +69,33 @@ use std::{convert::Infallible, fmt, marker::PhantomData};
 ///     .route("/get", get(get_secret));
 /// # let app: Router<_> = app;
 /// ```
+///
+/// If you have been using `Arc<AppState>` you cannot implement `FromRef<Arc<AppState>> for Key`.
+/// You can use a new type instead: `struct AppState(Arc<InnerState>);` and implement `Deref`.
```
davidpdrsn, line 74:
```suggestion
/// You can use a new type instead:
```

axum-extra/src/extract/cookie/signed.rs
```diff
@@ -87,6 +87,32 @@ use std::{convert::Infallible, fmt, marker::PhantomData};
 ///     .route("/me", get(me));
 /// # let app: Router<_> = app;
 /// ```
+/// If you have been using `Arc<AppState>` you cannot implement `FromRef<Arc<AppState>> for Key`.
+/// You can use a new type instead: `struct AppState(Arc<InnerState>);` and implement `Deref`.
```
davidpdrsn, line 91:
```suggestion
///
/// If you have been using `Arc<AppState>` you cannot implement `FromRef<Arc<AppState>> for Key`.
/// You can use a new type instead:
```

davidpdrsn (approved):
Thanks!

davidpdrsn:
@woile seems the doc tests failed. Do you wanna take a look?

PR #1573, opened 2022-11-25
docs: add axum-otel-metrics to axum ecosystem

ttys3:
a http metrics middleware for axum which 

uses the opentelemetry Rust sdk auto record http requests metrics

and export the metrics data to Prometheus at `/metrics` endpoint.

it include two major metrics: a request duration `histogram` metric and a request `counter` metric

https://github.com/ttys3/axum-otel-metrics


davidpdrsn:
Thanks!

PR #1580, opened 2022-11-26
Expand the docs for `Router::with_state`

davidpdrsn:
We've heard from a few users who have issues with getting the state types right, especially if they're returning a router (and thus have to name the type) and calling `with_state`.

This expands the docs for `Router::with_state` to hopefully provide examples that help but also provide some more intuition for how things work.

### DIFF
--- a/axum-core/src/extract/from_ref.rs
+++ b/axum-core/src/extract/from_ref.rs
@@ -5,10 +5,9 @@
 ///
 /// See [`State`] for more details on how library authors should use this trait.
 ///
-/// This trait can be derived using `#[derive(axum_macros::FromRef)]`.
+/// This trait can be derived using `#[derive(FromRef)]`.
 ///
 /// [`State`]: https://docs.rs/axum/0.6/axum/extract/struct.State.html
-/// [`#[derive(axum_macros::FromRef)]`]: https://docs.rs/axum-macros/latest/axum_macros/derive.FromRef.html
 // NOTE: This trait is defined in axum-core, even though it is mainly used with `State` which is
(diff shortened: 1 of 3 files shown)


axum-core/src/extract/from_ref.rs
```diff
@@ -5,10 +5,9 @@
 ///
 /// See [`State`] for more details on how library authors should use this trait.
 ///
-/// This trait can be derived using `#[derive(axum_macros::FromRef)]`.
+/// This trait can be derived using [`#[derive(FromRef)]`](https://docs.rs/axum-macros/latest/axum_macros/derive.FromRef.html).
```
davidpdrsn, line 8: For some reason this link didn't work with a reference at the bottom.

axum-core/src/extract/from_ref.rs
```diff
@@ -5,10 +5,9 @@
 ///
 /// See [`State`] for more details on how library authors should use this trait.
 ///
-/// This trait can be derived using `#[derive(axum_macros::FromRef)]`.
+/// This trait can be derived using `#[derive(FromRef)]`.
```
davidpdrsn, line 8: The link on this didn't work for some reason but I don't actually think we need it since the macro is exported at the same path as `FromRef` itself.

jplatte: Well, there were no square brackets around the code block.

jplatte (approved):

axum-core/src/extract/from_ref.rs
davidpdrsn: Even with I couldn't get it working 🤷

Issue #1325, opened 2022-08-25
Hyper 1.0 support


davidpdrsn:
Hyper is aiming to ship 1.0 this year and in the process making a lot of changes to the public API which will impact axum. We can discuss the changes here.

Off the top of my head the biggest changes will probably be
- Hyper dropping `tower::Service` support since tower-service wont be 1.0 in time.
- Moving `hyper::Server` into hyper-util. I don't think we should have a public dependency on hyper-util since breaking releases would cascade into axum.
- Removing `hyper::Body` (or possibly moving it into hyper-util). I think we should consider giving axum-core its own body type. Relevant to https://github.com/tokio-rs/axum/issues/1110

I haven't thought much about this yet but once 0.6.0 is out we start that work.

Relevant links
- [Hyper 1.0 project board](https://github.com/orgs/hyperium/projects/1/views/6)
- [Hyper's roadmap](https://github.com/hyperium/hyper/blob/master/docs/ROADMAP.md)

genusistimelord:
I guess the newer FromRequestParts and IntoResponseParts functions and State help Cover the loss of tower::Service. But this will heavily affect every single layer based library when and if you do update to the newest Hyper. So maybe we can improve and prepare all the Parts changes to make migration easier later on.  I am willing to switch my libraries over to using them so they can be used as Examples for other library makers to change theirs over too.

I'm guess hyper-util will be a faster change with more api breaking branch?

I think Axum should also have its own Body type. This would make it easier to Abstract the Body from different source and allow you more flexibility over them.

davidpdrsn:
I've started the work in https://github.com/tokio-rs/axum/pull/1882 but it's currently blocked by tower-http getting updated. That is blocked by figuring out how to implement the (de?)compression middleware.

davidpdrsn:
An update on where we're at with regards to hyper 1.0 support:

- Sean is planning to ship hyper 1.0 on November 15th 2023. This is super exciting! I can't guarantee that axum will be ready on the same day but that's my goal.
- The majority of the work to port axum is done. See https://github.com/tokio-rs/axum/pull/1882
- I've moved all issues, except this one, that were on the 0.7 milestone to a new 0.8 milestone. The focus for 0.7 is hyper 1.0 support. Other API changes will have to wait for the next major version.
- tower-http also needs to be updated. Most of that work is also done. See https://github.com/tower-rs/tower-http/pull/348. The only thing missing is compression and decompression middleware.

[...]

PR #1583, opened 2022-11-27
WIP - v0.7.0

davidpdrsn:
I wanna start slowly working towards 0.7.0. The biggest change for that will be [hyper 1.0 compatibility](https://github.com/tokio-rs/axum/issues/1325).

Other PRs for 0.7.0 should be merged into this one. Think we should keep `main` free of breaking changes for a bit.

Some things off the top of my head we need (might file separate issues on these):
- [ ] Remove re-exports of `hyper::Server`, `http_body::Full`, `http_body::Empty`, and possibly others. Those are being moved to `*-util` crates which will have breaking changes so we shouldn't have them in our public API.
- [ ] Add an `axum::serve(addr, make_service)` function. I think our recommended server should be the one in `hyper-util`. However I also think we should provide a very simple server function for "hello world" stuff. I think that should just be a single function without any configuration.
- [ ] Provide our own body type. `hyper::Body` is going away and it is using a new [`Incoming`](https://docs.rs/hyper/1.0.0-rc.1/hyper/body/struct.Incoming.html) which is much smaller in scope. I think this could lead to much improved ergonomics and perhaps we could even remove the generic `B` type everywhere (related to https://github.com/tokio-rs/axum/issues/1110)


jplatte:
`http_body` will also not be stable enough to depend on? I thought the concern was just with `hyper-util`.

davidpdrsn:
> `http_body` will also not be stable enough to depend on publically? I thought the concern was just with hyper-util.

`http_body` is going 1.0 as well with most things, except the `Body` trait, being moved to `http-body-util`. So similar treatment as hyper. So we can depend on the `http_body::Body` trait but not things like `http_body_util::Full`.

jplatte:
I'm a little confused about the first sentence of your reply, but the rest makes sense ^^

jplatte:
Also, are you working on upgrading axum to the rc's? If you could use some help, I'm interested in looking into (parts of) that.

davidpdrsn:
> Also, are you working on upgrading axum to the rc's? If you could use some help, I'm interested in looking into (parts of) that.

Not right now. I'm mostly doing small prep things. Also have a PR in the works for a simple server function that uses hyper 1.0-rc.1.

You're very welcome to help out! Its not quite clear to me if we even can update to hyper rc.1 as there might be some things missing. But attempting to update is probably the best way to figure that out 😅

Issue #1579, opened 2022-11-26
Strange `debug_assert!` and runtime panics


Pure-Peace:
### Version

`0.6.0-rc.5`, `0.6.0`

### Platform

Windows 10 x64

### Crates

axum

### Description

I am writing a service that selects a router based on the hostname and executes it using `router.oneshot(req)`

The code is similar to below:
```rust
Router::new().route("/*path", any(move |host: Host, req: Request<Body>| {
    let router = select_router_by_host(host);
    router.oneshot(req)
})
```

But I get the following error:
```bash
thread 'tokio-runtime-worker' panicked at 'assertion failed: matches!(extensions.remove :: < MatchedPath > (), None)', .cargo\registry\src\github.com-1ecc6299db9ec823\axum-0.6.0\src\extract\matched_path.rs:146:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

https://github.com/tokio-rs/axum/blob/8d62697c7224a4bb699913f1203dbf6839875f13/axum/src/extract/matched_path.rs#L142-L146

The problem occurs on line 146
https://github.com/tokio-rs/axum/blob/8d62697c7224a4bb699913f1203dbf6839875f13/axum/src/extract/matched_path.rs#L146

This `debug_assert!` confuses me, why does it exist? Or what is it used for?

Currently I avoid triggering this assertion while debugging by adding `req.extensions_mut().remove::<axum::extract::MatchedPath>();`

```rust
Router::new().route("/*path", any(move |host: Host, mut req: Request<Body>| {
    // Temporary solution
    let matched = req.extensions_mut().remove::<axum::extract::MatchedPath>();
    let router = select_router_by_host(host);
    router.oneshot(req)
})
```

And `matched` will get: `Some(MatchedPath("/*path"))`

I wrote a simple code sample and it didn't throw this error, so this exception is very strange.

Pure-Peace:
Below is a code sample, but this assertion is not triggered:

```rust
use axum::{
    body::{Body, BoxBody},
    extract::Host,
    http::{Request, StatusCode},
    response::{IntoResponse, Response},
    routing::{any, get},
    Router,
};
use std::net::SocketAddr;
use tower::ServiceExt;

#[tokio::main]
async fn main() {
    let app = Router::new()
        .route("/", any(|| async { "hello, world" }))
        .route(
            "/*path",
            any(move |host: Host, req: Request<Body>| any_path(host, req)),
        );

[...]

PR #1585, opened 2022-11-27
Remove wrong `debug_assert` in `MatchedPath`

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1579

### DIFF
--- a/axum/src/routing/tests/mod.rs
+++ b/axum/src/routing/tests/mod.rs
@@ -1,7 +1,7 @@
 use crate::{
     body::{Bytes, Empty},
     error_handling::HandleErrorLayer,
-    extract::{self, DefaultBodyLimit, FromRef, Path, State},
+    extract::{self, DefaultBodyLimit, FromRef, MatchedPath, Path, State},
     handler::{Handler, HandlerWithoutStateExt},
     response::IntoResponse,
     routing::{delete, get, get_service, on, on_service, patch, patch_service, post, MethodFilter},
@@ -19,9 +19,13 @@ use std::{
     task::{Context, Poll},
     time::Duration,
 };
-use tower::{service_fn, timeout::TimeoutLayer, util::MapResponseLayer, ServiceBuilder};
+use tower::{
+    service_fn, timeout::TimeoutLayer, util::MapResponseLayer, ServiceBuilder, ServiceExt,
+};
 use tower_http::{auth::RequireAuthorizationLayer, limit::RequestBodyLimitLayer};
 use tower_service::Service;
+
+use super::any;
 
 mod fallback;
 mod get_to_head;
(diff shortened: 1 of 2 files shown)


axum/src/routing/tests/mod.rs
```diff
@@ -799,3 +803,23 @@ async fn layer_response_into_response() {
...
+    let app = Router::new().route(
+        "/*path",
+        any(|req| Router::new().nest("/", inner).oneshot(req)),
+    );
+
+    let client = TestClient::new(app);
+
+    let res = client.get("/foobar").send().await;
+    assert_eq!(res.status(), StatusCode::OK);
+    assert_eq!(res.text().await, "/*path/foobar");
```
davidpdrsn, line 824: I'm not really sure what the matched path logically should be. Feels weird saying it is `/*path/foobar` since the request is just `/foobar`, so where does `/*path` come from?

However it kinda matches the way the router is setup 🤔

davidpdrsn: cc @Pure-Peace

davidpdrsn: In general I think a more correct router setup would be

```rust
async fn handler(req: Request<Body>) -> crate::response::Response {
    let inner = Router::new().route(
        "/foobar",
        get(|matched_path: MatchedPath| async move { matched_path.as_str().to_owned() }),
    );

    Router::new().nest("/", inner).oneshot(req).await.unwrap()
}

// run it with
handler.into_make_service();
```

Pure-Peace: generally like this

I'm using axum to make some framework like

```rust
pub struct App {
    pub cfg: Arc<GatewayConfig>,
}

impl Application for App {
    fn frame_cfg(&self) -> &ApiFrameConfig {
        &self.cfg.frame_cfg
    }

    fn router(&self) -> Router {
        Router::new()
            .route("/", get(peace_api::responder::app_root))
            .nest("/bancho", bancho::routers::bancho_client_routes())
    }

    fn apidocs(&self) -> utoipa::openapi::OpenApi {
        GatewayApiDocs::openapi()
    }

    fn match_hostname(
        &self,
        Host(hostname): Host,
        req: &Request<Body>,
    ) -> Option<Router> {
        match hostname {
            n if self.cfg.bancho_hostname.contains(&n) => {
                Some(bancho::routers::bancho_client_routes())
            },
            _ => None,
        }
    }
}

#[tokio::main]
pub async fn main() {
    let app = App::new(GatewayConfig::get());
    http::serve(app).await;
}
```

In general a reverse proxy can be used to direct the request to a certain path, which may be the router of a certain service

I would like to allow users to enable both `general router` and `hostname based router`. This may need to use `/*path`

possible to reuse some routers

davidpdrsn:
Superseded by https://github.com/tokio-rs/axum/pull/1935

PR #1587, opened 2022-11-27
Start porting to hyper 1.0.0-rc.1

jplatte:


axum-core/src/body.rs
```diff
@@ -37,58 +38,6 @@ where
...
-// all copies or substantial portions of the Software.
-
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-// THE SOFTWARE.
-pub(crate) async fn to_bytes<T>(body: T) -> Result<Bytes, T::Error>
```
Oliboy50, line 60: 👋 What's the replacement for this function?

(I've been hoping axum would expose this function some day, so we (axum users) won't have to depend on hyper anymore just for it... That's why I'm interested)

davidpdrsn: Its been added to http-body-util as https://docs.rs/http-body-util/0.1.0-rc.2/http_body_util/trait.BodyExt.html#method.collect

davidpdrsn:
I think I'll close for now since quite a few things have changed in the v0.7.0 branch and in hyper's release candidates. There are some smaller steps I think we can take before trying to port axum wholesale.

PR #1589, opened 2022-11-29
Fix wrong `From` impl for `Resource`

davidpdrsn:
Really not sure how this got past the tests.

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning].
 
 # Unreleased
 
-- None.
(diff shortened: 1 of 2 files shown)


jplatte (approved):

jplatte:
It wasn't wrong, just needlessly restrictive 😉 

Note that this is technically a breaking change, as people could have relied on inference figuring out `S = ()` from a `.into()` call.

davidpdrsn:
> Note that this is technically a breaking change, as people could have relied on inference figuring out `S = ()` from a `.into()` call.

Yeah true but I think we should categorize it as acceptable breakage because it was hard/impossible to actually use previously. I ran into it with https://github.com/EmbarkStudios/server-framework/pull/47

PR #1596, opened 2022-11-29
Refactor initializing tracing-subscriber

tottoto:
Refactoring examples.

### DIFF
--- a/examples/chat/src/main.rs
+++ b/examples/chat/src/main.rs
@@ -22,7 +22,6 @@ use std::{
     sync::{Arc, Mutex},
 };
 use tokio::sync::broadcast;
-use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
 // Our shared state
 struct AppState {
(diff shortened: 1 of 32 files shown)


davidpdrsn (commented):
Maybe its just me but I actually prefer using `registry()`

davidpdrsn:
@tottoto isn't that what were doing before? 🤔

tottoto:
I think we could use `EnvFilter::try_from_default_env()` in stead of reading `RUST_LOG` maually.

tottoto:
Like this https://github.com/tokio-rs/axum/pull/1596/commits/7fdd414f6e70b7f4d140622ec444131f3b520778.

This change uses [`Envfilter::try_from_default_env()`](https://docs.rs/tracing-subscriber/0.3.16/tracing_subscriber/filter/struct.EnvFilter.html#method.try_from_default_env) reading `RUST_LOG` environment variable.

davidpdrsn:
Ah right yes! Thats a bit nicer indeed :(

tottoto:
Done.

davidpdrsn (approved):
Thanks!

Really loving these refactoring PRs you've been doing ❤️

PR #1597, opened 2022-11-29
Fix clippy warning

tottoto:
Follows to the solution https://github.com/tokio-rs/prost/issues/661#issuecomment-1156606409.


davidpdrsn (approved):
Thanks!

PR #1598, opened 2022-11-30
Update to tokio-tungstenite-0.18.0 and use sha1

tottoto:
Updates to tokio-tungstenite 0.18.0.

This updates tungstenite version to 0.18.0 as dependency.

https://github.com/snapview/tokio-tungstenite/blob/v0.18.0/CHANGELOG.md#0180

This version of tungstenite uses sha1 instead of sha-1 because sha1 is transferred to https://github.com/RustCrypto (https://github.com/RustCrypto/hashes/pull/350) who owned sha-1 so it may be good to use sha1 in this change to avoid duplication.


davidpdrsn (approved):
Thanks!

PR #1613, opened 2022-12-02
Add missing import of `http` in docs example

cemremengu:
Wouldn't compile without importing `self` for `http`


davidpdrsn:
Wait, how why? We run doc tests on CI so why did this pass before?

davidpdrsn:
Ah of course its because axum depends in http itself, nvm :P

davidpdrsn (approved):

PR #1616, opened 2022-12-03
Use let-else in middleware docs example

cemremengu:
Use let-else in middleware docs example to simplify


cemremengu:
sorry never mind MSRV is `1.60`

PR #1621, opened 2022-12-06
refactor(json): json content type checker

henriklundgren:
It started with me thinking I would minimize the code and I guess arguable write it more rusty.

It turns out it become a bit more code and I ... no string matching, no flag behavior and no branching.
I like it.

In the end it is about twice as fast,
measuring using basic `Instant::now`.

### DIFF
--- a/axum/src/json.rs
+++ b/axum/src/json.rs
@@ -111,7 +111,7 @@ where
     type Rejection = JsonRejection;
 
     async fn from_request(req: Request<B>, state: &S) -> Result<Self, Self::Rejection> {
-        if json_content_type(req.headers()) {
+        if json_content_type(req.headers()).is_some() {
             let bytes = Bytes::from_request(req, state).await?;
             let deserializer = &mut serde_json::Deserializer::from_slice(&bytes);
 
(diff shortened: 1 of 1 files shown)


axum/src/json.rs
```diff
@@ -144,29 +144,42 @@ where
...
+
+        [
+            subtype,
+            suffix,
+        ]
+        .iter()
+        .any(|x| x.is_some())
+        .then_some(type_)
+        .flatten()
+    })
```
davidpdrsn, line 182: > In the end it is about twice as fast,

Why?

I think readability is also worse now since it doesn't use `?`. All the combinator functions are hard to follow. Also not sure we need to return an `Option<_>` since we're not using it.

henriklundgren: Hmm, ok, valid, but, "?" was not used before so I dont know what that means (the return value was a boolean).

Yeah, about readability, its a matter of taste I guess, I like the more functional approach.

You quoted the "twice as fast",
well I ran the `Instant::now`,  with about 20 random headers and its about 50% faster. (dunno if that is a good measurement).

davidpdrsn: > Hmm, ok, valid, but, "?" was not used before so I dont know what that means (the return value was a boolean).

I know we didn't use `?` before, but the way you wrote it is less readable IMO without `?`.

> You quoted the "twice as fast", well I ran the Instant::now, with about 20 random headers and its about 50% faster. (dunno if that is a good measurement).

I'm curious if you did any digging into why its faster. That isn't clear to me and I'm curious to know.

henriklundgren: > I'm curious if you did any digging into why its faster. That isn't clear to me and I'm curious to know.

No clue, isn't this like compiler things. Maybe its just something local (debug build) and in production its optimized. Or maybe its my naive perf test that isn't done right.

---

I guess one option of this change is that the `fn from_request` could go without the branching using `?` with something like,

`json_content_type(req.headers()).ok_or(MissingJsonContentType)?;`

ttys3:
> In the end it is about twice as fast, measuring using basic `Instant::now`.

any benchmark about this ? and compare each version ?

I this not see the point of this PR.

IMHO, also I do not see the code readability got improved, seems your version is more complicated than the original.

axum/src/json.rs
NateLing: I checked the asm output, The asm code generated by this pr is much less than the original, is really a strange problem.

henriklundgren:
I did a perf with `criterion`.

```
content_type/using branching/1
                        time:   [77.511 ns 80.400 ns 83.865 ns]
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) high mild
  5 (5.00%) high severe
content_type/using combinators/1
                        time:   [59.092 ns 59.323 ns 59.585 ns]
Found 9 outliers among 100 measurements (9.00%)
  5 (5.00%) high mild
  4 (4.00%) high severe
content_type/using branching/2
                        time:   [87.546 ns 88.881 ns 90.101 ns]
content_type/using combinators/2
                        time:   [72.148 ns 73.404 ns 74.857 ns]
Found 6 outliers among 100 measurements (6.00%)
  6 (6.00%) high mild
```

PR #1625, opened 2022-12-07
Multipart constraints POC

### LINKED ISSUE #1623: Allow setting multer::Multipart constraints
### Motivation

I'm running into a `length limit exceeded` error from multer when trying to read a file coming in from a Multipart form. Multer supports [`Constraints`](https://docs.rs/multer/latest/multer/struct.Constraints.html) to avoid this, but axum's multipart extractor does not allow forwarding these.

### Proposal

I'm not super sure how this could be implemented, since the constraints can't be passed through the parameter typing. Either there might need to be some documented recommended way of parsing the request manually inside the handler, or maybe the multipart extractor needs to be a bit fancier to allow passing constraints.

### Alternatives

For now, I'll "vendor" the Multipart extractor, since it seems simple enough, to add that functionality, but it seems like something axum's native extractor should support.

jplatte:
## Motivation

Could fix #1623.

## Solution

Add a (defaulted) const generic parameter to `Multipart`.

### DIFF
--- a/axum/src/extract/multipart.rs
+++ b/axum/src/extract/multipart.rs
@@ -50,12 +50,12 @@ use std::{
 /// ```
 #[cfg_attr(docsrs, doc(cfg(feature = "multipart")))]
 #[derive(Debug)]
-pub struct Multipart {
+pub struct Multipart<const C: Constraints = { Constraints::DEFAULT }> {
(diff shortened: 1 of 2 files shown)


davidpdrsn:
Think I'll just close this for now. When the necessary compiler features stabilize we can revisit this.

PR #1637, opened 2022-12-13
Updated websocket example to improve clarity and to be more distinct from chat example

alexpyattaev:
## Motivation

The web sockets example was not looking nice, it was not user-friendly at all (took a while to figure out what it was even supposed to be able to do). Furthermore, it was not actually showcasing anything about web sockets that the chat example did not already demonstrate better.  Either it had to go, or it had to be made more distinct, so I took the time to make it more "involved". 

## Solution

-  all parts of the example have received additional comments
- the sample webpage has received hints that the user should open browser's console to see messages (it was just blank)
- added a Python script that demonstrates some more tricky features such as ping/pong handling, and allows to quickly connect with multiple "users" to the test server
- added the logic to push unsolicited messages to the client form the server (which is not obvious at all though easy to actually do)
- added comments explaining that one should not try to implement pong responder (as axum already has one builtin, and thus even though you can see the ping messages, you should not actually respond to them)


davidpdrsn:
Thanks! I skimmed the changes and they generally look good. Will do a more thorough review later. However we cannot merge Python code. I don't know Python so don't wanna take that on. Can you write the client in Rust instead?

alexpyattaev:
All right, now there is a proper Rust client, which is properly illustrating the concurrency features of axum websockets.
Code-wise it is actually 90% identical to the server, which is pretty cute if someone wants to eventually have both sides of the socket in Rust.
Feel free to kill the Python client.py if you feel like it deserves it =)

examples/websockets/client.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python3
```
davidpdrsn, line 1: Yep the python should be deleted.

examples/websockets/assets/script.js
```diff
@@ -1,4 +1,4 @@
-const socket = new WebSocket('ws://localhost:3000/ws');
+socket = new WebSocket('ws://localhost:3000/ws');
```
davidpdrsn, line 1: Why did you remove the `const`?

examples/websockets/assets/script.js
```diff
@@ -7,3 +7,23 @@ socket.addEventListener('open', function (event) {
 socket.addEventListener('message', function (event) {
     console.log('Message from server ', event.data);
 });
+
+window.addEventListener("load", (event) => {
+   console.log("page is fully loaded");
+
+});
```
davidpdrsn, line 14: What is this for?

examples/websockets/src/main.rs
```diff
@@ -59,58 +76,169 @@ async fn main() {
     let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
     tracing::debug!("listening on {}", addr);
     axum::Server::bind(&addr)
-        .serve(app.into_make_service())
+        .serve(app.into_make_service_with_connect_info::<SocketAddr>())
         .await
         .unwrap();
 }
 
+/// A formally required handler for the request.
```
davidpdrsn, line 84: What does "formally" mean here?

examples/websockets/src/main.rs
```diff
@@ -59,58 +76,169 @@ async fn main() {
...
-                Message::Close(_) => {
-                    println!("client disconnected");
-                    return;
-                }
+            if process_message(msg, who) {
+                return;
             }
         } else {
-            println!("client disconnected");
+            println!("{} abruptly disconnected", who);
```
davidpdrsn, line 121: Why "abruptly"? Isn't `client disconnected` or `{who} disconnected` fine?

examples/websockets/src/main.rs
```diff
@@ -59,58 +76,169 @@ async fn main() {
...
             return;
         }
     }
 
-    loop {
+    // we can also send messages to client by simply sleeping,
+    // since each client gets individual statemachine. Thus, waiting for this
+    // client to finish getting his greetings does not prevent other clients form
+    // connecting to server.
```
davidpdrsn, line 129: I think this should be reworded a bit, you're not sending messages by sleeping.

Maybe it should say something about not worrying about blocking messages being sent to other connected clients.

examples/websockets/src/main.rs
```diff
@@ -59,58 +76,169 @@ async fn main() {
...
+        rv_b = (&mut recv_task) =>{
+            match rv_b {
+                Ok(b)=>println!("Received {} messages", b),
+                Err(b)=>println!("Error receiving messages {:?}", b)
+            }
+            send_task.abort();
+        }
+    }
+
+    // returning from the handler destroys the websocket context
```
davidpdrsn, line 208:
```suggestion
    // returning from the handler closes the websocket connection
```

examples/websockets/src/main.rs
```diff
@@ -59,58 +76,169 @@ async fn main() {
...
+    println!("Websocket context {} destroyed", who);
+}
+
+/// helper to print contents of messages to stdout. special treatment for Close
+fn process_message(msg: Message, who: SocketAddr) -> bool {
```
davidpdrsn, line 213: Could probably use https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html instead rather than a `bool`.

examples/websockets/src/main.rs
```diff
@@ -59,58 +76,169 @@ async fn main() {
...
+            }
+            return true;
+        }
+
+        Message::Pong(v) => {
+            println!(">>> {} sent pong with {:?}", who, v);
+        }
+        // You should never need to manually handle these, as tungstenite websocket library
+        // will do so for you automagically by replying with Pong and copying the v according to
+        // spec. But if you need the contents of the pings you can see them here.
```
davidpdrsn, line 238: Either we should explain what tungstenite is (the websocket library axum uses internally) or say that axum handles pings.

examples/websockets/src/main.rs
```diff
@@ -59,58 +76,169 @@ async fn main() {
...
+            recv_task.abort();
+        },
+        rv_b = (&mut recv_task) =>{
+            match rv_b {
+                Ok(b)=>println!("Received {} messages", b),
+                Err(b)=>println!("Error receiving messages {:?}", b)
+            }
+            send_task.abort();
+        }
+    }
```
davidpdrsn, line 206: So odd formatting going on here. Sadly rustfmt doesn't like macros 😕

examples/websockets/src/main.rs
```diff
@@ -59,58 +76,169 @@ async fn main() {
...
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+    }
+
+    // By splitting socket we can send and receive at the same time. In this example we will send
+    // unsolicited messages to client based on some sort of server's internal event (i.e .timer).
+    let (mut sender, mut receiver) = socket.split();
+
+    // Spawn a task that will push 100 messages to the client (does not matter what client does)
+    let mut send_task = tokio::spawn(async move {
+        let mut i = 20;
```
davidpdrsn, line 148: I think this should be `for n in 0..20 { ... }` to do something 20 times.

examples/websockets/src/main.rs
```diff
@@ -59,58 +76,169 @@ async fn main() {
...
+            }
+
+            tokio::time::sleep(std::time::Duration::from_millis(300)).await;
+            i -= 1;
+        }
+        if i == 0 {
+            println!("Sending close to {}...", who);
+            if let Err(e) = sender
+                .send(Message::Close(Some(CloseFrame {
+                    code: 1000,
```
davidpdrsn, line 166:
```suggestion
                    code: axum::extract::ws::close_code::NORMAL,
```

examples/websockets/src/main.rs
```diff
@@ -59,58 +76,169 @@ async fn main() {
...
+    let mut send_task = tokio::spawn(async move {
+        let mut i = 20;
+        while i > 0 {
+            // In any websocket error, break loop.
+            if sender
+                .send(Message::Text(format!("{} messages left...", i)))
+                .await
+                .is_err()
+            {
+                break;
```
davidpdrsn, line 156:
```suggestion
                return;
```

That way you don't need to check `i` after the loop because you'll return from the task.

examples/websockets/src/main.rs
```diff
@@ -1,9 +1,15 @@
...
-//! Run with
+//! Run the server with
 //!
 //! ```not_rust
-//! cd examples && cargo run -p example-websockets
+//! cargo run -p example-websockets
+//! firefox http://localhost:3000
+//! ```
+//!
+//! Alternatively you can run rust client with
```
davidpdrsn, line 10:
```suggestion
//! Alternatively you can run the rust client with
```

examples/websockets/src/client.rs
```diff
@@ -0,0 +1,165 @@
+//! Based on tokio-tungstenite example websocket client, but with multiple
```
davidpdrsn, line 1: Hm I'm kinda torn on having a client example here. Like what is it trying to demonstrate? How to write a websocket client in rust? Maybe just linking to [tungstenite's examples](https://github.com/snapview/tungstenite-rs/tree/master/examples) is enough for that. I generally prefer axum's examples to focus on axum related things, which I'd argue this isn't.

examples/websockets/src/client.rs
```diff
@@ -0,0 +1,165 @@
...
+//stream splitting
+use futures_util::{SinkExt, StreamExt};
+
+#[derive(Parser)]
+struct Args {
+    #[arg(short, long, default_value_t = String::from("ws://127.0.0.1:3000/ws"))]
+    url: String,
+    #[arg(short, long, default_value_t = 2)]
+    number_of_clients: usize,
+}
```
davidpdrsn, line 33: I don't think we need all this argument parsing stuff, its not important to what the example is trying to show.

Could just hardcode the url and have `let number_of_clients = 10;` variable.

examples/websockets/Cargo.toml
```diff
@@ -8,6 +8,21 @@ publish = false
...
 tokio = { version = "1.0", features = ["full"] }
+tower = { version = "0.4", features = ["util"] }
 tower-http = { version = "0.3.0", features = ["fs", "trace"] }
 tracing = "0.1"
+futures = "0.3"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+futures-util = { version = "0.3", default-features = false, features = ["sink", "std"] }
+tokio-tungstenite = "0.18.0"
+url = "2.0.0"
+futures-channel = "0.3"
```
davidpdrsn, line 19: Can probably just use the channels in tokio.

```suggestion
```

examples/websockets/src/client.rs
```diff
@@ -0,0 +1,165 @@
...
+
+#[tokio::main]
+async fn main() {
+    let args = Args::parse();
+    let url = Url::parse(&args.url).expect("Invalid URL supplied!");
```
davidpdrsn, line 38: If we hardcode the url then we can drop the dependency on `url` since [`IntoClientRequest`](tungstenite::client::IntoClientRequest) is also implemented for `&str`.

examples/websockets/assets/script.js
alexpyattaev: JS always conflicts with common sens in my head, calling a .close() on a const object just felt wrong. But you are correct, it should be const there.

examples/websockets/assets/script.js
alexpyattaev: Well following your logic that examples must be about axum only this serves no purpose at the moment. I'll trim it.

examples/websockets/client.py
alexpyattaev: got it, will do.

examples/websockets/assets/script.js
davidpdrsn: > examples must be about axum only

I didn't say that. I said they should be related to axum. For example a while ago someone submitted a PR with an sqlx example that included a whole setup for migrations. But migrations aren't related to axum so they don't need to be in the example.

I think there is value in being able to open the browser and see data coming in from the websocket. Having a 9 line JS file for that is totally fine.

examples/websockets/src/client.rs
alexpyattaev: TL;DR version:
- tungstenite example is very "shallow" and only shows how *one* client would interact with the server, which is already demonstrated well enough by the browser example.
- I wanted specifically do demonstrate how one could lay groundwork for more complex/non-standard state machines with multiple concurrent flows, rather than the beaten to death chat example. It would be tough to demonstrate with a silly client.

The entire idea I had with conceptual improvements to the example  was to really show good reasons one would want to use Axum implementation in particular if compared to, for example, actix. For me personally the key reason is specifically how one can have independent send and receive logic for the websockets, and how all of that can be 100% async-aware, and with minimal boilerplate. For contrast, in Actix the expectation is that one either

-  handles messages as "copy types", which means they are fully buffered before feeding them into a handler using their Actor model, or, alternatively, 
- or uses their low-level API, which is "fun", and you'd have to handle things like ping/pong yourself which is ugly. [https://github.com/actix/examples/blob/master/websockets/chat-actorless/src/handler.rs#L54](as they show in their own examples)

examples/websockets/src/client.rs
alexpyattaev: will do.

examples/websockets/src/client.rs
alexpyattaev: good idea. got carried away.

examples/websockets/src/main.rs
alexpyattaev: Well every websocket open is an HTTP request, so before we can get to websocket stuff proper we need to have a handler for the original HTTP request, which that function is. Which is why it is the last place where HTTP headers can be seen, and after that we are essentially talking in another protocol. In this sense it is a "formality" if all you wanted was websockets. I can try to rephrase this.

examples/websockets/src/client.rs
davidpdrsn: Alright I guess that fair. I'm okay with keeping it but I still my earlier comments about trimming it down (removing argument parsing etc) make sense. wdyt?

examples/websockets/src/main.rs
alexpyattaev: I think that in order to get into that branch the client would need to have disconnected in some *very strange way*. So by no means it would be a "normal business as usual event". However, in all my attempts I was unable to ever get into that error branch, no matter how I tortured the client or network (which included pulling network interface and killing client), so I do not even know how one would get that to occur.

examples/websockets/src/main.rs
alexpyattaev: I've noticed a few odd things about it too. At the very least it is consistently odd.

examples/websockets/src/main.rs
davidpdrsn: Ah I see, makes sense. Maybe something like "handle upgrading the connection to use websockets" or something?

examples/websockets/src/main.rs
davidpdrsn: Alright I see. We can just keep "abruptly" I guess then.

examples/websockets/assets/script.js
davidpdrsn: > JS always conflicts with common sens in my head

yeah same 😅

examples/websockets/src/main.rs
davidpdrsn: This feels in the spirit for what rustfmt would do

```rust
    tokio::select! {
        rv_a = (&mut send_task) => {
            match rv_a {
                Ok(a) => println!("{} messages remaining to send", a),
                Err(a) => println!("Error sending messages {:?}", a)
            }
            recv_task.abort();
        },
        rv_b = (&mut recv_task) => {
            match rv_b {
                Ok(b) => println!("Received {} messages", b),
                Err(b) => println!("Error receiving messages {:?}", b)
            }
            send_task.abort();
        }
    }
```

examples/websockets/assets/script.js
alexpyattaev: Exactly, and that message in particular was pure bloat, just on a smaller scale.

examples/websockets/src/client.rs
alexpyattaev: totally agree, will address them once I have some spare time.

examples/websockets/src/main.rs
avdb13: "switching from HTTP to websocket protocol" wouldn't be bad either.

examples/websockets/src/main.rs
alexpyattaev: Yes, figured a way to rewrite that without while. Looks better now. Will add that to the bigger commit.

examples/websockets/src/main.rs
alexpyattaev: Yep, that is way less ugly. Thanks!

examples/websockets/src/main.rs
alexpyattaev: Maybe a bit verbose, but I think it is correct now:
```
/// The handler for the HTTP request (this gets called when the HTTP GET lands at the start
/// of websocket negotiation. After this completes, the actual switching from HTTP to
/// websocket protocol will occur.
/// This is the last point where we can extract TCP/IP metadata such as IP address of the client
/// as well as things from HTTP headers such as user-agent of the browser etc.
```

alexpyattaev:
Ok the fat trimming is done, I hope I have not missed anything.

examples/websockets/src/main.rs
```diff
@@ -59,58 +76,173 @@ async fn main() {
...
+            println!(">>> {} sent pong with {:?}", who, v);
+        }
+        // You should never need to manually handle Message::Ping, as axum's websocket library
+        // will do so for you automagically by replying with Pong and copying the v according to
+        // spec. But if you need the contents of the pings you can see them here.
+        Message::Ping(v) => {
+            println!(">>> {} sent ping with {:?}", who, v);
+        }
     }
+    return ControlFlow::Continue(());
```
davidpdrsn, line 247:
```suggestion
    ControlFlow::Continue(())
```

davidpdrsn (commented):
Great! Almost there. Just a couple of minor nitpicks here and there.

examples/websockets/src/client.rs
```diff
@@ -0,0 +1,162 @@
...
+async fn main() {
+    let start_time = Instant::now();
+    //spawn several clients that will concurrently talk to the server
+    let clients: Vec<JoinHandle<_>> = (0..N_CLIENTS)
+        .into_iter()
+        .map(|cli| tokio::spawn(async move { spawn_client(cli).await }))
+        .collect();
+
+    //wait for all our clients to exit
+    futures::future::join_all(clients).await;
```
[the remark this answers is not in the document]
alexpyattaev: Thanks! The move was actually a "leftover" from the previous iteration where url was passed in, and move was actually necessary there. As for unordered futures I did not know that this option exists, its a nice optimization.

alexpyattaev:
Thanks for the tips! I've re-tested everything (there was a missing import and one unneeded import) and the last commit should contain a fully functional example with all your suggestions incorporated.

alexpyattaev:
Bump =)

davidpdrsn:
No need for bumps. We'll get to this when we get to it. I'm currently on Christmas holiday playing video games.

davidpdrsn (approved):
Thanks for working on this!

FrankReh:
I'm sure many people are going to benefit from this example, including the command-line client. Big thank you to both!

I was wondering how to run the example and then I saw the comments in main.rs (of course). Small nit, at least for me to get them running:
```diff
diff --git a/examples/websockets/src/main.rs b/examples/websockets/src/main.rs
index 26a31a2..6d598b1 100644
--- a/examples/websockets/src/main.rs
+++ b/examples/websockets/src/main.rs
@@ -3,13 +3,13 @@
 //! Run the server with
 //!
 //! ```not_rust
-//! cargo run -p example-websockets
+//! cargo run -p example-websockets --bin example-websockets
 //! firefox http://localhost:3000
 //! ```
 //!
 //! Alternatively you can run the rust client with
 //! ```not_rust
-//! cargo run -p example-client
+//! cargo run -p example-websockets --bin example-client
 //! ```
```

davidpdrsn:
@FrankReh good catch! Wanna submit a PR?

PR #1646, opened 2022-12-16
Fix typo

dkbay:
## Motivation
Saw a spelling mistake decided to make a PR.

## Solution
Fix the spelling mistake.

Added `'t`


davidpdrsn (approved):
Thanks!

jplatte:
I don't think it was accidentally saying *can* before, it's just that there's different conventions around what to write for the expect string. But maybe this is less confusing 🤷🏼

PR #1649, opened 2022-12-16
docs/middleware: Add missing `Clone` derive on `MyLayer`

Turbo87:
## Motivation

Without this `Clone` implementation the compiler was showing errors on the `router.layer(middleware)` call. 

## Solution

Adding the `Clone` derive fixed the compiler error for me.

I've also removed an unnecessary `mut` keyword, that the compiler was warning about.

### DIFF
--- a/axum/src/docs/middleware.md
+++ b/axum/src/docs/middleware.md
@@ -229,6 +229,7 @@ use futures::future::BoxFuture;
 use tower::{Service, Layer};
 use std::task::{Context, Poll};
 
+#[derive(Clone)]
 struct MyLayer;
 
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
Thanks!

Turbo87:
looks like CI is unhappy, but that seems to be unrelated to my change 😅

davidpdrsn:
Yep. That's not your fault 😛

Turbo87:
@davidpdrsn anything I can do to make CI green and this PR mergeable?

also, happy new year! :)

davidpdrsn:
Merging the latest main should fix CI.

PR #1655, opened 2022-12-21
Support streaming `Multipart` fields without buffering

davidpdrsn:
This adds `Multipart::into_stream` and `Field::into_stream` which makes it possible to forward the contents elsewhere without buffering the fields all at once.

That was hard previously because `Field` borrows the `Multipart`.

### DIFF
--- a/axum-core/CHANGELOG.md
+++ b/axum-core/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- **added** Implement IntoResponse for &'static [u8; N] and [u8; N] ([#1690])
+
+[#1690]: https://github.com/tokio-rs/axum/pull/1690
 
 # 0.3.1 (9. January, 2023)
 
@@ -124,7 +126,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **added:** Add `response::ErrorResponse` and `response::Result` for
   `IntoResponse`-based error handling ([#921])
 
-[#921]: https://github.com/tokio-rs/axum/pull/921 
+[#921]: https://github.com/tokio-rs/axum/pull/921
 
 # 0.2.2 (19. April, 2022)
 
--- a/axum-core/src/response/into_response.rs
+++ b/axum-core/src/response/into_response.rs
@@ -348,6 +348,18 @@ impl IntoResponse for &'static [u8] {
     }
 }
 
+impl<const N: usize> IntoResponse for &'static [u8; N] {
+    fn into_response(self) -> Response {
+        self.as_slice().into_response()
+    }
+}
+
+impl<const N: usize> IntoResponse for [u8; N] {
+    fn into_response(self) -> Response {
+        self.to_vec().into_response()
+    }
+}
+
 impl IntoResponse for Vec<u8> {
     fn into_response(self) -> Response {
         Cow::<'static, [u8]>::Owned(self).into_response()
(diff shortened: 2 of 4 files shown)


jplatte (commented):
Can't `Multipart` implement `Stream<Item = Result<Field<'static>, MultipartError>>` directly? AFAICT, the `into_stream` you're adding here would then be the same as `TryStreamExt::map_ok`.

davidpdrsn:
Hm I don't think. Then you could have two `Field`s referencing the same `Multipart` live at the same time

```rust
let field_1 = multipart.next().await??;
let field_2 = multipart.next().await??;

field_2.bytes().await;
field_1.bytes().await;
```

This way you can get a stream that maps each field but you cannot just return the fields themselves. Specifically `multipart.into_stream(|field| field)` doesn't compile.

jplatte:
> Specifically `multipart.into_stream(|field| field)` doesn't compile.

I think that is because you made the bound `F: FnMut(Field<'_>) -> T + Clone + Send + 'static` when it could be `F: FnMut(Field<'static>) -> T + Clone + Send + 'static` (since we use `multer::Multipart<'static>`.

davidpdrsn:
I'm not sure that would work regardless. While it make a stream that is
`'static` it still allows for overlapping field references which multer doesn't
allow.

At least if you implement it like this:

```rust
#[derive(Debug)]
pub struct Field<'a> {
    inner: multer::Field<'static>,
    // make this an option because we cannot produce a `&'static mut Multipart`
    _multipart: Option<&'a mut Multipart>,
}

impl Multipart {
    // a dedicated method like this using `impl Stream` is easier lifetime wise
    // and still demonstrates the issue
    pub fn into_stream(
        self,
    ) -> impl Stream<Item = Result<Field<'static>, MultipartError>> + Send + 'static {
        futures_util::stream::try_unfold(self, move |mut multipart| async move {
            let field = match multipart
                .inner
                .next_field()
                .await
                .map_err(MultipartError::from_multer)?
            {
                Some(field) => field,
                None => return Ok(None),
            };

            Ok(Some((
                Field {
                    inner: field,
                    _multipart: None,
                },
                multipart,
            )))
        })
    }
}
```

This compiles but panics at runtime:

```rust
async fn handler(multipart: Multipart) {
    let stream = multipart.into_stream();
    futures_util::pin_mut!(stream);

    let a = stream.next().await.unwrap().unwrap();
    let b = stream.next().await.unwrap().unwrap();

    b.text().await.unwrap();
    a.text().await.unwrap();
}
```

Message is

```
thread 'extract::multipart::tests::stream_fields_directly' panicked at 'called
`Result::unwrap()` on an `Err` value: MultipartError { source: failed to lock
multipart state }', axum/src/extract/multipart.rs:404:50
```

jplatte:
Aha, just found the [docs](https://docs.rs/multer/latest/multer/struct.Multipart.html#field-exclusivity) for this, there is some runtime tracking of lifetimes. I was just looking at the lifetime. So basically, using `Field<'_>` instead of `Field<'static>` is a small hack that enforces the consumption of the `Field` before the next one is obtained, to make this runtime requirement a compiletime requirement.

jplatte (approved):

davidpdrsn:
I've realized that this API can still be misused as you can do this:

```rust
async fn handle(multipart: Multipart) -> impl IntoResponse {
    let stream = multipart.into_stream(|field| field.into_stream());
    futures_util::pin_mut!(stream);

    let mut a = stream.next().await.unwrap().unwrap();
    // 💥💀
    let mut b = stream.next().await.unwrap().unwrap();
}
```

I'll convert this to a draft while I explore other paths.

davidpdrsn:
I think we should move to checking lifetimes at runtime similar to what `multer` is doing. I can't find a nice way to make it work otherwise, and our current API is too limiting. I'll make a new PR for that.

PR #1656, opened 2022-12-21
prefer `Query<BTreeMap>` to `Query<HashMap>`, for consistent order across requests

liigo:
## Motivation

improve docs

## Solution

improve docs

### DIFF
--- a/axum/src/docs/extract.md
+++ b/axum/src/docs/extract.md
@@ -65,14 +65,14 @@ use axum::{
     Router,
 };
 use serde_json::Value;
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 
 // `Path` gives you the path parameters and deserializes them. See its docs for
 // more details
 async fn path(Path(user_id): Path<u32>) {}
 
 // `Query` gives you the query parameters and deserializes them.
-async fn query(Query(params): Query<HashMap<String, String>>) {}
+async fn query(Query(params): Query<BTreeMap<String, String>>) {}
 
 // `HeaderMap` gives you all the headers
 async fn headers(headers: HeaderMap) {}


davidpdrsn:
Why does this matter for this example?

liigo:
For a request such as `/foo?a=1&b=2`:
If I iterate query params in a handler using `Query<HashMap<(String, String)>>`, I maybe get: (1) `a` then `b`, or (2) `b` then `a`. When using `Query<BTeeMap<(String, String)>>`, there is only one result ( one of (1) and (2) ).

ttys3:
I do not see the point.

as long as you can get the value by key, users just don't care.

for special cases, the user should know what to use. the example should not be particular.

liigo:
```shell
root@ecs-395009:~# curl 'http://127.0.0.1/foo?a=1&b=2&c=3'
  <div>a: 1</div>
  <div>c: 3</div>
  <div>b: 2</div>
root@ecs-395009:~# curl 'http://127.0.0.1/foo?a=1&b=2&c=3'
  <div>b: 2</div>
  <div>c: 3</div>
  <div>a: 1</div>
root@ecs-395009:~# curl 'http://127.0.0.1/foo?a=1&b=2&c=3'
  <div>b: 2</div>
  <div>c: 3</div>
  <div>a: 1</div>
root@ecs-395009:~# curl 'http://127.0.0.1/foo?a=1&b=2&c=3'
  <div>c: 3</div>
  <div>a: 1</div>
  <div>b: 2</div>
root@ecs-395009:~# curl 'http://127.0.0.1/foo?a=1&b=2&c=3'
  <div>c: 3</div>
  <div>a: 1</div>
  <div>b: 2</div>
root@ecs-395009:~# curl 'http://127.0.0.1/foo?a=1&b=2&c=3'
  <div>a: 1</div>
  <div>c: 3</div>
  <div>b: 2</div>
```

```rust
async fn foo(Query(params): Query<HashMap<String,String>>) -> Html<String> {
    let mut output = String::new();
    for (name,value) in params {
        output.push_str(format!("  <div>{name}: {value}</div>\n").as_str());
    }
    Html(output)
}
```

liigo:
> for special cases, the user should know what to use. the example should not be particular. I prefer NOT do any changes.

Maybe experts knows why and what to do, but beginners don't.

`BTreeMap` works for both (experts and beginners). HashMap gives you more memory usage, unstable ordered items, and no speed benefits in this case. Do you have any other reason to stay use `HashMap`?

jamesrampton:
> This example is in the docs, it teaches beginners how to use Query

This is a good reason to continue to use HashMap. It's in the rust book as the primary thing you use for a key, value store.

Seeing the possibly-unknown BTreeMap in the docs for something you're already unfamiliar with could raise the barrier to entry.

davidpdrsn:
I'm sorry but I agree with jamesrampton and ttys3. I think we should keep it as is.

Alost:
hello, i want to know how to keep the query order, the follow code's params is HashMap，can not IndexMap
let Query(params) = parts.extract().await.map_err(IntoResponse::into_response)?;

Alost:
> hello, i want to know how to keep the query order, the follow code's params is HashMap，can not IndexMap let Query(params) = parts.extract().await.map_err(IntoResponse::into_response)?;

```
pub struct IndexQuery(pub IndexMap<String, String>);
impl<S> FromRequestParts<S> for IndexQuery
where
    S: Send + Sync,
{
    type Rejection = Response;
    async fn from_request_parts(
        parts: &mut http::request::Parts, _state: &S,
    ) -> Result<Self, Self::Rejection> {
        let query = parts.uri.query().unwrap_or_default();
        let params: IndexMap<_, _> = form_urlencoded::parse(query.as_bytes())
            .map(|(k, v)| (k.into_owned(), v.into_owned()))
            .collect();
        Ok(IndexQuery(params))
    }
}

let IndexQuery(params) = parts.extract().await.map_err(IntoResponse::into_response)?;
```

jplatte:
If you need help, please use the [Q&A discussions section](https://github.com/tokio-rs/axum/discussions/categories/q-a). Don't post to old issues or PRs.

Issue #1662, opened 2022-12-28
`extract::Path` cannot be made optional by wrapping in `Option`


sukhmel:
- [x] I have looked for existing issues (including closed) about this

### Version

```shell
$ cargo tree | grep axum
├── axum v0.5.17
│   ├── axum-core v0.2.9
│   │   │       ├── axum v0.6.1
│   │   │       │   ├── axum-core v0.3.0
│   ├── axum v0.5.17 (*)
```

### Platform

```shell
$ uname -a
Darwin VladislavSukhmel-MBP 21.6.0 Darwin Kernel Version 21.6.0: Sun Nov  6 23:31:16 PST 2022; root:xnu-8020.240.14~1/RELEASE_X86_64 x86_64
```

### Description

I tried to use `extract::Path` with `Option` as suggested by [documentation](https://docs.rs/axum/latest/axum/extract/index.html#optional-extractors):

> If you wish to make an extractor optional you can wrap it in Option

but the resulting route rejects calls with a missing path parameter returning status 404 and behaves the same as one without an `Option`  wrapper

I tried this code:
```rust
let listener = TcpListener::bind("0.0.0.0:0").expect("Unable to bind ephemeral socket address");
let router = axum::Router::new()
    .route(
        "/test/:id",
        axum::routing::get(move |_: Option<Path<String>>| async move { StatusCode::OK }),
    );                           // ^^^^^^ this has no effect and may be removed without affecting behaviour
let server = axum::Server::from_tcp(listener)
    .expect("Unable to start Axum web server instance")
    .serve(router.into_make_service());
println!("{}", server.local_addr());
server.await
```

then I called it with cURL:
```shell
$ curl http://localhost:62025/test -w '%{response_code}\n'
```

I expected to see this happen:
```shell
# id is not present but the handler is still called
$ curl http://localhost:62025/test -w '%{response_code}\n'
200

# id is present, that's as usual
$ curl http://localhost:62025/test/id -w '%{response_code}\n'
200
```

Instead, this happened:
```shell
# id is not present, the handler is not called, early rejection?
$ curl http://localhost:62025/test -w '%{response_code}\n'
404

# id is present, that's as usual
$ curl http://localhost:62025/test/id -w '%{response_code}\n'
200
```

[...]

PR #1663, opened 2022-12-29
Example of optional Path extractor added

sukhmel:
An example of how a route handler with an `Option<Path<_>>` should be used is added. This should allow for a better understanding of how to idiomatically use multiple paths with the same handler, and allow new users not to stumble upon rejections for a seemingly optional parameter missing

Refs: #1662

## Motivation

For me as a new Axum user, the documentation was not clear as to how to define an optional path parameter.  After creating an issue and seeing a response from maintainers on how this should have been done, it became apparent that having a path parameter extractor optional is not enough as we'd need to somehow guess what the route would look like in such a case. Nevertheless, I find it possible that there will be others like me, who will lack the necessary insight and be puzzled by rejected requests.

I assume that this could be mitigated by stating the fact that Path extractors are a special case of optional extractors, and new users could benefit from an example of how those should be used. 

## Solution

A basic example of how a handler with an optional path parameter extractor could be used for handling several routes

### DIFF
--- a/axum/src/docs/extract.md
+++ b/axum/src/docs/extract.md
@@ -250,31 +250,11 @@ let app = Router::new().route("/users", post(create_user));
 # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
 # };
 ```
+
 For `Path` extractors you would need to define separate routes with the same 
-handler to benefit from optional extractor:
-```rust,no_run
-use axum::{
-    extract::Path,
-    routing::get,
-    Router,
-};
-
-async fn get_user(id: Option<Path<String>>) {
-    if let Some(Path(id)) = id {
-        // route with a path parameter was used
-    } else {
-        // route without a path parameter was used
-    }
-}
-
-let app = Router::new()
-    .route("/users/:id", get(get_user))
-    .route("/users", get(get_user));
-
-# async {
-# axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
-# };
-```
+handler to benefit from the optional extractor, but doing so with an `Option`
+wrapper is discouraged since the extractor may fail in other ways than because
+of a missing parameter. See below for an example of how to handle this.
 
 Wrapping extractors in `Result` makes them optional and gives you the reason
 the extraction failed:
@@ -313,6 +293,43 @@ async fn create_user(payload: Result<Json<Value>, JsonRejection>) {
 }
 
 let app = Router::new().route("/users", post(create_user));
+# async {
+# axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
+# };
+```
+
+Using separate routes to allow for the same function to handle both present
+and missing `Path` parameters, as well as bad requests:
+
+```rust,no_run
+use axum::{
+    extract::{Path, rejection::PathRejection},
+    routing::get,
+    Router,
+};
+
+async fn get_user(id: Result<Path<u128>, PathRejection>) {
+    match parameter {
+        Ok(Path(_)) => {
+            // Route with a valid path parameter was used
+        }
+        Err(PathRejection::MissingPathParams(_)) => {
+            // Route without a path parameter was used
+        }
+        Err(PathRejection::FailedToDeserializePathParams(_)) => {
+            // Route with an invalid path parameter was used
+        }
+        Err(_) => {
+            // `PathRejection` is marked `#[non_exhaustive]` so match must
+            // include a catch-all case.
+        }
+    }
+}
+
+let app = Router::new()
+    .route("/users/:id", get(get_user))
+    .route("/users", get(get_user));
+
 # async {
 # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
 # };
--- a/axum/src/docs/extract.md
+++ b/axum/src/docs/extract.md
@@ -251,10 +251,8 @@ let app = Router::new().route("/users", post(create_user));
 # };
 ```
 
-For `Path` extractors you would need to define separate routes with the same 
-handler to benefit from the optional extractor, but doing so with an `Option`
-wrapper is discouraged since the extractor may fail in other ways than because
-of a missing parameter. See below for an example of how to handle this.
+`Path` extractors have a specific behaviour and should only be used with `Result`
+wrapper, see [`path`](`Path`#optional-path-parameters) module doc for details.
 
 Wrapping extractors in `Result` makes them optional and gives you the reason
 the extraction failed:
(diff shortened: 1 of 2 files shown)


davidpdrsn (commented):
Thanks! Though I think it should be moved to the `Path` extractor, since its specific to that.

jplatte:
For a `String` path parameter it isn't really relevant, but `Path` extraction can fail in another way than the route not having the required parameter(s): argument deserialization (first variant of [`PathRejection`](https://docs.rs/axum/latest/axum/extract/rejection/enum.PathRejection.html)). Because of that, I'm not sure whether we should encourage this pattern in general.

davidpdrsn:
Hm yeah I was thinking the same. I thought it was more relevant to middleware maybe.

sukhmel:
@jplatte as for deserialization errors, I had something like this in mind, since we have a numeric IDs and a special case of non-numeric ID to represent the current user. So having it fail and then try another route with that special case is exactly what is needed. As for a String parameter in an example, maybe I should replace it with something that must be parsed like u128 instead?

@davidpdrsn maybe it's better to move the example to Path, but still a hint in the Optional extractors section is necessary, I think. I for one would not try searching for way to make a specific extractor optional in it's doc, if there exist a dedicated doc section about the general case of optional extractors.

jplatte:
> @jplatte as for deserialization errors, I had something like this in mind, since we have a numeric IDs and a special case of non-numeric ID to represent the current user. So having it fail and then try another route with that special case is exactly what is needed.

You are misunderstanding. axum never tries another route when an extractor fails. Instead, for two routes for `/user/me` and `/user/:user_id`, the former is determined to be more specific and its handler will be called on that exact path. However, your handler will still be called for `/user/not_an_integer` which also matches `/user/:user_id`, and because `Path<IntegerType>` extraction will fail with a deserialization error, your `Option<Path<IntegerType>>` extractor parameter will be `None`.

sukhmel:
thanks for the explanation, that will certainly help me when defining routes

regarding the concern of encouraging this pattern, I find it possible that the user could wrap it in Result instead, to be able to check whether this was a missing parameter or a parameter, that failed to parse. Maybe I could rewrite the example to reflect that?

jplatte:
Maybe the new docs could be on the `Path` page but linked to from the extract module page?

sukhmel:
@jplatte ok, I'll try to device a way to phrase and link it that way

sukhmel:
not sure if I need to tag anyone after I've committed changes, please do tell me if I should, as I am not used to this workflow

I have moved an example and added a link, not sure if it is worded in a proper way, but it seems ready for a review

axum/src/extract/path/mod.rs
```diff
@@ -143,6 +143,52 @@ use std::{
...
+/// }
+///
+/// let app = Router::new()
+///     .route("/users/:id", get(get_user))
+///     .route("/users", get(get_user));
+///
+/// # async {
+/// # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
+/// # };
+/// ```
```
davidpdrsn, line 189: I'm not sure we should show examples of code we _don't_ recommend users write 🤔

I think this is a nicer way of doing it

```rust
async fn with_params(Path(id): Path<u32>) -> Response {
    do_thing(Some(id)).await
}

async fn without_params() -> Response {
    do_thing(None).await
}

async fn do_thing(param: Option<u32>) -> Response { ... }

Router::new()
    .route("/:id", get(with_params))
    .route("/", get(without_params));
```

I'd be okay with including something like that in the docs for `Path`.

axum/src/docs/extract.md
```diff
@@ -251,6 +251,9 @@ let app = Router::new().route("/users", post(create_user));
 # };
 ```
 
+`Path` extractors have a specific behaviour and should only be used with `Result`
+wrapper, see [`path`](`Path`#optional-path-parameters) module doc for details.
```
davidpdrsn, line 255: Hm I don't think I'd say that `PathRejection` is special. All rejections work like this. For example `Option<Json<T>>` doesn't tell you whether the `content-type: application/json` was missing or whether there was a deserialization error.

I think I'd just leave out this paragraph.

axum/src/extract/path/mod.rs
sukhmel: I thought the point was not to use the `Option<Path>` rather than not to use `Result<Path>`. Maybe it would be OK to add both this to the example and `Result<Path>` as a manual handling example?

axum/src/docs/extract.md
sukhmel: Well, what I meant is that a missing path parameter will be rejected early and not reach the handler at all. Maybe I can rephrase this to emphasize that this is about rejection based on a route before any handler is called

axum/src/extract/path/mod.rs
davidpdrsn: I don't think `Result<Path>` (or `Option<Path>`) should be used to implement optional path params. So I don't think we should have an example that shows that. `Result<Path>` can be used to customize deserialization errors but thats already covered by the existing extractor docs.

axum/src/docs/extract.md
davidpdrsn: But that is also not special for `Path`. Any rejection will cause the handler to not be called.

sukhmel: I seem to misunderstand how this works, but I can't get a clear picture. Let's take the optional extractor example, there we have `Option<Json<Value>>` and missing json does not mean that we will get a rejection without calling the handler.

In the case of Path extractor, on the other hand, the parameters should match the route path first and only then they are passed to handler, that might also reject the result based on its own parameters.

axum/src/extract/path/mod.rs
sukhmel: I see, it makes sense to rewrite this example then, will do

axum/src/docs/extract.md
davidpdrsn: > In the case of Path extractor, on the other hand, the parameters should match the route path first and only then they are passed to handler, that might also reject the result based on its own parameters.

Thats not correct. If you do `.route("/", get(|params: Option<Path<String>>| async {}))`, i.e a route without any path params using `Option<Path<_>>` then the handler is still always called. `params` will always be `None`.

A mismatch between the route and `Path` triggers a rejection which `Option` discards and yields a `None` (https://github.com/tokio-rs/axum/blob/main/axum-core/src/extract/mod.rs#L153).

sukhmel: So `Option<Json<T>>` being None is not a rejection, or is that something else I am missing?

davidpdrsn: What do you mean by " is not a rejection"? `Option<Json<_>>` is an extractor whose rejection is `Infallible` (ie never rejects)

davidpdrsn: I feel like you're misunderstanding what I'm trying to say.

> if I do `.route("/", get(|params: Option<Path<String>>| async {}))` it will handle the missing path parameter as expected, because there is no parameter in the path, nowhere to get it from

In this case the handler will _always_ be called and `params` will be `None`. No rejections will happen.

> if I do `.route("/:parameter/suffix", get(|params: Option<Path<String>>| async {}))` then there is no way to get this handler called

Here the handler will also _always_ be called and `params` will be a `Some`. Again no rejections. If the route matches then the path parameter can be extracted and `String` cannot fail deserialization.

> even if I do `.route("/:parameter", get(|params: Option<Path<String>>| async {}))` it will end up rejected both without trailing slash, and with it (because with trailing slash there will be a redirection ending with rejection, too)

Same here as the previous example. It'll always work.

davidpdrsn:
I'm sorry but I think I'll close this. I worry that adding this to the docs is just gonna confused people and optional path parameters is probably not something we wanna encourage anyway. Feel free to ask follow up questions though!

axum/src/docs/extract.md
sukhmel: I was talking about those routes being called without parameter, like `//suffix`

well, anyway, maybe this issue will at least pop up in search for someone. This week I've already explained to a colleague why they shouldn't use `Option<Path>` directly in axum handler and rather create a shim like you showed, so this is not something that's not going to be used just because it is not clearly mentioned in the docs.

I'd argue that it would be better to change the relevant part as per your comments to show the way it should be used rather than have everyone come up with their own ideas of what to do in such a case

PR #1665, opened 2022-12-29
add more comments to chat example

hmeine:
## Motivation

As discussed with Programatik and Ferry on Discord, people have lots of
questions about the chat example, so some more code comments might be
helpful.

## Solution

As a new axum user myself (target audience of this example), I just tried to
express what I learned from carefully reading this example and implementing
something similar with slightly more words than before.  A particular focus was
on the spawned tasks (hopefully useful for readers not too familiar with async
rust yet).

### DIFF
--- a/examples/chat/src/main.rs
+++ b/examples/chat/src/main.rs
@@ -26,8 +26,10 @@ use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
 // Our shared state
 struct AppState {
-    user_set: Mutex<HashSet<String>>, // Currently taken user names (must be unique)
-    tx: broadcast::Sender<String>, // Receivers can be created via tx.subscribe()
+    // We require unique usernames. This tracks which usernames have been taken.
+    user_set: Mutex<HashSet<String>>,
+    // Channel used to send messages to all connected clients.
+    tx: broadcast::Sender<String>,
 }
 
 #[tokio::main]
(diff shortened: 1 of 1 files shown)


examples/chat/src/main.rs
```diff
@@ -26,8 +26,8 @@ use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
 // Our shared state
 struct AppState {
-    user_set: Mutex<HashSet<String>>,
-    tx: broadcast::Sender<String>,
+    user_set: Mutex<HashSet<String>>, // Currently taken user names (must be unique)
+    tx: broadcast::Sender<String>, // Receivers can be created via tx.subscribe()
```
davidpdrsn, line 30:
```suggestion
    // We require unique usernames. This tracks which usernames have been taken.
    user_set: Mutex<HashSet<String>>,
    // Channel used to send messages to all connected clients.
    tx: broadcast::Sender<String>,
```

davidpdrsn (commented):
Thanks!

A very minor nit but would you change the `:` at the end of the comments back to either `.` or just nothing? We don't use that convention anywhere else in axum.

examples/chat/src/main.rs
hmeine: Oh, sure. I must admit that I found the full stop annoying ;), but I can see that consistency is most important. Will change that again.

davidpdrsn (approved):
Thanks!

PR #1671, opened 2023-01-03
Enable the runtime feature of hyper when using tokio

paolobarbolini:
## Motivation

Currently axum never enables the `runtime` feature of hyper, which gates functions like [`Server::http1_header_read_timeout`](https://docs.rs/hyper/latest/hyper/server/struct.Builder.html#method.http1_header_read_timeout)

## Solution

This makes axum enable the runtime `feature`. The alternative is to import hyper and enable the feature but then Rust Analyzer imports things from hyper instead of axum, which is very annoying.


jplatte:
Well since the runtime feature is described as

> Enables convenient integration with `tokio`, providing connectors and acceptors for TCP, and a default executor.

and this is about what features our own `tokio` dependency activates, it seems fine to me.

davidpdrsn (approved):

PR #1676, opened 2023-01-03
fix: warnings for cloning references

### LINKED ISSUE #1674: Rustc and clippy warnings generated by derive(FromRef)

### Version

- axum v0.6.1
- axum-core v0.3.0
- axum-macros v0.3.0 (proc-macro)

### Platform

Linux alpha 6.1.1 #1-NixOS SMP PREEMPT_DYNAMIC Wed Dec 21 16:48:12 UTC 2022 x86_64 GNU/Linux

### Crates

axum-macros

### Description

When deriving `FromRef` for `&'static Something`, the following lints are triggered by the generated code:
- `clippy::clone_double_ref` (deny by default)
- `noop_method_call` (allow by default)

I expect the generated code to cause no errors, possibly by suppressing the ones mentioned above.

Minimal example:
```rs
#![warn(noop_method_call)]

#[derive(axum::extract::FromRef)]
struct State {
    foo: &'static str,
}
```

alexander-jackson:
## Motivation

In #1674, it was pointed out that `axum-macros` causes some `clippy` warnings (and some `rustc` lints to fail) as the `FromRef` macro generates code which clones reference types unnecessarily.

## Solution

This seems to be when generating the body of the trait implementation, but since we have the type information from `syn` at that point it's relatively simple to resolve these.

This change:
* Avoids using `.clone` if we are generating the implementation for a reference type field
* Adds a simple UI test to ensure this compiles with the `rustc` lint set to deny

Fixes #1674.

### DIFF
--- a/axum-macros/src/from_ref.rs
+++ b/axum-macros/src/from_ref.rs
@@ -3,7 +3,7 @@ use quote::quote_spanned;
 use syn::{
     parse::{Parse, ParseStream},
     spanned::Spanned,
-    Field, ItemStruct, Token,
+    Field, ItemStruct, Token, Type,
 };
 
 use crate::attr_parsing::{combine_unary_attribute, parse_attrs, Combine};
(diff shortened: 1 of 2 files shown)


davidpdrsn (approved):
Thanks!

davidpdrsn:
Wanna add a note to changelog as well?

alexander-jackson:
@davidpdrsn I don't mind, happy to add one if it's worth it 😅 not the most exciting of changes though, I presume `rustc` optimises away the `.clone` anyway at compile time

Issue #1654, opened 2022-12-19
Enable Routing by Content-Type Header


dawid-nowak:
- [v ] I have looked for existing issues (including closed) about this

## Feature Request
Enable Content Based Routing

### Motivation
Currently Axum routes by path only so in order to handle different types of content an approch shown [in example](https://github.com/tokio-rs/axum/tree/main/examples/parse-body-based-on-content-type) has to be implemented in all handlers resulting in extra boilerplate and complexity that could be handled by the framework.

Other frameworks can route by path and content type. See examples for [dotnet](https://learn.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-7.0) and [springboot](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/bind/annotation/RequestMapping.html)

### Proposal
One approach would be to extend the existing Router API to allow the user to pass the content type that the handler is processing. 
For example:
```
let app = Router::new()
    .route("/path", post(path_handler_for_json, "application/json")
    .route("/path", post(path_handler_for_xml,"application/xml")
```

### Alternatives
Create a completely separate ContentRouter service

jplatte:
I think this is definitely something we could explore inside `axum-extra`. However, that would require first making `.route` more generic. @davidpdrsn wdyt about axum growing its own `Service` & `Layer` trait equivalents with blanket implementations for any type implementing the tower ones? These could be generic over the state type so we can get rid of the `{route,nest,..}` vs. `{route,nest,..}_service` distinction in the public API again. Then axum-extra could provide `MethodAndContentTypeRouter`.

davidpdrsn:
I think it's worth exploring!

dawid-nowak:
Looking through the code i can see how this is can be "E-hard" :) 

Perhaps another option would be to do something like this ?
```
let app = Router::new()
    .route(("/path", "application/json"), post(path_handler_for_json)
    .route("/path"), post(path_handler_for_default))

```
I think that would only require changes to Node and instead of str we woudl have a tuple (str, mime)

jplatte:
There are definitely ways to solve this that aren't that hard, but if it's going to be part of axum-extra and not a thirdparty project, I want it to be very close to the existing API and not something else just because that might make the implementation simpler 😉

davidpdrsn:
Yeah I agree it should feel very close to the existing API. I also think how axum doesn't allow overlapping routes is an important feature and any additional routing we support should do the same.

[...]

PR #1679, opened 2023-01-05
Enable Routing by Content-Type Header

dawid-nowak:
Continuation from [Enable Routing by Content-Type Header](https://github.com/tokio-rs/axum/issues/1654)


davidpdrsn:
I'll close this for now. I'm not convinced this additional complexity is worth it. We can continue the discussion in https://github.com/tokio-rs/axum/issues/1654.

Issue #1378, opened 2022-09-16
Wrong query string should yield 400, not 422


jblachly:
### Version

axum v0.5.0-rc.2

### Platform

Darwin mbp16.local 21.6.0 Darwin Kernel Version 21.6.0: Wed Aug 10 14:28:23 PDT 2022; root:xnu-8020.141.5~2/RELEASE_ARM64_T6000 arm64

### Description

The Query extractor, if unable to serialize results, yields 422 Unprocessable entity.

However, I expect it should return 400, Bad Request.

422 is appropriate when the body is well-formed, but semantically incorrect, and 422 is commonly returned from other frameworks for JSON deserialization errors.

However, I have never seen a bad query string yield 422; typically a bad query string would yield 400 ("The client should not repeat this request without modification.")

Ultimately my suggestion is to keep current behavior for Body extractors such as Json, but return 400 for failed Query extractor.

**references:**

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400

**Some prior discussion on 422**
https://github.com/tokio-rs/axum/issues/865

jplatte:
Please add more detail to your issue, only posting a title with no explanation / description is disrespectful of the maintainer's time.

jblachly:
> Please add more detail to your issue, only posting a title with no explanation / description is disrespectful of the maintainer's time.

Sorry, I had hit "cmd+enter" when filling out the template! It's filled out now -- i raced against time to get it filled out before you commented, but you were too quick

davidpdrsn:
422 still seems correct to me. Why is JSON different?

jblachly:
> 422 still seems correct to me. Why is JSON different?

Great question. A reading of the specs reveals this difference in semantics:

422 deals with the the request body in particular and informs us that the "server understands the _content type_ of the request entity, and )the syntax of the request entity is correct,_ " (emphasis mine)

while 400 indicates that "the server cannot or will not process the request due to something that is perceived to be a client error (for example, **malformed request syntax**" (emphasis mine)

davidpdrsn:
Ah yes I see that. I agree, that seems correct to use 400 instead of 422. I've filed https://github.com/tokio-rs/axum/pull/1387

PR #1681, opened 2023-01-06
Use 422 Unprocessable Entity for form deserialization errors

### LINKED ISSUE #1680: `Form` extractor should return 422 Unprocessable Entity for deserialization errors

### Motivation

I noticed this when upgrading a project from axum 0.5 to 0.6, which led to some failing tests of invalid form data: deserialization errors from the [`Form`](https://docs.rs/axum/latest/axum/struct.Form.html) extractor now convert into `400 Bad Request` responses, when previously they were `422 Unprocessable Entity` responses. Per the semantics discussed in #1378 it seems like `422 Unprocessable Entity` would be the more appropriate response for syntactically valid form data that could not be deserialized.

### Proposal

From doing some digging around, it looks like this change occurred to fix #1378. At that time, the rejection for a `Form` deserialization error was also `FailedToDeserializeQueryString`: https://github.com/tokio-rs/axum/blob/74105fc7904411df27f116c958f9266007f4650b/axum/src/extract/rejection.rs#L141-L151

Later, in #1496, `Form` gets a dedicated `FailedToDeserializeForm` variant: https://github.com/tokio-rs/axum/blob/351df649ab05fb30c382167b2158828a4f48e4a3/axum/src/extract/rejection.rs#L99-L109

However, this rejection is still set to return a `400 Bad Request`: https://github.com/tokio-rs/axum/blob/211de92d24a0d6e1ffc0c76b75c26c53c7eb09d3/axum/src/extract/rejection.rs#L73-L79

Since the `Form` and `Query` extractors now have dedicated rejection types, it seems we could simply update the `status` for `FailedToDeserializeForm` to `UNPROCESSABLE_ENTITY` without disturbing the original fix in #1378. I'd be happy to make that PR if it's desirable and I'm not missing anything. This would be a breaking change (assuming the rejection statuses are covered by the versioning policy, and this isn't considered a bug).

connec:
Fixes #1680.

## Motivation

Per #1680, we would prefer to use `422 Unprocessable Entity` responses for non-syntactic deserialization errors from the `Form` extractor. This is now trivial since the `Form` and `Query` rejection types were separated in #1496.

## Solution

- Update the `status` for `axum::extract::rejection::FailedToDeserializeForm` from `BAD_REQUEST` to `UNPROCESSABLE_ENTITY`.
- The same change has been applied to `axum_extra::extract::FormRejection::FailedToDeserializeForm` for consistency.

### DIFF
--- a/axum-extra/src/extract/form.rs
+++ b/axum-extra/src/extract/form.rs
@@ -92,7 +92,7 @@ impl IntoResponse for FormRejection {
         match self {
             Self::RawFormRejection(inner) => inner.into_response(),
             Self::FailedToDeserializeForm(inner) => (
-                StatusCode::BAD_REQUEST,
+                StatusCode::UNPROCESSABLE_ENTITY,
                 format!("Failed to deserialize form: {}", inner),
             )
                 .into_response(),
--- a/axum/src/extract/rejection.rs
+++ b/axum/src/extract/rejection.rs
@@ -71,7 +71,7 @@ define_rejection! {
 }
 
 define_rejection! {
-    #[status = BAD_REQUEST]
+    #[status = UNPROCESSABLE_ENTITY]
     #[body = "Failed to deserialize form"]
     /// Rejection type used if the [`Form`](super::Form) extractor is unable to
     /// deserialize the form into the target type.


connec:
Sure, I was a bit lazy and left it out since nothing was presently testing it, but happy to add one. I'll have to sort it a little bit later though.

davidpdrsn:
That's alright. I think having a test would be good. Then its harder to break by accident.

connec:
Ah, it looks like I spoke too soon with my assumption that the `Form` and `Query` extractors were fully separated. You can presently use the `Form` extractor with `GET` requests and it will read from the URL query instead: https://github.com/tokio-rs/axum/blob/211de92d24a0d6e1ffc0c76b75c26c53c7eb09d3/axum/src/extract/raw_form.rs#L47-L64

As such, with this change, a `GET` request with semantically invalid `Form` data in the URL would return a `422`, which #1378 establishes is invalid.

I'm not sure what would be the preferred solution to this? Some options I can think of:

1. Remove the special-casing for `GET` requests (and perhaps direct people to `Query` in the documentation?). This seems like it *would* be a breaking change though.

2. Pipe through some information about the source of the form data, and use separate `composite_rejection` variants for semantically invalid queries vs. bodies. Adding a separate variant would not be breaking, but `RawForm` is a public type with public fields, so adding any more info to the struct would also be a breaking change.

3. The `FromRequest` implementations of `RawForm` and `Form` could be fully decoupled (i.e. copy the `RawForm` body implementation into the `Form` implementation). This would functionally be a breaking change, but the documentation for `Form` currently suggests that only request bodies are processed, so it could be argued to be a bug fix 😅 https://github.com/tokio-rs/axum/blob/211de92d24a0d6e1ffc0c76b75c26c53c7eb09d3/axum/src/form.rs#L15-L21

Perhaps there are other options?

davidpdrsn:
Good catch! Another reason testing is important. I think adding another variant to `FormRejection` is the simplest approach. I went ahead and implemented it in https://github.com/tokio-rs/axum/pull/1683.

davidpdrsn:
Thanks for working on this but I think we'll move forward with https://github.com/tokio-rs/axum/pull/1683

connec:
Of course, sorry I didn't have to follow it through!

davidpdrsn:
No worries! You're welcome to give https://github.com/tokio-rs/axum/pull/1683 a look though :)

PR #1683, opened 2023-01-07
Use `422 Unprocessable Entity` for `Form` deserialization errors, except `GET` and `HEAD` requests

davidpdrsn:
For `GET` and `HEAD` requests `Form` should use `400 Bad Request` (`422` is only for request bodies).

Fixes https://github.com/tokio-rs/axum/issues/1680

## TODO

- [x] Changelog
- [x] Update docs for `Form` to mention that it handles both query and body params.


axum/src/extract/rejection.rs
```diff
@@ -104,6 +112,7 @@ composite_rejection! {
     pub enum FormRejection {
         InvalidFormContentType,
         FailedToDeserializeForm,
```
jplatte, line 114: This should have a `Query` prefix then in the next breaking-change release, I guess?

jplatte (approved):

axum/src/extract/rejection.rs
davidpdrsn: Yes I think so!

PR #1692, opened 2023-01-13
Add `axum_extra::extract::Multipart`

davidpdrsn:
It is is similar to `axum::extract::Multipart` except that it enforces field exclusivity at runtime instead of compile time.

I kinda wish we didn't have to do it honestly. Tracking lifetimes at compile feels much more correct but we've learned that it causes lots of headaches, especially because `Field` not being `'static`.

I'm thinking for 0.7 we can move this into axum but keep it in axum-extra in the meantime.


jplatte:
Are we certain the things people tried to do and ran into compile errors with ultimately worked? I got the impression that most or all of them would have led to runtime panics instead if converted to this API.

davidpdrsn:
This is mainly for fixing things like https://github.com/tokio-rs/axum/discussions/1644 where you wanna stream the contents of the fields somewhere else, perhaps to upload the data somewhere.

The use cases we've heard about where people wanna find a field by name and do things with it would most likely result in runtime errors. But thats probably the wrong way to approach the problem in the first place.

netthier:
> Are we certain the things people tried to do and ran into compile errors with ultimately worked? I got the impression that most or all of them would have led to runtime panics instead if converted to this API.

I had a case where I wanted to stream a field of a multipart upload directly to S3 and got a similar problem as the user in the discussion linked by David. 
As I only ever held onto one field at a time it shouldn't cause runtime errors with the new solution.

davidpdrsn:
@jplatte should we merge this?

jplatte (approved):
Haven't reviewed in detail, but sure.

Issue #1678, opened 2023-01-05
`extract::Path` doesn't work with types using `serde::Deserializer::deserialize_any`


paolobarbolini:
### Motivation

`PathDeserializer` doesn't support `deserialize_any` [^1] so it doesn't work with things like `time::Date` [^2]

### Proposal

Add support for `deserialize_any`. I might be able to look into it in the following days.

### Alternatives

Create a wrapper around the type and force it to use a specific deserializer

[^1]: https://github.com/tokio-rs/axum/blob/87a80ec47bfd2c39ac4ea3d1acffb5d1ff540f63/axum/src/extract/path/de.rs#L59
[^2]: https://github.com/time-rs/time/blob/d97594b9f5cfeab0335c16687e26a8b51c56a526/time/src/serde/mod.rs#L124

davidpdrsn:
What would such an implementation look like? Afaik URL params aren't a self describing format so when you encounter some key/value pair how do you know what to serialize that into?

jplatte:
I think _if_ we support it, `deserialize_str` is the right implementation.

paolobarbolini:
`serde_urlencoded` fowards it to `deserialize_map` and it seems to work
https://github.com/nox/serde_urlencoded/blob/d8bc15d16adf6b3ba6ae46d199d7109cf5079efa/src/de.rs#L102

The playground doesn't enable the right time features for it to work, but you can test it locally:

## Cargo.toml

```toml
[package]
name = "test123"
version = "0.1.0"
edition = "2021"

[dependencies]
serde = { version = "1", features = ["derive"] }
time = { version = "0.3", features = ["serde-human-readable", "parsing"] }
serde_urlencoded = "0.7"
```

## main.rs

```rust
use serde::Deserialize;
use time::Date;

fn main() {
    #[derive(Debug, Deserialize)]
    struct Stuff {
        date: Date,
    }

    let deserialized: Stuff = serde_urlencoded::from_str("date=2022-01-01").unwrap();
    println!("{deserialized:?}");
}
```

## stdout

```
Stuff { date: 2022-01-01 }
```

davidpdrsn:
It seems `serde_urlencoded` also just calls `deserialize_str` for values, so thats probably also what axum should do (https://github.com/tokio-rs/axum/pull/1693)

PR #1693, opened 2023-01-13
Support `deserialize_any` for `Path`

davidpdrsn:
Forwarding to `deserialize_str` is probably gonna do the correct thing in most cases. It is also what `serde_urlencoded` does for values (https://github.com/nox/serde_urlencoded/blob/master/src/de.rs#L194-L202)

Fixes https://github.com/tokio-rs/axum/issues/1678

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -8,8 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 # Unreleased
 
 - - **added** Implement IntoResponse for &'static [u8; N] and [u8; N] ([#1690])
+- **fixed:** Make `Path` support types uses `serde::Deserializer::deserialize_any` ([#1693])
 
 [#1690]: https://github.com/tokio-rs/axum/pull/1690
+[#1693]: https://github.com/tokio-rs/axum/pull/1693
 
 # 0.6.2 (9. January, 2023)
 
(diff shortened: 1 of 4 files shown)


jplatte (commented):
Hmmm thinking about this more, what happens when there are multiple path params?

davidpdrsn:
Oh good catch. That didn't "just work". Not sure why but I'll look into it.

jplatte:
Actually I think we should use forward to the sequence or map deserialize if there's more than one param... Which calls into question whether str for a single parameter really makes that much sense.

davidpdrsn:
I don't quite understand what you mean 🤔 but I think I was able to make all the cases work.

jplatte:
So what do we end up with when using a truly "catch-all" type like `serde_json::Value`, for the different cases (one param / multiple)?

davidpdrsn:
You just get strings. So `/:a` with `Path<Value>` gives `Value::String("foo")`. Same for multiple params. Thats why I was a bit on the fence about this in the first place but I don't think there is much we can do (since path params aren't a self describing format) and forwarding to `deserialize_str` seems like the most useful thing.

jplatte:
What does that mean for multiple parameters? You get them concatenated into one string?

davidpdrsn:
If you do `/:a/:b` you have to use `Path<(Value, Value)>`. `Path<Value>` doesn't work and gives a "wrong number of arguments" error.

jplatte:
Right, I guess that is fine.

davidpdrsn:
I've added a test for that as well.

jplatte (approved):

jplatte:
(one could kind of expect it to return `Ok(Value::Array(_))` or `Ok(Value::Object(_))` but it's really not that important I think)

Issue #1624, opened 2022-12-07
`Router::nest` has poor performance on 0.6


nicolaspernoud:
## Bug Report (more of a questions actually)

I know it is kind of a blind shot, but maybe someone can give some insight into this...

### Version

├── axum v0.6.1
│   ├── axum-core v0.3.0
├── axum-extra v0.4.2
│   ├── axum v0.6.1 (*)
├── axum-server v0.4.2
│   ├── axum-server v0.4.2 (*)

### Platform

Linux *** 5.19.0-26-generic #27-Ubuntu SMP PREEMPT_DYNAMIC Wed Nov 23 20:44:15 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

### Crates

axum = { version="0.6.1", features = ["query", "json", "http2", "tokio", "headers"], default-features = false }
axum-extra = { version = "0.4.2", features = ["cookie-private"], default-features = false }

### Description

I migrated my app from axum 0.5 to axum 0.6 as it is detailed here : https://github.com/nicolaspernoud/atrium/commit/d002700df371eb002c44504038e87ddcb6e14715#diff-28504f0a0ed99657d5f5bee11b72eb6ef2db4faf68b0f6e94bfe28c77e34e35a .
The old code manages approximately 24 500 req/sec in a reverse proxy use case.
The new code and so on manage approximately 16 500 req/sec, which is a huge drop.
The new code also use state instead of extensions, but I am not sure that is the problem, it seems to be related to axum itself : a temp branch kept extensions for configuration and it was kind of the same performance.

Any idea or advice on how to improve that ?

### Test code

The test code is here : https://github.com/nicolaspernoud/go_vs_rust_reverse_proxy_benchmark/blob/main/test_atrium.sh .

By the way, in a super simple use case as shown here : https://github.com/nicolaspernoud/go_vs_rust_reverse_proxy_benchmark/blob/main/test.sh ; that is not the case, axum 0.6 is actually slightly faster.

I suppose that is related to the way axum extract something from the request (?).

davidpdrsn:
That is hard to say with such a big amount of code. Are you able to make a smaller repro?

nicolaspernoud:
Well, that will take some work (and time), but I will try...

davidpdrsn:
We do have some benchmarks [here](https://github.com/tokio-rs/axum/blob/main/axum/benches/benches.rs) that you try to poke at and see if you notice anything.

[...]

PR #1694, opened 2023-01-13
Add internal macro to make tests of `nest` easier to write

davidpdrsn:
This might be considered a dirty hack but stick with me.

There have been talk about changing `nest` back to how it worked in 0.5, where it would "flatten" the routes into the outer router. That should fix the [performance regression](https://github.com/tokio-rs/axum/issues/1624) and the [`MatchedPath` issues](https://github.com/tokio-rs/axum/issues/1441) people have reported.

However the reason I change `nest` in 0.6 to just call `nest_service` was to more the subtle differences between the two methods. So if we're gonna back I wanna make sure we aren't introducing new differences.

One way to do that is for each test that uses `nest`, write the same test but use `nest_service`. That obviously kinda tedious to do manually but easy to do in a macro. So this PR adds such a macro.

### DIFF
--- a/axum-macros/Cargo.toml
+++ b/axum-macros/Cargo.toml
@@ -11,6 +11,10 @@ readme = "README.md"
 repository = "https://github.com/tokio-rs/axum"
 version = "0.3.1" # remember to also bump the version that axum and axum-extra depends on
 
+[features]
+default = []
+__private = ["syn/visit-mut"]
+
 [lib]
 proc-macro = true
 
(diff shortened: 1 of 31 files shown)


jplatte (commented):
Hm, how does this not cause compilation errors for cases where the inner Router has a non-unit state type?

davidpdrsn:
It does so for such test we wouldn't be able to use it.

jplatte:
Do we not have such tests yet? It seems to me like you just search&replace'd all the `tokio::test`s, why not be a little more selective?

jplatte (approved):
Other than that the macro is maybe overused a little, LGTM.

davidpdrsn:
> Do we not have such tests yet? 

Apparently not 😅 I think those tests are doctests actually.

> It seems to me like you just search&replace'd all the tokio::tests, why not be a little more selective?

Yeah I did. Thought that was easier and shouldn't have any impact when used on functions that don't use `nest`.

PR #1696, opened 2023-01-14
Add `option_layer`

davidpdrsn:
Tower already has `option_layer` for conditionally applying a `Layer` but since it uses `tower::util::Either` it always changes the error type to `BoxError`. That requires using `HandleErrorLayer` which is inconvenient for axum users.

That has been changed in tower
(https://github.com/tower-rs/tower/pull/637) but still has some issues (https://github.com/tower-rs/tower/issues/665). Its also a breaking change so hasn't been released yet.

In the meantime I figure we can provide our own thing in axum-extra, since we already have an `Either` type there.

### DIFF
--- a/axum-extra/src/middleware.rs
+++ b/axum-extra/src/middleware.rs
@@ -1,0 +1,44 @@
+//! Additional middleware utilities.
+
+use crate::either::Either;
+use tower_layer::Identity;
+
+/// Convert an `Option<Layer>` into a [`Layer`].
+///
+/// If the layer is a `Some` it'll be applied, otherwise not.
+///
+/// # Example
+///
+/// ```
+/// use axum_extra::either::option_layer;
+/// use axum::{Router, routing::get};
+/// use std::time::Duration;
+/// use tower_http::timeout::TimeoutLayer;
+///
+/// # let option_timeout = Some(Duration::new(10, 0));
+/// let timeout_layer = option_timeout.map(TimeoutLayer::new);
+///
+/// let app = Router::new()
+///     .route("/", get(|| async {}))
+///     .layer(option_layer(timeout_layer));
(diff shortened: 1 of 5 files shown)


axum-extra/src/middleware.rs
```diff
@@ -0,0 +1,44 @@
...
+/// use axum::{Router, routing::get};
+/// use std::time::Duration;
+/// use tower_http::timeout::TimeoutLayer;
+///
+/// # let option_timeout = Some(Duration::new(10, 0));
+/// let timeout_layer = option_timeout.map(TimeoutLayer::new);
+///
+/// let app = Router::new()
+///     .route("/", get(|| async {}))
+///     .layer(option_layer(timeout_layer));
```
Oliboy50, line 23: Just to be sure, if we pass `.layer(Identity::new())`, it means zero additional cost for each incoming request?
Just like we wouldn't write `.layer(...)` at all?

Oliboy50 (approved):
🙇

axum-extra/src/middleware.rs
davidpdrsn: It does have additional cost. It'll create a new [`Route`](https://github.com/tokio-rs/axum/blob/main/axum/src/routing/route.rs#L30) which requires allocation and a dynamic dispatch. I doubt that impact will be measurable in a real world app however but if you really wanna avoid it you have to use an `if`.

davidpdrsn: And by the way this is why we recommend adding multiple middleware with [`ServiceBuilder`](https://docs.rs/axum/latest/axum/middleware/index.html#applying-multiple-middleware).

jplatte (approved):

PR #1699, opened 2023-01-15
StreamBody responds with Bytes, not Result

hamza1311:
## Motivation

When streaming body in response, errors from the stream cannot be sent, as request's status code and headers are already sent. The `Err` in the `StreamBody`'s `Result` was silently ignored. 

## Solution

Replace `Result<Bytes, Error>` with `Bytes`, which this PR does.

## Ref

Discord discussion: https://discord.com/channels/500028886025895936/870760546109116496/1064231913289683016

### DIFF
--- a/axum/src/body/stream_body.rs
+++ b/axum/src/body/stream_body.rs
@@ -1,11 +1,11 @@
 use crate::{
     body::{self, Bytes, HttpBody},
     response::{IntoResponse, Response},
-    BoxError, Error,
+    Error,
 };
 use futures_util::{
     ready,
-    stream::{self, TryStream},
+    stream::{self, Stream},
 };
 use http::HeaderMap;
 use pin_project_lite::pin_project;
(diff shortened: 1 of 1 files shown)


axum/src/body/stream_body.rs
```diff
@@ -35,8 +35,8 @@ pin_project! {
     /// use futures::stream::{self, Stream};
     /// use std::io;
     ///
-    /// async fn handler() -> StreamBody<impl Stream<Item = io::Result<&'static str>>> {
-    ///     let chunks: Vec<io::Result<_>> = vec![
+    /// async fn handler() -> StreamBody<impl Stream<Item = &'static str>> {
+    ///     let chunks: Vec<_> = vec![
```
davidpdrsn, line 39: Need to update the items in the vec to not be `Ok(_)`.

davidpdrsn:
Also seems there are some other compile errors in axum-extra.

hamza1311:
How would/should errors be handled for `axum_extra::body::AsyncReadBody` and the like? `tokio_util::io::ReaderStream` understandably returns `Result<Bytes, std::io::Error>` but we can't return the error. It seems wrong to silently ignore any errors from the stream. Same is the case with `axum_extra::json_lines::JsonLines`. What if we can't serialize the value as JSON? There's no way to return that error

davidpdrsn:
Hm thats a good question 🤔 I'm not sure. I suppose we need to log the error or maybe provide a callback so the user can do their own logging, like we do for [`WebSocketUpgrade`](https://docs.rs/axum/latest/axum/extract/struct.WebSocketUpgrade.html#method.on_failed_upgrade). I'll try and think about it.

davidpdrsn:
Closing in favor of https://github.com/tokio-rs/axum/pull/1766. Let me know what you think!

PR #1702, opened 2023-01-17
Fix some comments in websockets example

FrankReh:
A few comments in the websockets example weren't complete. A few had trivial typos. And one comment change with a gender-neutral seems to be more idiomatic with tokio comments.

### DIFF
--- a/examples/websockets/src/main.rs
+++ b/examples/websockets/src/main.rs
@@ -1,15 +1,19 @@
 //! Example websocket server.
 //!
 //! Run the server with
-//!
 //! ```not_rust
-//! cargo run -p example-websockets
+//! cargo run -p example-websockets --bin example-websockets
+//! ```
+//!
+//! Run a browser client with
+//! ```not_rust
 //! firefox http://localhost:3000
 //! ```
 //!
-//! Alternatively you can run the rust client with
+//! Alternatively you can run the rust client (showing two
+//! concurrent websocket connections being established) with
 //! ```not_rust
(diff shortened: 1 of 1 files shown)


FrankReh:
Let me know if you would accept even more trivial changes to the comments to fix capitalization and punctuation, just to make them consistent for the two rust source files.

examples/websockets/src/main.rs
```diff
@@ -131,7 +134,7 @@ async fn handle_socket(mut socket: WebSocket, who: SocketAddr) {
 
     // Since each client gets individual statemachine, we can pause handling
     // when necessary to wait for some external event (in this case illustrated by sleeping).
-    // Waiting for this client to finish getting his greetings does not prevent other clients form
+    // Waiting for this client to finish getting their greetings does not prevent other clients from
```
jplatte, line 137: I think this should be "its".

FrankReh:
Also a handful of clippy warnings.

examples/websockets/src/main.rs
```diff
@@ -63,8 +67,7 @@ async fn main() {
                     )
                 }),
         )
-        // routes are matched from bottom to top, so we have to put `nest` at the
-        // top since it matches all routes
+        // routes are matched from bottom to top
```
davidpdrsn, line 70: This is no longer true actually and haven't been for a while. So we can just remove this comment.

```suggestion
```

davidpdrsn (approved):
Thanks!

PR #1711, opened 2023-01-20
Change `Router::nest` to flatten the routes

davidpdrsn:
This changes `Router::nest` to "flatten" the routes. By flattening I mean just
moving the routes from one router to another by calling
`.route(nest_path, nested_endpoint)`. That means there aren't any `Router`s
nested inside each other and instead we just have one flattened `Router` which
performs better.

The main tricky part was fallback inheritance. The main reason we didn't flatten
routes in 0.6 previously was specifically because of fallback inheritance.

The solution I went for here was to make an internal `PathRouter` type. It has
most routing functions from `Router`, except fallbacks. A `Router` then
contains two `PathRouter`s: One for the actual routes and another for
fallbacks. Then when you add a fallback to the root router that adds `/` and
`/*` routes to the fallback `PathRouter`. Nesting is then pretty straight
forward: We just nest the routes from each `PathRouter`.

Fixes https://github.com/tokio-rs/axum/issues/1624
Fixes https://github.com/tokio-rs/axum/issues/1441

## TODO

- [x] How to handle nested fallbacks
- [x] Update tests for `MatchedPath`
- [x] Changelog

Don't believe there are any docs changes to make.


axum/src/routing/route.rs
```diff
@@ -171,16 +168,6 @@ where
...
 
-        if res
-            .extensions()
-            .get::<AlreadyPassedThroughRouteFuture>()
-            .is_some()
-        {
-            return Poll::Ready(Ok(res));
-        } else {
-            res.extensions_mut().insert(AlreadyPassedThroughRouteFuture);
-        }
```
davidpdrsn, line 182: tbh I'm not totally sure why everything still worked after removing this. Doesn't seem related to this change because it also works on `main` if you remove this. So must be something else that changed 🤷 

https://github.com/tokio-rs/axum/pull/755 did add a regression test and I also tried the repro from https://github.com/tokio-rs/axum/issues/747. Both worked correctly.

davidpdrsn:
@jplatte I finally got fallback inheritance working properly 🎉

axum/src/routing/mod.rs
```diff
@@ -744,3 +721,30 @@ fn traits() {
     use crate::test_helpers::*;
     assert_send::<Router<(), ()>>();
 }
+
+fn path_for_nested_route<'a>(prefix: &'a str, path: &'a str) -> Cow<'a, str> {
+    debug_assert!(prefix.starts_with('/'));
+    debug_assert!(path.starts_with('/'));
+
+    if prefix.ends_with('/') {
+        format!("{prefix}{}", path.trim_start_matches('/')).into()
```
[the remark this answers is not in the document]
davidpdrsn: Like if you do `.nest("////", _)`?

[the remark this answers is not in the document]
jplatte: Yeah, that or `.nest("///foo/", _)`.

axum/src/routing/mod.rs
```diff
@@ -44,25 +42,32 @@ pub use self::method_routing::{
...
 };
 
+macro_rules! panic_on_err {
+    ($expr:expr) => {
+        match $expr {
+            Ok(x) => x,
+            Err(err) => panic!("{err}"),
+        }
+    };
+}
```
jplatte, line 52: This is just `.unwrap()`, no?

axum/src/routing/mod.rs
```diff
@@ -420,86 +290,42 @@ where
...
-
-                match endpont {
-                    Endpoint::MethodRouter(method_router) => {
-                        method_router.call_with_state(req, state)
+                match self.fallback_router.call_with_state(req, state) {
+                    Ok(future) => future,
+                    Err((_req, _state)) => {
+                        // the default fallback added in `Router::new` matches
+                        // everything
+                        unreachable!()
```
jplatte, line 323:
```suggestion
                        unreachable!(
                            "the default fallback added in `Router::new` \
                             matches everything"
                        )
```

axum/src/extract/matched_path.rs
```diff
@@ -262,6 +282,23 @@ mod tests {
         assert_eq!(res.status(), StatusCode::OK);
     }
 
+    #[tokio::test]
+    async fn can_extract_nested_matched_path_in_middleware_via_extension_using_nest() {
```
jplatte, line 286: 🙌🏼

axum/src/routing/mod.rs
davidpdrsn: Almos:

```
# unwrap
❯ cargo run
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: E', src/main.rs:5:7
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

# unwrap
❯ cargo run
thread 'main' panicked at 'something went wrong', src/main.rs:7:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I prefer the latter.

`unwrap` also prints the debug representation.

axum/src/routing/mod.rs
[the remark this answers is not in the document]
davidpdrsn: You've nerd sniped me 😜 I'll make a note and do some tests.

axum/src/routing/not_found.rs
```diff
@@ -29,6 +29,7 @@ where
     }
 
     fn call(&mut self, _req: Request<B>) -> Self::Future {
+        println!("NotFound hit");
```
yanns, line 32: Should this stay?

MidasLamb:
Thanks for this! I think this is responsible for fixing an issue we had where the telemetry we had for nested routers didn't get their `http.route` set properly

davidpdrsn:
Yep that most likely https://github.com/tokio-rs/axum/issues/1441

PR #1713, opened 2023-01-20
Add `RawPathParams`

davidpdrsn:
For library authors that are building on top of axum it might be useful to get the raw path parameters, without deserializing, and thus saving some allocations.


jplatte (approved):
Really good idea!

davidpdrsn:
Credit goes to @carllerche for this idea :)

PR #1736, opened 2023-02-07
Document gotchas with `impl IntoResponse`

davidpdrsn:
We sometimes hear from people who are confused by why `impl IntoResponse` doesn't work.

Figured it makes sense to document. That way we can link to this when we get questions about it :P

### DIFF
--- a/axum/src/docs/response.md
+++ b/axum/src/docs/response.md
@@ -4,6 +4,7 @@ Types and traits for generating responses.
 
 - [Building responses](#building-responses)
 - [Returning different response types](#returning-different-response-types)
+- [Regarding `impl IntoResponse`](#regarding-impl-intoresponse)
 
 # Building responses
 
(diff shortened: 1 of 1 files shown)


axum/src/docs/response.md
```diff
@@ -218,6 +219,108 @@ fn something_else() -> bool {
...
+    # Ok(())
+    // ...
+}
+```
+
+This is because `?` supports using the [`From`] trait to convert to a different
+error type but it doesn't know which type to convert to, because we only
+specified `impl IntoResponse` as the return type.
+
+`Result<impl IntoResponse, impl IntoResponse>` unfortunately doesn't work either:
```
jplatte, line 289: It's not like it never works, just in *some* common cases such as the example where the second `impl IntoResponse` can't be inferred. Maybe that can be clarified.

jplatte (approved):
Looks good, just one nitpick

axum/src/docs/response.md
davidpdrsn: Thats true. I've tweaked the wording a bit.

Issue #1740, opened 2023-02-10
Add Query Parameter functionality to TypedPath


erlandsona:
- [x] I have looked for existing issues (including closed) about this

### Motivation

There are a number of frameworks that include the concept of "typed" or "checked" URL generation for use in templating systems returning Html specifically in like links `<a href={route_for_path}` or form tags.

See prior art:
In Rust:
https://rocket.rs/v0.5-rc/guide/upgrading/#more-typed-uris
In Haskell (servant)
https://docs.servant.dev/en/stable/cookbook/generic/Generic.html#links
As an example
```hs
-- Type Directed API derivation (client, server, links, swagger, et al)
data Site_ r = Site_
  { _index :: r :- Get '[HTML] Html
  , _smile :: r :- "smile" :> Get '[HTML] Html
  , _wow   :: r :- "wow" :> Capture "stuff" Int :> Get '[HTML] Html
  , static :: r :- Raw
  }
  deriving Generic

smile :: Html
smile =
  btn
    -- Typesafe Routing via Servant Links!
    ! hxGet (_wow route 1) -- _wow :: Site_ -> Int -> String basically.
    ! hxSwap "outerHTML"
    $ "Click me :)"
```

### Proposal

As I posted on Discord there's a few directions I think this could go in Axum:
Context / Problem description:
```rust
    // Old Way w/ Format string can go stale super easily.
    let formated = format!(r#"/colleagues/social.html?locked={}"#, (!opts.show_locked).to_string());
```
Hacked together approach leveraging TypedPath
```rust
    // Newer-ish way is less susceptable to the pathname being changed because it's compile checked here.
    let url = SocialHtmlPath.to_string();
    let params = serde_urlencoded::to_string(Params {
        locked: Some(!opts.show_locked),
    })
    .unwrap();
    let url_and_params = url + "?" + &params;
```

But _IDEALLY_ I'm thinking of being able to use a helper function like what I've defined for my codebase.
```rust
// Look mah, sorta type-safe link generation!
fn path_and_params<P: TypedPath, Q: Serialize>(p: P, q: Q) -> String {
    p.to_string()
        + &serde_urlencoded::to_string(q)
            .map(|q| "?".to_string() + &q)
            .unwrap_or("".to_string())
}
// And later in the rsx! template
<button
    hx-get={
        // Look Mah! Type Safe links!
[...]

PR #1744, opened 2023-02-11
Add `TypedPath::with_query_params`

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1740

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -32,7 +32,7 @@ json-lines = [
 protobuf = ["dep:prost"]
 query = ["dep:serde", "dep:serde_html_form"]
 spa = ["tower-http/fs"]
-typed-routing = ["dep:axum-macros", "dep:serde", "dep:percent-encoding"]
+typed-routing = ["dep:axum-macros", "dep:serde", "dep:percent-encoding", "dep:serde_urlencoded", "dep:form_urlencoded"]
 
 [dependencies]
 axum = { path = "../axum", version = "0.6.0", default-features = false }
@@ -55,6 +55,8 @@ prost = { version = "0.11", optional = true }
 serde = { version = "1.0", optional = true }
 serde_html_form = { version = "0.1", optional = true }
 serde_json = { version = "1.0.71", optional = true }
+serde_urlencoded = { version = "0.7", optional = true }
+form_urlencoded = { version = "1.1.0", optional = true }
 tokio-stream = { version = "0.1.9", optional = true }
 tokio-util = { version = "0.7", optional = true }
 
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -50,13 +50,13 @@ tower-service = "0.3"
 # optional dependencies
 axum-macros = { path = "../axum-macros", version = "0.3.1", optional = true }
 cookie = { package = "cookie", version = "0.16", features = ["percent-encode"], optional = true }
+form_urlencoded = { version = "1.1.0", optional = true }
 percent-encoding = { version = "2.1", optional = true }
 prost = { version = "0.11", optional = true }
 serde = { version = "1.0", optional = true }
 serde_html_form = { version = "0.1", optional = true }
 serde_json = { version = "1.0.71", optional = true }
 serde_urlencoded = { version = "0.7", optional = true }
-form_urlencoded = { version = "1.1.0", optional = true }
 tokio-stream = { version = "0.1.9", optional = true }
 tokio-util = { version = "0.7", optional = true }
 
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -32,7 +32,7 @@ json-lines = [
 protobuf = ["dep:prost"]
 query = ["dep:serde", "dep:serde_html_form"]
 spa = ["tower-http/fs"]
-typed-routing = ["dep:axum-macros", "dep:serde", "dep:percent-encoding", "dep:serde_urlencoded", "dep:form_urlencoded"]
+typed-routing = ["dep:axum-macros", "dep:serde", "dep:percent-encoding", "dep:serde_html_form", "dep:form_urlencoded"]
 
 [dependencies]
 axum = { path = "../axum", version = "0.6.0", default-features = false }
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -56,7 +56,6 @@ prost = { version = "0.11", optional = true }
 serde = { version = "1.0", optional = true }
 serde_html_form = { version = "0.1", optional = true }
 serde_json = { version = "1.0.71", optional = true }
-serde_urlencoded = { version = "0.7", optional = true }
 tokio-stream = { version = "0.1.9", optional = true }
 tokio-util = { version = "0.7", optional = true }
 
--- a/axum-extra/src/routing/typed.rs
+++ b/axum-extra/src/routing/typed.rs
@@ -386,20 +386,13 @@ impl_second_element_is!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, 
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::routing::TypedPath;
+    use serde::Deserialize;
 
+    #[derive(TypedPath, Deserialize)]
+    #[typed_path("/users/:id")]
     struct UsersShow {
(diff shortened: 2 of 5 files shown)


axum-extra/src/routing/typed.rs
```diff
@@ -229,6 +232,88 @@ pub trait TypedPath: std::fmt::Display {
...
+        let mut urlencoder = form_urlencoded::Serializer::new(&mut out);
+        self.params
+            .serialize(serde_urlencoded::ser::Serializer::new(&mut urlencoder))
+            .unwrap_or_else(|err| {
+                panic!(
+                    "failed to URL encode value of type `{}`: {}",
+                    type_name::<T>(),
+                    err
+                )
+            });
```
davidpdrsn, line 304: This panic is a little unfortunate but maybe its okay 🤔

axum-extra/src/routing/typed.rs
```diff
@@ -295,3 +382,63 @@ impl_second_element_is!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13);
...
+        // according to [the spec] starting the params with `?&` is allowed specifically:
+        //
+        // > If bytes is the empty byte sequence, then continue.
+        //
+        // [the spec]: https://url.spec.whatwg.org/#urlencoded-parsing
+        assert_eq!(uri, "/users/1?&foo=foo&bar=123&baz=true");
+    }
+
+    #[test]
+    fn with_params_called_multiple_times() {
```
erlandsona, line 431: We should also address how query params are merged? EG:
```rust
 let path = UsersShow { id: 1 }
            .with_query_params(Params {
                foo: "foo",
                bar: 123,
                baz: true,
            })
            .with_query_params([("foo", 1337)]); // does this override `foo` or get merged into a `foo[]="foo"&foo[]=1337`?
```

axum-extra/src/routing/typed.rs
erlandsona: This only happens in the context of debugging tho right? Like we wouldn't expect this to happen in a deployment at runtime (unless a dev left a debug statement in the prod release right?) and they should be using like tracing & tracing-subscribers, right?

erlandsona: This only happens in the context of debugging tho right? Like we wouldn't expect this to happen in a deployment at runtime (unless a dev left a debug statement in the prod release right?) and they should be using like tracing & tracing-subscribers, right?

erlandsona (approved):
This is what I was requesting :raised_hands: :bow: lookin' forward to the merge and tinkerin' with it myself! Much better than the helper functions I've been cobbling together!

axum-extra/src/routing/typed.rs
erlandsona: The bit that's confusing here is _when_ does a `serde_urlencoded` return an `Err("how does it error")`?
Wouldn't `encode("") -> ""`? Does it bail on certain unicode characters or something?

To me a nicer failure mode might be to default to the non-fallable `path.to_string()` eg: if we hit this error maybe just return `self` unmodified? :thinking: 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9457bd28f9ac8cdd5953848e36c76553
image


Okay worked some of this out for myself. Given the failure mode of this lib, I feel like maybe serde_urlencoded is maybe _not_ the best solution. 

It shouldn't take as input things that can't be turned into a url as in the above?

Can we not make a better function that only takes structures that are able to be serialized into the url spec properly so the encoder can just return `Uri` or `String` instead of the `Result<Uri, Wtf>`?

erlandsona: The bit that's confusing here is _when_ does a `serde_urlencoded` return an `Err("how does it error")`?
Wouldn't `encode("") -> ""`? Does it bail on certain unicode characters or something?

To me a nicer failure mode might be to default to the non-fallable `path.to_string()` eg: if we hit this error maybe just return `self` unmodified? :thinking: 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9457bd28f9ac8cdd5953848e36c76553
image


Okay worked some of this out for myself. Given the failure mode of this lib, I feel like maybe serde_urlencoded is maybe _not_ the best solution. 

It shouldn't take as input things that can't be turned into a url as in the above?

Can we not make a better function that only takes structures that are able to be serialized into the url spec properly so the encoder can just return `Uri` or `String` instead of the `Result<Uri, Wtf>`?

axum-extra/src/routing/typed.rs
```diff
@@ -295,3 +382,63 @@ impl_second_element_is!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13);
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    struct UsersShow {
+        id: i32,
+    }
+
+    // can't derive `TypedPath` because that references `::axum_extra` :(
+    impl fmt::Display for UsersShow {
```
erlandsona, line 395: Can this test module not be moved into a separate file where both could be imported without creating a cycle?

davidpdrsn: No. Its not because there is a cycle but because the generated code contains `::axum_extra::routing::TypedPath`, i.e. it access axum_extra from the root. I don't think its a big deal for this test.

axum-extra/src/routing/typed.rs
davidpdrsn: Multiple values for the same query is valid, in fact that is what HTML forms do. Hence the existence of https://docs.rs/serde_html_form/latest/serde_html_form/index.html#

axum-extra/src/routing/typed.rs
erlandsona: 🤷👍

axum-extra/src/routing/typed.rs
erlandsona: 🤘

axum-extra/src/routing/typed.rs
```diff
@@ -229,6 +232,90 @@ pub trait TypedPath: std::fmt::Display {
...
+    T: Serialize,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut out = self.path.to_string();
+        if !out.contains('?') {
+            out.push('?');
+        }
+        let mut urlencoder = form_urlencoded::Serializer::new(&mut out);
+        self.params
+            .serialize(serde_urlencoded::ser::Serializer::new(&mut urlencoder))
```
jplatte, line 299: Why `serde_urlencoded` over `serde_html_form` here?

axum-extra/src/routing/typed.rs
jplatte: We can solve this by adding `extern crate self as axum_extra;` to the crate root. Then the test can use the proc-macro and will be closer to a real-world scenario.

davidpdrsn: TIL!

axum-extra/src/routing/typed.rs
davidpdrsn: No reason, just forgot.

jplatte (approved):

axum-extra/src/routing/typed.rs
davidpdrsn: I'll refer to our discussion on the issue (https://github.com/tokio-rs/axum/issues/1740). This is because of how serde works and not serde_urlencoded's fault.

axum-extra/Cargo.toml
```diff
@@ -50,11 +50,13 @@ tower-service = "0.3"
 # optional dependencies
 axum-macros = { path = "../axum-macros", version = "0.3.1", optional = true }
 cookie = { package = "cookie", version = "0.16", features = ["percent-encode"], optional = true }
+form_urlencoded = { version = "1.1.0", optional = true }
 percent-encoding = { version = "2.1", optional = true }
 prost = { version = "0.11", optional = true }
 serde = { version = "1.0", optional = true }
 serde_html_form = { version = "0.1", optional = true }
 serde_json = { version = "1.0.71", optional = true }
+serde_urlencoded = { version = "0.7", optional = true }
```
jplatte, line 59:
```suggestion
```

jplatte (changes requested):

jplatte (approved):

erlandsona:
Hot dog! 🙌

PR #1745, opened 2023-02-11
Avoid unnecessary clones of state in MethodRouter

semicoleon:
## Motivation

I noticed that a state type was getting cloned a lot more than I expected it to be on a server with just two endpoints. This turned out to be due to MethodRouter cloning the state for every single possible method endpoint, even those which didn't exist. Since it's relatively uncommon for every single route to have a handler for every single HTTP method, this is just wasted work.

Of course generally state will be wrapped in something like an Arc, so the amount of wasted work will be small, but reducing the number of clones is easy enough to justify here I think.

## Solution

MethodEndpoint is private, so changing its with_state method to take a reference it can clone lazily is easy and not a breaking change. This still results in 4 clones for 2 endpoints (one for each MethodRouter, and one for each MethodEndpoint), but that's much better than the 18 I was seeing previously

### DIFF
--- a/axum/src/routing/method_routing.rs
+++ b/axum/src/routing/method_routing.rs
@@ -734,14 +734,14 @@ where
     /// Provide the state for the router.
     pub fn with_state<S2>(self, state: S) -> MethodRouter<S2, B, E> {
         MethodRouter {
-            get: self.get.with_state(state.clone()),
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
Great find! Thanks 😃

PR #1746, opened 2023-02-11
Upgrade dependencies

jplatte:
Changelogs:

- https://github.com/jplatte/serde_html_form/blob/main/CHANGELOG.md#020
- base64 doesn't have one


davidpdrsn (commented):
Should we write something in axum-extra's changelog since its a breaking change?

jplatte:
What's a breaking change here? None of these are public dependencies, right?

jplatte:
Undid the cookie upgrade so it's not a breaking change.

davidpdrsn (approved):
I'm okay with shipping a breaking change to axum-extra. I can do a separate PR.

Issue #1737, opened 2023-02-08
#[derive(FromRef)] with fields that are Copy generates Clippy warnings


tonyfinn:
- [X] I have looked for existing issues (including closed) about this

### Version

```
axum v0.6.4
    ├── axum-core v0.3.2
    ├── axum-macros v0.3.2 (proc-macro)
```

### Crates

axum-macros v0.3.2

### Description

Similar to https://github.com/tokio-rs/axum/issues/1674 but this time with types implementing `Copy` instead of simple references. If you `#[derive(FromRef)]` on a struct containing fields which are custom types implementing `Copy`, you get a clippy warning.

```rust
use axum::extract::FromRef;

#[derive(Debug, Clone, Copy)]
enum SimpleEnum {
    A
}

#[derive(Clone, Debug, FromRef)]
struct ForState {
    enu: SimpleEnum,
    text: String
}

fn main() {
    let x = ForState {
        enu: SimpleEnum::A,
        text: "A".to_string(),
    };
    println!("{x:?}");
}
```

I would not expect this to have any clippy warnings, and in fact if you remove the FromRef automatic derivation there is none. However, if you include that derive, you get the following warning:

```
warning: using `clone` on type `SimpleEnum` which implements the `Copy` trait
  --> src/main.rs:10:10
   |
10 |     enu: SimpleEnum,
   |          ^^^^^^^^^^ help: try removing the `clone` call: `enu: SimpleEnum`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy
   = note: `#[warn(clippy::clone_on_copy)]` on by default
```

Using rust 1.67 and `clippy 0.1.67 (fc594f1 2023-01-24)`

jplatte:
Note to anybody wanting to work on this: It probably makes sense to just fix this by adding `#[allow(clippy::clone_on_copy)]` to the generated output. proc-macro's don't have type information, so it's not possible to detect from the macro code whether any types of the struct fields implement `Copy` or not.

PR #1749, opened 2023-02-12
[Fix] Allow `clone_on_copy` for `FromRef`

JulesGuesnon:
There was a clippy warning if you derived a struct with `FromRef` that contains a type that derives `Copy` 

## Solution

As suggested in the issue, I added `#[allow(clippy::clone_on_copy)]` to the macro output

Fixes: #1737


davidpdrsn (approved):
Thanks!

PR #1750, opened 2023-02-12
Release axum, axum-extra, and axum-macros

davidpdrsn:


axum/Cargo.toml
```diff
@@ -52,7 +52,7 @@ tower-layer = "0.3.2"
 tower-service = "0.3"
 
 # optional dependencies
-axum-macros = { path = "../axum-macros", version = "0.3.3", optional = true }
+axum-macros = { path = "../axum-macros", version = "0.3.4", optional = true }
```
jplatte, line 55: Hm, didn't we use to have an exact version dependency here? I guess it's harder with there being two crates that depend on this, but as-is we must be careful to never make new versions of axum-macros rely on things from axum / axum-extra that were introduced in a patch release of the latest minor (breaking) one. (i.e. we need to keep in mind that axum-macros 0.3.4 can be used with axum 0.6.0 and axum-extra 0.4.0)

jplatte (approved):

axum/Cargo.toml
davidpdrsn: > Hm, didn't we use to have an exact version dependency here?

Don't think so 🤔 but yeah thats true.

PR #1751, opened 2023-02-12
Remove `B` type param

davidpdrsn:
Since the beginning all axum services have been generic over the request body type, the `B` in `Router<S, B>`. However I've been thinking that I'd like to remove that for a few reasons:

- https://github.com/tokio-rs/axum/issues/1110 is the biggest reason. If `Router` wasn't generic over the request body then `.layer(RequestBodyLimit::new(...))` wouldn't change the router type and it would still be `Router<S>`. I quite like that.
- Its one less for new users to learn and worry about.

The core idea is:

```rust
impl<B> Service<Request<B>> for Router<()>
where
    B: HttpBody<Data = bytes::Bytes> + Send + 'static,
    B::Error: Into<axum_core::BoxError>,
{
    fn call(&mut self, req: Request<B>) -> Self::Future {
        let req = req.map(Body::new);
        self.call_with_state(req, ())
    }
}
```

I.e. `Router` implements `Service` for any `B` but immediately converts that into a `Body`, rather than carrying the `B` type forward.

The performance impact should be negligible since `Body` prevents double boxing (https://github.com/tokio-rs/axum/blob/v0.7.0/axum-core/src/body.rs#L103)

I have changed all the code in a separate branch but its quite a lot of changes so this PR contains the core changes that everything else follow from. I think we should discuss the overall change here and I'll make smaller PRs that are easier to review and we can merge them into this branch.

## TODO

- [ ] changelog


jplatte (approved):
Seems like a good idea. I feel like it might be worth a review from somebody closer to tower / hyper though.

davidpdrsn:
Hm I think we need to find a solution for methods from `ServiceExt` not getting inferred. Users will run into it in tests:

```rust
    Checking example-testing v0.1.0 (/Users/david.pedersen/code/axum/examples/testing)
error[E0698]: type inside `async` block must be known in this context
   --> testing/src/main.rs:199:28
    |
199 |         let response = app.ready().await.unwrap().call(request).await.unwrap();
    |                            ^^^^^ cannot infer type for type parameter `B`
    |
```

Having to fully qualify this is gonna be very tedious. I'll try and think about it 🤔

axum/src/routing/mod.rs
```diff
@@ -498,6 +499,66 @@ where
...
+    ///     routing::get,
+    ///     http::Request,
+    ///     body::Body,
+    /// };
+    /// use tower::{Service, ServiceExt};
+    ///
+    /// # async fn async_main() -> Result<(), Box<dyn std::error::Error>> {
+    /// let mut router = Router::new().route("/", get(|| async {}));
+    /// let request = Request::new(Body::empty());
+    /// let response = router.as_service().ready().await?.call(request).await?;
```
jplatte, line 554: So this is still generic, just here the `B` can be inferred? I'm a little confused. Do you understand why the compiler accepts this, but not the `compile_fail` example?

davidpdrsn: > So this is still generic, just here the B can be inferred?

Yes

> Do you understand why the compiler accepts this, but not the compile_fail example?

No. I suppose it's a limitation (maybe bug?) in the type inference.

PR #1759, opened 2023-02-16
Add shared mutable state documentation

ThanePatrol:
## Motivation
I spent a lot of time trying to find documentation for creating a mutable state. I eventually found discussion #629 

## Solution

Added documentation for creating mutable state and wrote a small trivial example using a `Vec<String>`

### DIFF
--- a/axum/src/docs/routing/with_state.md
+++ b/axum/src/docs/routing/with_state.md
@@ -105,14 +105,6 @@ receives. That means it is not suitable for holding state derived from a
 request, such as authorization data extracted in a middleware. Use [`Extension`]
 instead for such data.
 
-# Mutable state
-
-As state is used across routes it is not mutable by default. To mutate a state
-you will need to use an `Arc<Mutex>` or similar. Accessing mutable IO such as 
-a database connection requires an async mutex or similar. See the 
-[tokio discussion on sync vs async mutex](https://docs.rs/tokio/1.25.0/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use). 
-
-
 # What `S` in `Router<S>` means
 
 `Router<S>` means a router that is _missing_ a state of type `S` to be able to
--- a/axum/src/docs/routing/with_state.md
+++ b/axum/src/docs/routing/with_state.md
@@ -108,37 +108,11 @@ instead for such data.
 # Mutable state
 
 As state is used across routes it is not mutable by default. To mutate a state
-you will need to use an `Arc<Mutex>` or similar. Note if accessing a mutable IO such as 
-a database connection using an async mutex is required. See 
-<a href="https://docs.rs/tokio/1.25.0/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use">this</a>
-discussion for a more detailed conversation on the differences between an async mutex 
-and synchronous mutex. 
+you will need to use an `Arc<Mutex>` or similar. Accessing mutable IO such as 
+a database connection requires an async mutex or similar. See the
+<a href="https://docs.rs/tokio/1.25.0/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use">
+tokio discussion on sync vs async mutex</a>. 
 
-For example:
-
-```rust
-# use axum::{Router, routing::get, extract::State, response::IntoResponse, http::StatusCode};
-use std::sync::{Arc, Mutex};
-// This AppState consists of a vec of mutable strings 
-#[derive(Clone)]
-struct AppState { mutable_state: Vec<String>, }
-async fn mutate_state(State(state): State<Arc<Mutex<AppState>>>) -> impl IntoResponse {
-    // Wait to get the state
-    let mutable_state = state.lock();
-    // Now the AppState can be mutated  
-    StatusCode::OK
-}
-// A mutable AppState
-let state = Arc::new(Mutex::new(AppState { mutable_state: Default::default(), } ));
-let router = Router::new()
-    .route("/", get(mutate_state))
-    .with_state(state);
-# async {
-axum::Server::bind( & "0.0.0.0:3000".parse().unwrap())
-    .serve(router.into_make_service())
-    .await;
-# };
-```
 
 # What `S` in `Router<S>` means
 
--- a/axum/src/docs/routing/with_state.md
+++ b/axum/src/docs/routing/with_state.md
@@ -105,6 +105,40 @@ receives. That means it is not suitable for holding state derived from a
 request, such as authorization data extracted in a middleware. Use [`Extension`]
 instead for such data.
 
+# Mutable state
+
+As state is used across routes it is not mutable by default. To mutate a state
+you will need to use an `Arc<Mutex>` or similar. Note that the Mutex must
+support async locks.
+
+For example:
+
+```rust
+# use axum::{Router, routing::get, extract::State, response::IntoResponse, http::StatusCode};
+// Note the use of the tokio Mutex
+use tokio::sync::Mutex;
+use std::sync::Arc;
+// This AppState consists of a vec of mutable strings 
+#[derive(Clone)]
+struct AppState { mutable_state: Vec<String>, }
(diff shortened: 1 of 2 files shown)


axum/src/docs/routing/with_state.md
```diff
@@ -105,6 +105,40 @@ receives. That means it is not suitable for holding state derived from a
 request, such as authorization data extracted in a middleware. Use [`Extension`]
 instead for such data.
 
+# Mutable state
+
+As state is used across routes it is not mutable by default. To mutate a state
+you will need to use an `Arc<Mutex>` or similar. Note that the Mutex must
+support async locks.
```
FSMaxB, line 112: You don't actually need an asynchronous mutex. Synchronous is fine and even recommended. An async mutex is only needed if the lock is held across `.await` points.

See also https://docs.rs/tokio/1.25.0/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

davidpdrsn: Yes. There are many caveats when it comes to shared mutable state in async. Therefore I'm hesitant to include examples of it in our docs since users might see it as the "axum approved way" without understanding what they're getting themselves into.

ThanePatrol: Thanks for the feedback and the link. I mistakenly made the assumption about blocking mutex being required as my own project did require it.

davidpdrsn (commented):
I still don't feel this is appropriate to have in axum's docs 🤷

jplatte:
Maybe we should link to tokio's "Which kind of mutex should you use?" docs though?

davidpdrsn:
Yeah that'd be fine with me.

axum/src/docs/routing/with_state.md
```diff
@@ -105,6 +105,15 @@ receives. That means it is not suitable for holding state derived from a
 request, such as authorization data extracted in a middleware. Use [`Extension`]
 instead for such data.
 
+# Mutable state
+
+As state is used across routes it is not mutable by default. To mutate a state
+you will need to use an `Arc<Mutex>` or similar. Accessing mutable IO such as 
+a database connection requires an async mutex or similar. See the
+<a href="https://docs.rs/tokio/1.25.0/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use">
+tokio discussion on sync vs async mutex</a>.
```
jplatte, line 114: Why an html link rather than markdown syntax?

ThanePatrol: Apologies. I'm more familiar with HTML than markdown. Will update commit

axum/src/docs/routing/with_state.md
```diff
@@ -105,6 +105,14 @@ receives. That means it is not suitable for holding state derived from a
 request, such as authorization data extracted in a middleware. Use [`Extension`]
 instead for such data.
 
+# Mutable state
+
+As state is used across routes it is not mutable by default. To mutate a state
+you will need to use an `Arc<Mutex>` or similar. Accessing mutable IO such as 
+a database connection requires an async mutex or similar. See the 
+[tokio discussion on sync vs async mutex](https://docs.rs/tokio/1.25.0/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use).
```
davidpdrsn, line 113: This makes it sounds like sharing a database connection behind a mutex is common, which I don't think it is. Normally you share a connection pool which deals with all this internally, so users don't need a mutex.

You generally don't wanna share IO objects with mutexes, its a deadlock waiting to happen. In that case you wanna use [an actor](https://ryhl.io/blog/actors-with-tokio).

So I think we should phrase this differently somehow.

Vagelis-Prokopiou:
@ThanePatrol check the referenced pull request I crested #1777.
I am using a string, and I have added compiling example code.

ThanePatrol:
@Vagelis-Prokopiou Thanks for letting me know. I've share the repo with you and incorporated your changes.

Vagelis-Prokopiou:
This PR looks ready. Kindly waiting on the owners @davidpdrsn @jplatte.

axum/src/extract/state.rs
```diff
@@ -295,6 +295,44 @@ use std::{
 ///
 /// In general however we recommend you implement `Clone` for all your state types to avoid
 /// potential type errors.
+///
+/// # When mutable shared state is needed
+/// As state is used across routes it is not mutable by default. To mutate the state
+/// you will need to use an `Arc<Mutex>` or similar.
+/// You may have to use `tokio::sync::Mutex` instead of `std::sync::Mutex`,
+/// since `std::sync::Mutex` is not safe for use across threads, but this comes
```
FSMaxB, line 303: `std::sync::Mutex` **is** safe to use across threads. If you don't have multiple threads, there's even no point in using it at all.

The issue is using it across `.await` points because that is extremely likely to cause an intermittend deadlock, and that is an issue even on single threaded runtimes, no only multithreaded ones.

FSMaxB: By "using across await points" what I mean is: Keeping it locked across await points, holding a `MutexGuard` while `.await`ing.

Vagelis-Prokopiou: @FSMaxB fixed in commit 856ac49.

axum/src/extract/state.rs
```diff
@@ -300,9 +300,10 @@ use std::{
 /// As state is used across routes it is not mutable by default. To mutate the state
 /// you will need to use an `Arc<Mutex>` or similar.
 /// You may have to use `tokio::sync::Mutex` instead of `std::sync::Mutex`,
-/// since `std::sync::Mutex` is not safe for use across threads, but this comes
-/// with performance cost. See the
-// [tokio discussion on sync vs async mutex](https://docs.rs/tokio/1.25.0/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use).
+/// since `std::sync::Mutex` cannot be used across `.await` points,
```
jplatte, line 303:
```suggestion
/// since `std::sync::Mutex` should not be locked across `.await` points,
```

Vagelis-Prokopiou:
@jplatte suggested change applied.

davidpdrsn:
I had lots of nitpicky suggestions so I've rewritten it myself. @jplatte wanna give it a final look?

axum/src/extract/state.rs
```diff
@@ -295,6 +295,49 @@ use std::{
...
+/// # Shared mutable state
+///
+/// [As state is global within a `Router`][global] you can't directly get a mutable reference to
+/// the state.
+///
+/// The most basic solution is to use an `Arc<Mutex<_>>`. Which kind of mutex you need depends on
+/// your use case. See [the tokio docs] for more details.
+///
+/// Note that holding a locked `std::sync::Mutex` across `.await` points will result in `!Send`
+/// futures which are incompatible with axum. If you need to hold a mutex across `.await` points
```
jplatte, line 308:
```suggestion
/// futures which are incompatible with axum. If you need to hold a mutex across `.await` points,
```

axum/src/extract/state.rs
```diff
@@ -295,6 +295,49 @@ use std::{
...
+/// ## Example
+///
+/// ```
+/// use axum::{Router, routing::get, extract::State};
+/// use std::sync::{Arc, Mutex};
+///
+/// #[derive(Clone)]
+/// struct AppState {
+///     data: Arc<Mutex<String>>,
+/// }
```
jplatte, line 320: I'm a little worried that people will do fine-grained locking where more coarse locking is the right solution, based on this example. I'm not sure just showing an `Arc<Mutex<_>>` with everything inside locked at once is better either though...

jplatte (approved):

axum/src/extract/state.rs
Vagelis-Prokopiou: @jplatte  Maybe something like the following can be added, in order to make your point super explicit?
```
#[derive(Clone)]
struct AppState {
      // This is just an example usage, not a best practice.
      // Choose a setup that matches you use specific use case.
      data: Arc<Mutex<String>>,
}
```

PR #1761, opened 2023-02-16
Try `cargo public-api-crates` on CI

davidpdrsn:
Its a little [cargo command](https://github.com/davidpdrsn/cargo-public-api-crates) I'm experimenting with. Something I've wanted for a while. I wanna give it a shot here and see how its working.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -74,6 +74,22 @@ jobs:
     - name: cargo hack check
       run: cargo hack check --each-feature --no-dev-deps --all
 
+  cargo-public-api-crates:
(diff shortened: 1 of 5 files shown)


jplatte:
Interesting, does this temporarily rewrite the `Cargo.toml` to use the experimental public / private dependency feature of Cargo nightly?

davidpdrsn:
Nope. Though I'm really looking forward to that. Should try that as well.

It builds the docs with `--output-format json`, parses it with https://docs.rs/rustdoc-types/0.20.0/rustdoc_types/index.html, and checks all paths.

PR #1766, opened 2023-02-17
Require `StreamBody` to be infallible, add `TryStreamBody`

davidpdrsn:
I've been thinking some more about #1699 and I think this approach makes sense.

This changes `StreamBody` to always require the underlying stream to be infallible, i.e. it must yield `impl Into<Bytes>`. `Result`s are not allowed. That is because hyper will silently ignore body errors, so its kind of a footgun.

Additionally I think we can provide a `TryStreamBody` that allows `Result`s but also a callback to call on errors. Then we can use `TryStreamBody` in `JsonLines` and `AsyncReadBody` which don't have a good way to surface errors otherwise.

### DIFF
--- a/axum/src/body/stream_body.rs
+++ b/axum/src/body/stream_body.rs
@@ -1,15 +1,18 @@
+//! Streaming bodies.
+
 use crate::{
     body::{self, Bytes, HttpBody},
     response::{IntoResponse, Response},
-    BoxError, Error,
+    Error,
 };
 use futures_util::{
     ready,
-    stream::{self, TryStream},
+    stream::{self, Stream, TryStream},
 };
 use http::HeaderMap;
 use pin_project_lite::pin_project;
 use std::{
+    convert::Infallible,
     fmt,
     pin::Pin,
     task::{Context, Poll},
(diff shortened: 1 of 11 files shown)


axum/src/body/stream_body.rs
```diff
@@ -109,22 +107,178 @@ impl<S> fmt::Debug for StreamBody<S> {
...
+    /// # axum::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
+    /// # };
+    /// ```
+    ///
+    /// [`Stream`]: futures_util::stream::Stream
+    pub struct TryStreamBody<S, T = DefaultOnError> {
+        #[pin]
+        stream: SyncWrapper<S>,
+        on_error: T,
+    }
```
davidpdrsn, line 193: I don't love the name `TryStreamBody` here because it makes it sound the body produces errors, which it doesn't. Its the stream the might fail. Its really `(TryStream)Body` 😅

davidpdrsn: Hm maybe. Although there is a convention to use "try" (`TryFuture`, `TryStream`, etc)

jplatte:
I'm finding it hard to justify taking time for reviewing this – it's not my area of expertise and I have little free time to dedicate to axum at the moment. I think somebody else should review.

davidpdrsn:
I'll close this for now but added it to the 0.7.0 todo list. I think now that we have our own body type that implements stream directly we might not need these wrappers. Can instead have `impl Stream for Body` and `Body::from_stream`. Only question is how to deal with errors.

PR #1767, opened 2023-02-17
Add `MockConnectInfo`

davidpdrsn:
Using the `ConnectInfo` extractor during tests was a little tricky. You can't just call `Router::into_make_service_with_connect_info` and then send it requests, because that produces a `MakeService` and not a `Service<Request>`.

The solution was to set `ConnectInfo` in an extension yourself however that relies on implementation details which feels bad.

This introduces `MockConnectInfo` which basically does the same thing but now with a proper API, so the implementation details are encapsulated.


jplatte (approved):
Nice!

PR #1772, opened 2023-02-20
Remove `B` type param: axum-core

davidpdrsn:
This is the next step in removing the `B` type param as outlined in https://github.com/tokio-rs/axum/pull/1751. This PR updates axum-core.

The changes are entirely mechanical. Stop using `Request<B>` and use `Request<axum_core::body::Body>` instead.

I think a `type Request<B = axum_core::body::Body> = http::Request<B>` alias would also be nice, similarly to what we have for `Response`. I'll probably add that when https://github.com/tokio-rs/axum/pull/1751 is merged.

### DIFF
--- a/axum-core/src/extract/default_body_limit.rs
+++ b/axum-core/src/extract/default_body_limit.rs
@@ -41,7 +41,7 @@ use tower_layer::Layer;
 ///         post(|request: Request<Body>| async {}),
 ///     )
 ///     .layer(DefaultBodyLimit::max(1024));
-/// # let _: Router<()> = app;
+/// # let _: Router = app;
 /// ```
 ///
 /// ```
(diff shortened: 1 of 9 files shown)


axum-core/src/extract/default_body_limit.rs
```diff
@@ -41,7 +41,7 @@ use tower_layer::Layer;
 ///         post(|request: Request<Body>| async {}),
 ///     )
 ///     .layer(DefaultBodyLimit::max(1024));
-/// # let _: Router<(), _> = app;
+/// # let _: Router<()> = app;
```
jplatte, line 44: Does the state type of `Router` have a default actually? Seems like it would be nice in situations like this where you want to annotate a binding type.

axum-core/src/response/into_response.rs
```diff
@@ -119,9 +119,7 @@ use std::{
 ///
 /// // `MyBody` can now be returned from handlers.
 /// let app = Router::new().route("/", get(|| async { MyBody }));
-/// # async {
-/// # hyper::Server::bind(&"".parse().unwrap()).serve(app.into_make_service()).await.unwrap();
-/// # };
+/// # let _: Router = app;
```
jplatte, line 122: I guess that answers my question 😄

jplatte (approved):

axum-core/src/response/into_response.rs
davidpdrsn: Hehe yes. I ran a quick search and replace to catch at least some of them.

Issue #1770, opened 2023-02-19
Allow Header from 405 response is lost through service layers unless the route is `.nest`ed


taylor1791:
### Versions

axum 0.6.7
axum-core 0.3.2
axum-macros v0.3.4

### Prior Art
- [X] I have looked for existing issues (including closed) about this
Issues: #1525 #731 
PRs: #1526 #733 

### Platform

Linux hostname 5.15.94 # 1-NixOS SMP Tue Feb 14 18:18:05 UTC 2023 x86_64 GNU/Linux

### Description

While flattening some of my `.nest`ed routers, the router stopped responding with the Allow header on 405 responses. I simplified my code down to the following failing test.

```rust
#[crate::test]
async fn fail_1() {
    async fn hello() -> &'static str {
        "Hello World"
    }

    let middleware = tower::ServiceBuilder::new();
    let app = Router::<(), Body>::new().route("/hello", get(hello)).layer(middleware);
    let client = TestClient::new(app);

    let res = client.patch("/hello").send().await;
    assert_eq!(res.status(), StatusCode::METHOD_NOT_ALLOWED);
    assert_eq!(res.headers()[ALLOW], "GET,HEAD");
}
```

The interesting thing is that the following tests both pass!
```rust
#[crate::test]
async fn pass_1() {
    async fn hello() -> &'static str {
        "Hello World"
    }

    let app = Router::<(), Body>::new().route("/hello", get(hello));
    let client = TestClient::new(app);

    let res = client.patch("/hello").send().await;
    assert_eq!(res.status(), StatusCode::METHOD_NOT_ALLOWED);
    assert_eq!(res.headers()[ALLOW], "GET,HEAD");
}

#[crate::test]
async fn pass_2() {
    async fn hello() -> &'static str {
        "Hello World"
    }

    let middleware = tower::ServiceBuilder::new();
    let hello = Router::<(), Body>::new().route("/hello", get(hello));
    let app = Router::<(), Body>::new().nest("/", hello).layer(middleware);

    let client = TestClient::new(app);

    let res = client.patch("/hello").send().await;
    assert_eq!(res.status(), StatusCode::METHOD_NOT_ALLOWED);
    assert_eq!(res.headers()[ALLOW], "GET,HEAD");
}
```

[...]

PR #1773, opened 2023-02-20
Fix missing `Allow` when middleware are applied to `MethodRouter`

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1770

To be totally honest I don't understand why this fixes the issue. I also noticed some weirdness in this piece of the code in https://github.com/tokio-rs/axum/pull/1711/files#r1082406463. Removing it doesn't seem to break anything else, and the bug it was originally added to fix doesn't occur anymore.

The reason its pretty hard to follow I think is that we wrap things in `Route` quite a bit, which also wraps the response future in `RouteFuture`. So there are several layers of `RouteFuture` being called and previously mutating the response.

### DIFF
--- a/axum/src/routing/route.rs
+++ b/axum/src/routing/route.rs
@@ -155,9 +155,6 @@ where
 
     #[inline]
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        #[derive(Clone, Copy)]
-        struct AlreadyPassedThroughRouteFuture;
-
         let this = self.project();
 
         let mut res = match this.kind.project() {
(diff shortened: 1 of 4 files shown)


axum/src/routing/route.rs
```diff
@@ -155,9 +155,6 @@ where
 
     #[inline]
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        #[derive(Clone, Copy)]
-        struct AlreadyPassedThroughRouteFuture;
```
jplatte, line 159: Maybe dig up the PR that introduced this via git blame?

jplatte (approved):

axum/src/routing/route.rs
davidpdrsn: I did that [here](https://github.com/tokio-rs/axum/pull/1711/files#r1082406463)

> tbh I'm not totally sure why everything still worked after removing this. Doesn't seem related to this change because it also works on `main` if you remove this. So must be something else that changed 🤷 
> 
> https://github.com/tokio-rs/axum/pull/755 did add a regression test and I also tried the repro from https://github.com/tokio-rs/axum/issues/747. Both worked correctly.

and tested specifically the repro from the issue and everything worked fine. So something else must have changed that made `AlreadyPassedThroughRouteFuture` unnecessary.

davidpdrsn:
Oh yeah theres a failing tests. That interesting 😅

axum/src/routing/route.rs
jplatte: Ah, great 🙂

PR #1774, opened 2023-02-20
Remove `B` type param: `Router`, `MethodRouter`, `Handler`

davidpdrsn:
These are the interesting parts of axum. The rest is fairly trivial so I'll do a separate PR for those.

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -67,7 +67,7 @@ pub struct Router<S = ()> {
     fallback: Fallback<S>,
 }
 
-impl<S, B> Clone for Router<S, B> {
+impl<S> Clone for Router<S> {
     fn clone(&self) -> Self {
         Self {
             routes: self.routes.clone(),
@@ -77,9 +77,8 @@ impl<S, B> Clone for Router<S, B> {
     }
 }
 
-impl<S, B> Default for Router<S, B>
+impl<S> Default for Router<S>
 where
-    B: HttpBody + Send + 'static,
     S: Clone + Send + Sync + 'static,
 {
     fn default() -> Self {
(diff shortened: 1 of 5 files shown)


axum/src/routing/mod.rs
```diff
@@ -311,14 +305,13 @@ where
...
-        L::Service: Service<Request<NewReqBody>> + Clone + Send + 'static,
-        <L::Service as Service<Request<NewReqBody>>>::Response: IntoResponse + 'static,
-        <L::Service as Service<Request<NewReqBody>>>::Error: Into<Infallible> + 'static,
-        <L::Service as Service<Request<NewReqBody>>>::Future: Send + 'static,
-        NewReqBody: HttpBody + 'static,
+        L: Layer<Route> + Clone + Send + 'static,
+        L::Service: Service<Request<Body>> + Clone + Send + 'static,
+        <L::Service as Service<Request<Body>>>::Response: IntoResponse + 'static,
+        <L::Service as Service<Request<Body>>>::Error: Into<Infallible> + 'static,
+        <L::Service as Service<Request<Body>>>::Future: Send + 'static,
```
davidpdrsn, line 314: I was initially a little worried that requiring `L::Service: Service<Request<Body>>` wouldn't work for [`RequestBodyLimit`](https://docs.rs/tower-http/latest/tower_http/limit/struct.RequestBodyLimit.html#impl-Service%3CRequest%3CReqBody%3E%3E-for-RequestBodyLimit%3CS%3E) but `RequestBodyLimit` still implements `Service<Request<B>>` which matches `Service<Request<Body>>` so its all fine.

axum/src/routing/mod.rs
```diff
@@ -508,10 +500,7 @@ where
     }
 }
 
-impl<B> Router<(), B>
-where
-    B: HttpBody + Send + 'static,
-{
+impl Router<()> {
```
jplatte, line 503:
```suggestion
impl Router {
```

jplatte (approved):

PR #1775, opened 2023-02-20
Remove `B` type param: axum-extra

davidpdrsn:
Just mechanical changes.

### DIFF
--- a/axum-extra/src/extract/cookie/mod.rs
+++ b/axum-extra/src/extract/cookie/mod.rs
@@ -255,7 +255,7 @@ mod tests {
                     custom_key: CustomKey(Key::generate()),
                 };
 
-                let app = Router::<_>::new()
+                let app = Router::new()
                     .route("/set", get(set_cookie))
                     .route("/get", get(get_cookie))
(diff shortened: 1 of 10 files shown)


axum-extra/src/extract/cookie/mod.rs
```diff
@@ -255,7 +255,7 @@ mod tests {
                     custom_key: CustomKey(Key::generate()),
                 };
 
-                let app = Router::<_, Body>::new()
+                let app = Router::<_>::new()
```
jplatte, line 258: In expression position, generic parameter defaults don't apply so this can be

```suggestion
                let app = Router::new()
```

axum-extra/src/extract/cookie/mod.rs
```diff
@@ -352,7 +352,7 @@ mod tests {
             custom_key: CustomKey(Key::generate()),
         };
 
-        let app = Router::<_, Body>::new()
+        let app = Router::<_>::new()
```
jplatte, line 355:
```suggestion
        let app = Router::new()
```

jplatte (approved):

PR #1785, opened 2023-02-25
docs: Clarify that state parameters need to come before body too

robinst:
## Motivation

I used `axum` for the first time recently. I had a `state` parameter after `body` and got an inscrutable compile error (that I wrongly assumed was because of my complex return type) and it took me a good while to figure out what was wrong.

## Solution

I had read the section about order of extractors but in my mental model state wasn't extracting anything from the request. 

Mentioning it explicitly like I did in this PR would have helped me, and potentially others in the future.

### DIFF
--- a/axum/src/docs/extract.md
+++ b/axum/src/docs/extract.md
@@ -168,20 +168,26 @@ handler takes.
 For example
 
 ```rust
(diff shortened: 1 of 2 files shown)


davidpdrsn (approved):
Thanks!

PR #1787, opened 2023-02-25
update tokio dep to fix potential security vulnerability

spencerbart:
Update tokio version to fix potential security vulnerability.

https://deps.rs/crate/axum/0.6.8


davidpdrsn (commented):
Should probably bump it in axum-core as well as libraries might not depend on axum.

PR #1795, opened 2023-02-27
Update documentation for `FailedToDeserializeQueryString` response type

robertwayne:
I believe this was just an oversight from #1387.

Changes the stated return type from `422 Unprocessable Entity` to `400 Bad Request`.


davidpdrsn (approved):
Thanks!

PR #1800, opened 2023-03-02
Document the fact that `debug_handler` doesn't work within impl blocks

jesjos:
## Motivation

When upgrading from from axum 0.5.x to 0.6.x I encountered a breaking change to how the `debug_handler` macro works. 

In 0.5.x, an associated function (not a method) could be used with `debug_handler`. In 0.6.x this yields confusing compilation errors. 

As per @davidpdrsn on Discord: 

> ah yes I see why now. Basically in 0.6 FromRequest got a new (secret) type param M, which is used to work around some otherwise overlapping impls. But because of this we need to generate slightly different code in debug_handler, which doesn't work we're inside an impl would be good to document. 

## Solution

This PR attempts to document this behavior.

### DIFF
--- a/axum-macros/src/lib.rs
+++ b/axum-macros/src/lib.rs
@@ -542,14 +542,17 @@ pub fn derive_from_request_parts(item: TokenStream) -> TokenStream {
 ///
 /// # Limitations
 ///
-/// This macro does not work for associated functions — functions defined in an impl block.
+/// This macro does not work for associated functions — functions defined in an `impl` block that
+/// don't take `self`:
 ///
-/// ```
+/// ```compile_fail
 /// use axum::{debug_handler, extract::Path};
+///
 /// struct App {}
-/// #[debug_handler]
+///
 /// impl App {
-///     pub async fn handler(Path(_): Path(String)) {}
+///     #[debug_handler]
+///     async fn handler(Path(_): Path<String>) {}
 /// }
 /// ```
 ///
--- a/axum-macros/src/lib.rs
+++ b/axum-macros/src/lib.rs
@@ -540,6 +540,29 @@ pub fn derive_from_request_parts(item: TokenStream) -> TokenStream {
 /// }
 /// ```
 ///
+/// # Limitations
+///
+/// This macro does not work for associated functions — functions defined in an impl block.
+///
+/// ```
+/// use axum::{debug_handler, extract::Path};
+/// struct App {}
+/// #[debug_handler]
+/// impl App {
+///     pub async fn handler(Path(_): Path(String)) {}
+/// }
+/// ```
+///
+/// This will yield an error similar to this:
+///
+/// ```text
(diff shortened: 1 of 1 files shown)


axum-macros/src/lib.rs
```diff
@@ -540,6 +540,29 @@ pub fn derive_from_request_parts(item: TokenStream) -> TokenStream {
 /// }
 /// ```
 ///
+/// # Limitations
+///
+/// This macro does not work for associated functions — functions defined in an impl block.
```
davidpdrsn, line 545: I'm not sure everyone is gonna be familiar with what "associated functions" means. What do you think about clarifying it to say that it does work on methods that take `self`?

axum-macros/src/lib.rs
```diff
@@ -540,6 +540,29 @@ pub fn derive_from_request_parts(item: TokenStream) -> TokenStream {
...
+///
+/// ```
+/// use axum::{debug_handler, extract::Path};
+/// struct App {}
+/// #[debug_handler]
+/// impl App {
+///     pub async fn handler(Path(_): Path(String)) {}
+/// }
+/// ```
```
davidpdrsn, line 554:
```suggestion
/// ```compile_fail
/// use axum::{debug_handler, extract::Path};
///
/// struct App {}
///
/// impl App {
///     #[debug_handler]
///     pub async fn handler(Path(_): Path(String)) {}
/// }
/// ```
```

davidpdrsn (approved):

axum-macros/src/lib.rs
```diff
@@ -540,6 +540,32 @@ pub fn derive_from_request_parts(item: TokenStream) -> TokenStream {
 /// }
 /// ```
 ///
+/// # Limitations
+///
+/// This macro does not work for associated functions — functions defined in an `impl` block that
+/// don't take `self`:
```
jesjos, line 546:
```suggestion
/// This macro does not work for associated functions that aren't methods — functions defined in an `impl` block that
/// don't take `self`:
```

davidpdrsn: But an associated function is never a method 🤔

jesjos: I added this clarification @davidpdrsn — maybe I'm splitting hairs, but I think it pays to try and get it right

jesjos: https://doc.rust-lang.org/reference/items/associated-items.html#methods

> Associated functions whose first parameter is named self are called methods and may be invoked using the method call operator, for example, x.foo(), as well as the usual function call notation.

jplatte:
```suggestion
/// This macro does not work functions in an `impl` block that
/// don't have a `self` parameter:
```

jplatte: Oh, already merged 😅

PR #1801, opened 2023-03-02
Replace weird doc paragraph by compiler-checked must_use attribute

jplatte:
Very lazy fix for making it harder to accidentally discard the `Response` generated by `WebSocketUpgrade::on_upgrade`. Feel free to improve by prodiving a message (`#[must_use = "message"]` or restoring a more useful version of the docs I deleted.

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -282,7 +282,7 @@ impl<F> WebSocketUpgrade<F> {
 
     /// Finalize upgrading the connection and call the provided callback with
     /// the stream.
-    #[must_use]
+    #[must_use = "to setup the WebSocket connection, this response must be returned"]
     pub fn on_upgrade<C, Fut>(self, callback: C) -> Response
     where
         C: FnOnce(WebSocket) -> Fut + Send + 'static,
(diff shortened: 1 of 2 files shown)


zombiepigdragon:
I don't think it's necessary to remove the comment, especially without adding a reason to the `#must_use`. 

Also, is there anywhere else in Axum that this footgun can occur? I was planning to search for it and open a PR of my own later today before finding this, but since you've got a PR already it's probably worth checking out.

jplatte:
I found the paragraph I removed confusing. Otherwise I would not have removed it. Let's see what David thinks.

Also, I don't intend to update the PR to cover more places that could use a `#[must_use]`, so if you want to investigate, don't let the existence of this PR stop you :)

axum/src/extract/ws.rs
```diff
@@ -282,10 +282,7 @@ impl<F> WebSocketUpgrade<F> {
 
     /// Finalize upgrading the connection and call the provided callback with
     /// the stream.
-    ///
-    /// When using `WebSocketUpgrade`, the response produced by this method
-    /// should be returned from the handler. See the [module docs](self) for an
-    /// example.
+    #[must_use]
```
davidpdrsn, line 285: What do you think about

```suggestion
    #[must_use = "to setup the WebSocket connection this response must be returned"]
```

jplatte: Looks good, with extra comma :)

davidpdrsn (approved):

PR #1784, opened 2023-02-24
Remove `SpaRouter`

davidpdrsn:
`ServeDir` and `ServeFile` are easy enough to use that `SpaRouter` doesn't add much value.

PR #1802, opened 2023-03-02
Remove remnants of spa feature

jtdowney:
## Motivation

I saw that `SpaRouter` was removed in #1784, but the `spa` feature flag was still referenced in a few places, including doc comments.

## Solution

I cleaned up the remnants of the `spa` feature flag that I could find.


davidpdrsn (approved):
Thanks!

Also damnit... removing the feature requires another breaking release 😅

jtdowney:
Is it better to remove the comment and the example code and leave the feature for the next breaking release?

davidpdrsn:
Nah I don't think so. The whole point of axum-extra is that we can make breaking changes. [We just did one recently](https://github.com/tokio-rs/axum/blob/main/axum-extra/CHANGELOG.md#060-24-february-2022) but doing another one isn't the end of the world. Can probably do one this week or next.

davidpdrsn (approved):

PR #1804, opened 2023-03-03
Fixed small typo

sharkspear94:
## Motivation
``` rust
async fn handler(Path(p): Path<i32>) -> String {
    format!("{p}")
}
```
The above handler returns:

Invalid URL: Cannot parse value at index 0 with value "123123123123123" to a **i16**

It should return:

Invalid URL: Cannot parse value at index 0 with value "123123123123123" to a **i32**


jplatte (approved):
What a fun bug 😅

davidpdrsn:
Dunno what's up with GitHub actions but I ran the tests locally and they all pass. Thanks!

Issue #1803, opened 2023-03-02
dylib routing issue caused by RouteId


Curid:
- [x] I have looked for existing issues (including closed) about this

### Version

├── axum v0.6.9
│   ├── axum-core v0.3.2

### Platform

Linux

### Description

Adding routes to a router from a shared library causes routes to point to the wrong handlers.   I'm guessing [this](https://github.com/tokio-rs/axum/blob/1a78a3f2243f67e7401c2c6fad5c5e713876df52/src/routing/mod.rs#L45) is the issue.

Route `/a` should respond with `A` instead of `C`

[minimal example repo](https://github.com/Curid/axum_minimal_dylib_issue)

#### main.rs
```
#[tokio::main]
async fn main() {
    let router = Router::new()
        .route("/a", get(|| async { "A" }))
        .route("/b", get(|| async { "B" }));

    let dylib = unsafe { Library::new("./target/debug/libdylib.so").expect("load library") };
    let add_route: Symbol<fn(router: Router) -> Router> =
        unsafe { dylib.get(b"add_route") }.expect("load symbol");

    let router = add_route(router);

    axum::Server::bind(&"0.0.0.0:3000".parse().unwrap())
        .serve(router.into_make_service())
        .await
        .unwrap();
}
```
#### dylib.rs
```
#[no_mangle]
pub fn add_route(router: Router) -> Router {
    router.route("/c", get(|| async { "C" }))
}
```

davidpdrsn:
I haven't really worked much with dynamic libraries but I suppose the solution would be to use a random number for the ids instead of incrementing an atomic counter. Wdyt?

jplatte:
I would worry about introducing that kind of non-determinism. It could still lead to duplicate usage of an ID, even if very unlikely. Couldn't we just have a regular `u64` inside `Router` that is used for the `RouteId` and incremented in the two places we currently use `RouteId::next()`?

davidpdrsn:
I used to think that wouldn't work for cases like

```rust
let one = Router::new().route("/one", ...);
let two = Router::new().route("/two", ...);

one.merge(two)
```

but since `merge` just calls `route` which creates a new id I actually think it should work!

Curid:
Thanks for the quick fix! I wasn't sure if this even was a supported use case, since core Tokio has a few similar issues that can't really be fixed without breaking changes.

https://github.com/tokio-rs/tokio/pull/4940

PR #1806, opened 2023-03-03
Fix routing issues when loading a `Router` via a dynamic library

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1803

I also tested it with the repro script and does indeed fix the problem.

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -502,7 +502,8 @@ where
     }
 
     fn next_route_id(&mut self) -> RouteId {
-        let next_id = self.prev_route_id
+        let next_id = self
+            .prev_route_id
             .0
             .checked_add(1)
             .expect("Over `u32::MAX` routes created. If you need this, please file an issue.");
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -52,7 +52,7 @@ pub struct Router<S = (), B = Body> {
     routes: HashMap<RouteId, Endpoint<S, B>>,
     node: Arc<Node>,
     fallback: Fallback<S, B>,
-    prev_route_id: Option<RouteId>,
+    prev_route_id: RouteId,
 }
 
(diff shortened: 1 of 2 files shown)


davidpdrsn:
@Curid do you wanna test this to double check that it works?

axum/src/routing/mod.rs
```diff
@@ -47,24 +47,12 @@ pub use self::method_routing::{
...
-        Self(id)
-    }
-}
-
 /// The router type for composing handlers and services.
 pub struct Router<S = (), B = Body> {
     routes: HashMap<RouteId, Endpoint<S, B>>,
     node: Arc<Node>,
     fallback: Fallback<S, B>,
+    prev_route_id: Option<RouteId>,
```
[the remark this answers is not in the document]
davidpdrsn: Oh yeah we could just default to `RouteId(0)` 👍

axum/src/routing/mod.rs
```diff
@@ -506,6 +500,16 @@ where
...
     }
+
+    fn next_route_id(&mut self) -> RouteId {
+        let next_id = self
+            .prev_route_id
+            .0
+            .checked_add(1)
+            .expect("Over `u32::MAX` routes created. If you need this, please file an issue.");
+        self.prev_route_id = RouteId(next_id);
+        self.prev_route_id
```
jplatte, line 511: I still we should name this `next_route_id` and return `RouteId(0)` as the first one as before, not `RouteId(1)`. But that's very nitpicky at this point.

jplatte (approved):

axum/src/routing/mod.rs
davidpdrsn: Hehe I think I'll just leave it like this.

davidpdrsn:
I'll just merge this. @Curid let us know if it doesn't work for you :)

PR #1808, opened 2023-03-03
Implement `Clone` for `CookieJar`, `PrivateCookieJar` and `SignedCookieJar`

danhandrea:
## Motivation

Derive `Clone` missing from `CookieJar`.

## Solution

Add `Clone` to `CookieJar`

### DIFF
--- a/axum-extra/src/extract/cookie/private.rs
+++ b/axum-extra/src/extract/cookie/private.rs
@@ -102,7 +102,6 @@ use std::{convert::Infallible, fmt, marker::PhantomData};
 ///     }
 /// }
 /// ```
-#[derive(Clone)]
 pub struct PrivateCookieJar<K = Key> {
     jar: cookie::CookieJar,
     key: Key,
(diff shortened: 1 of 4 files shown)


davidpdrsn (commented):
I guess we should do the same for `SignedCookieJar` and `PrivateCookieJar`.

Also, do you wanna add a note to change log?

danhandrea:
Implemented for `SignedCookieJar` and `PrivateCookieJar`.
Also updated the change log.

axum-extra/src/extract/cookie/private.rs
```diff
@@ -102,6 +102,7 @@ use std::{convert::Infallible, fmt, marker::PhantomData};
 ///     }
 /// }
 /// ```
+#[derive(Clone)]
```
davidpdrsn, line 105: This will actually derive an impl like this

```rust
impl<K: Clone> Clone for PrivateCookieJar<K> {
    fn clone(&self) -> Self {
        Self {
            jar: self.jar.clone(),
            key: self.key.clone(),
            _marker: self._marker,
        }
    }
}
```

i.e. where `K` must also be `Clone`. But that is not strictly necessary since the struct doesn't actually contain a `K`. It's just references the type via `PhantomData`.

So a more correct implementation would be

```rust
impl<K> Clone for PrivateCookieJar<K> {
    fn clone(&self) -> Self {
        Self {
            jar: self.jar.clone(),
            key: self.key.clone(),
            _marker: self._marker,
        }
    }
}
```

Same goes for `SignedCookieJar`

danhandrea: Sorry, got confused by **K = Key**

davidpdrsn (approved):
Thanks!

jplatte (approved):
Nice!

PR #1826, opened 2023-03-07
Improve debug_handler message for generic request-consuming extractors

### LINKED ISSUE #1738: `#[debug_handler]` could suggest `Deserialize` for `Json`

### Motivation

While trying to teach someone axum I suggested the use of the `#[debug_handler]` macro in order to find out why their handler wasn't working. The error message did not however provide enough information for them to figure out the issue on their own, so I had to step in and explain what was wrong.

Following scenario:
```rust
struct NotDeserialize;

#[debug_handler]
async fn handler(_: Json<NotDeserialize>) -> impl IntoResponse {
	StatusCode::OK
}
```

The error message was:
```
error[E0277]: the trait bound `Json<NotDeserialize>: FromRequest<(), Body, _>` is not satisfied
  --> src/main.rs:20:21
   |
20 | async fn handler(_: Json<NotDeserialize>) -> impl IntoResponse {
   |                     ^^^^ the trait `FromRequest<(), Body, _>` is not implemented for `Json<NotDeserialize>`
   |
   = help: the trait `FromRequest<S, B>` is implemented for `Json<T>`
note: required by a bound in `__axum_macros_check_handler_0_from_request_check`
  --> src/main.rs:20:21
   |
20 | async fn handler(_: Json<NotDeserialize>) -> impl IntoResponse {
   |                     ^^^^ required by this bound in `__axum_macros_check_handler_0_from_request_check`
```

### Proposal

It would be great if it didn't just tell you that `FromRequest` isn't implemented but also that this is the case because of a missing implementation of `Deserialize`.

Not sure how feasible this is, given that I haven't really looked at how `debug_handler` is actually implemented so far. But wanted to share this because a new users first impression is often valuable for UX.

jplatte:
## Motivation

Fixes #1738.

## Solution

Set the generic parameter to `ViaRequest` in the debug_handler check function for the last parameter if its type is known to consume the request.

### DIFF
--- a/axum-macros/CHANGELOG.md
+++ b/axum-macros/CHANGELOG.md
@@ -7,7 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- **fixed:** Improve `#[debug_handler]` message for known generic
+  request-consuming extractors ([#1826])
+
+[#1826]: https://github.com/tokio-rs/axum/pull/1826
 
 # 0.3.5 (03. March, 2023)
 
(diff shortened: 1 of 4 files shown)


davidpdrsn (approved):
Nice!

PR #1828, opened 2023-03-08
Add low level OpenSSL example

Firstyear:
This adds an OpenSSL low level example similar to the rustls example.

## Motivation

People have various needs for cryptograhic libraries, and providing a wider range of examples can assist people to get started with libraries like axum.

## Solution

This example is modeled from the rustls low level equivalent, but with OpenSSL instead.


davidpdrsn (approved):
Thanks!

Firstyear:
@davidpdrsn No problem. I just added https://github.com/programatik29/axum-server/pull/71 so there will be an axum-server version of this to add soon too.

PR #1832, opened 2023-03-09
Fix static-file-server example instructions

Moises-Herradon-Cueto:
## Motivation

Running `cd examples && cargo run -p example-static-file-server` won't actually work, because the assets directory is inside the `static-file-server` folder. This had me stumped for way too long.

## Solution

I changed the comment. Another possible solution would be to move the assets folder to `examples` but I think the subfolders of examples are meant to be one example each.


davidpdrsn (approved):
Thanks!

PR #1834, opened 2023-03-09
 the trait `std::convert::From<axum::http::HeaderValue>` is not imple…

eademir:
…mented for `AnyOr<Origin>`

`.allow_origin("http://localhost:3000".parse::<HeaderValue>().unwrap());` gives error. Instead of parse::<HeaderValue>, Origin::exact() works just fine.

## Motivation

I countered a problem and fixed it. 

## Solution

I changed parse::<HeaderValue> as Origin::exact()

### DIFF
--- a/examples/cors/src/main.rs
+++ b/examples/cors/src/main.rs
@@ -5,13 +5,13 @@
 //! ```
 
 use axum::{
-    http::{HeaderValue, Method},
+    http::Method,
     response::{Html, IntoResponse},
     routing::get,
     Json, Router,
 };
 use std::net::SocketAddr;
-use tower_http::cors::CorsLayer;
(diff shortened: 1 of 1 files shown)


davidpdrsn:
Works fine for me without this change

```
❯ cargo check -p example-cors
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
```

We also check examples on CI so it couldn't have been broken by accident.

eademir:
> Works fine for me without this change
> 
> ```
> ❯ cargo check -p example-cors
>     Finished dev [unoptimized + debuginfo] target(s) in 0.08s
> ```
> 
> We also check examples on CI so it couldn't have been broken by accident.

Sorry. My bad. I was running on an older version of tower-http.

PR #1835, opened 2023-03-10
Add workaround for inference errors after removing `B` type param

davidpdrsn:
This adds `Router::as_service` which fixes the request body type and helps with type inference. Its a workaround for the issues like

```
    Checking example-testing v0.1.0 (/Users/david.pedersen/code/axum/examples/testing)
error[E0698]: type inside `async` block must be known in this context
   --> testing/src/main.rs:199:28
    |
199 |         let response = app.ready().await.unwrap().call(request).await.unwrap();
    |                            ^^^^^ cannot infer type for type parameter `B`
    |
```

that we encountered [here](https://github.com/tokio-rs/axum/pull/1751#issuecomment-1450182222).

I'm not really fond of having to do this but I don't see any other solutions 😞


davidpdrsn:
I think I'll move forward with this. I don't really like it but I think it's worth it considering the ergonomics we get from removing the `B` type param.

PR #1836, opened 2023-03-10
Don't require `S: Debug` for `impl Debug for Router<S>`

davidpdrsn:
This bound wasn't necessary since the routes are boxed and don't
actually print the state.

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -77,10 +77,7 @@ where
     }
 }
 
-impl<S, B> fmt::Debug for Router<S, B>
-where
-    S: fmt::Debug,
-{
+impl<S, B> fmt::Debug for Router<S, B> {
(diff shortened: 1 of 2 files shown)


axum/src/routing/mod.rs
```diff
@@ -77,10 +77,7 @@ where
     }
 }
 
-impl<S, B> fmt::Debug for Router<S, B>
-where
-    S: fmt::Debug,
-{
+impl<S, B> fmt::Debug for Router<S, B> {
```
jplatte, line 80: Do we want to print the new `prev_route_id` field in the `Debug` impl?

jplatte (approved):

axum/src/routing/mod.rs
davidpdrsn: Sure! Added in 5c11c5c44d0e0

PR #1837, opened 2023-03-10
Clone state a bit less

davidpdrsn:
This reduces the state clones by 1 for each request. Not much but better than nothing. Had to inline the `call_route` to get around some borrowing issues.

The remaining clones are
1. https://github.com/tokio-rs/axum/blob/main/axum/src/handler/service.rs#L171
    - I'm gonna invest making `Handler::call` take `S` by reference rather than value. Not sure what the implications are. Think the response future should still be `'static` so might not be possible.
3. https://github.com/tokio-rs/axum/blob/main/axum-core/src/extract/from_ref.rs#L23
    - This one we can't do much about. Its the `State` extractor via `FromRef`. That needs to produce an owned state.
4. https://github.com/tokio-rs/axum/blob/main/axum/src/routing/route.rs#L48
    - Because of backpressure `Route::call` uses `oneshot` which requires `clone`. Not sure its possible to fix but worth investigating.
5. https://github.com/tokio-rs/axum/blob/main/axum/src/test_helpers/test_client.rs#L32 (when hyper calls `call` on `Shared` that clones the service)
    - Nothing we can do about this. Hyper needs an owned service to serve the connection.

Part of https://github.com/tokio-rs/axum/issues/1827

### DIFF
--- a/axum/src/routing/tests/mod.rs
+++ b/axum/src/routing/tests/mod.rs
@@ -822,7 +822,7 @@ async fn state_isnt_cloned_too_much() {
 
     impl Clone for AppState {
         fn clone(&self) -> Self {
-            #[rustversion::stable]
+            #[rustversion::since(1.65)]
             #[track_caller]
             fn count() {
                 if SETUP_DONE.load(Ordering::SeqCst) {
@@ -838,7 +838,7 @@ async fn state_isnt_cloned_too_much() {
                 }
             }
 
-            #[rustversion::not(stable)]
+            #[rustversion::not(since(1.65))]
             fn count() {
                 if SETUP_DONE.load(Ordering::SeqCst) {
                     COUNT.fetch_add(1, Ordering::SeqCst);
--- a/axum/src/routing/tests/mod.rs
+++ b/axum/src/routing/tests/mod.rs
@@ -821,27 +821,32 @@ async fn state_isnt_cloned_too_much() {
     struct AppState;
 
     impl Clone for AppState {
-        #[rustversion::stable]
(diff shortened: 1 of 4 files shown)


axum/src/routing/tests/mod.rs
```diff
@@ -812,3 +812,43 @@ fn method_router_fallback_with_state() {
...
+        fn clone(&self) -> Self {
+            if SETUP_DONE.load(Ordering::SeqCst) {
+                let bt = std::backtrace::Backtrace::force_capture();
+                let bt = bt
+                    .to_string()
+                    .lines()
+                    .filter(|line| line.contains("axum") || line.contains("./src"))
+                    .collect::<Vec<_>>()
+                    .join("\n");
+                println!("AppState::Clone:\n===============\n{}\n", bt);
```
davidpdrsn, line 833: Wanna leave this in for easier debugging in the future.

axum/src/routing/tests/mod.rs
```diff
@@ -812,3 +812,43 @@ fn method_router_fallback_with_state() {
...
+        .with_state(AppState);
+
+    let client = TestClient::new(app);
+
+    // ignore clones made during setup
+    SETUP_DONE.store(true, Ordering::SeqCst);
+
+    client.get("/").send().await;
+
+    assert_eq!(COUNT.load(Ordering::SeqCst), 4);
```
jplatte, line 853: Maybe assert `<=` instead of equals?

jplatte (approved):

PR #1838, opened 2023-03-10
Change tower-http to optional

tottoto:
As axum depends on tower-http only in document, we might be able to move it to optional dependencies.

### DIFF
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -78,7 +78,6 @@ pin-project-lite = "0.2.7"
 serde = "1.0"
 sync_wrapper = "0.1.1"
 tower = { version = "0.4.13", default-features = false, features = ["util"] }
-tower-http = { version = "0.4", features = ["util", "map-response-body"] }
 tower-layer = "0.3.2"
 tower-service = "0.3"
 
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
Good catch!

davidpdrsn:
Hm now I'm actually wondering if this is gonna break intra docs links (https://github.com/tokio-rs/axum/blob/main/axum/Cargo.toml#L30) 🤔

tottoto:
In my testing, when the document is generated by `cargo doc`, this change will break links to tower-http. However when generated by `cargo doc --features __private_docs`, it won't.

jplatte:
So we had this dependency *exclusively* for docs? In any case, we should update the private feature to depend on `dep:tower-http` and `tower-http?/feature` for all `feature`s, so we don't have an implicit `tower-http` feature on the axum crate.

davidpdrsn:
> So we had this dependency exclusively for docs?

I think we used to use it (the fact that we enable `map-response-body` rings a bell) but I think some refactorings made it no longer used.

I'll make sure to check it out next time we publish a release! Otherwise our users are gonna notice quickly 😅

PR #1844, opened 2023-03-13
Docs: router wildcards - mention how to capture root path

imbolc:
Mention `Option<Path>` extractor

### DIFF
--- a/axum/src/docs/routing/route.md
+++ b/axum/src/docs/routing/route.md
@@ -54,6 +54,9 @@ Wildcard captures can also be extracted using [`Path`](crate::extract::Path).
 Note that the leading slash is not included, i.e. for the route `/foo/*rest` and
 the path `/foo/bar/baz` the value of `rest` will be `bar/baz`.
 
+If you want root path to be bound to the same handler, add both routes (e.g.
+`/assets/` and `/assets/*path`) and use `Option<Path<T>>` extractor.
+
 # Accepting multiple methods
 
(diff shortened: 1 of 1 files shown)


jplatte (commented):
I think we are intentionally *not* suggesting this because `Path` extraction can fail in other ways than the route not containing the right number of placeholders (for example non-UTF8 bytes in a captured segment).

davidpdrsn:
> I think we are intentionally _not_ suggesting this because `Path` extraction can fail in other ways than the route not containing the right number of placeholders (for example non-UTF8 bytes in a captured segment).

Exactly. See https://github.com/tokio-rs/axum/pull/1663.

My opinion is still that we shouldn't encourage this pattern. I think its better to have

```rust
Router::new()
    .route("/assets/", get(one))
    .route("/assets/*path", get(two));

async fn one() {
    some_shared_function(None).await
}

async fn two(Path(params): Path<String>) {
    some_shared_function(Some(params)).await
}

async fn some_shared_function(params: String) {}
```

imbolc:
Thank you, I didn't notice the discussion. Though the issue seems to be common, I guess you already thought about `Path<Option<T>>`?

davidpdrsn:
`Path` doesn't support deserializing `Option` (https://github.com/tokio-rs/axum/blob/main/axum/src/extract/path/de.rs#L60). Not sure what that would mean and it would probably have the same footgun.

imbolc:
Thank you @jplatte, I'll look into it, closing this one

PR #1846, opened 2023-03-13
Added information about MultipartError

myyrakle:
## Motivation

Personally, I am writing a file server. When a `MultipartError` occurs, there is a limit to finding out the error information, so the function was proposed. 
    
For example, I have to do the following.

## Solution

So, I added the `into_multer` function to get `multer::Error` information from `MultipartError` and branch it, and re-exported the `multer` module. 

Also, the existing `from_multer` function has been changed to `pub` so that `MultipartError` can be reproduced.

### DIFF
--- a/axum/src/extract/multipart.rs
+++ b/axum/src/extract/multipart.rs
@@ -206,8 +206,14 @@ pub struct MultipartError {
 }
 
 impl MultipartError {
-    fn from_multer(multer: multer::Error) -> Self {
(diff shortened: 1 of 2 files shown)


davidpdrsn (changes requested):
Sorry but we can't do this. It puts `multer` in axum's public api which I don't think we should.

The solution is to get the source error and downcast it like this https://docs.rs/axum/0.6.11/axum/extract/index.html#accessing-inner-errors

PR #1848, opened 2023-03-13
Add `Body::from_stream`

davidpdrsn:
This replaces `StreamBody` with `Body::from_stream`.


jplatte (approved):

waynr:
@davidpdrsn @jplatte did y'all mean to leave the the `axum/src/body/stream_body.rs` module dangling (now that it's not pulled into `axum/src/body/mod.rs` it doesn't appear to be used anywhere else.

PR #1850, opened 2023-03-13
Move `TypedHeader` to axum-extra

davidpdrsn:
This moves `TypedHeader` into axum-extra to protect axum against breaking changes to headers. Turned out to require quite a few changes since it was used in a bunch of docs.

I opted to keep the headers re-export. Probably fine in axum-extra and is nice that users don't have to worry about getting the correct version of headers. That'd otherwise lead to confusing errors.

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -32,6 +32,7 @@ json-lines = [
 multipart = ["dep:multer"]
 protobuf = ["dep:prost"]
 query = ["dep:serde", "dep:serde_html_form"]
+typed-header = ["dep:headers"]
 typed-routing = ["dep:axum-macros", "dep:serde", "dep:percent-encoding", "dep:serde_html_form", "dep:form_urlencoded"]
 
 [dependencies]
@@ -51,6 +52,7 @@ tower-service = "0.3"
 axum-macros = { path = "../axum-macros", version = "0.3.5", optional = true }
 cookie = { package = "cookie", version = "0.17", features = ["percent-encode"], optional = true }
 form_urlencoded = { version = "1.1.0", optional = true }
+headers = { version = "0.3.8", optional = true }
 multer = { version = "2.0.0", optional = true }
 percent-encoding = { version = "2.1", optional = true }
 prost = { version = "0.11", optional = true }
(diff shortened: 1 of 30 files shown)


axum-extra/Cargo.toml
```diff
@@ -84,6 +86,7 @@ allowed = [
     "cookie",
     "futures_core",
     "futures_util",
+    "headers_core",
```
jplatte, line 89: Why only the core crate? The entire `headers` crate is being re-exported.

axum-extra/src/typed_header.rs
```diff
@@ -1,10 +1,16 @@
...
+use axum::{
+    async_trait,
+    extract::FromRequestParts,
+    response::{IntoResponse, IntoResponseParts, Response, ResponseParts},
+};
+use headers::Header;
+use http::{request::Parts, HeaderValue};
 use std::{convert::Infallible, ops::Deref};
 
+pub use headers;
```
jplatte, line 12: I think it would be more convenient for this to be `axum_extra::headers` instead of `axum_extra::typed_header::headers`. Then the `typed_header` module also doesn't need to be public either, I think.

axum-extra/src/typed_header.rs
```diff
@@ -168,12 +172,55 @@ impl std::error::Error for TypedHeaderRejection {
     }
 }
 
+// copied from https://docs.rs/headers/latest/src/headers/map_ext.rs.html#22-52
+fn typed_insert<H>(headers: &mut http::HeaderMap, header: H)
```
jplatte, line 176: Why is it copied, instead of being reused?

axum-extra/src/typed_header.rs
jplatte: Also should probably be `#[doc(no_inline)]` like before.

davidpdrsn: I'm okay with exporting `headers` and `TypedHeader` from the root but I wanna keep the module public because it contains a bunch more types (like the rejection). I'd rather not expose those from the root.

Time to update all the docs again 😅

axum-extra/src/typed_header.rs
davidpdrsn: I originally tried to just depend on headers-core so had to copy this but can just be removed now!

axum-extra/Cargo.toml
davidpdrsn: That is fixed now.

davidpdrsn:
The UI tests pass locally but not on CI 😭

PR #1852, opened 2023-03-14
Fix typo in extractors docs

fredr:
## Motivation

Fixes a small typo in https://docs.rs/axum/latest/axum/extract/trait.FromRequest.html and https://docs.rs/axum/latest/axum/extract/trait.FromRequestParts.html

## Solution

extra**x**tors => extra**c**tors


davidpdrsn (approved):
Thanks!

Issue #475, opened 2021-11-06
can you please add one diesel + postgres example?


kushwahashiv:
specially one covering transactions. how to achieve that using axum.

davidpdrsn:
I haven't used diesel in a very long time and never from async code. So I'm not sure what the best way to do it is. I guess `tokio::task::spawn_blocking` and you can use diesel as par usual but I'm not sure about transactions.

Vesafary:
Maybe this crate will help you: https://docs.rs/bb8-diesel/0.2.1/bb8_diesel/
Here's an example of the crate: https://github.com/overdrivenpotato/bb8-diesel/blob/master/examples/usage.rs

And together with this Axum example you should be able to make it work :)
https://github.com/tokio-rs/axum/blob/main/examples/tokio-postgres/src/main.rs

dark-fusion:
I haven't used `diesel` in a very long time, either, but I understand how to put together an example for it.

If there's demand and you'd like a hand with this, I can create an example that shows axum with diesel + postgresql (including transactions).

davidpdrsn:
That would be great! Depending on the size of the example we could also consider linking to it from [here](https://github.com/tokio-rs/axum/blob/main/ECOSYSTEM.md).

dark-fusion:
Awesome 🙂! I'll get started on it as soon as possible. I don't anticipate that it will take too long to finish. I'll try to keep the example small, but not at the expense of clarity.

I'm okay with whichever location you think suits the example best (examples / community projects / etc).

davidpdrsn:
Sounds good! Just make a PR with our stuff and then we can discuss it 😊

dark-fusion:
@kushwahashiv I understand you would like an example that includes `diesel` and SQL transactions. Are you interested in `diesel` with the `postgres` feature? I just want to clarify before I finish putting things together.

I'm operating on the assumption that "diesel + postgres" == `diesel` w/ `postgres` feature enabled. If there are additional constraints or requirements, please let me know.

Thanks

kushwahashiv:
> @kushwahashiv I understand you would like an example that includes `diesel` and SQL transactions. Are you interested in `diesel` with the `postgres` feature? I just want to clarify before I finish putting things together.
> 
> I'm operating on the assumption that "diesel + postgres" == `diesel` w/ `postgres` feature enabled. If there are additional constraints or requirements, please let me know.
> 
> Thanks

This would be very helpful to understand how the transactions work.

[...]

PR #1853, opened 2023-03-14
Add a diesel and diesel-async example

weiznich:
## Motivation

Diesel is a popular ORM that is often used in combination with web frameworks like axum. I've noticed that there are no examples for this in the example directory yet. 

## Solution

Add two examples, one based on the main (sync) diesel crate and one based on the slightly experimental diesel-async version.

### DIFF
--- a/examples/diesel-async-postgres/src/main.rs
+++ b/examples/diesel-async-postgres/src/main.rs
@@ -1,0 +1,143 @@
+//! Run with
+//!
+//! ```sh
+//! export DATABASE_URL=postgres://localhost/your_db
+//! diesel migration run
+//! cargo0 run -p example-diesel-postgres//!
+//! ```
+//!
+//! Checkout the [diesel webpage](https://diesel.rs) for
+//! longer guides about diesel
+//!
+//! Checkout the [crates.io source code](https://github.com/rust-lang/crates.io/)
+//! for a real world application using axum and diesel
+
+use axum::{
+    async_trait,
+    extract::{FromRef, FromRequestParts, State},
+    http::{request::Parts, StatusCode},
+    response::Json,
+    routing::{get, post},
+    Router,
+};
+use diesel::prelude::*;
+use diesel_async::{
+    pooled_connection::AsyncDieselConnectionManager, AsyncPgConnection, RunQueryDsl,
+};
+use std::net::SocketAddr;
+use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
+
+// normally part of your generated schema.rs file
+table! {
+    users (id) {
+        id -> Integer,
+        name -> Text,
+        hair_color -> Nullable<Text>,
+    }
+}
+
+#[derive(serde::Serialize, Selectable, Queryable)]
+struct User {
+    id: i32,
+    name: String,
+    hair_color: Option<String>,
+}
+
+#[derive(serde::Deserialize, Insertable)]
+#[diesel(table_name = users)]
+struct NewUser {
+    name: String,
+    hair_color: Option<String>,
+}
+
+type Pool = bb8::Pool<AsyncDieselConnectionManager<AsyncPgConnection>>;
+
+#[tokio::main]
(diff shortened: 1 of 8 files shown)


weiznich:
This addresses https://github.com/tokio-rs/axum/issues/475. Any change to get a review for this?

examples/diesel-async-postgres/src/main.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+type Pool = bb8::Pool<AsyncDieselConnectionManager<AsyncPgConnection>>;
+
+#[tokio::main]
+async fn main() {
+    tracing_subscriber::registry()
+        .with(
+            tracing_subscriber::EnvFilter::try_from_default_env()
+                .unwrap_or_else(|_| "example_tokio_postgres=debug".into()),
```
davidpdrsn, line 60:
```suggestion
                .unwrap_or_else(|_| "example_diesel_async_postgres=debug".into()),
```

examples/diesel-postgres/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+                .returning(User::as_returning())
+                .get_result(conn)
+        })
+        .await
+        .map_err(internal_error)?
+        .map_err(internal_error)?;
+    Ok(Json(res))
+}
+
+#[axum_macros::debug_handler]
```
davidpdrsn, line 99:
```suggestion
```

examples/diesel-postgres/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+    // run it with hyper
+    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
+    tracing::debug!("listening on {}", addr);
+    axum::Server::bind(&addr)
+        .serve(app.into_make_service())
+        .await
+        .unwrap();
+}
+
+#[axum_macros::debug_handler]
```
davidpdrsn, line 80:
```suggestion
```

examples/diesel-postgres/src/main.rs
```diff
@@ -0,0 +1,119 @@
+//! Run with
+//!
+//! ```not_rust
+//! cargo run -p example-diesel-postgres
```
davidpdrsn, line 4: Do we need a command here to run migrations?

examples/diesel-async-postgres/src/main.rs
```diff
@@ -0,0 +1,143 @@
...
+    // run it with hyper
+    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
+    tracing::debug!("listening on {}", addr);
+    axum::Server::bind(&addr)
+        .serve(app.into_make_service())
+        .await
+        .unwrap();
+}
+
+#[axum_macros::debug_handler]
```
weiznich, line 86:
```suggestion
```

examples/diesel-postgres/src/main.rs
weiznich: Yes that was missing.
Rather then requiring a separate step there I just used `diesel_migration` to automatically run the migrations on application startup. 
See https://github.com/tokio-rs/axum/pull/1853/commits/0abd85cf9e79eb4fa0066c0f0ce2880d71042817

(That's not possible with diesel-async yet, so no change there)

davidpdrsn (approved):
Thanks!

davidpdrsn:
Alright thanks! It used to work though but I'll give that a shot.

davidpdrsn:
`brew install libpq` didn't fix it 😞

weiznich:
Hmm, that's interesting. The targets indicates you use an arm m1/m2 mac? In that case I would start with checking whether the arm version of libpq is installed or the x86 version. If it's the arm version you can try setting `PQ_LIB_DIR` so that it points to the directory that contains `libpq.so`. That should normally fix the issue, although I'm a bit surprised that it doesn't pick up the correct lib by default (that normally works on linux/mac os)

Issue #1851, opened 2023-03-13
Implement `IntoResponse` for `MultipartError`


davidpdrsn:
See https://github.com/tokio-rs/axum/discussions/1847 for context.

Should be done both for the `Multipart` in axum and axum-extra.

PR #1861, opened 2023-03-17
Implement `IntoResponse` for `MultipartError`

davidpdrsn:
So you can return `MultipartError` from handlers without having to
downcast the errors yourself.

@AcrylicShrimp what do you think about this?

Fixes https://github.com/tokio-rs/axum/issues/1851

### DIFF
--- a/axum-extra/src/extract/multipart.rs
+++ b/axum-extra/src/extract/multipart.rs
@@ -274,7 +274,6 @@ fn status_code_from_multer_error(err: &multer::Error) -> StatusCode {
         multer::Error::FieldSizeExceeded { .. } | multer::Error::StreamSizeExceeded { .. } => {
             StatusCode::PAYLOAD_TOO_LARGE
         }
-        multer::Error::LockFailure => StatusCode::INTERNAL_SERVER_ERROR,
         multer::Error::StreamReadFailed(err) => {
             if let Some(err) = err.downcast_ref::<multer::Error>() {
                 return status_code_from_multer_error(err);
@@ -291,7 +290,7 @@ fn status_code_from_multer_error(err: &multer::Error) -> StatusCode {
 
             StatusCode::INTERNAL_SERVER_ERROR
         }
-        _ => StatusCode::BAD_REQUEST,
+        _ => StatusCode::INTERNAL_SERVER_ERROR,
     }
 }
 
--- a/axum-extra/src/extract/multipart.rs
+++ b/axum-extra/src/extract/multipart.rs
@@ -12,9 +12,10 @@ use axum::{
 use futures_util::stream::Stream;
 use http::{
     header::{HeaderMap, CONTENT_TYPE},
-    Request,
+    Request, StatusCode,
 };
 use std::{
+    error::Error,
     fmt,
     pin::Pin,
     task::{Context, Poll},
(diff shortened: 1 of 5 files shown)


AcrylicShrimp:
Great! What a nice work. Your implementation looks fine to me. Let's note that this kind of handling may not work(or make no sense) in future, because of changes of actual error structures. But I think it's not big deal.

davidpdrsn:
> Let's note that this kind of handling may not work(or make no sense) in future, because of changes of actual error structures. But I think it's not big deal.

That's always the case and why multer is a private dependency of axum. Then we can change things without breakage.

axum-extra/src/extract/multipart.rs
```diff
@@ -246,6 +247,58 @@ impl MultipartError {
...
+        | multer::Error::NoMultipart
+        | multer::Error::IncompleteStream => StatusCode::BAD_REQUEST,
+        multer::Error::FieldSizeExceeded { .. } | multer::Error::StreamSizeExceeded { .. } => {
+            StatusCode::PAYLOAD_TOO_LARGE
+        }
+        multer::Error::LockFailure => StatusCode::INTERNAL_SERVER_ERROR,
+        multer::Error::StreamReadFailed(err) => {
+            if let Some(err) = err.downcast_ref::<multer::Error>() {
+                return status_code_from_multer_error(err);
+            }
```
jplatte, line 281: Why would `multer::Error` recursively contain another `multer::Error` for the `StreamReadFailed` variant?

AcrylicShrimp: @jplatte Because of below chain.

1. https://github.com/rousan/multer-rs/blob/a49e03645a307110691f90493070c2e6ddfa5d00/src/error.rs#L60
1. https://github.com/rousan/multer-rs/blob/a49e03645a307110691f90493070c2e6ddfa5d00/src/field.rs#L342
1. https://github.com/rousan/multer-rs/blob/a49e03645a307110691f90493070c2e6ddfa5d00/src/buffer.rs#L50
1. https://github.com/rousan/multer-rs/blob/a49e03645a307110691f90493070c2e6ddfa5d00/src/multipart.rs#L126
1. https://github.com/hyperium/http-body/blob/6f722a14f7c6488cd53dcc4126a84e6ee201964f/src/limited.rs#L52

So, final error looks like:

```rust
axum::extract::multipart::MultipartError {
    source: multer::Error::StreamReadFailed(
        Box::new(multer::Error::StreamReadFailed(
            Box::new(http_body::LengthLimitError)
        ))
    )
}
```

axum-extra/src/extract/multipart.rs
```diff
@@ -246,6 +247,58 @@ impl MultipartError {
...
+        status_code_from_multer_error(&self.source)
+    }
+}
+
+fn status_code_from_multer_error(err: &multer::Error) -> StatusCode {
+    match err {
+        multer::Error::UnknownField { .. }
+        | multer::Error::IncompleteFieldData { .. }
+        | multer::Error::IncompleteHeaders
+        | multer::Error::ReadHeaderFailed(..)
```
AcrylicShrimp, line 267: I'm not sure if this `multer::Error::ReadHeaderFailed` is always thrown by the client side. How do you think? Is it OK to response as `400`?

axum-extra/src/extract/multipart.rs
```diff
@@ -246,6 +247,58 @@ impl MultipartError {
...
+        | multer::Error::DecodeHeaderName { .. }
+        | multer::Error::DecodeContentType(..)
+        | multer::Error::NoBoundary
+        | multer::Error::DecodeHeaderValue { .. }
+        | multer::Error::NoMultipart
+        | multer::Error::IncompleteStream => StatusCode::BAD_REQUEST,
+        multer::Error::FieldSizeExceeded { .. } | multer::Error::StreamSizeExceeded { .. } => {
```
AcrylicShrimp, line 274: This kind of error will never thrown, because `axum` does not support constraints of `multer`.

axum-extra/src/extract/multipart.rs
```diff
@@ -246,6 +247,58 @@ impl MultipartError {
...
+                .and_then(|err| err.source())
+                .and_then(|err| err.downcast_ref::<http_body::LengthLimitError>())
+                .is_some()
+            {
+                return StatusCode::PAYLOAD_TOO_LARGE;
+            }
+
+            StatusCode::INTERNAL_SERVER_ERROR
+        }
+        _ => StatusCode::BAD_REQUEST,
```
AcrylicShrimp, line 294: I think we should respond as `500` for not-yet-handled errors or safety.

AcrylicShrimp (approved):
Thank you for hard working!

jplatte (approved):

axum-extra/src/extract/multipart.rs
davidpdrsn: But we still need an exhaustive match.

axum-extra/src/extract/multipart.rs
davidpdrsn: Feels like if the headers cannot be parsed then that's a client issue. The inner error is [`httparse::Error`](https://docs.rs/httparse/1.8.0/httparse/enum.Error.html)

axum-extra/src/extract/multipart.rs
davidpdrsn: Yeah that's probably better.

PR #1862, opened 2023-03-18
Update to syn 2

davidpdrsn:
🚀

### DIFF
--- a/axum-macros/src/attr_parsing.rs
+++ b/axum-macros/src/attr_parsing.rs
@@ -62,7 +62,7 @@ where
 {
     attrs
         .iter()
-        .filter(|attr| attr.path.is_ident(ident))
+        .filter(|attr| attr.meta.path().is_ident(ident))
         .map(|attr| attr.parse_args::<T>())
         .try_fold(T::default(), |out, next| out.combine(next?))
 }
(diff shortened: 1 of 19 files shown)


deny.toml
```diff
@@ -23,6 +23,8 @@ skip-tree = [
     { name = "windows-sys" },
     # old version pulled in by rustls via ring
     { name = "spin" },
+    # lots still pulls in syn 1.x
+    { name = "syn" },
```
davidpdrsn, line 27: I tried to remove duplicate syn versions but still lots of crates that need to be updated. I'm gonna submit some patches but don't it should block us from updating.

jplatte (approved):

axum-macros/src/attr_parsing.rs
```diff
@@ -62,7 +62,7 @@ where
 {
     attrs
         .iter()
-        .filter(|attr| attr.path.is_ident(ident))
+        .filter(|attr| attr.meta.path().is_ident(ident))
```
jplatte, line 65: Having just upgraded another project to syn 2, I realized you used a more verbose solution than necessary here. `Attribute` also has a `path` method, so this can be `attr.path()`.

PR #1866, opened 2023-03-20
Add docs in mdBook format

liamwh:
## Motivation

Improve documentation accessibility and consumeability. 

#1811 

## Solution

I ported the docs to an mdBook, and added a pipeline to publish them Github pages.

image

## Additional notes

- This is just a PoC to stimulate discussion and give people something to criticize, I did not put care into arranging the content.
- To see the book for yourself locally, install mdBook, then in the docs folder at the root of the repo run `mdbook watch -o`
- Even if we go to a more beautiful site in the future, using something more consumable than docs.rs is likely to help people in the interim until a dedicated site is online.


davidpdrsn:
We don't need the api docs in a book as well.

PR #1868, opened 2023-03-20
Add `serve` function and remove `Server` re-export

davidpdrsn:
Still work in progress.

-------

axum current re-exports `hyper::Server` but in hyper 1.0 that is going away, so we need a replacement.

Its likely that `hyper::Server` will be moved to hyper-util, in some form or another, but axum can't have hyper-util in its public API because hyper-util will receive breaking changes. That's the whole point of that crate, similar to axum-extra.

My proposal is to make a very simple `async fn serve()` function for basic use cases. It can't be configured and only handles http1. For other use cases we can tell people to use hyper-util. We could also wrap whatever ends up in hyper-util but then we have to work to keep those in sync and possibly still need to make breaking changes.

Using `axum::serve` looks like this:

```rust
axum::serve(listener, router).await.unwrap();
```

It still supports `.into_make_service()` and `.into_make_service_with_connect_info()`.

If the server in hyper-util doesn't end up using `MakeService` then we might have to remove `.into_make_service_with_connect_info()` but that's a separate issue.

## TODO 

- [ ] docs
- [ ] changelog

### DIFF
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -50,6 +50,15 @@ tower = { version = "0.4.13", default-features = false, features = ["util"] }
 tower-layer = "0.3.2"
 tower-service = "0.3"
 
+hyper1 = { package = "hyper", version = "1.0.0-rc.3", features = [
+    "server",
+    "http1",
+] }
+tower-hyper-http-body-compat = { version = "0.1.3", features = [
+    "server",
+    "http1",
+] }
+
 # optional dependencies
 axum-macros = { path = "../axum-macros", version = "0.3.6", optional = true }
 base64 = { version = "0.21.0", optional = true }
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -50,14 +50,9 @@ tower = { version = "0.4.13", default-features = false, features = ["util"] }
 tower-layer = "0.3.2"
 tower-service = "0.3"
 
-hyper1 = { package = "hyper", version = "1.0.0-rc.3", features = [
-    "server",
-    "http1",
-] }
-tower-hyper-http-body-compat = { version = "0.1.3", features = [
-    "server",
-    "http1",
-] }
+# wont need this when axum uses http-body 1.0
+hyper1 = { package = "hyper", version = "1.0.0-rc.3", features = ["server", "http1"] }
+tower-hyper-http-body-compat = { version = "0.1.3", features = ["server", "http1"] }
 
 # optional dependencies
 axum-macros = { path = "../axum-macros", version = "0.3.6", optional = true }
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -52,7 +52,7 @@ tower-service = "0.3"
 
 # wont need this when axum uses http-body 1.0
 hyper1 = { package = "hyper", version = "1.0.0-rc.3", features = ["server", "http1"] }
-tower-hyper-http-body-compat = { version = "0.1.3", features = ["server", "http1"] }
+tower-hyper-http-body-compat = { version = "0.1.4", features = ["server", "http1"] }
 
 # optional dependencies
 axum-macros = { path = "../axum-macros", version = "0.3.6", optional = true }
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -190,6 +190,7 @@ allowed = [
     "http_body",
     "hyper",
     "serde",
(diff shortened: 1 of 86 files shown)


axum/Cargo.toml
```diff
@@ -50,6 +50,10 @@ tower = { version = "0.4.13", default-features = false, features = ["util"] }
 tower-layer = "0.3.2"
 tower-service = "0.3"
 
+# wont need this when axum uses http-body 1.0
+hyper1 = { package = "hyper", version = "1.0.0-rc.3", features = ["server", "http1"] }
+tower-hyper-http-body-compat = { version = "0.1.3", features = ["server", "http1"] }
```
davidpdrsn, line 55: Just a stepping stone until we've ported axum to hyper 1.0. We won't actually ship 0.7 with this.

axum/src/serve.rs
```diff
@@ -0,0 +1,132 @@
...
+use futures_util::{future::poll_fn, FutureExt};
+use hyper1::server::conn::http1;
+use tokio::net::{TcpListener, TcpStream};
+use tower_hyper_http_body_compat::{HttpBody04ToHttpBody1, HttpBody1ToHttpBody04};
+use tower_service::Service;
+
+/// TODO(david): docs
+pub async fn serve<M, S>(tcp_listener: TcpListener, mut make_service: M) -> io::Result<()>
+where
+    M: for<'a> Service<&'a (TcpStream, SocketAddr), Error = Infallible, Response = S>,
```
davidpdrsn, line 18: note to self: Probably should make a struct instead of `&'a (TcpStream, SocketAddr)`. It's not clear that the `SocketAddr` is `local_addr`.

axum/src/serve.rs
```diff
@@ -0,0 +1,132 @@
...
+        });
+
+        tokio::task::spawn(async move {
+            match http1::Builder::new()
+                .serve_connection(tcp_stream, service)
+                .await
+            {
+                Ok(()) => {}
+                Err(_err) => {
+                    // TODO(david): how to handle this error?
```
davidpdrsn, line 81: 👀

jplatte:
I think losing http2 support would be unfortunate. How about making it only configurable at compile time? I.e. `serve` enables all http versions activated via cargo feature? I'm pretty sure it's the same thing we have now with the common form of `Server` usage.

davidpdrsn:
Hyper doesn't have any built in way to handle both http 1 and 2 at the same time. The plan is to add something like that to hyper-util but it's quite complicated https://github.com/hyperium/hyper-util/pull/11

I've added a note of it [here](https://github.com/tokio-rs/axum/pull/1583). Then we can see how things land in hyper-util.

davidpdrsn:
@programatik29 do you have time to review this?

axum/src/serve.rs
programatik29: Either we should create a mechanism that allows user to log the error or we should just ignore it.

axum/src/serve.rs
```diff
@@ -0,0 +1,237 @@
...
+            match service.poll_ready(&mut cx) {
+                Poll::Ready(Ok(())) => {}
+                Poll::Ready(Err(err)) => match err {},
+                // ...otherwise load shed
+                Poll::Pending => {
+                    let mut res = Response::new(HttpBody04ToHttpBody1::new(Body::empty()));
+                    *res.status_mut() = http::StatusCode::SERVICE_UNAVAILABLE;
+                    return std::future::ready(Ok(res)).left_future();
+                }
+            }
```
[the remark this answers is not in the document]
davidpdrsn: Oh yeah I forgot about that!

axum/src/serve.rs
davidpdrsn: Yeah... I'm leaning towards just ignoring it but that also isn't great. I just don't wanna add more type parameters or arguments to this function. We could also log it.

programatik29 (approved):

davidpdrsn:
@jplatte do you wanna give this a look as well or should we just :shipit: ?

alper:
Which TcpListener is this supposed to be?

```
help: consider importing one of these structs
   |
6  + use std::net::TcpListener;
   |
6  + use tokio::net::TcpListener;
```

alper:
OK. I see it here in the example: https://docs.rs/axum/latest/axum/serve/struct.Serve.html#example

PR #1869, opened 2023-03-21
Make tracing-aka-logging example more realistic

jplatte:
## Motivation

People often have questions about tracing and this example didn't previously help that much.

## Solution

Use `make_span_with` in the example and show a few nice features of tracing's macros.


davidpdrsn (approved):
Thanks!

PR #1877, opened 2023-03-22
Remove `IntoResponse` for http-body types

davidpdrsn:
These types are being moved to http-body-util as part of hyper 1.0 so they can't be in axum's public API.


jplatte (approved):
Never used any of those impls.

PR #1880, opened 2023-03-23
Use State extractor in WS module docs

zaynetro:
## Motivation

`Extension` is outdated mechanism for accessing app state. Switch to newer `State` extractor.


jplatte (approved):
`Extension` is not outdated, but I agree `State` is preferable here. Will wait for David's opinion before merging though.

davidpdrsn (approved):
Yes extension still has valid uses but this probably should use State.

Thanks!

PR #1882, opened 2023-03-23
Update to latest versions of hyper and http-body

davidpdrsn:
Part of https://github.com/tokio-rs/axum/issues/1325

Mostly just mechanical changes. Nothing too complicated.

Overall changes:

- Removes the dependency on hyper 0.14
- Removes usage of `hyper::Server` in docs
- Updates to the latest release candidates for hyper, http-body, and http-body-util
- Updates body types to the new `Body` trait
- Replaces usages of `CompressionLayer` with someone else, where appropriate
- Tests that specifically needed compression were commented out
- Examples that don't yet support hyper 1.0 were commented out
- Make sure we don't expose things from http-body-util (such as `Limited`) in our public API

Even though we're still relying on patches in `Cargo.toml` it would be nice to merge this so we can make smaller patches going forward.

### Blocked by downstreams updating

- https://github.com/rousan/multer-rs/pull/59
- New release of headers, technically just for axum-extra but would be nice to ship everything at the same time

### When tower-http's compression stuff is done

- [x] Bring back the `head_with_middleware_applied` test. It uses tower-http's compression middleware which haven't been upgraded yet.
- [x] Fix link to `CompressionLayer` in `axum/src/docs/middleware.md`


jplatte:
So, will tower-http be much work?

davidpdrsn:
I'm gonna find out tomorrow 😅 but I don't think so. Probably just a lot of code to port.

Cargo.toml
```diff
@@ -5,3 +5,10 @@ default-members = ["axum", "axum-*"]
 # Example has been deleted, but README.md remains
 exclude = ["examples/async-graphql"]
 resolver = "2"
+
+[patch.crates-io]
+# https://github.com/tower-rs/tower-http/pull/348
+tower-http = { git = "https://github.com/tower-rs/tower-http", rev = "8734fc4438828" }
+
+# for `Frame::map_data`
+http-body = { git = "https://github.com/hyperium/http-body", rev = "7bf321acbb422" }
```
davidpdrsn, line 14: 👀

axum-core/src/body.rs
```diff
@@ -182,8 +121,16 @@ impl Stream for Body {
...
 
     #[inline]
-    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        self.poll_data(cx)
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        loop {
+            match futures_util::ready!(self.as_mut().poll_frame(cx)?) {
+                Some(frame) => match frame.into_data() {
+                    Ok(data) => return Poll::Ready(Some(Ok(data))),
+                    Err(_frame) => {}
```
jplatte, line 129: This is for streaming headers and trailers, right? Or just trailers? Is it really okay to discard these in the `Stream` impl? What is this impl even used for? :sweat_smile:

axum-core/src/ext_traits/request.rs
```diff
@@ -259,12 +258,18 @@ pub trait RequestExt: sealed::Sealed + Sized {
     /// Apply the [default body limit](crate::extract::DefaultBodyLimit).
     ///
     /// If it is disabled, return the request as-is in `Err`.
-    fn with_limited_body(self) -> Result<Request<Limited<Body>>, Request>;
+    ///
+    /// Note that while the `Ok` and `Err` types are the same, [`http_body_util::Limited`] will have
+    /// been applied in the `Ok` case.
+    fn with_limited_body(self) -> Result<Request, Request>;
```
jplatte, line 264: I think we should return `Result` now that the types are the same anyways. Same for `into_limited_body`. If we want people to be able to detect whether this would do anything, that could be a separate method.

jplatte (commented):
Screenshot_2023-11-15_194020

axum/CHANGELOG.md
```diff
@@ -66,6 +66,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **added:** Implement `IntoResponse` for `(R,) where R: IntoResponse` ([#2143])
 - **changed:** For SSE, add space between field and value for compatibility ([#2149])
 - **added:** Add `NestedPath` extractor ([#1924])
+- **breaking:** `impl<T> IntoResponse(Parts) for Extension<T>` now requires `T:
+  Clone`. As that is required by the http crate ([#1882])
```
jplatte, line 70:
```suggestion
  Clone`, as that is required by the http crate ([#1882])
```

axum-core/src/extract/rejection.rs
```diff
@@ -19,11 +19,18 @@ impl FailedToBufferBody {
     where
         E: Into<BoxError>,
     {
+        // two layers of boxes where because `with_limited_body`
```
jplatte, line 22:
```suggestion
        // two layers of boxes here because `with_limited_body`
```

axum/src/routing/tests/mod.rs
```diff
@@ -17,10 +17,11 @@ use crate::{
 use axum_core::extract::Request;
 use futures_util::stream::StreamExt;
 use http::{
-    header::CONTENT_LENGTH,
-    header::{ALLOW, HOST},
+    header::ALLOW,
+    header::{CONTENT_LENGTH, HOST},
```
jplatte, line 21: How about `header::{ALLOW, CONTENT_LENGTH, HOST}`? 😄

axum-core/src/body.rs
davidpdrsn: Yeah `Err(_frame)` is if the body returns a non-data frame, which can currently only be trailers.

http-body-util provides [a stream adapter which streams frames](https://docs.rs/http-body-util/latest/http_body_util/struct.StreamBody.html). This impl just streams the bytes in data frames which I think is more generally useful. Makes things like streaming request bodies to files easier.

It is a bit iffy to ignore trailers but [`BodyStream` in 0.6](https://docs.rs/axum/latest/src/axum/extract/request_parts.rs.html#143-149) does it too so I think its okay.

davidpdrsn: Been thinking a little more about this and I think adding a dedicated method to turn the body into a data stream is worth it. So I've added `Body::into_data_stream` and `Body` no longer implements `Stream` directly.

davidpdrsn:
@jplatte it sounds like the next major version of multer (using http 1.0) could be a bit far off https://github.com/rousan/multer-rs/pull/59#issuecomment-1823493658 :/

To work around this in our multipart I think we should just eagerly convert the headers in the field to http 1.0. I've added that in b9983b6bc6d20. It is a bit unfortunate having to do that because of the extra allocations but at least its just temporary and should enable more people to upgrade.

axum-core/src/ext_traits/request.rs
```diff
@@ -258,13 +257,13 @@ pub trait RequestExt: sealed::Sealed + Sized {
 
     /// Apply the [default body limit](crate::extract::DefaultBodyLimit).
     ///
-    /// If it is disabled, return the request as-is in `Err`.
-    fn with_limited_body(self) -> Result<Request<Limited<Body>>, Request>;
+    /// If it is disabled, the request is return as-is.
```
jplatte, line 260:
```suggestion
    /// If it is disabled, the request is returned as-is.
```

axum-extra/src/extract/multipart.rs
```diff
@@ -118,7 +115,22 @@ impl Multipart {
...
 
         if let Some(field) = field {
-            Ok(Some(Field { inner: field }))
+            // multer still uses http 0.2 which means we cannot directly expose
+            // `multer::Field::headers`. Instead we have to eagerly convert the headers into http
+            // 1.0
+            //
+            // When the next major version of multer is published we can remove this.
+            let mut headers = HeaderMap::with_capacity(field.headers().len());
+            headers.extend(field.headers().clone().into_iter().map(|(name, value)| {
```
jplatte, line 124: No need to create a new `HeaderMap` and then immediately turn it into a destructuring iterator, right?

```suggestion
            headers.extend(field.headers().iter().cloned().map(|(name, value)| {
```

axum/CHANGELOG.md
```diff
@@ -61,11 +61,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - **fixed:** Allow unreachable code in `#[debug_handler]` ([#2014])
 - **change:** Update tokio-tungstenite to 0.19 ([#2021])
-- **change:** axum's MSRV is now 1.65 ([#2021])
+- **change:** axum's MSRV is now 1.66 ([#1882])
 - **added:** Implement `Handler` for `T: IntoResponse` ([#2140])
 - **added:** Implement `IntoResponse` for `(R,) where R: IntoResponse` ([#2143])
 - **changed:** For SSE, add space between field and value for compatibility ([#2149])
 - **added:** Add `NestedPath` extractor ([#1924])
+- **breaking:** `impl<T> IntoResponse(Parts) for Extension<T>` now requires `T:
+  Clone`, as that is required by the http crate ([#1882])
```
jplatte, line 70: Do multiline inline code blocks work in markdown? Seems safer without:

```suggestion
- **breaking:** `impl<T> IntoResponse(Parts) for Extension<T>` now requires
  `T: Clone`, as that is required by the http crate ([#1882])
```

axum/src/docs/routing/layer.md
```diff
@@ -32,20 +32,21 @@ If you only want middleware on some routes you can use [`Router::merge`]:
...
 
 let with_tracing = Router::new()
     .route("/foo", get(|| async {}))
     .layer(TraceLayer::new_for_http());
 
-let with_compression = Router::new()
+let with_timeout = Router::new()
     .route("/bar", get(|| async {}))
-    .layer(CompressionLayer::new());
+    .layer(TimeoutLayer::new(Duration::from_secs(10)));
```
jplatte, line 44: Since compression has been ported, do you want to revert the changes in this file?

axum/src/extract/multipart.rs
```diff
@@ -7,11 +7,12 @@ use crate::body::Bytes;
...
 use axum_core::__composite_rejection as composite_rejection;
 use axum_core::__define_rejection as define_rejection;
-use axum_core::body::Body;
 use axum_core::response::{IntoResponse, Response};
 use axum_core::RequestExt;
 use futures_util::stream::Stream;
 use http::header::{HeaderMap, CONTENT_TYPE};
+use http::HeaderName;
+use http::HeaderValue;
 use http::StatusCode;
```
jplatte, line 16: I know we don't maintain a consistent import granularity, but one line per import still looks weird 😄 

```suggestion
use http::{HeaderName, HeaderValue, StatusCode};
```

axum/src/extract/multipart.rs
```diff
@@ -91,8 +89,21 @@ impl Multipart {
             .map_err(MultipartError::from_multer)?;
 
         if let Some(field) = field {
+            // multer still uses http 0.2 which means we cannot directly expose
+            // `multer::Field::headers`. Instead we have to eagerly convert the headers into http
+            // 1.0
+            //
+            // When the next major version of multer is published we can remove this.
+            let mut headers = HeaderMap::with_capacity(field.headers().len());
+            headers.extend(field.headers().clone().into_iter().map(|(name, value)| {
```
jplatte, line 98:
```suggestion
            headers.extend(field.headers().iter().cloned().map(|(name, value)| {
```

axum/src/lib.rs
```diff
@@ -308,16 +308,12 @@
...
 //! tower = "<latest-version>"
 //! ```
 //!
-//! The `"full"` feature for hyper and tokio isn't strictly necessary but it's
-//! the easiest way to get started.
-//!
-//! Note that [`hyper::Server`] is re-exported by axum so if that's all you need
-//! then you don't have to explicitly depend on hyper.
+//! The `"full"` feature for tokio isn't strictly necessary but it's the easiest way to get
+//! started.
```
jplatte, line 316:
```suggestion
//! The `"full"` feature for tokio isn't necessary but it's the easiest way to get started.
```

axum/src/lib.rs
```diff
@@ -444,7 +440,7 @@ pub mod handler;
 pub mod middleware;
 pub mod response;
 pub mod routing;
-#[cfg(feature = "tokio")]
+#[cfg(all(feature = "tokio", feature = "http1"))]
 pub mod serve;
```
jplatte, line 444: http2-only is not supported?

axum/src/serve.rs
```diff
@@ -158,6 +131,49 @@ where
     }
 }
 
+#[derive(Debug, Copy, Clone)]
+struct TowerToHyperService<S> {
+    service: S,
+}
```
jplatte, line 137: Would be nice if this was part of an axum-independent utility crate (it would be applicable for stuff like the caniuse.rs xtask as well, right?)

jplatte (approved):

axum/src/extract/multipart.rs
davidpdrsn: I'm eagerly awaiting rustfmt stable to support that

axum/src/serve.rs
davidpdrsn: Yeah my plan is to upstream it to hyper-utils.

axum/src/lib.rs
davidpdrsn: Oh yes good catch! Change it to `#[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]`

PR #1888, opened 2023-03-27
Update allowed crates in public API

davidpdrsn:
`cargo-public-api-crates` now [detects re-exports](https://github.com/davidpdrsn/cargo-public-api-crates/pull/2) meaning we had to tweak some things here.

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -80,6 +80,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 allowed = [
     "axum",
     "axum_core",
+    "axum_macros",
     "bytes",
     "cookie",
     "futures_core",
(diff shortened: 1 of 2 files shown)


axum-extra/Cargo.toml
```diff
@@ -80,14 +80,14 @@ rustdoc-args = ["--cfg", "docsrs"]
...
     "axum_core",
+    "axum_macros",
     "bytes",
     "cookie",
     "futures_core",
     "futures_util",
     "http",
     "http_body",
     "hyper",
-    "percent_encoding",
```
davidpdrsn, line 90: Not sure why `percent_encoding` was here before. It's `#[doc(hidden)]` so shouldn't show up in the docs which `cargo-public-api-crates` generates. Might have been left over from a previous bug 🤷

davidpdrsn: `cargo-public-api-crates` probably should detect crates in this list which _aren't_ in the public API.

PR #1889, opened 2023-03-27
Add OptionalPath extractor

### LINKED ISSUE #1884: OptionalPath extractor

### Motivation

We've been getting many people asking about an easy way of extracting a path segment value only if a placeholder exists at the route the handler was called from.

### Proposal

Add an extractor `struct OptionalPath<T>(Option<T>);` that succeeds with `None` if there are no placeholders for the given route, and if there are placeholders, attempts to deserialize the arguments to `T` and return `Some(T)` on success, rejecting the request on failure.

### Alternatives

🤷🏼

jplatte:
## Motivation

Resolves #1884.

## Solution

What was already written on the issue 😄 

I also had to add `kind()` to `FailedToDeserializePathParams`. I don't think we should really have both `kind()` and `into_kind()` but I wasn't too sure what it should look like and went for roughly the path of least resistance (it would have been easier to return `&ErrorKind` but I'm guessing we don't want to guarantee that the rejection holds an `ErrorKind`).

### DIFF
--- a/axum-extra/src/extract/optional_path.rs
+++ b/axum-extra/src/extract/optional_path.rs
@@ -1,0 +1,103 @@
+use axum::{
+    async_trait,
+    extract::{path::ErrorKind, rejection::PathRejection, FromRequestParts, Path},
+    RequestPartsExt,
+};
+use serde::de::DeserializeOwned;
+
+/// Extractor that extracts path arguments the same way as [`Path`], except if there aren't any.
+///
+/// This extractor can be used in place of `Path` when you have two routes that you want to handle
+/// in mostly the same way, where one has a path parameter and the other one doesn't.
+///
+/// # Example
+///
+/// ```
+/// use std::num::NonZeroU32;
+/// use axum::{
+///     response::IntoResponse,
+///     routing::get,
+///     Router,
+/// };
+/// use axum_extra::extract::OptionalPath;
+///
+/// async fn render_blog(OptionalPath(page): OptionalPath<NonZeroU32>) -> impl IntoResponse {
+///     // Convert to u32, default to page 1 if not specified
+///     let page = page.map_or(1, |param| param.get());
+///     // ...
+/// }
+///
+/// let app = Router::new()
+///     .route("/blog", get(render_blog))
+///     .route("/blog/:page", get(render_blog));
+/// # let app: Router = app;
+/// # app.into_make_service();
+/// ```
+#[derive(Debug)]
+pub struct OptionalPath<T>(pub Option<T>);
+
+#[async_trait]
+impl<T, S> FromRequestParts<S> for OptionalPath<T>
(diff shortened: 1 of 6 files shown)


axum/src/extract/path/mod.rs
```diff
@@ -380,7 +380,14 @@ impl fmt::Display for ErrorKind {
 pub struct FailedToDeserializePathParams(PathDeserializationError);
 
 impl FailedToDeserializePathParams {
+    /// Get the underlying error kind.
+    pub fn kind(&self) -> ErrorKind {
+        self.0.kind.clone()
+    }
```
davidpdrsn, line 386: Any reason to not return a reference here? I guess it means we change the internals of `FailedToDeserializePathParams` but maybe that's okay

```suggestion
    pub fn kind(&self) -> &ErrorKind {
        &self.0.kind
    }
```

jplatte: If we commit to `FailedToDeserializePathParams` containing `ErrorKind`, do we keep `into_kind` or deprecate / later remove it? If we keep the `Clone` impl I added for making the current implementation of this method possible, `.kind().clone()` would be semantically equivalent to the current `.into_kind()` (though theoretically less efficient).

jplatte: Ping @davidpdrsn

davidpdrsn: I'm fine with committing to `ErrorKind` for now.

jplatte: Err, I don't see how that answers my question.. I guess I will just do the easiest thing, keep `into_kind`, remove the `Clone` impl and not deprecate anything.

davidpdrsn: Ah sorry. I thought you were asking about committing to having an `ErrorKind` inside `FailedToDeserializePathParams`, not whether to keep `into_kind` and `kind`. I'm fine with keeping both methods, like you've implemented it here.

axum-extra/src/extract/optional_path.rs
```diff
@@ -0,0 +1,103 @@
...
+///     // Convert to u32, default to page 1 if not specified
+///     let page = page.map_or(1, |param| param.get());
+///     // ...
+/// }
+///
+/// let app = Router::new()
+///     .route("/blog", get(render_blog))
+///     .route("/blog/:page", get(render_blog));
+/// # let app: Router = app;
+/// # app.into_make_service();
```
davidpdrsn, line 34: If the previous like works then so should this. Can probably just omit it.

```suggestion
```

davidpdrsn (commented):
Changelog?

davidpdrsn:
I'll do the last view changes now.

davidpdrsn (approved):

PR #1902, opened 2023-04-01
Update `rest-grpc-multiplex` example to include reflection

davidpdrsn:
This updates the `rest-grpc-multiplex` example to also include setup for
`tonic-reflection`. Unless you're familiar with all the pieces its not very
easy to do so figured it makes sense to include in the example.

Context: https://github.com/tokio-rs/axum/discussions/1840


jplatte (approved):
I know nothing about grpc / tonic, but doesn't look crazy 😅

PR #1919, opened 2023-04-09
Fix a typo inside handlers_intro.md

escwxyz:
## Motivation

Fix a small typo in https://docs.rs/axum/latest/axum/#handlers.

## Solution

`is` => `are`


davidpdrsn (approved):
Thanks!

PR #1920, opened 2023-04-09
Add WASM response type

LiamGallagher737:
## Description

Added a `Wasm` response type to `axum-extra` that automatically gets the `Content-Type: application/wasm` header.

## Motivation

I was returning WASM from an API and had to implment this myself.

## Solution

Pretty much the same as the `Html` response type but with the `Content-Type: application/wasm` header.

## Additional Info

As there is no `application/wasm` in the mime crate and as the crate is no longer maintained I decided to just go with a string literal.

### DIFF
--- a/axum-extra/src/response/wasm.rs
+++ b/axum-extra/src/response/wasm.rs
@@ -1,0 +1,26 @@
+use axum::response::{IntoResponse, Response};
+use bytes::Bytes;
+use http::header;
+use http_body::Full;
+
+/// A WASM response.
+///
+/// Will automatically get `Content-Type: application/wasm`.
+#[derive(Clone, Copy, Debug)]
+#[must_use]
+pub struct Wasm<T>(pub T);
+
+impl<T> IntoResponse for Wasm<T>
+where
+    T: Into<Full<Bytes>>,
+{
+    fn into_response(self) -> Response {
+        ([(header::CONTENT_TYPE, "application/wasm")], self.0.into()).into_response()
+    }
+}
+
+impl<T> From<T> for Wasm<T> {
+    fn from(inner: T) -> Self {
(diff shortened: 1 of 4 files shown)


davidpdrsn (commented):
Not sure this is worth having in the main axum crate. We can put it in axum-extra instead.

davidpdrsn (commented):
Nice! Wanna update the changelog as well?

LiamGallagher737:
Thank you for replying so quickly, don't think I've ever been reviewed so quick

axum-extra/src/response/wasm.rs
```diff
@@ -0,0 +1,26 @@
...
+/// A WASM response.
+///
+/// Will automatically get `Content-Type: application/wasm`.
+#[derive(Clone, Copy, Debug)]
+#[must_use]
+pub struct Wasm<T>(pub T);
+
+impl<T> IntoResponse for Wasm<T>
+where
+    T: Into<Full<Bytes>>,
```
jplatte, line 15: Do we have the same bound in another place already? Just wondering where it comes from.

LiamGallagher737: Copied from the Html IntoResponse implementation
https://github.com/tokio-rs/axum/blob/946d8c32530ae301bf26928ab22090b33366a748/axum/src/response/mod.rs#L45-L48

davidpdrsn: I've actually been thinking that we could make these types work for any `T: IntoResponse`. Then it wouldn't be required to have the contents in memory.

I can that in a separate PR.

davidpdrsn:
Superseded by https://github.com/tokio-rs/axum/pull/1921. Thanks!

PR #1921, opened 2023-04-09
Add `Html`, `Css`, `JavaScript`, and `Wasm` response types

davidpdrsn:
I think using `T: IntoResponse` is more flexible than [the previous bound](https://github.com/tokio-rs/axum/blob/main/axum/src/response/mod.rs#L47). For example you can response with the contents of a file with

```rust
Router::new().route(
    "/",
    get(|| async {
        let Ok(file) = tokio::fs::File::open("index.html").await else {
            return Err(StatusCode::INTERNAL_SERVER_ERROR);
        };
        Ok(Html(AsyncReadBody::new(file)))
    }),
);
```

Also added css, javascript, and wasm, because why not.

We can move `Html` into axum in 0.7.

Supersedes https://github.com/tokio-rs/axum/pull/1920

### DIFF
--- a/axum-extra/src/response/mod.rs
+++ b/axum-extra/src/response/mod.rs
@@ -9,3 +9,82 @@ pub use erased_json::ErasedJson;
 #[cfg(feature = "json-lines")]
 #[doc(no_inline)]
 pub use crate::json_lines::JsonLines;
+
+macro_rules! mime_response {
+    (
+        $(#[$m:meta])*
+        $ident:ident,
+        $mime:ident,
+    ) => {
+        mime_response! {
+            $(#[$m])*
+            $ident,
+            mime::$mime.as_ref(),
+        }
+    };
+
+    (
+        $(#[$m:meta])*
+        $ident:ident,
+        $mime:expr,
+    ) => {
+        $(#[$m])*
+        #[derive(Clone, Copy, Debug)]
+        #[must_use]
+        pub struct $ident<T>(pub T);
+
+        impl<T> axum::response::IntoResponse for $ident<T>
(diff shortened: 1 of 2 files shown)


LiamGallagher737:
I just made a crate with a bunch of content types, maybe I could add them here. Most of them are too specific to have here but most of the image ones and files like pdf would be nice to have.
https://github.com/LiamGallagher737/more_axum_responses/blob/main/src/lib.rs

davidpdrsn:
> but most of the image ones and files like pdf would be nice to have

As long they're just wrappers around plain data and don't require other dependencies I'm fine with it. But yeah some of those are quite specialized and probably not worth having.

Can you make a follow up pr when this is merged?

axum-extra/src/response/mod.rs
```diff
@@ -9,3 +9,82 @@ pub use erased_json::ErasedJson;
...
+
+        impl<T> axum::response::IntoResponse for $ident<T>
+        where
+            T: axum::response::IntoResponse,
+        {
+            fn into_response(self) -> axum::response::Response {
+                (
+                    [(
+                        http::header::CONTENT_TYPE,
+                        http::HeaderValue::from_static($mime),
```
jplatte, line 44: Maybe put in a `const` so it's instantiated at compile time?

davidpdrsn: Can't do that since `$mime` isn't `const` because we it from [`mime.as_ref()`](https://docs.rs/mime/latest/mime/struct.Mime.html#impl-AsRef%3Cstr%3E-for-Mime).

Issue #1858, opened 2023-03-17
Support relative redirect or reading its ancestor path from inside a nested `Router` or `handler`


dulguun0225:
### Discussed in https://github.com/tokio-rs/axum/discussions/1854

<div type='discussions-op-text'>

<sup>Originally posted by **dulguun0225** March 15, 2023</sup>
I wanna redirect to one of my own paths without hardcoding what my ancestor path is.

```Rust
use std::net::SocketAddr;

use axum::{response::Redirect, routing::get, Router};

fn users_router() -> Router {
    Router::new()
        .route("/", get(|| async { "users index" })) 
        .route("/fallback", get(|| async { "users fallback"}))
        // make it redirect without hardcoding /users part.
        .fallback(|| async { Redirect::to("/users/fallback") })
}

#[tokio::main]
async fn main() {
    let app = Router::new().nest("/users", users_router());
    
    axum::Server::bind(&SocketAddr::from(([127, 0, 0, 1], 5000)))
        .serve(app.into_make_service())
        .await
        .unwrap()
}
```

Maybe I could pass ancestor path to the `users_router` function. But it could get tedious when things are nested deeply.</div>

dulguun0225:
Maybe a path extractor would do? How one would implement it?

davidpdrsn:
I guess nesting a router could set some extension, which handlers can then extract. Similar to how `MatchedPath` works.

ttys3:
this is important when impl a otel tracing middleware.

https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#http-server-semantic-conventions

now we can not get route name in nested route.

related issue: https://github.com/davidB/axum-tracing-opentelemetry/pull/54#issuecomment-1493376105

Issue #1731, opened 2023-02-02
ServeDir in nested route causes invalid redirects


yu-re-ka:
- [x] I have looked for existing issues (including closed) about this

### Version
```
├── axum v0.5.13
│   ├── axum-core v0.2.7
```

```
│   ├── tower-http v0.3.4
└── tower-http v0.3.4 (*)
```
-->

### Platform

`Linux yaya 6.1.7 #1-NixOS SMP PREEMPT_DYNAMIC Wed Jan 18 10:58:34 UTC 2023 x86_64 GNU/Linux`

### Description

https://yaya.yuka.dev/ is running a axum based webserver with a `axum::Router`. It has a `nest` route for `/public` that contains a `ServeDir` service with path `/mnt/public/`.

Now, in `/mnt/public/`, there exists an empty directory `/mnt/public/foo`.
I request `https://yaya.yuka.dev/public/foo`
And I get:
```
< HTTP/2 307 
< location: https://yaya.yuka.dev/foo/
```
while I would have expected a redirect to `https://yaya.yuka.dev/public/foo/`

This redirect is created [here](https://github.com/tower-rs/tower-http/blob/master/tower-http/src/services/fs/serve_dir/future.rs#L102) in ServeDir, and it uses the stripped path instead of the original uri to determine the redirect location.

Unsure how to solve this, maybe the StripPrefix middleware and OriginalUri could be moved to tower_http and then tower_http can try to use the OriginalUri first and then fall back to the req.uri()

davidpdrsn:
> maybe the StripPrefix middleware and OriginalUri could be moved to tower_http and then tower_http can try to use the OriginalUri first and then fall back to the req.uri().

That seems like the only viable solution to me. It just feels wrong to move axum specific stuff into tower-http to solve an, arguably, axum bug. `StripPrefix` also supports handling URL params for things like `.nest("/:id", _)`. tower-http doesn't have anything like that currently. So I'd like to avoid moving things if at all possible. Will require some more thinking.

jplatte:
Yeah, this is definitely hard. We couldn't turn `OriginalUri` and the `http::Request` path modifications on its head either (putting the modified, rather than original path in an extension instead), because then `ServeDir` won't be able to find its files.

Maybe the best solution to be had is `ServeDir` getting a new knob where you can tell it that it's mounted at some non-root path?

davidpdrsn:
[...]

PR #1924, opened 2023-04-11
Add `NestedPath`

davidpdrsn:
This adds `NestedPath` which is an extractor that gives the path the matched route is nested at.

I consider this mostly a lower level building block. Wanna add a convenience helper to `Redirect` that handles it more conveniently. Perhaps also a middleware that automatically adds it to relative redirects, that could be used with `ServeDir`.

Fixes https://github.com/tokio-rs/axum/issues/1858
Part of https://github.com/tokio-rs/axum/issues/1731


jplatte (approved):
I guess this was written before the fix for `MatchedPath` not being available on middleware that was applied to a nested router, right? I think there might be some code to share between that and this feature. Anyways, code and tests look good.

davidpdrsn:
I've redone this on top of the changes on main.

PR #1929, opened 2023-04-12
Enable `println` warning

davidpdrsn:
Turns out there is a clippy lint for usage of `println!`. We should probably enable that.

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -106,6 +106,8 @@ where
     /// Unless you add additional routes this will respond with `404 Not Found` to
     /// all requests.
     pub fn new() -> Self {
+        println!("prints not allowed. Build should fail");
+
         let mut this = Self {
             path_router: Default::default(),
             fallback_router: Default::default(),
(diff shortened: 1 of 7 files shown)


axum/src/routing/mod.rs
```diff
@@ -106,6 +106,8 @@ where
     /// Unless you add additional routes this will respond with `404 Not Found` to
     /// all requests.
     pub fn new() -> Self {
+        println!("prints not allowed. Build should fail");
```
davidpdrsn, line 109: Gonna verify that this causes CI to fail.

davidpdrsn:
We have that enabled as well but probably could be allowed in tests.

CI didn't fail though so have to tweak something.

jplatte:
Does it fail with `dbg!`? Maybe you "just" have to update your muscle memory to that? 😅

davidpdrsn:
`dbg!` didn't fail CI either because we deny warnings. But I think the setup I have now is pretty decent. `dbg!` and `println!` are just warnings locally, but always allowed in tests. Any outside of tests will make CI fail.

I do like having `dbg!` and `println!` left in tests for easy debugging if a test starts failing.

jplatte (approved):

PR #1939, opened 2023-04-15
Use axum-macros/rust-toolchain file in ci

tottoto:
## Motivation

Improves the maintainability of the ci.   

## Solution

Uses the axum-macros/rust-toolchain file in the ci.


davidpdrsn (approved):
Didn't know you could do this. Thanks!

PR #1956, opened 2023-04-22
Don't inherit fallbacks when using `Router::nest_service`

davidpdrsn:
I think it makes sense to only inherit fallbacks for routers nested with
`Router::nest` and not `Router::nest_service`. It also simplifies the
code and avoids some unfortunate extension work arounds.


jplatte (approved):
I thought this was how we already did fallback inheritance 😅

Makes a lot of sense not to do it for `nest_service`.

PR #1970, opened 2023-04-28
Fix typo in json_lines.rs

eric-seppanen:
I *think* that this meant to say "Marker type", not "Maker type".

## Motivation

It's in the public crate docs, so it's more visible than your average typo.

## Solution

This is my guess as to what it meant to say.  Feel free to close if I'm wrong.


jplatte (approved):

davidpdrsn (approved):
Thanks!

PR #1984, opened 2023-05-03
Fix localhost vs 0.0.0.0 discrepancy

mattfbacon:
#1982

### DIFF
--- a/axum-macros/src/lib.rs
+++ b/axum-macros/src/lib.rs
@@ -426,7 +426,7 @@ pub fn derive_from_request_parts(item: TokenStream) -> TokenStream {
 /// async fn main() {
 ///     let app = Router::new().route("/", get(handler));
 ///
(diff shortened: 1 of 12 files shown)


davidpdrsn:
I don't think we should change it everywhere. When you bind on a cloud host you often have to bind to `0.0.0.0` and `127.0.0.1` won't work. I've had that experience on Heroku and heard the same from others. So because people copy/paste our examples I don't think we should change it everywhere.

jplatte:
I think we should actually not change the listen address anywhere, instead we should update the comment(s).

davidpdrsn:
I agree!

mattfbacon:
I agree that people often copy example code. For this reason I find 0.0.0.0 to be an insecure default.

That said, changing the comments is fine too, and maybe adding a clear message about the implications would be enough to resolve the aforementioned issue.

mattfbacon:
Thanks for squashing

PR #2025, opened 2023-05-29
docs: add note about large file multipart uploads

wowkster:
## Motivation

When using the Multipart extractor in a project, I quickly ran into the problem of my requests failing due to the upload size being too large. To find a solution I had to google around to find that the default body limit is set to 2MB. I think that mentioning this in the documentation for the Multipart extractor would be beneficial and provide some clarity for those not familiar with DefaultBodyLimit or aware that it even exists.

## Solution

The Multipart extractor is commonly used for file uploads, but it is not mentioned directly in the relevant documentation that these uploads are affected by the global request body limit. This pull request adds a note in the Multipart documentation that links to DefaultBodyLimit.

### DIFF
--- a/axum/src/extract/multipart.rs
+++ b/axum/src/extract/multipart.rs
@@ -53,8 +53,8 @@ use std::{
 /// 
 /// # Large Files
 /// 
-/// For security reasons, by default, all axum request bodies (including multipart uploads) are limited to 2MB. 
-/// To upload files larger than this, use [DefaultBodyLimit][default-body-limit].
+/// For security reasons, by default, `Multipart` limits the request body size to 2MB. 
+/// See [`DefaultBodyLimit`][default-body-limit] for how to configure this limit.
 /// 
 /// [default-body-limit]: crate::extract::DefaultBodyLimit
 #[cfg_attr(docsrs, doc(cfg(feature = "multipart")))]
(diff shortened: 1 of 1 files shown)


axum/src/extract/multipart.rs
```diff
@@ -50,6 +50,13 @@ use std::{
 /// let app = Router::new().route("/upload", post(upload));
 /// # let _: Router = app;
 /// ```
+/// 
+/// # Large Files
+/// 
+/// For security reasons, by default, all axum request bodies (including multipart uploads) are limited to 2MB. 
+/// To upload files larger than this, use [DefaultBodyLimit][default-body-limit].
+/// 
+/// [default-body-limit]: crate::extract::DefaultBodyLimit
```
davidpdrsn, line 59: I think I'd rephrase this slightly. "all axum request bodies" isn't technically true since extractors themselves have to enforce the limit. For example extracting `Request<Body>` and then consuming the body wont consider the default limit. You have to do that yourself.

Multipart enforces the limit [here](https://github.com/tokio-rs/axum/blob/main/axum-extra/src/extract/multipart.rs#L102-L105)

```suggestion
/// # Large Files
/// 
/// For security reasons, by default, `Multipart` limits the request body size to 2MB. 
/// See [`DefaultBodyLimit`][default-body-limit] for how to configure this limit.
/// 
/// [default-body-limit]: crate::extract::DefaultBodyLimit
```

davidpdrsn (approved):

PR #2030, opened 2023-06-02
Fix `.source()` of composite rejections

SabrinaJewson:
## Motivation

Currently the error chain of composite rejections has the issue of containing multiple items that `Display` to the same thing, which looks weird in error reports (e.g. “Expected request with `Content-Type: application/json`: Expected request with `Content-Type: application/json`”).

## Solution

Forward `.source()` to the inner type instead of giving the inner error directly.


SabrinaJewson:
Doesn’t look like CI failures are related?

davidpdrsn (approved):
Thanks!

I'll look at CI separately.

PR #2035, opened 2023-06-08
Implement `IntoResponse` for boxed slices

alvra:
This change makes it possible to create responses from boxed slices: `Box<[u8]>` and `Box<str>`.


davidpdrsn (commented):
Thanks!

Code looks good. Do you wanna update the changelog as well?

Issue #2046, opened 2023-06-18
anyhow-based error handling example should be advertised more broadly


vi:
- [X] I have looked for existing issues (including closed) about this

### Motivation

Primary place where I end up when looking how to make fallible handlers is [`axum::error_handling` module documentation page](https://docs.rs/axum/latest/axum/error_handling/index.html). It even contains mentions of "anyhow" in "Routing to fallible services" section, but it is not clear how just upgrade existing infallible handlers to anyhow-fallible ones. 

Fortunately, there is [already an example](https://github.com/tokio-rs/axum/pull/1398) of how to wrap Anyhow errors for Axum consumption (even with a properly working `?` operator), but it is not obvious how to get there.

### Proposal

Include a link to [anyhow-error-response](https://github.com/tokio-rs/axum/blob/main/examples/anyhow-error-response/src/main.rs) in `axum::error_handling` module documentation, e.g. in "axum’s error handling model" section.

### Alternatives

* Inline the example of anyhow-wrapping `AppError` directly in the docs (instead of a link).
* Maybe also introduce "Routing to fallible handlers" section after "model", before "fallible services" which is oriented on practical approaches of making fallible handlers, with only theoretical things retained in the "model" section.

davidpdrsn:
Like described in the error handling docs axum does _not_ consider a handler returning `Result<_, _>` as fallible. Thats just something that'll be converted into a response. Only service level errors are considered errors since they don't send back a response at all and instead hyper terminates the connection.

With that said I think it's fine to link to the anyhow example from the docs. Do you want to submit a PR for that?

vi:
> Like described in the error handling docs axum does not consider a handler returning Result<_, _> as fallible. Thats just something that'll be converted into a response. Only service level errors are considered errors since they don't send back a response at all and instead hyper terminates the connection.

Do I understand that there can be two meanings of the word "fallible" in this context:

* "fallible service", in `tower_service::Service<Error != Infallible>` sense. This is banned for `axum::handler::Handler`s, as they do not have error type inside. So "real", hard errors should be already handled within a Hander before it reaches the router.
* "fallible endpoint/handler", meaning request that can fail in user code (because it uses fallible operations inside and contains `?` operators). Router does not "see" those errors, as they are converted to responses just like successful outcomes.

> With that said I think it's fine to link to the anyhow example from the docs. Do you want to submit a PR for that?

Maybe, but I want to ensure I understand it properly myself first.  
[...]

PR #2049, opened 2023-06-20
docs: Add links to more examples of Result-returning handlers.

vi:
Resovles #2046.

r? @davidpdrsn

### DIFF
--- a/axum/src/docs/error_handling.md
+++ b/axum/src/docs/error_handling.md
@@ -38,6 +38,16 @@ It doesn't matter whether you return `Err(StatusCode::NOT_FOUND)` or
 `Err(StatusCode::INTERNAL_SERVER_ERROR)`. These are not considered errors in
 axum.
 
+Instead of a direct `StatusCode`, it makes sense to use intermediate error type
+that can ultimately be converted to `Reponse`. This allows using `?` operator
+in handlers. See those examples:
+
+* [`anyhow-error-response`][anyhow] for generic boxed errors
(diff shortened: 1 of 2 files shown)


axum/src/handler/mod.rs
```diff
@@ -30,6 +30,9 @@
 //!         Err(StatusCode::BAD_REQUEST)
 //!     }
 //! }
+//!
+//! // See "Error handling" section of the documentation for more examples of
+//! // of Result-returning handlers.
```
davidpdrsn, line 35: I think we should just link to the two examples again here, instead of making people go find it themselves. People might also get confused about "service level errors" which might be relevant to them yet.

vi: Copied the text, now it is inline in handler docs.

davidpdrsn (approved):
Thanks!

axum/src/docs/error_handling.md
```diff
@@ -38,6 +38,16 @@ It doesn't matter whether you return `Err(StatusCode::NOT_FOUND)` or
 `Err(StatusCode::INTERNAL_SERVER_ERROR)`. These are not considered errors in
 axum.
 
+Instead of a direct `StatusCode`, it makes sense to use intermediate error type
+that can ultimately be converted to `Reponse`. This allows using `?` operator
```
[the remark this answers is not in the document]
davidpdrsn: Thanks for the heads up! Fixed in cc9629fe88d7f

PR #2051, opened 2023-06-20
docs: update router::route multiple methods docs

mohad12211:
## Motivation
In the early days you had to add handlers at the same time if they are for the same route with different methods,
but that hasn't been the case for a while.
Now you can do both but the docs are outdated.

## Solution
Change `must` to `can`, and add a note that the one-by-one way is also possible.

[Relevant discussion](https://github.com/tokio-rs/axum/discussions/2050)


davidpdrsn (approved):
Thanks!

Issue #2048, opened 2023-06-20
Consider adding TryFrom<&Uri> for Query


CosmicHorrorDev:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

Consider adding `impl TryFrom<&Uri> for axum::extract::Query`

### Motivation

I was porting a project from `rocket` and `axum` and accidentally regressed some endpoints from differences in query parsing. It would have been nice to have a simple way to test the query parsing in isolation, but the simplest solution I could come up with was to just copy the internals of `Query`'s `FromRequestParts`

### Proposal

Move the logic from `Query`'s `FromRequestParts` implementation into `TryFrom<&Uri> for Query` and call `Query::try_from(&parts.uri)` in `from_request_parts()`

Only drawback I can think of is that this adds some slight noise to the API

### Alternatives

None that seem compelling

davidpdrsn:
Sure! That sounds like a fine addition to me.

davidpdrsn:
Fixed in https://github.com/tokio-rs/axum/pull/2058

PR #2058, opened 2023-06-24
Add `axum::extract::Query::try_from_uri`

dobecad:
## Motivation

Quoting from #2048:
"I was porting a project from rocket and axum and accidentally regressed some endpoints from differences in query parsing. It would have been nice to have a simple way to test the query parsing in isolation, but the simplest solution I could come up with was to just copy the internals of Query's FromRequestParts"

## Solution

The solution is the same as the `FromRequestParts for Query` impl, as recommended by @CosmicHorrorDev.


axum/src/extract/query.rs
```diff
@@ -137,4 +151,35 @@ mod tests {
...
+        #[derive(Deserialize)]
+        struct TestQueryParams {
+            foo: String,
+            bar: u32,
+        }
+        let uri: Uri = "http://example.com/path?foo=hello&bar=42".parse().unwrap();
+        let query: Result<Query<TestQueryParams>, _> = Query::try_from(&uri);
+        assert!(query.is_ok());
+
+        let query = query.unwrap();
```
CosmicHorrorDev, line 166:
```suggestion
        let query: Query<TestQueryParams> = Query::try_from(&uri).unwrap();
```

Nit: an `.unwrap()` is essentially equivalent to `assert!(query.is_ok())`

CosmicHorrorDev (commented):
Just a really small nit, but everything else LGTM

Thanks for picking up this change!

axum/src/extract/query.rs
dobecad: Gotcha, that makes sense

jplatte:
I'm wondering if this should really be a `TryFrom` impl, rather than an inherent function. It doesn't feel like a straight-up "conversion" to me, and I don't see why one would ever want to use `Query<_>` for a generic function accepting `TryFrom<&Uri>` (or why one would write such a function in the first place).

I think we may also want to add inherent sync functions for other extractors that don't do any `.await`ing in their `FromRequestParts` implementation (`Path` and `axum-extra`s `Query` come to mind). But that can be left up to future PRs.

davidpdrsn:
> I'm wondering if this should really be a TryFrom impl, rather than an inherent function. It doesn't feel like a straight-up "conversion" to me, and I don't see why one would ever want to use Query<_> for a generic function accepting TryFrom<&Uri> (or why one would write such a function in the first place).

I agree with that. I think an inherent function would be more appropriate.

dobecad:
Just to check if I am understanding correctly, rather than a `TryFrom`, this functionality should just be a function within an `impl Query` block?

davidpdrsn:
Exactly.

axum/src/extract/query.rs
```diff
@@ -62,6 +62,37 @@ where
...
+    /// let result: Query<ExampleParams> = Query::from_uri(&uri).unwrap();
+    /// assert_eq!(result.foo, String::from("hello"));
+    /// assert_eq!(result.bar, 42);
+    /// ```
+    pub fn from_uri(value: &Uri) -> Result<Self, QueryRejection> {
+        let query = value.query().unwrap_or_default();
+        let params: T =
+            serde_urlencoded::from_str(query).map_err(FailedToDeserializeQueryString::from_err)?;
+        Ok(Query(params))
+    }
```
davidpdrsn, line 93: I think we should still include the `try_` prefix in the name.

```suggestion
    pub fn try_from_uri(value: &Uri) -> Result<Self, QueryRejection> {
        let query = value.query().unwrap_or_default();
        let params =
            serde_urlencoded::from_str(query).map_err(FailedToDeserializeQueryString::from_err)?;
        Ok(Query(params))
    }
```

We should also call this from `Query::from_request`.

dobecad: Just to check if I understand, do you also want me to implement the `FromRequest` trait for `Query`?

davidpdrsn: Sorry no, `FromRequestParts`. The implementation is currently duplicated. It shouldn't be.

davidpdrsn (commented):
Code looks good! Wanna add a note to the changelog as well? Then I think its good to merge :)

axum/CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **breaking:** Only inherit fallbacks for routers nested with `Router::nest`.
   Routers nested with `Router::nest_service` will no longer inherit fallbacks ([#1956])
 - **fixed:** Don't remove the `Sec-WebSocket-Key` header in `WebSocketUpgrade` ([#1972])
+- **added:** Extract `axum::extract::Query` from `&http::Uri` ([#2058])
```
davidpdrsn, line 56: I think this should just say

```suggestion
- **added:** Add `axum::extract::Query::try_from_uri` ([#2058])
```

Since "extract" refers to extractors and `Query` was already an extractor.

davidpdrsn (approved):
Thanks!

Issue #2039, opened 2023-06-09
Remove `unwrap`s from examples, where feasible


erlangparasu:
### Motivation
For production ready sample

davidpdrsn:
Which examples?

erlangparasu:
@davidpdrsn All examples :)

Ptrskay3:
Well, I'm not sure I agree.

The examples generally are to show you how does a particular feature works, and not supposed to teach you Rust basics about error handling or other miscellaneous topics. Making all the examples production ready and robust would take the focus away from the actual feature it tries to present. 

If you're looking for a more advanced code sample, take a look at [realworld-axum-sqlx](https://github.com/launchbadge/realworld-axum-sqlx) for example.

Altair-Bueno:
> The examples generally are to show you how does a particular feature works, and not supposed to teach you Rust basics about error handling or other miscellaneous topics. Making all the examples production ready and robust would take the focus away from the actual feature it supposed to present.

Unfortunately, people are used to just copy/paste and carry on with their lives. [We already make sure the right status codes are being used for this exact reason](https://github.com/tokio-rs/axum/pull/1276#discussion_r949514713), so I would consider this request sensible. Also, a more rusty (but basic) error handling can be archived with `Box<dyn Error>` and the question mark operator. At least the compiler would be able to check silly panics in case of copy/paste.

Thankfully there are "only" 271 instances of unwrap/expect around the examples dir. rustdoc examples should be changed too, which im afraid I don't have a regex around to count them. I would consider this to be a low priority, good first issue to leave open and let someone pick up.

<details><summary>Grep command for examples dir</summary>
<p>

```sh
$ grep -rnw 'examples' -e '\(unwrap\|expect\)(' | wc -l
271
```

</p>
</details>

davidpdrsn:
The goal of the examples is not to showcase production ready setups. Like people have mentioned that's would make the examples much less focused on what they're actually trying to show.

With that said, there are probably some unwraps that could be removed. Happy to take PRs that cleans that up!

davidpdrsn:
I have removed some unnecessary unwraps from example-stream-to-file and example-reverse-proxy. There are still some left in example-oauth that would be nice to clean up. So if someone wants to submit a PR that'd be much appreciated.

The remaining examples are fine so I'll close this for now.

@erlangparasu if there are any specific examples you want clarity on please open a new issue and be specific about what you'd like to see clarified.

[...]

PR #2069, opened 2023-07-02
Remove unwraps via '?' with anyhow crate for example-oauth

rsdlt:
Refs: #2039

## Motivation
Substitutes `unwrap` calls with proper error handling via the `anyhow` crate for `example-oauth`.

## Solution
Substitute `unwrap`s with `?` via `anyhow`. Using `if let` or `match` is very verbose and makes the example lose focus. Using `?` keeps the example readable, focused and shows better practices.  

Validated solution by compiling and testing via `CLIENT_ID=REPLACE_ME CLIENT_SECRET=REPLACE_ME cargo run -p example-oauth`

### DIFF
--- a/examples/oauth/src/main.rs
+++ b/examples/oauth/src/main.rs
@@ -31,7 +31,7 @@ use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 static COOKIE_NAME: &str = "SESSION";
 
 #[tokio::main]
-async fn main() -> Result<(), AppError> {
+async fn main() {
     tracing_subscriber::registry()
         .with(
             tracing_subscriber::EnvFilter::try_from_default_env()
@@ -42,7 +42,7 @@ async fn main() -> Result<(), AppError> {
 
     // `MemoryStore` is just used as an example. Don't use this in production.
     let store = MemoryStore::new();
-    let oauth_client = oauth_client()?;
+    let oauth_client = oauth_client().unwrap();
     let app_state = AppState {
         store,
         oauth_client,
@@ -73,8 +73,6 @@ async fn main() -> Result<(), AppError> {
         .await
         .context("failed to serve service")
         .unwrap();
-
-    Ok(())
 }
 
 #[derive(Clone)]
@@ -169,14 +167,9 @@ async fn logout(
     State(store): State<MemoryStore>,
     TypedHeader(cookies): TypedHeader<headers::Cookie>,
 ) -> Result<impl IntoResponse, AppError> {
-    let cookie = match cookies.get(COOKIE_NAME) {
-        Some(cookie) => cookie,
-        None => {
-            return Err(AppError(anyhow::Error::msg(
-                "unexpected error getting cookie name",
-            )))
-        }
-    };
+    let cookie = cookies
+        .get(COOKIE_NAME)
+        .context("unexpected error getting cookie name")?;
 
     let session = match store
         .load_session(cookie.to_string())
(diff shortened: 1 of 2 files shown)


examples/oauth/src/main.rs
```diff
@@ -55,11 +56,11 @@ async fn main() {
...
-    let listener = tokio::net::TcpListener::bind("127.0.0.1:3000")
-        .await
-        .unwrap();
-    tracing::debug!("listening on {}", listener.local_addr().unwrap());
-    axum::serve(listener, app).await.unwrap();
+    let listener = tokio::net::TcpListener::bind("127.0.0.1:3000").await?;
+    tracing::debug!("listening on {}", listener.local_addr()?);
+    axum::serve(listener, app).await?;
+
+    Ok(())
```
davidpdrsn, line 63: I'd still keep the unwraps in main. That way you get a backtrace that points exactly to which line failed. You're also gonna see the errors when you run the app immediately, so returning a result doesn't make much difference.

examples/oauth/src/main.rs
```diff
@@ -256,3 +269,30 @@ where
...
+// For a simplified example of using anyhow in axum check /examples/anyhow-error-response
+#[derive(Debug)]
+struct AppError(anyhow::Error);
+
+// Tell axum how to convert `AppError` into a response.
+impl IntoResponse for AppError {
+    fn into_response(self) -> Response {
+        (
+            StatusCode::INTERNAL_SERVER_ERROR,
+            format!("Something went wrong: {}", self.0),
```
davidpdrsn, line 283: I don't think we should return the error here. That could reveal private details about the application. Instead we should return a generic "Something went wrong" body and log the error.

I also think the error should be logged as `tracing::error!("{:#}", self.0)`, specifically using `{:#}` as documented [here](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations).

davidpdrsn (changes requested):
Thanks for working on this!

In general I think we should also use `.context("...")?` to make it clear where the error is coming from. Otherwise it can be hard to debug later.

examples/oauth/src/main.rs
rsdlt: Solid feedback, appreciated!

examples/oauth/src/main.rs
```diff
@@ -256,3 +304,32 @@ where
...
+struct AppError(anyhow::Error);
+
+// Tell axum how to convert `AppError` into a response.
+impl IntoResponse for AppError {
+    fn into_response(self) -> Response {
+        tracing::error!("Application error: {:#}", self.0);
+
+        (
+            StatusCode::INTERNAL_SERVER_ERROR,
+            "Something went wrong".to_string(),
```
davidpdrsn, line 320: nit: You don't need to call `.to_string()` here since `IntoResponse` is implemented for `&'static str`.

```suggestion
            "Something went wrong",
```

examples/oauth/src/main.rs
```diff
@@ -191,26 +223,42 @@ async fn login_authorized(
...
+        .await
+        .context("failed to store session")?
+    {
+        Some(cookie) => cookie,
+        None => {
+            return Err(AppError(async_session::Error::msg(
+                "unexpected error retrieving cookie value",
+            )))
+        }
+    };
```
davidpdrsn, line 249: You can actually call `.context()` on an `Option` to turn it into a result, so this should work

```suggestion
        .context("failed to store session")?
        .context("unexpected error retrieving cookie value")?;
```

examples/oauth/src/main.rs
```diff
@@ -151,17 +168,32 @@ async fn protected(user: User) -> impl IntoResponse {
...
-    let session = match store.load_session(cookie.to_string()).await.unwrap() {
+) -> Result<impl IntoResponse, AppError> {
+    let cookie = match cookies.get(COOKIE_NAME) {
+        Some(cookie) => cookie,
+        None => {
+            return Err(AppError(anyhow::Error::msg(
+                "unexpected error getting cookie name",
+            )))
+        }
+    };
```
davidpdrsn, line 179:
```suggestion
    let cookie = match cookies.get(COOKIE_NAME).context("unexpected error getting cookie name")?;
```

examples/oauth/src/main.rs
```diff
@@ -30,7 +31,7 @@ use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 static COOKIE_NAME: &str = "SESSION";
 
 #[tokio::main]
-async fn main() {
+async fn main() -> Result<(), AppError> {
```
davidpdrsn, line 34: I still think we should `unwrap` the results in main and not return a `Result` at all.

```suggestion
async fn main() {
```

davidpdrsn (approved):
Thanks!

PR #2073, opened 2023-07-04
Remove `bitflags` crate dependency

rsdlt:
Implement our own `MethodFilter` and `EventFlags` types and remove `bitflags` dependency.

## Motivation
Remove dependency from `bitflags` crate by creating our own types as discussed [here](https://github.com/tokio-rs/axum/pull/2067#issuecomment-1616483907)

## Solution

- Remove `bitflags` crate from `axum/Cargo.toml`
- Create `method_filter/MethodFilter` and `response/EventFlag` types following a similar structure as that of the `bitflags` macro expansion
- No other code modifications
- Implement only required methods to compile and pass tests 

Fixes https://github.com/tokio-rs/axum/issues/2066

### DIFF
--- a/axum/src/routing/method_filter.rs
+++ b/axum/src/routing/method_filter.rs
@@ -1,4 +1,3 @@
-// use bitflags::bitflags;
 use http::Method;
 use std::{
     fmt,
--- a/axum/src/routing/method_filter.rs
+++ b/axum/src/routing/method_filter.rs
@@ -28,24 +28,18 @@ impl MethodFilter {
 
     const fn bits(&self) -> u16 {
         let bits = self;
-        {
-            bits.0
-        }
+        bits.0
     }
 
     const fn from_bits(bits: u16) -> Self {
         let bits = bits;
-        {
-            Self(bits)
-        }
+        Self(bits)
     }
 
     pub(crate) const fn contains(&self, other: Self) -> bool {
         let same = self;
         let other = other;
-        {
-            same.bits() & other.bits() == other.bits()
-        }
+        same.bits() & other.bits() == other.bits()
     }
 }
 impl std::ops::BitOr for MethodFilter {
--- a/axum/src/routing/method_filter.rs
+++ b/axum/src/routing/method_filter.rs
@@ -41,12 +41,10 @@ impl MethodFilter {
         let other = other;
         same.bits() & other.bits() == other.bits()
     }
-}
-impl std::ops::BitOr for MethodFilter {
-    type Output = Self;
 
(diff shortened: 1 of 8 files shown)


axum/src/routing/method_filter.rs
```diff
@@ -1,29 +1,59 @@
-use bitflags::bitflags;
+// use bitflags::bitflags;
```
jplatte, line 1: Should be removed.

jplatte (approved):
Thanks! I won't merge yet since David might have a different opinion on how this should be done, but I think this looks good.

axum/src/routing/method_filter.rs
```diff
@@ -1,29 +1,58 @@
...
+        }
+    }
+
+    pub(crate) const fn contains(&self, other: Self) -> bool {
+        let same = self;
+        let other = other;
+        {
+            same.bits() & other.bits() == other.bits()
+        }
+    }
```
davidpdrsn, line 49: Mostly because I'm curious but why the extra scopes on these functions?

axum/src/routing/method_filter.rs
```diff
@@ -1,29 +1,58 @@
...
+}
+impl std::ops::BitOr for MethodFilter {
```
davidpdrsn, line 51: Having this impl almost leaks an implementation detail. What do you think about replacing it with a normal public method? That could probably also have a better name and easier to find docs.

axum/src/routing/method_filter.rs
rsdlt: I forgot to clean those, they're not necessary. I just removed them in the latest commit.

axum/src/routing/method_filter.rs
rsdlt: I think that's a good idea. 

I tried it like this, and compiles and passes the tests:

In the type impl:
```rust 
    impl MethodFilter

    ...

    pub const fn or(self, other: Self) -> Self {
        Self(self.0 | other.0)
    }
```
When calling in `form.rs`, for example:
```rust
            "/",
            on(
                MethodFilter::or(MethodFilter::GET, MethodFilter::POST),
                |_: Form<Payload>| async {},
            ),
        );
```
Please let me know your thoughts if you think this approach works. If so, I can submit another commit removing the `impl std::ops::BitOr`. Thanks!

davidpdrsn: That looks good to me, although I think the example should use

```rust
MethodFilter::GET.or(MethodFilter::POST)
```

instead of the fully qualified `MethodFilter::or(...)`

PR #2089, opened 2023-07-14
Implemented support for multiple handlers on the same endpoint.

davidv1992:
## Motivation

There are times where I want to accept both JSON and Form data on the same path/method combination, or when I want to choose between handlers based on available query parameters. This pull request allows an easy syntax for that.

## Solution

This PR adds a can_accept(_*) method to `FromRequest`, `FromRequestParts` and `Handler` that is intended to provide an optimistic fast-path check whether the handling is likely to succeed for that `FromRequest` or `Handler`.

This allows us to make tuples of handlers also a handler, which then handles the request with the first handler in the tuple whose `can_accept` returns true for the request, or the last if non can.

Code example:
```
use axum::{
    extract,
    routing::post,
    Router,
};
use serde::Deserialize;

#[derive(Deserialize)]
struct CreateUser {
    email: String,
    password: String,
}

async fn create_user_json(extract::Json(payload): extract::Json<CreateUser>) {
    // payload is a `CreateUser`
}

async fn create_user_form(extract::Form(payload): extract::Form<CreateUser>) {
    // payload is a `CreateUser`
}

let app = Router::new().route("/users", post((create_user_json, create_user_form)));
```

### DIFF
--- a/axum-core/src/extract/mod.rs
+++ b/axum-core/src/extract/mod.rs
@@ -56,6 +56,12 @@ pub trait FromRequestParts<S>: Sized {
 
     /// Perform the extraction.
     async fn from_request_parts(parts: &mut Parts, state: &S) -> Result<Self, Self::Rejection>;
+
+    /// Optimistic estimation of whether this extractor can accept this request.
+    /// Used to allow multiple handlers on a single route.
+    fn can_accept_parts(_req: &Request, _state: &S) -> bool {
+        true
+    }
 }
 
 /// Types that can be created from requests.
@@ -83,6 +89,12 @@ pub trait FromRequest<S, M = private::ViaRequest>: Sized {
 
     /// Perform the extraction.
     async fn from_request(req: Request, state: &S) -> Result<Self, Self::Rejection>;
+
+    /// Optimistic estimation of whether this extractor can accept this request.
+    /// Used to allow multiple handlers on a single route.
+    fn can_accept_request(_req: &Request, _state: &S) -> bool {
+        true
+    }
 }
 
 #[async_trait]
@@ -96,6 +108,10 @@ where
     async fn from_request(req: Request, state: &S) -> Result<Self, Self::Rejection> {
         let (mut parts, _) = req.into_parts();
         Self::from_request_parts(&mut parts, state).await
+    }
+
+    fn can_accept_request(req: &Request, state: &S) -> bool {
+        Self::can_accept_parts(req, state)
     }
 }
 
--- a/axum-core/src/extract/tuple.rs
+++ b/axum-core/src/extract/tuple.rs
@@ -42,6 +42,16 @@ macro_rules! impl_from_request {
 
                 Ok(($($ty,)* $last,))
             }
+
+            fn can_accept_parts(req: &Request, state: &S) -> bool {
+                let mut can_accept = true;
+
+                $(
+                    can_accept = can_accept && $ty::can_accept_parts(req, state);
+                )*
+
+                can_accept && $last::can_accept_parts(req, state)
+            }
         }
 
         // This impl must not be generic over M, otherwise it would conflict with the blanket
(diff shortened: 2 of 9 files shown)


davidv1992:
I specifically made this a draft as it is not yet 100% complete. However, before putting in the effort to push it over the line I would like to ask if this is something axum is interested in, and is this the right approach?

davidpdrsn:
Hm this is an interesting suggestion! I hadn't thought about this.

It sorta feels like a step in the direction of content-type based routing (https://github.com/tokio-rs/axum/issues/1654). The currently recommended way of doing that is writing your own extractor (https://github.com/tokio-rs/axum/blob/main/examples/parse-body-based-on-content-type/src/main.rs#L51-L79) but this of course has the advantage that you don't need to do that.

I think my primary question is how often you truly need to do different things based on the content-type or whether you're actually better saved by having a single handler + extractor that inspects the content-type.

davidpdrsn:
It's sorta similar to https://docs.rs/axum-extra/latest/axum_extra/handler/trait.HandlerCallWithExtractors.html#method.or by the way. Not sure if you've seen that.

davidv1992:
Hadn't seen the `HandlerCallWithExtractors`, that is a good one to keep in mind for my own applications in the future. This serves a similar purpose indeed, although with different trade-offs. Biggest that I see are:
- this approach would give extractors more control over the acceptance process, potentially making it more efficient. That comes with the cost of extractors potentially still failing even when they says they can accept
- this supports handlers with `FromRequest` extractors. As far as I can tell, `HandlerCallWithExtractors::or` doesn't.
- an earlier handler's extractors can't modify the request for later handlers.
Perhaps this could also play nicer with middleware, though I am not aware enough of how middleware exactly works to be sure of that.

As for content-type based routing, this would pretty much eliminate the need for separate support for that, that could be done entirely through this if all relevant types properly implement `can_accept`. For how often that comes up, I have done it twice in applications, once for dealing with different encryption types and once to keep a public facing API backwards compatible. The first one is kind of niche, but I can see the second one coming up quite regularly.

I will probably have some time next week, so trying to implement some further small parts would be an option if that would be helpfull for making a decision (perhaps toying around a bit with how this interacts with middleware would be a good idea). Just let me know.

AlexCzar:
I have encountered this requirement several times when I was working in FinTech. The requirement usually was: implement an API that works with both JSON and XML supported by all the resource paths.

davidpdrsn:
@AlexCzar you can already do that like this https://github.com/tokio-rs/axum/blob/main/examples/parse-body-based-on-content-type/src/main.rs

AlexCzar:
Yes, maybe it is subjective, but I find FormOrJson aproach less desirable because
* When new format is added, existing impl needs to be modified, instead of just adding a new extractor at the top level.
* in good humour, I'm conditioned by years of code rieviews to dislike things named AorB, CandD, etc, :)

davidpdrsn:
I don’t follow. The approach in this PR requires changing all paths to handle a new format. You need to add a new handler to all the tuples.

The example I linked doesn’t have that issue.

AlexCzar:
Right, I didn't think about that. 
I come from a framework (in java ecosystem) where extractors are fully abstracted away if a custom extractor is needed, it's implemented and registered with the framework, it is never referenced in the application code directly, so there Open-Closed principle (loosely speaking) violation in FormOrJson is unnecessary, as FormExtractor and JsonExtractor are registered with the framework and automatically checked against whent the framework is parsing the request.
With the explicit aproach of Axum extractors, FormOrJson makes sense.
Forget I said anything :)

davidpdrsn:
I think I'll close this for now. While I do think the idea is interesting it is not clear to me that this approach is better than the extractor approach [I linked](https://github.com/tokio-rs/axum/blob/main/examples/parse-body-based-on-content-type/src/main.rs). So think I'd rather not introduce new changes that burden people who write extractors.

PR #2094, opened 2023-07-15
Update to hyper 1.0.0-rc.4

davidpdrsn:


axum/src/lib.rs
```diff
@@ -418,7 +418,9 @@
 )]
 #![deny(unreachable_pub, private_in_public)]
 #![allow(elided_lifetimes_in_paths, clippy::type_complexity)]
-#![forbid(unsafe_code)]
+// can't be `forbid` since we've vendored code from hyper-util that contains `unsafe`
+// when hyper-util is on crates.io we can stop vendoring it and go back to `forbid`
+#![deny(unsafe_code)]
```
jplatte, line 423: Why not a git dependency?

examples/hyper-1-0/Cargo.toml
```diff
@@ -6,9 +6,10 @@ publish = false
 
 [dependencies]
 axum = { path = "../../axum" }
-hyper = { version = "1.0.0-rc.3", features = ["full"] }
+hyper = { version = "=1.0.0-rc.4", features = ["full"] }
+hyper-util = { git = "https://github.com/hyperium/hyper-util", features = ["full"] }
```
jplatte, line 10: I'd recommend pinning git dependencies to a specific rev, especially if there's no git-tracked lockfile.

axum/src/lib.rs
davidpdrsn: I'm considering doing an rc release of 0.7 when tower-http has been updated so would like to avoid git dependencies for that reason.

jplatte (approved):

PR #2116, opened 2023-07-22
Update tokio-tungstenite 0.20

alexheretic:
## Motivation
More optimal websocket `SinkExt::feed` performance by using recent upstream versions with better flushing behaviour.

## Solution
* Update _tokio-tungstenite_ to `0.20`. See [changelog](https://github.com/snapview/tokio-tungstenite/blob/master/CHANGELOG.md#0200), [tungstenite changelog](https://github.com/snapview/tungstenite-rs/blob/master/CHANGELOG.md#0200).
* **breaking:** Remove `WebSocketUpgrade::max_send_queue`.
* Add `WebSocketUpgrade::write_buffer_size`, `max_write_buffer_size` (new upstream).
  _Note: I added all upstream docs, feel free to trim them down if you prefer._

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -151,6 +151,7 @@ impl<F> std::fmt::Debug for WebSocketUpgrade<F> {
 impl<F> WebSocketUpgrade<F> {
     /// The target minimum size of the write buffer to reach before writing the data
     /// to the underlying stream.
+    ///
     /// The default value is 128 KiB.
     ///
     /// If set to `0` each message will be eagerly written to the underlying stream.
@@ -164,6 +165,7 @@ impl<F> WebSocketUpgrade<F> {
 
     /// The max size of the write buffer in bytes. Setting this can provide backpressure
     /// in the case the write buffer is filling up due to write errors.
+    ///
     /// The default value is unlimited.
     ///
     /// Note: The write buffer only builds up past [`write_buffer_size`](Self::write_buffer_size)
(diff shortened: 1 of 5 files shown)


axum/src/extract/ws.rs
```diff
@@ -149,9 +149,37 @@ impl<F> std::fmt::Debug for WebSocketUpgrade<F> {
 }
 
 impl<F> WebSocketUpgrade<F> {
-    /// Set the size of the internal message send queue.
-    pub fn max_send_queue(mut self, max: usize) -> Self {
-        self.config.max_send_queue = Some(max);
+    /// Does nothing, instead use `max_write_buffer_size`.
+    #[deprecated]
+    pub fn max_send_queue(self, _: usize) -> Self {
+        self
```
davidpdrsn, line 155: You say it’s deprecated upstream but that should mean we can still call it right?

alexheretic: We can. However, upstream it does absolutely nothing since the send-queue concept was removed.
The field was kept & deprecated to smoothen the churn. It should be removed with the next tungstenite breaking change.

So with no reason to set it, we can just deprecate and have the method do nothing (which also even more clearly demonstrates that it does nothing to axum users). Or as a breaking change remove it.

jplatte: My 2c: Merge deprecation into `v0.6.x` (and do another point release when releasing 0.7, if not sooner), remove on `main`.

davidpdrsn: That sounds good to me!

@alexheretic do you wanna change this PR to include the breaking changes and make another one into the `v0.6.x` branch?

alexheretic: Done. I raised #2123 with the non-breaking update to `v0.6.x`.

jplatte (approved):

davidpdrsn (approved):
Thanks!

PR #2117, opened 2023-07-25
Remove unnecessary allow(dead_code) from testing example

rnestler:
## Motivation

I saw this while looking through the examples and wanted to fix it

## Solution

Fix it :slightly_smiling_face:


davidpdrsn (approved):
Thanks!

PR #2121, opened 2023-07-26
docs: delete MatchedPath nesting section

Kinrany:
## Motivation

`MatchedPath` used to have limitations when used together with [`Router::nest`](https://docs.rs/axum/latest/axum/routing/struct.Router.html#method.nest). This is no longer the case.

Relevant changelog for April 2023: https://github.com/tokio-rs/axum/blob/main/axum/CHANGELOG.md#0613-11-april-2023

Test that covers the new use case: https://github.com/tokio-rs/axum/blob/5567d665d7c5f0ed09e1d5c588c18c28751f23e1/axum/src/extract/matched_path.rs#L248-L266

Discord discussion: https://discord.com/channels/500028886025895936/870760546109116496/1133784088336154624

## Solution

This PR removes the outdated section of the docs on `MatchedPath`, [Matched path in nested routers](https://docs.rs/axum/latest/axum/extract/struct.MatchedPath.html#matched-path-in-nested-routers).

### DIFF
--- a/axum/src/extract/matched_path.rs
+++ b/axum/src/extract/matched_path.rs
@@ -52,43 +52,6 @@ use std::{collections::HashMap, sync::Arc};
 ///     );
 /// # let _: Router = app;
 /// ```
-///
-/// # Matched path in nested routers
-///
-/// Because of how [nesting] works `MatchedPath` isn't accessible in middleware on nested routes:
-///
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
Thanks!

PR #2122, opened 2023-07-26
Don't depend on axum-macros in examples

davidpdrsn:
All macros are re-exported by axum so that's the recommended way to get the macros.

PR #2124, opened 2023-07-27
Fix typo on `customize-extractor-error` example

tuhanayim:
Fix https://github.com/tokio-rs/axum/pull/2122#discussion_r1276265733


davidpdrsn (approved):
Thanks!

Issue #2126, opened 2023-07-28
Docs and tests should cover empty path segment capture


kriswuollett:
- [x] I have looked for existing issues (including closed) about this

### Motivation

It wasn't clear from the docs nor test cases that path segment capture supports empty paths which are valid in RFC 3986. For example it would be nice to inform users that they could implement "optional parameters" in an empty string.

### Proposal

Update docs and test cases to inform about this use case, e.g., both `/:key/method` and `/method/:key` can capture empty strings for `:key`.

### Alternatives

N/A

davidpdrsn:
Sure! Do you wanna submit a PR?

kriswuollett:
> Sure! Do you wanna submit a PR?

Already had one ready to go to embed the fixes in the commit message. :-)

PR #2127, opened 2023-07-28
Add empty path capture docs and tests

kriswuollett:
Fixes: #2126

## Motivation

Inform users that empty path segment capture is possible.

## Solution

Update path segment capture docs and add unit tests to demo and confirm it works.

### DIFF
--- a/axum/src/extract/path/mod.rs
+++ b/axum/src/extract/path/mod.rs
@@ -629,7 +629,7 @@ mod tests {
     }
 
     #[crate::test]
-    async fn captures_dont_match_empty_segments() {
+    async fn captures_dont_match_empty_path() {
         let app = Router::new().route("/:key", get(|| async {}));
 
         let client = TestClient::new(app);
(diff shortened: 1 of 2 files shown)


axum/src/extract/path/mod.rs
```diff
@@ -641,6 +641,41 @@ mod tests {
...
+
+        let res = client.get("/method/abc").send().await;
+        assert_eq!(res.text().await, "abc");
+
+        let res = client.get("/method/").send().await;
+        assert_eq!(res.text().await, "");
+
+        let res = client.get("/method").send().await;
+        assert_eq!(res.status(), StatusCode::NOT_FOUND);
+    }
```
davidpdrsn, line 677: Nice!

Wanna add a similar test for `/method/:key/`? So with a trailing slash.

kriswuollett: Or do you mean another route handler for that?

davidpdrsn: Yeah that's what I meant. A new test with `/method/:key/`.

davidpdrsn (approved):
Thanks!

Issue #2128, opened 2023-07-28
Path parsing doesn't handle `:` nor rfc3986 `sub-delims`


kriswuollett:
- [x] I have looked for existing issues (including closed) about this

The closest issue I could find is #261.

### Version

```shell
kris@slate axum % cargo tree | grep axum
axum v0.6.16 (/Users/kris/code/kriswuollett/axum/axum)
├── axum-core v0.3.4 (/Users/kris/code/kriswuollett/axum/axum-core)
│   ├── axum v0.6.16 (/Users/kris/code/kriswuollett/axum/axum) (*)
│   ├── axum-extra v0.7.4 (/Users/kris/code/kriswuollett/axum/axum-extra)
│   │   ├── axum v0.6.16 (/Users/kris/code/kriswuollett/axum/axum) (*)
│   │   ├── axum-core v0.3.4 (/Users/kris/code/kriswuollett/axum/axum-core) (*)
│   │   ├── axum-macros v0.3.7 (proc-macro) (/Users/kris/code/kriswuollett/axum/axum-macros)
│   │   │   ├── axum v0.6.16 (/Users/kris/code/kriswuollett/axum/axum) (*)
│   │   │   ├── axum-extra v0.7.4 (/Users/kris/code/kriswuollett/axum/axum-extra) (*)
│   │   ├── axum v0.6.16 (/Users/kris/code/kriswuollett/axum/axum) (*)
├── axum-macros v0.3.7 (proc-macro) (/Users/kris/code/kriswuollett/axum/axum-macros) (*)
├── axum-macros v0.3.7 (proc-macro) (/Users/kris/code/kriswuollett/axum/axum-macros) (*)
axum-core v0.3.4 (/Users/kris/code/kriswuollett/axum/axum-core) (*)
axum-extra v0.7.4 (/Users/kris/code/kriswuollett/axum/axum-extra) (*)
axum-macros v0.3.7 (proc-macro) (/Users/kris/code/kriswuollett/axum/axum-macros) (*)
kris@slate axum % git log -1
commit 64c566cd1c7ff9458a69de2274ae4833cda82666 (HEAD -> main, origin/main, origin/HEAD)
Author: tuhana <github@tuhana.me>
Date:   Thu Jul 27 17:03:10 2023 +0300
```

### Platform

```shell
kris@slate axum % uname -a
[...]

PR #2129, opened 2023-07-28
Add test cases to cover colon and percent encoding issues

kriswuollett:
They currently do not pass, but should be addressed when #2128 is fixed.

## Motivation

Currently cannot figure out how to use path segments with embedded `:` characters.

## Solution

Add test cases that will need to be fixed to show that the use case works.


davidpdrsn:
Don't think we need failing tests for this yet. Can just wait for matchit to fix it. We track their releases anyway.

PR #2132, opened 2023-07-30
Update main.rs - Correct method in comments

Calorus:
Correct method in comment on the example usage

## Motivation
Reduce confusion for new users

## Solution
Comment updated.


davidpdrsn (approved):
Thanks!

PR #2135, opened 2023-08-01
axum-extra: Remove unused tower-http dependency

jplatte:
## Motivation

The dependency is not used.

## Solution

Remove it.

## Notes

This PR is for the 0.6.x branch. I didn't check whether `main` is also affected.


davidpdrsn (approved):
\o/

PR #2140, opened 2023-08-02
Implement `Handler` for `T: IntoResponse`

davidpdrsn:


axum/src/handler/mod.rs
```diff
@@ -224,6 +224,23 @@ macro_rules! impl_handler {
 
 all_the_tuples!(impl_handler);
 
+mod private {
+    // Marker type for `impl<T: IntoResponse> Handler for T`
+    #[allow(missing_debug_implementations)]
+    pub enum IntoResponseHandler {}
```
jplatte, line 230: I don't think this needs to have `pub` visibility? At least on my v0.6.x based branch, things compiled with a private marker type.

davidpdrsn: It didn't actually work when I tried using it

```
error: type `handler::IntoResponseHandler` is private
  --> examples/hello-world/src/main.rs:12:40
   |
12 |     let app = Router::new().route("/", get(StatusCode::OK));
   |                                        ^^^ private type
```

jplatte: What, at the usage site?? 😱

davidpdrsn: yeah that confused me as well. Haven't seen that before.

davidpdrsn:
@jplatte I've added some docs.

jplatte (approved):
Awesome! 🎉 

I wouldn't mind you taking care of the backporting as well x)

davidpdrsn:
Go for it!

PR #2142, opened 2023-08-02
Implement Clone for ErasedJson

jplatte:
## Motivation

Makes `ErasedJson` a `Handler`. Also kinda expected, maybe? All of `Json`, `Html` and so on implement `Clone`.

## Solution

Implement `Clone`. Requires the serde_json error to be wrapped in `Arc` because it's not `Clone`, but failing to serialize should be an extremely rare edge-case.

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -7,9 +7,11 @@ and this project adheres to [Semantic Versioning].
 
 # Unreleased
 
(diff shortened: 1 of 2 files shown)


davidpdrsn (approved):

jplatte:
Hope you don't mind me drive-by removing the duplicate "added" from the changelog ^^

davidpdrsn (approved):

PR #2143, opened 2023-08-02
Implement `IntoResponse` for `(R,) where R: IntoResponse`

davidpdrsn:
I just happened to notice we were missing this impl. Probably isn't the most useful thing but still nice to have for consistency and maybe some cursed macro stuff.


jplatte (approved):
Yay for cursed macro stuff 😄

PR #2144, opened 2023-08-03
Replace :emoji: codes with the actual emoji

jplatte:
## Motivation

Just saw this on https://crates.io/crates/axum-extra:

Screenshot_2023-08-03_101135

## Solution

What the PR title says.


davidpdrsn (approved):
🎈🎈🎈

PR #2149, opened 2023-08-04
Add space between SSE field and value for compatibility

ttys3:
according to https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation using `field: value` style is OK.

> Collect the characters on the line after the first U+003A COLON character (:), and let value be that string. If value starts with a U+0020 SPACE character, remove it from value.

## Motivation

other client side tools may detect the `data` field rely on the space, 

like this one https://github.com/openai/openai-python/blob/b82a3f7e4c462a8a10fa445193301a3cefef9a4a/openai/api_requestor.py#L106 

and this one: https://github.com/sashabaranov/go-openai/blob/71a24931dbc5b7029901ff963dc4d0d2509aa7ed/stream_reader.go#L14

in this case, axum powered http server responsed sse event will not be correctly parsed by theses libs,
and they can not receive any message from axum powered sse server.

## Solution

add a U+0020 SPACE character, between `field` and `value`

## Related 

https://github.com/openai/openai-python/pull/559

https://github.com/openai/openai-python/pull/502

https://github.com/tokio-rs/axum/discussions/2146

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -64,6 +64,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **change:** axum's MSRV is now 1.63 ([#2021])
 - **added:** Implement `Handler` for `T: IntoResponse` ([#2140])
 - **added:** Implement `IntoResponse` for `(R,) where R: IntoResponse` ([#2143])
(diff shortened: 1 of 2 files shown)


davidpdrsn (approved):
Thanks!

PR #2154, opened 2023-08-05
Clarify documentation about `route_layer`

Maaxxs:
## Motivation

improve the documentation.

## Solution

I guess these were copy and paste typos? I think `route` should refer to `route_layer` in the `route_layer` documentation.

### DIFF
--- a/axum/src/docs/method_routing/route_layer.md
+++ b/axum/src/docs/method_routing/route_layer.md
@@ -2,8 +2,9 @@ Apply a [`tower::Layer`] to the router that will only run if the request matches
 a route.
 
 Note that the middleware is only applied to existing routes. So you have to
-first add your routes (and / or fallback) and then call `layer` afterwards. Additional
-routes added after `layer` is called will not have the middleware added.
+first add your routes (and / or fallback) and then call `route_layer`
+afterwards. Additional routes added after `route_layer` is called will not have
+the middleware added.
 
 This works similarly to [`MethodRouter::layer`] except the middleware will only run if
 the request matches a route. This is useful for middleware that return early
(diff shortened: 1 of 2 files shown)


axum/src/docs/method_routing/route_layer.md
```diff
@@ -19,10 +20,9 @@ use axum::{
 };
 use tower_http::validate_request::ValidateRequestHeaderLayer;
 
-let app = Router::new().route(
-    "/foo",
-    get(|| async {})
-        .route_layer(ValidateRequestHeaderLayer::bearer("password"))
+let app = Router::new()
+    .route("/foo", get(|| async {})
+    .route_layer(ValidateRequestHeaderLayer::bearer("password"))
```
davidpdrsn, line 25: This change calls `Router::route_layer` but these docs are for `MethodRouter::route_layer`. So the code was internally like the way it was before.

Maaxxs: ah sorry, I didn't mention this. Yes, that line did not change, I only "formatted" it to make it look like the many other code examples in the documentation. 
I only meant the wording in the description/text of the documentation itself where it says `.route` but I think it should say `.route_layer` 

If you want I can remove that "formatting" change from that commit or put it in its own commit

Maaxxs: Oh, now I see what you mean. I didn't notice this actually. I guess it would be nice to have a correct example where `MethodRouter::route_layer` is called :sweat_smile:

Maaxxs: nevermind, I was confused. So the "formatting" change does not change the functionality. It still calls the `route_layer` on the `.get(..)` method, however, the new formatting makes that less obvious which is bad.

davidpdrsn (approved):
Thanks!

axum/src/docs/method_routing/route_layer.md
Maaxxs: already done :sweat_smile:

PR #2156, opened 2023-08-06
Add sqlx-postgresql-stream example

belovaf:
## Motivation
It is not straightforward  how to stream large data sets with sqlx.
Sqlx requires a borrowed pool or connection which is not compatible with axum `StreamBody`.
You can do this with `SelfRefStream` trick.

I found the relevant discussion https://github.com/tokio-rs/axum/discussions/400, but it doesn't contain good example.

May be you decide to include code from this PR to axum core to simplify such use cases.

## Solution
Working example with postgres streaming.

### DIFF
--- a/examples/sqlx-postgres-stream/Cargo.toml
+++ b/examples/sqlx-postgres-stream/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "example-sqlx-postgres-stream"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+axum = { path = "../../axum" }
+futures = "0.3"
+ouroboros = "0.15"
+sqlx = { version = "0.7", features = ["postgres", "runtime-tokio"] }
+tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
+tracing = "0.1"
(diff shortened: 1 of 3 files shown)


davidpdrsn:
I'm not sure this deserves its own examples. The examples are meant to show how to use certain features of axum or how to use axum with other crates. This really just feels like it shows a workaround for a sqlx quirk. Feels like it belongs more in their docs than ours.

davidpdrsn:
I did some testing and it seems you can achieve the same with the `async-stream` crate. That allows you to easily move the pool into the stream

```rust
async fn using_self_ref_stream(State(pool): State<PgPool>) -> impl IntoResponse {
    let stream = async_stream::stream! {
        let stream = sqlx::query_scalar::<_, String>(
            r#"SELECT col1 FROM (VALUES ('hello'), ('from'), ('postgres')) AS q (col1)"#,
        )
        .fetch(&pool);

        for await item in stream {
            yield item
        }
    };

    (
        StatusCode::OK,
        [(header::CONTENT_TYPE, "text/plain")],
        StreamBody::new(stream),
    )
}
```

belovaf:
OK, thank you!
So I close this PR.

PR #2159, opened 2023-08-08
Fix typos in `examples/diesel-postgres/src/main.rs`

bnzone:
## Motivation
While looking at the Diesel Postgres example I noticed a few typos and decided to fix 

## Solution

N/A


jplatte (approved):
Thanks!

PR #2162, opened 2023-08-08
Improve CI

jplatte:
cc @weiznich re. first commit.


weiznich:
Thanks for the ping. I currently work on https://github.com/rust-lang/rust/pull/114452 which should replace the internal attribute and hopefully is usable on stable someday.

PR #2166, opened 2023-08-08
Add auto-reload example

jplatte:
## Motivation

Auto-reload is very convenient for development, but few people are aware how easy it is to implement.

## Solution

Add an example!

### DIFF
--- a/examples/auto-reload/README.md
+++ b/examples/auto-reload/README.md
@@ -1,0 +1,18 @@
+# auto-reload
+
+This example shows how you can set up a development environment for your axum
+service such that whenever the source code changes, the app is recompiled and
+restarted. It uses `listenfd` to be able to migrate connections from an old
+version of the app to a newly-compiled version.
+
+## Setup
+
+```sh
+cargo install cargo-watch systemfd
+```
+
+## Running
+
+```sh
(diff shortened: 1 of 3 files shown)


examples/auto-reload/README.md
```diff
@@ -0,0 +1,18 @@
...
+## Setup
+
+```sh
+cargo install cargo-watch systemfd
+```
+
+## Running
+
+```sh
+systemfd --no-pid -s http::3000 -- cargo watch -x run
```
hi-rustin, line 17: Should we mention that you only can run this command under the `examples/auto-reload` dir?

hi-rustin (approved):
Looks good to me.

davidpdrsn (approved):

matx64:
Hi @jplatte , is there any advantage between using `listenfd` compared to just `cargo-watch -c -q -w . -x "run"` ?

jplatte:
IIRC, stopping and restarting an application that binds to a certain address / port within a small timeframe can result in the OS not allowing the new bind. Additionally, listenfd running while the Rust app is still compiling means that a client (e.g. browser) can already connect and will just wait for the server to start responding, rather than producing a cannot-connect error which often requires the user to retry explicitly once the server it up.

PR #2174, opened 2023-08-14
Don't create a `RouteId` if routes can be merged

danpls:
While looking through the code, I noticed that the function `PathRouter::route` unnecessarily creates a `RouteId` even if the correct route already exists, thus remaining unused in the code path. To prevent this, move the creation of a new `RouteId` further down to right before it is used.


davidpdrsn (approved):
Thanks!

Issue #2182, opened 2023-08-19
Router::into_make_service doesn't compile unless with_state is called last


selfhoster1312:
It took me quite a bit of hair pulling to figure this out. When calling `with_state` before `route` or `nest` the compilation fails with:

```
error[E0599]: no method named `into_make_service` found for struct `Router<Foo, _>` in the current scope
  --> src/main.rs:17:20
   |
17 |         .serve(app.into_make_service())
   |                    ^^^^^^^^^^^^^^^^^ help: there is a method with a similar name: `route_service`
```

The error message is not informative.

Doesn't work:

```
    let app = Router::new()
        .with_state(shared_state)
        .route("/", get(handler));
```

Works:

```
    let app = Router::new()
        .route("/", get(handler))
        .with_state(shared_state);
```

I don't think this is documented yet. I don't know if this should be fixed somehow.

Minimum reproduction case:

```
use axum::extract::State;

#[tokio::main]
async fn main() {
    let shared_state: u8 = 1;
    let app = axum::Router::new()
        .with_state(shared_state)
        .route("/", axum::routing::get(handler));

    axum::Server::bind(&"0.0.0.0:3000".parse().unwrap())
        .serve(app.into_make_service())
        .await
        .unwrap();
}

async fn handler(
    State(_state): State<u8>,
) {}
```

davidpdrsn:
Not sure I'd call this a bug. It's more a consequence of how the implementation works. The router doesn't store the state itself directly. Instead when you call `with_state` it provides the state to all handlers and boxes them as `Box<dyn Service>`. Therefore the routes must be added first.

I don't think we need to change this but its worth documenting for sure. You're welcome to make a PR.

PR #2183, opened 2023-08-19
Document that with_state needs to be last when building a router

selfhoster1312:
See #2182

### DIFF
--- a/axum/src/docs/routing/with_state.md
+++ b/axum/src/docs/routing/with_state.md
@@ -1,4 +1,4 @@
-Provide the state for the router.
+Provide the state for the router. This methods needs to be called last when building the router, or your code won't compile.
 
 ```rust
 use axum::{Router, routing::get, extract::State};
--- a/axum/src/extract/state.rs
+++ b/axum/src/extract/state.rs
@@ -51,6 +51,26 @@ use std::{
 ///
 /// [order-of-extractors]: crate::extract#the-order-of-extractors
 ///
+/// Note that `State` should be provided last when building a [`Router`]. This means this code won't compile:
+///
+/// ```compile_fail
+/// # use axum::{Router, routing::get, extract::State};
+/// # #[derive(Clone)]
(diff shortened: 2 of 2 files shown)


axum/src/docs/routing/with_state.md
```diff
@@ -1,4 +1,4 @@
-Provide the state for the router.
+Provide the state for the router. This methods needs to be called last when building the router, or your code won't compile.
```
davidpdrsn, line 1: This is not always true:

```rust
Router::new()
    .with_state(1)
    .route("/", get(|| async {}))
```

That works.

I think instead we should move the example from the state extractor here instead and remove the changes to this sentence.

axum/src/extract/state.rs
```diff
@@ -51,6 +51,26 @@ use std::{
...
+///     // Dont do this! Place `with_state` after `route`, `nest` or `merge`.
+///     .with_state(state)
+///     .route("/", get(handler));
+/// #
+/// # async fn handler(State(state): State<AppState>) {}
+///
+/// // Compilation will fail here
+/// let service = app.into_make_service();
+/// ```
+///
```
davidpdrsn, line 73: I think we should move this example to the docs for `Router::with_state` since that's the method that's important in the example.

selfhoster1312: I think it's easier to discover on that page since that's what's linked from the docs index, not the with_state method. Do you insist on this? If so i will move it, and simply place a shorter sentence on the State docs if that's alright with you.

davidpdrsn: Yes I think it should be moved.

axum/src/docs/routing/with_state.md
davidpdrsn: You resolved this conversation but it appears to not be fixed 🤔

davidpdrsn:
Gonna close this due to inactivity. Please re-open if you still think its important.

PR #2192, opened 2023-08-22
Fix CI errors

jplatte:
## Motivation

CI is failing on main.

## Solution

See individual commits.


jplatte:
cargo-public-api crates failure seems to be https://github.com/rust-lang/rust/issues/115062, I'm looking for a nightly that doesn't reproduce the ICE.

davidpdrsn (approved):

Issue #2173, opened 2023-08-14
debug_handler: Check IntoResponseParts / IntoResponse for return tuple element types


jplatte:
In Jon Gjengset's [Decrusting the axum crate](https://youtu.be/Wnb_n5YktO8) video, he tried using `#[debug_handler]` to get a more useful error message for `(StatusCode, Json<_>, AppendHeaders<_>)` not implementing `IntoResponse` as part of a handler, but that didn't work. I think we can make it work.

SpeedReach:
Is anyone working on this? I'd like to implement this if not.

davidpdrsn:
I don’t think so. You’re welcome to submit a PR!

SpeedReach:
Thanks for the opportunity! 
I improved the error message from
```
error[E0277]: the trait bound `(StatusCode, Json<&str>, AppendHeaders<[(HeaderName, &str); 1]>): IntoResponse` is not satisfied
 --> tests/debug_handler/fail/wrong_return_tuple_type.rs:5:23
  |
5 |   async fn handler() -> (
  |  _______________________^
6 | |     axum::http::StatusCode,
7 | |     axum::Json<&'static str>,
8 | |     axum::response::AppendHeaders<[( axum::http::HeaderName,&'static str); 1]>,
9 | | ) {
  | |_^ the trait `IntoResponse` is not implemented for `(StatusCode, Json<&str>, AppendHeaders<[(HeaderName, &str); 1]>)`
```
to
```
error[E0277]: the trait bound `Json<&'static str>: IntoResponseParts` is not satisfied
 --> tests/debug_handler/fail/wrong_return_tuple_type.rs:7:5
  |
7 |     axum::Json<&'static str>,
  |     ^^^^^^^^^^^^^^^^^^^^^^^^ the trait `IntoResponseParts` is not implemented for `Json<&'static str>`
```

It checks the
first element => StatusCode || Parts || IntoResponseParts
last element => IntoResponse
other elements => IntoResponseParts

I'll start to add tests if you guys think this is a good approach.

jplatte:
That sounds good, but I think for the extractors we also have special checks for Json and other body extractors of known names where we then print a custom error message if they're not last. Those kinds of messages would be helpful here as well, I think.

davidpdrsn:
Yes I agree that would be good but not essential and a PR without is also cool. We can always add that later.

SpeedReach:
[...]

PR #2201, opened 2023-09-01
Improve `debug_handler` on tuple response types

SpeedReach:
## Motivation
Refs: #2173

## Solution
add an check if the output is a tuple with 2 or more elements


axum-macros/src/debug_handler.rs
```diff
@@ -284,6 +286,169 @@ fn check_inputs_impls_from_request(item_fn: &ItemFn, state_ty: Type) -> TokenStr
...
+        ReturnType::Type(_, ty) => match &**ty {
+            Type::Tuple(tuple) => &tuple.elems,
+            _ => return None,
+        },
+        _ => return None,
+    };
+
+    if elements.len() < 2 {
+        return None;
+    }
```
davidpdrsn, line 307: This this special case? What if the return type is `(String,)`, its a bit exotic but if it doesn't require much special casing then it would be cool to support.

axum-macros/src/debug_handler.rs
```diff
@@ -284,6 +286,169 @@ fn check_inputs_impls_from_request(item_fn: &ItemFn, state_ty: Type) -> TokenStr
...
+                        "Response" => quote! {},
+                        "StatusCode" => {
+                            quote! {}
+                        }
+                        _ => {
+                            parts_amount += 1;
+                            check_into_response_parts(ty, parts_amount)
+                        }
+                    }
+                }
```
davidpdrsn, line 332: I think you can do `let typename = extract_clean_typename(ty).map(...)` instead.

axum-macros/src/debug_handler.rs
```diff
@@ -284,6 +286,169 @@ fn check_inputs_impls_from_request(item_fn: &ItemFn, state_ty: Type) -> TokenStr
...
+                "Json" => "Json<_>",
+                "String" => "String",
+                //todo add more types that implement IntoResponse but not IntoResponseParts
+                _ => return None,
+            };
+            Some(s)
+        }
+        None => None,
+    }
+}
```
davidpdrsn, line 374: We already have the function https://github.com/tokio-rs/axum/blob/4cd8498aeb723ba7eda85a4eaf7680173c085b6f/axum-macros/src/debug_handler.rs#L531 which does something similar. Could we use that instead?

axum-macros/tests/debug_handler/fail/wrong_return_tuple.rs
```diff
@@ -0,0 +1,11 @@
+
+#[axum::debug_handler]
+async fn handler() -> (
+    axum::http::StatusCode,
+    axum::Json<&'static str>,
+    axum::response::AppendHeaders<[( axum::http::HeaderName,&'static str); 1]>,
+) {
```
davidpdrsn, line 7: I think it would be nice to have a third test like this but with a type that isn't special cased. So a type made by the user that implements `IntoResponse` and isn't the last item. To see that the error is also good there.

axum-macros/src/debug_handler.rs
jplatte: Some of the types that that one handles don't implement `IntoResponse`, so I think we should make another helper function like that one, `response_body_provider_type_names` or whatever.

SpeedReach: Thanks for the review! I fixed all the problems except for this one, we may need a separate list for this one I think.

axum-macros/src/debug_handler.rs
SpeedReach: My thoughts were that the origin error message for this case is already specific enough.
ex: (String,) does not implement IntoResponse.
But it'd be a great idea to make a early return and the error more specific, so I added the check.

SpeedReach: Wait I just noticed that (String,)  does not implement IntoResponse.
Do you want a check for String if it implements IntoResponse and suggest removing the parentheses  ?

davidpdrsn: It does on main https://github.com/tokio-rs/axum/pull/2143

davidpdrsn: Generally the length of the tuple shouldn’t matter. (String,) should work with the same code path as any other tuple I think 🤔

davidpdrsn: Why do you need to special case 1 element tuples? Shouldn't the same code work for 1/2/3/... length tuples? I mean you're just doing loops and maps and stuff.

SpeedReach: I wasn't familiar with the WithPosition Api and didn't notice that when there's only one element it falls to Position::Only instead of Position::First, concerning it might cause unknown behaviors.
After discovering this , I deleted the length check . The code should be more readable rn , thanks for the comment!

axum-macros/src/debug_handler.rs
jplatte: > we may need a separate list for this one

Yes, that is what I was saying (unless I'm misunderstanding) 😅

SpeedReach: Yeah exactly, I forgot to tag your response😓.
I could test on some common used types and provide a more complete list for this if necessary.

SpeedReach:
Hi @davidpdrsn 

I hope this message finds you well. I wanted to check in on the status of this pull request. Is there anything else needed from my end, or any feedback you'd like me to address?

Thank you for your time!

davidpdrsn:
I’ll take another look at this week. Sorry for the delay. Been busy at work and traveling.

SpeedReach:
Sure, I fixed the tests and did some modification.

SpeedReach:
any updates

SpeedReach:
@davidpdrsn  any updates?

davidpdrsn:
Sorry for the delay. I think I'm gonna get this across the finish line.

davidpdrsn:
I've cleaned up the code a bit. @jplatte wanna give this a quick look?

axum-macros/src/debug_handler.rs
```diff
@@ -284,6 +289,167 @@ fn check_inputs_impls_from_request(item_fn: &ItemFn, state_ty: Type) -> TokenStr
...
+        ReturnType::Default => return quote! {},
+    };
+
+    let handler_ident = &item_fn.sig.ident;
+
+    match elems.len() {
+        0 => quote! {},
+        n if n > 17 => syn::Error::new_spanned(
+            &item_fn.sig.output,
+            "Cannot return tuples with more 17 elements",
```
jplatte, line 307:
```suggestion
            "Cannot return tuples with more than 17 elements",
```

axum-macros/src/debug_handler.rs
```diff
@@ -284,6 +289,167 @@ fn check_inputs_impls_from_request(item_fn: &ItemFn, state_ty: Type) -> TokenStr
...
+                Position::First(ty) => match extract_clean_typename(ty).as_deref() {
+                    Some("StatusCode" | "Response") => quote! {},
+                    Some("Parts") => check_is_response_parts(ty, handler_ident, idx),
+                    Some(_) | None => {
+                        if let Some(tn) = well_known_last_response_type(ty) {
+                            syn::Error::new_spanned(
+                                ty,
+                                format!(
+                                    "`{tn}` must be the last element \
+                                                in a response tuple"
```
jplatte, line 322: Second line should be un-indented.

axum-macros/src/debug_handler.rs
```diff
@@ -284,6 +289,167 @@ fn check_inputs_impls_from_request(item_fn: &ItemFn, state_ty: Type) -> TokenStr
...
+            .collect::<TokenStream>(),
+    }
+}
+
+fn check_into_response(handler: &Ident, ty: &Type) -> TokenStream {
+    let (span, ty) = (ty.span(), ty.clone());
+
+    let check_fn = format_ident!(
+        "__axum_macros_check_{}_into_response_check",
+        handler,
```
jplatte, line 353:
```suggestion
        "__axum_macros_check_{handler}_into_response_check",
```

axum-macros/src/debug_handler.rs
```diff
@@ -284,6 +289,167 @@ fn check_inputs_impls_from_request(item_fn: &ItemFn, state_ty: Type) -> TokenStr
...
+        span = span,
+    );
+
+    let call_check_fn = format_ident!(
+        "__axum_macros_check_{}_into_response_call_check",
+        handler,
```
jplatte, line 359:
```suggestion
        "__axum_macros_check_{handler}_into_response_call_check",
```

axum-macros/src/debug_handler.rs
```diff
@@ -284,6 +289,167 @@ fn check_inputs_impls_from_request(item_fn: &ItemFn, state_ty: Type) -> TokenStr
...
+            #from_request_bound,
+        {}
+
+        // we have to call the function to actually trigger a compile error
+        // since the function is generic, just defining it is not enough
+        #[allow(warnings)]
+        #[allow(unreachable_code)]
+        #[doc(hidden)]
+        fn #call_check_fn()
+        {
```
jplatte, line 385:
```suggestion
        fn #call_check_fn() {
```

axum-macros/src/debug_handler.rs
```diff
@@ -284,6 +289,167 @@ fn check_inputs_impls_from_request(item_fn: &ItemFn, state_ty: Type) -> TokenStr
...
+    );
+
+    let call_check_fn_body = quote_spanned! {span=>
+        #check_fn();
+    };
+
+    let from_request_bound = quote_spanned! {span=>
+        #ty: ::axum::response::IntoResponseParts
+    };
+    quote::quote_spanned! {span=>
```
jplatte, line 432:
```suggestion
    quote_spanned! {span=>
```

axum-macros/src/debug_handler.rs
```diff
@@ -284,6 +289,167 @@ fn check_inputs_impls_from_request(item_fn: &ItemFn, state_ty: Type) -> TokenStr
...
+            #from_request_bound,
+        {}
+
+        // we have to call the function to actually trigger a compile error
+        // since the function is generic, just defining it is not enough
+        #[allow(warnings)]
+        #[allow(unreachable_code)]
+        #[doc(hidden)]
+        fn #call_check_fn()
+        {
```
jplatte, line 447:
```suggestion
        fn #call_check_fn() {
```

axum-macros/src/debug_handler.rs
```diff
@@ -384,6 +553,25 @@ fn request_consuming_type_name(ty: &Type) -> Option<&'static str> {
     Some(type_name)
 }
 
+fn well_known_last_response_type(ty: &Type) -> Option<&'static str> {
+    let typename = match extract_clean_typename(ty) {
+        Some(tn) => tn,
+        None => return None,
+    };
```
jplatte, line 560:
```suggestion
    let typename = extract_clean_typename(ty)?;
```

axum-macros/tests/debug_handler/fail/output_tuple_too_many.rs
```diff
@@ -0,0 +1,29 @@
...
+    AppendHeaders<[(axum::http::HeaderName, &'static str); 1]>,
+    AppendHeaders<[(axum::http::HeaderName, &'static str); 1]>,
+    AppendHeaders<[(axum::http::HeaderName, &'static str); 1]>,
+    AppendHeaders<[(axum::http::HeaderName, &'static str); 1]>,
+    axum::http::StatusCode,
+) {
+    panic!()
+}
+
+fn main(){}
```
jplatte, line 29: Missing final newline.

axum-macros/tests/debug_handler/fail/single_wrong_return_tuple.rs
```diff
@@ -0,0 +1,12 @@
...
+struct NotIntoResponse;
+
+#[axum::debug_handler]
+async fn handler() -> (NotIntoResponse) {
+    panic!()
+}
+
+fn main(){
+
+}
```
jplatte, line 12: Same here.

axum-macros/tests/debug_handler/fail/wrong_return_tuple.rs
```diff
@@ -0,0 +1,30 @@
...
+#[axum::debug_handler]
+async fn custom_type() -> (
+    axum::http::StatusCode,
+    CustomIntoResponse,
+    axum::response::AppendHeaders<[( axum::http::HeaderName,&'static str); 1]>,
+) {
+    panic!()
+}
+
+fn main(){}
```
jplatte, line 30: … also here.

axum-macros/tests/debug_handler/fail/output_tuple_too_many.rs
jplatte: Also missing space.

axum-macros/src/debug_handler.rs
davidpdrsn: I was surprised by this too but we do actually allow 17 since the big T1..T16 impl as a final `R` at the end https://docs.rs/axum/latest/axum/response/trait.IntoResponse.html#impl-IntoResponse-for-(T1,+T2,+T3,+T4,+T5,+T6,+T7,+T8,+T9,+T10,+T11,+T12,+T13,+T14,+T15,+T16,+R)

jplatte: Right, edited the suggestion.

jplatte (approved):

PR #2208, opened 2023-09-06
Add tls-graceful-shutdown example

PGIII:
## Motivation
I couldn't find an example of graceful shutdown when using axum_server for TLS and hyper for redirect.

## Solution
This example shows how to use the axum_server handle to trigger a graceful shutdown.
While still sharing the future with the hyper server to use the built-in graceful_shutdown feature of hyper.
This is based off the existing tls-rustls example and existing graceful-shutdown example


examples/tls-graceful-shutdown/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "example-tls-graceful-shutdown"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+axum = { path = "../../axum" }
+axum-server = { version = "0.3", features = ["tls-rustls"] }
```
davidpdrsn, line 9: Does this actually work? I don't think axum-server supports the breaking changes on axum's main branch.

PGIII: I tested it and seems to be working properly
```
$ cargo run -p example-tls-graceful-shutdown

   Compiling example-tls-graceful-shutdown v0.1.0 (/dev/source/axum/examples/tls-graceful-shutdown)
    Finished dev [unoptimized + debuginfo] target(s) in 2.25s
     Running `target/debug/example-tls-graceful-shutdown`
2023-09-06T16:16:33.914537Z DEBUG example_tls_graceful_shutdown: listening on 127.0.0.1:7878
2023-09-06T16:16:33.916763Z DEBUG example_tls_graceful_shutdown: listening on 127.0.0.1:3000
^C2023-09-06T16:16:35.679814Z  INFO example_tls_graceful_shutdown: Received termination signal shutting down
```
Server runs and serves Hello world properly as well and the hyper server redirects

davidpdrsn: You're right. I think it works because `main` still depends on hyper 0.14. We're in progress updating fully to hyper 1.0 [here](https://github.com/tokio-rs/axum/pull/1882).

davidpdrsn (approved):
Thanks!

PR #2216, opened 2023-09-08
Use "set up" rather than "setup" when used as a verb

oalders:
## Motivation

"setup" was being written as a noun in some cases where it appears it should be
used as a verb "set up"

## Solution

See https://writingexplained.org/setup-vs-set-up-difference


jplatte (approved):
Nice, thanks!

PR #2217, opened 2023-09-08
Re-export Duration for Cookie set_max_age

marlonbaeten:
## Motivation

The `set_max_age` method of `Cookie` takes a `time::Duration` as argument. The `cookie` crate re-exports this struct, but axum-extra does not.

## Solution

Re-export the `Duration` struct to make the API accessible without the need to include `time` as a direct dependency.

### DIFF
--- a/axum-extra/src/extract/cookie/mod.rs
+++ b/axum-extra/src/extract/cookie/mod.rs
@@ -24,7 +24,7 @@ pub use self::private::PrivateCookieJar;
 #[cfg(feature = "cookie-signed")]
 pub use self::signed::SignedCookieJar;
 
-pub use cookie::{Cookie, Expiration, SameSite};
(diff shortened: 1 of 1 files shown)


davidpdrsn:
This has been request before but I don’t think it’s appropriate to re-export a dependency of a dependency. Just do “cargo add time” and you’re good to go.

jplatte:
I wonder why the `cookie` crate uses `time`'s `Duration` in the first place. Docs say it allows negative duration unlike std, but that doesn't sound like something that would be useful for cookies?

davidpdrsn:
Yeah that's probably worth asking them about.

jplatte:
https://github.com/SergioBenitez/cookie-rs/issues/109#issuecomment-1717218404

PR #2221, opened 2023-09-13
add a validator inspired by axum in ecosystem

tu6ge:
## Motivation

This is a functional validator, and it is ergonomics like as axum, it should be cooperate well with axum.

## Solution

none


davidpdrsn:
Sorry but I don't think this belongs on this list. Its seems like a general validation library and not something specific to axum.

davidpdrsn:
I still think I'll draw the line here and say we don't include this. Otherwise _every_ crate that could conceivably be used with axum goes on the list.

tu6ge:
I created this project based on inspiration from axum and I really want to share it. sorry to bother you.

PR #2229, opened 2023-09-17
Use tuples instead of `ServiceBuilder` internally

davidpdrsn:
Replace most uses of `ServiceBuilder` internally with tuples. A bit cleaner I think.

I still kept uses of `ServiceBuilder` in the docs because its more discoverable.

### DIFF
--- a/axum/src/routing/method_routing.rs
+++ b/axum/src/routing/method_routing.rs
@@ -1254,7 +1254,7 @@ mod tests {
     use axum_core::response::IntoResponse;
     use http::{header::ALLOW, HeaderMap};
     use std::time::Duration;
-    use tower::{timeout::TimeoutLayer, Service, ServiceBuilder, ServiceExt};
+    use tower::{timeout::TimeoutLayer, Service, ServiceExt};
     use tower_http::{services::fs::ServeDir, validate_request::ValidateRequestHeaderLayer};
 
     #[crate::test]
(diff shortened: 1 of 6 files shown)


axum/src/routing/method_routing.rs
```diff
@@ -1354,13 +1354,10 @@ mod tests {
...
-                    ServiceBuilder::new()
-                        .layer(HandleErrorLayer::new(|_| async {
-                            StatusCode::REQUEST_TIMEOUT
-                        }))
-                        .layer(TimeoutLayer::new(Duration::from_secs(10))),
-                ),
+                .layer((
+                    HandleErrorLayer::new(|_| async { StatusCode::REQUEST_TIMEOUT }),
+                    TimeoutLayer::new(Duration::from_secs(10)),
+                )),
```
jplatte, line 1360: Should we use the `tower_http` timeout layer here instead?

axum/src/routing/tests/merge.rs
```diff
@@ -127,13 +127,10 @@ async fn layer_and_handle_error() {
...
-            ServiceBuilder::new()
-                .layer(HandleErrorLayer::new(|_| async {
-                    StatusCode::REQUEST_TIMEOUT
-                }))
-                .layer(TimeoutLayer::new(Duration::from_millis(10))),
-        );
+        .layer((
+            HandleErrorLayer::new(|_| async { StatusCode::REQUEST_TIMEOUT }),
+            TimeoutLayer::new(Duration::from_millis(10)),
+        ));
```
jplatte, line 133: also here

axum/src/routing/tests/mod.rs
```diff
@@ -309,13 +301,10 @@ async fn wildcard_sees_whole_url() {
...
-            ServiceBuilder::new()
-                .layer(HandleErrorLayer::new(|_: BoxError| async move {
-                    StatusCode::REQUEST_TIMEOUT
-                }))
-                .layer(TimeoutLayer::new(Duration::new(0, 0))),
-        )
+        .layer((
+            HandleErrorLayer::new(|_: BoxError| async move { StatusCode::REQUEST_TIMEOUT }),
+            TimeoutLayer::new(Duration::new(0, 0)),
+        ))
```
jplatte, line 307: and here

PR #2230, opened 2023-09-17
Add `FixNestedRedirect` middleware

### LINKED ISSUE #1731: ServeDir in nested route causes invalid redirects

davidpdrsn:
This adds `FixNestedRedirectLayer` which uses `NestedPath` to change redirects from nested services to include the path they're nested at.

Fixes #1731

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -66,6 +66,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **added:** Implement `IntoResponse` for `(R,) where R: IntoResponse` ([#2143])
 - **changed:** For SSE, add space between field and value for compatibility ([#2149])
 - **added:** Add `NestedPath` extractor ([#1924])
+- **added:** Add `FixNestedRedirect` middleware
 
 [#1664]: https://github.com/tokio-rs/axum/pull/1664
 [#1751]: https://github.com/tokio-rs/axum/pull/1751
--- a/axum/src/extract/nested_path.rs
+++ b/axum/src/extract/nested_path.rs
@@ -45,6 +45,13 @@ impl NestedPath {
     pub fn as_str(&self) -> &str {
         &self.0
     }
+
+    pub(crate) fn extract(parts: &mut Parts) -> Result<Self, NestedPathRejection> {
+        match parts.extensions.get::<Self>() {
+            Some(nested_path) => Ok(nested_path.clone()),
+            None => Err(NestedPathRejection),
+        }
+    }
 }
 
 #[async_trait]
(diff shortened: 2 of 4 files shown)


georgmu:
Wouldn't it be better to fix this in tower-http?
The current implementation in tower-http doesn't allow scenarios to use ServeDir with a prefix (e.g. /files/*).
If such a mechanism would exist, a thin wrapper around ServeDir on axum side would fix this automatically.

davidpdrsn:
> Wouldn't it be better to fix this in tower-http?

That would require moving axum's `NestedPath` to tower-http or adding something similar there. I'm not a big fan of that.

davidpdrsn:
Converting to draft because I just remember that this (and `NestedPath`) doesn't support path params in the nested path. axum supports `.nest("/:foo", ...)` which this should probably support.

jplatte (commented):
I'm a bit surprised this is in `axum` as opposed to `axum-extra`. There seem to be quite a few design choices to be made, for instance I would not have thought of including `FixNestedRedirectOptOut`, but I was wondering whether there could be problems with people applying the middleware multiple times redundantly (more / less nested); if we wanted to solve this having `extract` *remove* the `NestedPath` or insert some other extension do tell that the middleware ran already could work?

georgmu:
> > Wouldn't it be better to fix this in tower-http?
> 
> That would require moving axum's `NestedPath` to tower-http or adding something similar there. I'm not a big fan of that.

I wouldn't add NestedPath to tower-http, but give ServeDir the option to specify a "mount point" in the URL tree (like "files", so that `GET /files/foo.txt` would search for `foo.txt`, not `files/foo.txt`). 

If ServeDir could do this, a thin wrapper around ServeDir in axum (let's call it `NestedServeDir` could use the `NestedPath` mechanism to apply to instanciate `ServeDir` with the correct settings.

davidpdrsn:
> I'm a bit surprised this is in axum as opposed to axum-extra.

Yeah I considered putting it in axum-extra. That probably is the better path. I'll move it.

> I wouldn't add NestedPath to tower-http, but give ServeDir the option to specify a "mount point" in the URL tree (like "files", so that GET /files/foo.txt would search for foo.txt, not files/foo.txt).

I think I'll move forward with the path in this PR and then see what happens as people use it and give feedback.

PR #2232, opened 2023-09-19
Use inline format args

nyurik:
## Motivation
Do a few format arg inlining to make the code a bit more concise and easier to read.  In some rare cases this also prevents double-refs, e.g. `format!("{}", &foo)`, which cases a small run-time perf penalty.

This PR also updates many docs examples for the same reason.

### DIFF
--- a/axum-core/src/macros.rs
+++ b/axum-core/src/macros.rs
@@ -193,7 +193,7 @@ macro_rules! __composite_rejection {
             fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
                 match self {
                     $(
-                        Self::$variant(inner) => write!(f, "{}", inner),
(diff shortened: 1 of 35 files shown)


axum-extra/src/extract/multipart.rs
```diff
@@ -43,7 +43,7 @@ use std::{
 ///         let name = field.name().unwrap().to_string();
 ///         let data = field.bytes().await.unwrap();
 ///
-///         println!("Length of `{}` is {} bytes", name, data.len());
+///         println!("Length of `{name}` is {} bytes", data.len());
```
davidpdrsn, line 46: For cases like this where there are some inline args and some not I prefer to no inline any of them.

nyurik: fixed

davidpdrsn (approved):
Thanks!

PR #2233, opened 2023-09-19
Chore: A few clippy inspired changes

nyurik:
* Combine identical `match` branches into one statement with a `|` symbol
* Use `_` in a 9-bit binary values (I didn't even realize they were 9-bit instead of the common 8-bit until I did this)
* Replaced a few `.map().unwrap_or_else()` with `.map_or_else()`
* Use specific type with `default()` instead of `Default::default()` to make it easier to read/understand

### DIFF
--- a/axum-core/src/ext_traits/request.rs
+++ b/axum-core/src/ext_traits/request.rs
@@ -305,7 +305,7 @@ impl RequestExt for Request {
         *req.uri_mut() = self.uri().clone();
         *req.headers_mut() = std::mem::take(self.headers_mut());
         *req.extensions_mut() = std::mem::take(self.extensions_mut());
-        let (mut parts, _) = req.into_parts();
+        let (mut parts, ()) = req.into_parts();
 
         Box::pin(async move {
             let result = E::from_request_parts(&mut parts, state).await;
--- a/axum-macros/src/from_request.rs
+++ b/axum-macros/src/from_request.rs
@@ -63,8 +63,7 @@ impl State {
     /// ```
     fn trait_generics(&self) -> impl Iterator<Item = Type> {
         match self {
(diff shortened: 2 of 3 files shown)


axum-extra/src/extract/with_rejection.rs
```diff
@@ -89,7 +89,7 @@ impl<E, R> Copy for WithRejection<E, R> where E: Copy {}
 
 impl<E: Default, R> Default for WithRejection<E, R> {
     fn default() -> Self {
-        Self(Default::default(), Default::default())
+        Self(E::default(), PhantomData)
```
davidpdrsn, line 92: I'd rather just keep `Default::default()` here and for the rest of these changes. I don't think it impacts maintainability much.

axum-extra/src/middleware.rs
```diff
@@ -38,7 +38,5 @@ use tower_layer::Identity;
 /// [`HandleErrorLayer`]: axum::error_handling::HandleErrorLayer
 /// [`Infallible`]: std::convert::Infallible
 pub fn option_layer<L>(layer: Option<L>) -> Either<L, Identity> {
-    layer
-        .map(Either::E1)
-        .unwrap_or_else(|| Either::E2(Identity::new()))
+    layer.map_or_else(|| Either::E2(Identity::new()), Either::E1)
```
davidpdrsn, line 41: I actually think `map().unwrap_or_else()` is easier to parse. Closures as the first argument is generally uncommon in Rust. So I'd rather keep this as is. Same goes for the other similar changes in this PR.

axum-extra/src/extract/with_rejection.rs
nyurik: fixed

davidpdrsn (approved):
Thanks!

PR #2236, opened 2023-09-21
Make tokio an optional dependency of axum-extra

davidpdrsn:
Requested by someone in Discord. That way you can use axum-extra in wasm, as long as you don't enable the features that require tokio. We already do this for axum but think we just forgot axum-extra.

@jplatte I'm right in this being a breaking change right? 🤔


jplatte (approved):
I don't see how this would be a breaking change. Maybe try `cargo-semver-checks` locally? By default it compares the current state against the crates.io release. It's still missing lots of lints AFAIK, but better to run it than not to.

PR #2237, opened 2023-09-21
Make sure axum-extra also supports being compiled to WASM

davidpdrsn:
Someone on Discord is having issues with compiling axum-extra to WASM so I'm adding that as a dependency to `example-simple-router-wasm`. Just so we don't accidentally break WASM compilation in the future by adding some non-WASM dependency as part of the default feature set.

### DIFF
--- a/examples/simple-router-wasm/Cargo.toml
+++ b/examples/simple-router-wasm/Cargo.toml
@@ -8,6 +8,9 @@ publish = false
 # `default-features = false` to not depend on tokio features which don't support wasm
 # you can still pull in tokio manually and only add features that tokio supports for wasm
 axum = { path = "../../axum", default-features = false }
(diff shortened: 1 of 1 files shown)


examples/simple-router-wasm/Cargo.toml
```diff
@@ -8,6 +8,9 @@ publish = false
 # `default-features = false` to not depend on tokio features which don't support wasm
 # you can still pull in tokio manually and only add features that tokio supports for wasm
 axum = { path = "../../axum", default-features = false }
+# we don't strictly need axum-extra for this example but wanna make sure that
```
jplatte, line 11:
```suggestion
# we don't strictly use axum-extra in this example but wanna make sure that
```

jplatte (approved):

PR #2241, opened 2023-09-26
Add HTTP/1 and HTTP/2 to `axum::serve`

### LINKED ISSUE #1953: Can `axum::serve` handle both http 1 and 2?
`axum::Server` was a re-export of `hyper::Server` which is going away for hyper 1.0. Therefore axum is getting `axum::serve(tcp_listener, service)` in 0.7 for simple use cases.

It currently only supports http 1. We should investigate if we can also support http 2. Might require https://github.com/hyperium/hyper-util/pull/11

rakshith-ravi:
## Motivation

Adding HTTP/2 support to `axum::serve`

## Solution

Currently this works by adding hyper-util as a dependency.

**This is still a work in progress** as `.with_upgrades` doesn't exist for `http2.serve_connection`. Need to figure out if we add that to hyper-util directly or if we make a copy of the code in axum and use our own version. (CC @seanmonstar)

Also, this PR seemed a tad bit too easy, given this is my first time contributing to this repo. Almost to the point where I feel like either I took up a really simple PR or I'm doing something horribly wrong. Please let me know if I'm doing something daft here 👀 

Fixes #1953


axum/src/serve.rs
```diff
@@ -15,7 +17,7 @@ use tower_service::Service;
 /// This method of running a service is intentionally simple and doesn't support any configuration.
 /// Use hyper or hyper-util if you need configuration.
```
hi-rustin, line 18: Do we need to update this comment as well?

rakshith-ravi: I wanted to keep it because we're still doing what _we_ believe to be sane defaults. I figured we'd leave this here to let the user know that they can configure custom stuff (TLS and what not) if they really want to

davidpdrsn: Yes I think this comment should stay. We still won't expose any customization for this function even if we're wrapping hyper-util.

rakshith-ravi:
I've put up the necessary changes in hyperium/hyper-util#38. Once that gets merged this can be un-`WIP`ed

rakshith-ravi:
Heyyyyyyy my PR got merged. WE GOOD TO GOOOOOOOO!!!!

(Sorry, I'm just wayy too excited today)

rakshith-ravi:
Okay hold on. `hyper-util` should be behind the `tokio` feature flag, right? Or should it just be there an a necessary dependency?

davidpdrsn:
I think it should be behind the tokio feature. `axum::serve` already is right?

rakshith-ravi:
Oh yeah. It is. I'll update it right away

rakshith-ravi:
Done

hi-rustin (approved):
Looks good to me. Thanks!

davidpdrsn (approved):
Thanks!

PR #2242, opened 2023-09-28
Ensure we include the port when parsing authority

bouk:
## Motivation

The 'Host' header should include the port number. When extracting the host from `uri` we use `uri.host()` however, which strips off the port number to only include the hostname. 

The existing test also didn't test the right thing, because the request would still include a Host header—it's added automatically by `hyper` and there's no way to turn that off from reqwest.

## Solution

Re-implement the authority parsing method directly. This code is copied from the `http` crate, just without the port stripping: https://github.com/hyperium/http/blob/1a04e715f49c9ac3fe8195583e8c9959cbd368d3/src/uri/authority.rs#L487-L505


waynr:
Hey @bouk long time no see! (we worked together very briefly a few years ago) What a coincidence that you would open your first PR here just before I open my first PR :)

One suggestion I have here is to add a test case to validate the split you are doing on `@` to exclude the `username:password` portion of the authority.

bouk:
Hey @waynr! Good suggestion, added another test case

davidpdrsn (commented):
Looks good!

bouk:
Added it to the changelog

ttys3:
maybe also add a link to the rfc ?

https://www.rfc-editor.org/rfc/rfc9110.html#name-host-and-authority

axum/src/extract/host.rs
```diff
@@ -130,8 +141,14 @@ mod tests {
...
-        let host = test_client().get("/").await.text().await;
-        assert!(host.contains("127.0.0.1"));
+        let mut parts = Request::new(()).into_parts().0;
+        parts.uri = "https://127.0.0.1:1234/image.jpg".parse().unwrap();
+        let host = Host::from_request_parts(&mut parts, &()).await.unwrap();
+        assert_eq!(host.0, "127.0.0.1:1234");
+
+        parts.uri = "http://cool:user@[::1]:456/file.txt".parse().unwrap();
+        let host = Host::from_request_parts(&mut parts, &()).await.unwrap();
+        assert_eq!(host.0, "[::1]:456");
```
jplatte, line 151: Why did you completely rewrite this example to not use the test client?

jplatte (commented):
Hey, just looking through old PRs and found this.

I fixed the merge conflicts so you don't have to, but I have another questions before merging this, see below.

axum/src/extract/host.rs
```diff
@@ -6,17 +6,21 @@ use async_trait::async_trait;
...
 ///
-/// Hostname is resolved through the following, in order:
+/// Host is resolved through the following, in order:
 /// - `Forwarded` header
 /// - `X-Forwarded-Host` header
 /// - `Host` header
-/// - request target / URI
+/// - Authority of the request URI
+///
+/// See https://www.rfc-editor.org/rfc/rfc9110.html#name-host-and-authority for the definition of
```
jplatte, line 22:
```suggestion
/// See <https://www.rfc-editor.org/rfc/rfc9110.html#name-host-and-authority> for the definition of
```

yanns:
@bouk would you have time to address the comments and rebase your changes? If not, I can take over.

bouk:
@yanns I don't, feel free to take over this PR

axum/src/extract/host.rs
yanns: I've re-introduced the test client in https://github.com/tokio-rs/axum/pull/2242/commits/e4a109ed5a713b100d6af07c6f75c392eb504df8

yanns (approved):

yanns:
@jplatte the PR is ready for review again

axum/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+- **fixed:** Include port number when parsing authority ([#2242])
```
jplatte, line 10: This should be in axum-extra's changelog, no? Also, it should mention that this affects the `Host` extractor, it's currently very hard to tell what this is about.

yanns: true, the merge moved the files into axum-extra, but not the changelog :) good catch!

yanns: https://github.com/tokio-rs/axum/pull/2242/commits/dfd438b933bf24f26f3c714231d43c8410989559

jplatte (approved):

PR #2245, opened 2023-09-28
axum::body no longer pulls in axum::body::stream_body

waynr:
I mentioned this in #1848 but I don't see any reason to keep this file in the
repo and it kind of misled me while I was trying to upgrade my code to work
with a local copy that includes #2244

Signed-off-by: wayne warren <wayne.warren.s@gmail.com>


davidpdrsn (approved):
Oh good catch!

PR #2246, opened 2023-09-29
Use the diagnostic namespace

weiznich:
## Motivation

This PR aims to remove the internal rustc attribute `#[rustc_on_unimplemented]` in favor of the (unstable) `#[diagnostic::on_unimplemented]` attribute. The later attribute has an accepted RFC and will be hopefully stable at some point, while the former attribute will likely remain unstable forever. 
In addition I personally want to get some feedback from different crates whether or not the new attribute fits it's usecase, so that we can work on stabilizing it. 

This PR is split into two commits. The first one changes the nightly version used for running the compile fail tests to a newer nightly that allows using the new attribute. It also fixes some new compiler warnings. The second commit changes the actual attributes.

## Solution

Switch from `#[rustc_on_unimplemented]` to `#[diagnostic::on_unimplemented]`


axum-macros/src/lib.rs
```diff
@@ -37,7 +37,7 @@
     missing_debug_implementations,
     missing_docs
 )]
-#![deny(unreachable_pub, private_in_public)]
+#![deny(unreachable_pub)]
```
davidpdrsn, line 40: Did you mean to remove `private_in_public`? I like having that 🤔

weiznich: The [compiler complained](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=45633c5e4070cc44032892418cb25545) about that lint being removed:

```
warning: lint `private_in_public` has been removed: replaced with another group of lints, see RFC <https://rust-lang.github.io/rfcs/2145-type-privacy.html> for more information
 --> src/main.rs:1:9
  |
1 | #![deny(private_in_public)]
  |         ^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(renamed_and_removed_lints)]` on by default
```

I can readd that if you want but it seems not to be helpful having these warnings in all the compile test outputs.

davidpdrsn: Ah I see. Thats probably okay. axum-macros just exposes proc-macros anyway.

davidpdrsn (approved):
Thanks and also thanks for cleaning up the macro tests!

PR #2285, opened 2023-10-22
Replace reference to multiple with singular in docs.

nectariner:
## Motivation
Correct documentation

## Solution
N/A


davidpdrsn (approved):
Thanks!

Issue #2294, opened 2023-10-29
Add special handling of `Extension` and `Clone` for `debug_handler`


davidpdrsn:
To help with cases like https://github.com/tokio-rs/axum/issues/2293

Basically generate `T: Clone` for `Extension<T>` arguments.

We already do something similar for argument ordering so should be possible to extend that.

loganbnielsen:
Macros operate on the syntax tree and therefore don't generally have knowledge as to whether a trait is implemented for a type. I'm not sure if it's possible to make a check for this using similar techniques from `debug_handler.rs`. AFAIK, we could only identify whether it's `Extension` and the `T` but not whether the `T` implements `Clone` (without doing something infeasible like registering all types that implement clone haha)

jplatte:
@loganbnielsen If you look at the existing checks in `debug_handler.rs`, we already have checks like "inner type of `Json` implements `Serialize` in return type position. It works by generating a function like `fn assert_serialize<T: Serialize>() {}` and calling it as `assert_serialize::<InnerType>();`.

loganbnielsen:
I can publish a PR for this today

jplatte:
Go ahead! This issue is `help wanted` so PRs are welcome :)

I'll also assign the issue to you so it's easy to see that it has somebody working on it.

loganbnielsen:
PR request [here](https://github.com/tokio-rs/axum/pull/2301).

PR #2301, opened 2023-11-01
Add special handling of Extension and Clone for debug_handler

loganbnielsen:
## Motivation

Addresses #2294 using the solution discussed there. ([Link](https://github.com/tokio-rs/axum/issues/2294))


loganbnielsen:
Am I supposed to do anything to request a review?

jplatte:
No. I can request David's review and maybe also take a look myself (though given that this is mostly stuff that David wrote, I don't want to merge w/o his review), but I don't think it will actually speed up things as there's probably a bandwidth problem, not an awareness problem.

loganbnielsen:
No problem, just wanted to make sure I wasn't dropping the ball

yanns (approved):
This can have a good value when someone is using `Extension`.

jplatte:
I couldn't see why we would need to special case `Extension` in particular when we already make sure to have good error messages for other extractors such as `Json` so I decided to see what the error message is on `main` with `#[debug_handler]`. Turns out that we don't actually need to do anything, see #2731.

PR #2305, opened 2023-11-08
Update ECOSYSTEM.md about Shuttle

orhun:
## Motivation

[Shuttle](https://www.shuttle.rs/) has out-of-the-box support for Axum and it makes hosting Axum projects easier for the users.

Actix Web docs also has a separate section that mentions Shuttle: https://actix.rs/docs/shuttle

I thought it would be nice to add it to Axum as well.

## Solution

<s>This PR adds a new example called `shuttle-service` for demonstrating how to create a Axum project and deploying it on the Shuttle platform.</s>

Additionally, updated the outdated Shuttle reference in `ECOSYSTEM.md` and added another example.

<s>I also added a separate section to README.md for listing possible hosting solutions.</s>


davidpdrsn:
Thanks for the PR though I don't think we can feature shuttle quite this prominently.

I think linking to stuff from ECOSYSTEM.md is fine but I don't think we should have a shuttle specific example or mention it in the readme. It feels too much like an official endorsement from tokio which I don't think is appropriate.

orhun:
Got it, reasonable. Updated the PR!

davidpdrsn (approved):
Thanks!

PR #2307, opened 2023-11-10
fix(ws): check connection by header_eq

bestgopher:
## Motivation

I think the `connection` header must be `upgrade`.

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -393,7 +393,7 @@ where
             return Err(MethodNotGet.into());
         }
 
-        if !header_contains(&parts.headers, header::CONNECTION, "upgrade") {
+        if !header_eq(&parts.headers, header::CONNECTION, "upgrade") {
             return Err(InvalidConnectionHeader.into());
         }
 


randomairborne (commented):
According to [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#field.connection), this header might contain a comma-seperated list of other header names, and may not always just be upgrade. [(this is delineated by the hashtag)](https://www.rfc-editor.org/rfc/rfc9110#name-syntax-notation)

bestgopher:
> According to [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#field.connection), this header might contain a comma-seperated list of other header names, and may not always just be upgrade. [(this is delineated by the hashtag)](https://www.rfc-editor.org/rfc/rfc9110#name-syntax-notation)

Thanks for your reply. But when we use `header_contains`, `Upgradeaaaa` is a valid value.

randomairborne:
Parsing the whole header is likely considered either overkill or not worth the benefit. is an incorrect or malicious client sending this going to cause issues anywhere?

bestgopher:
It's just a requirement of [RFC6455](https://datatracker.ietf.org/doc/html/rfc6455). Is it possible to have multiple values for the `upgrade` header?

randomairborne:
No, I don't believe so. However, the Upgrade header isn't really relevant here.

bestgopher:
> No, I don't believe so. However, the Upgrade header isn't really relevant here.

Okay, I will close the pr. But I still stick to my opinion.

PR #2310, opened 2023-11-12
Add OptionalQuery extractor

### LINKED ISSUE #2309: OptionalQuery extractor

### Motivation

Currently, there is no ergonomic way to express the "handler accepts either all query parameters or none" in the handler signature.

Signature like this 
```
pub async fn login_form(
    Query(query): Query<Option<QueryParams>>,
) ->impl IntoResponse {
```
results in an error message from the deserialization library on request without query parameters.

Signature like this
```
pub async fn login_form(
   query: Option<Query<QueryParams>>,
) ->impl IntoResponse {
```
works but silently swallows all parsing errors.

### Proposal

Add an extractor `OptionalQuery`, similar to `OptionalPath` proposed in #1884 that succeeds with None if there are no query parameters present in the request URI, and attempts to deserialize the query parameters to T and return Some(T) on success, rejecting the request on failure.

### Alternatives
All of the alternatives I can think of require the user of the library to perform verification logic or custom parsing. 

- Use of `Result` as an extractor and then matching on `Err` variant will work, but it is much less ergonomic and requires nested match statements

```
pub async fn login_form(
    query: Result<Query<QueryParams>, QueryRejection>>,
) ->impl IntoResponse {
```

- Implementing custom parsing with `RawQuery` extractor is much less ergonomic and will duplicate parsing logic from the library into user code.

- Defining `QueryParams` like this and adding verification logic onto the struct also seems undesirable

```
struct QueryParams {
    a: Option<String>,
    b: Option<String>,
}
```

mikhailantoshkin:
## Motivation

Resolves #2309 

## Solution

`from_request_parts` does an additional check if query parameters are supplied. If none are present the `OptionalQuery(None)` is returned, otherwise it falls back to logic identical to `axum_extra::extract::Query`.


axum-extra/src/extract/mod.rs
```diff
@@ -10,6 +10,9 @@ mod form;
 #[cfg(feature = "cookie")]
 pub mod cookie;
 
+#[cfg(feature = "query")]
+mod optional_query;
```
davidpdrsn, line 14: I think this should live in the `query` module, since we already have that.

axum-extra/src/extract/optional_query.rs
```diff
@@ -0,0 +1,189 @@
...
+/// struct Pagination {
+///     page: usize,
+///     per_page: usize,
+/// }
+///
+/// // This will parse query strings like `?page=2&per_page=30` into `Some(Pagination)` and
+/// // empty query string into `None`
+/// async fn list_things(OptionalQuery(pagination): OptionalQuery<Pagination>) {
+///     match pagination {
+///         Some(Pagination{page, per_page}) => { /* return specified page */ },
```
davidpdrsn, line 32:
```suggestion
///         Some(Pagination { page, per_page }) => { /* return specified page */ },
```

davidpdrsn (commented):
Some minor nit picks but otherwise looks good!

mikhailantoshkin:
@davidpdrsn, made clippy happy again, take a look, please

davidpdrsn:
I'll take a look later today!

davidpdrsn (approved):

PR #2311, opened 2023-11-12
example: update axum version to the latest one for sse example

Migorithm:
## Motivation
Current examples for sse are outdated, with the latest Axum(at the time of writing, it's 0.6.2), it doesn't work! 
This PR is to renew the update with slightly better example code.


jplatte:
The examples are checked in CI and always up to date with the state of the main crates in this repo. Please note the warning about breaking changes in the toplevel readme file.

Issue #2313, opened 2023-11-13
Path extractor documentation does not mention that multiple Path arguments should not be used


uckelman:
- [x] I have looked for existing issues (including closed) about this

### Version

```
├── axum v0.6.20
│   ├── axum-core v0.3.4
│   ├── axum-macros v0.3.8 (proc-macro)
```

### Platform

```
Linux scylla 6.5.10-200.fc38.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Nov  2 19:59:55 UTC 2023 x86_64 GNU/Linux
```

### Description

A few days ago on Discord somebody posted code like the following, wondering why it doesn't work:
```rust
async fn handler(
    Path(package): Path<String>,
    Path(path): Path<String>,
) -> Result<(), AppError> {
    ....
}

 let app = axum::Router::new()
        .route("/:package/v1/*path", get(handler));
```

It's natural to see the `Path` extractor and think that you need one as an argument in your handler for each path component you want to capture. When I saw this, I knew exactly what the problem was, _because I had exactly the same problem_ a few weeks ago. 

While the documentation for `Path` does say that it extracts multiple components to tuples, it never says explicitly that you should not try to extract multiple components each with its own `Path` extractor like above. Saying this explicitly would head off a common confusion for new users.

PR #2314, opened 2023-11-13
Improve documentation of Path

uckelman:
## Motivation

This improves the documentation for `Path` by specifically calling out that one `Path` extractor captures all components for a route. New users fall into the trap of thinking that each capture needs its own `Path` argument., which does not work.

See #2313 for more details.

## Solution

This PR adds two sentences to the docs to clarify that only one `Path` should be used.

Fixes https://github.com/tokio-rs/axum/issues/2313


davidpdrsn (approved):
Thanks!

PR #2316, opened 2023-11-15
explain that `Router::merge` only merges paths and fallback

Emilgardis:
## Motivation

The current documentation on `Router::merge` suggests to me in a naive reading that layers applied in a `Router` that is then merged with another `Router` gets applied to the "top-level" `Router`s paths, this is not the case.

## Solution

Explain that only paths and fallbacks are merged, hopefully to dispel any wrongful assumptions.

### DIFF
--- a/axum/src/docs/routing/merge.md
+++ b/axum/src/docs/routing/merge.md
@@ -1,4 +1,4 @@
-Merge the paths of two routers into a single [`Router`].
+Merge the paths and fallbacks of two routers into a single [`Router`].
 
 This is useful for breaking apps into smaller pieces and combining them
 into one.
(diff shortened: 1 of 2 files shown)


axum/src/docs/routing/merge.md
```diff
@@ -1,4 +1,4 @@
-Merge two routers into one.
+Merge the paths of two routers into a single [`Router`].
```
Emilgardis, line 1: I was thinking about writing this as
```suggestion
Merge the paths and fallbacks of two routers into a single [`Router`].
```

but that doesn't make sense since there can only be one fallback

davidpdrsn (approved):
Thanks!

PR #2326, opened 2023-11-17
Add a Redirect factory function for 302 Found

ledoyen:
Hello 👋 

## Motivation

I use the `302 Found` redirect in a project of mine, and today building something custom with
```rust
(StatusCode::FOUND, [(LOCATION, HeaderValue::from_static(uri))]).into_response()
```

It is a little cumbersome, considering that the `Redirect` struct seems to supply shorthands for other redirection codes (303, 308, etc.)

## Solution

Add a factory function for `302 Found` in the `Redirect` struct implementation.

Let me know if I missed something 😉

### DIFF
--- a/axum/src/response/redirect.rs
+++ b/axum/src/response/redirect.rs
@@ -69,6 +69,17 @@ impl Redirect {
         Self::with_status_code(StatusCode::PERMANENT_REDIRECT, uri)
     }
 
+    /// Create a new [`Redirect`] that uses a [`302 Found`][mdn] status code.
+    ///
+    /// # Panics
+    ///
+    /// If `uri` isn't a valid [`HeaderValue`].
+    ///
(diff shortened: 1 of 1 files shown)


davidpdrsn:
Is there a reason you don’t use 307 Temporary Redirect?

rakshith-ravi:
> Factory function

\* PTSD from Java times ensues *

ledoyen:
> Is there a reason you don’t use 307 Temporary Redirect?

Yes, because I want some POST request to be able to redirect with a GET verb, whereas with a 307, the verb is kept by the browser for the redirected call.

davidpdrsn:
> because I want some POST request to be able to be redirected with a GET verb

That’s what “See Other” does. It’s used after a POST request to submit a form to then redirect to another page that is requested with GET.

ledoyen:
Indeed, my bad.

PR #2327, opened 2023-11-19
Update Cargo.toml http = 1.0.0

xxaier:
http release 1.0.0


davidpdrsn:
Did you try to compile the code as well? If so you would have seen that lots of code need to be changed.

I’m updating the dependencies in https://github.com/tokio-rs/axum/pull/1882

PR #2338, opened 2023-11-23
Add example showing how to use hyper's low level API

davidpdrsn:
If `axum::serve` doesn't work for people we can point them to this example.


jplatte (approved):
Nice!

PR #2340, opened 2023-11-24
GitHub: Add a Q&A discussion template

jplatte:
## Motivation

We're about to release 0.7, it would be nice if new posts in the Q&A discussion sections would contain the version of axum the poster used.

## Solution

Add a template! The [docs](https://docs.github.com/en/discussions/managing-discussions-for-your-community/creating-discussion-category-forms) are quite rudimentary, so I have no clue howe well the one I created will work. Might need further iteration after merge.

PR #2341, opened 2023-11-24
Fix error in Q&A discussion template

jplatte:
## Motivation

#2340 didn't work. Luckily after opening the file in GitHub's web UI, I got a somewhat usable error message.

## Solution

Fix the error x)
It was about indentation of the validations.required thing, but also that was on the wrong item in the first place.


davidpdrsn (approved):
Yaml \o/

PR #2343, opened 2023-11-24
refactor deprecated Cookie::named fn

deid84:
The cookie crate is now at version 0.18

In its changelog i see a deprecated fn and after doing cargo test i saw that Cookie::named could be now substituted by cookie::from or cookie::build

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -52,7 +52,7 @@ tower-service = "0.3"
 
 # optional dependencies
 axum-macros = { path = "../axum-macros", version = "0.3.7", optional = true }
-cookie = { package = "cookie", version = "0.17", features = ["percent-encode"], optional = true }
+cookie = { package = "cookie", version = "0.18.0", features = ["percent-encode"], optional = true }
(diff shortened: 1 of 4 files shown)


jplatte:
Good timing, given we're pretty close to releasing a breaking-change version and this is a public dependency! :)

jplatte (approved):
The cookie crate's [changelog](https://github.com/SergioBenitez/cookie-rs/blob/master/CHANGELOG.md#version-018) mentions that it now uses `impl Into<Cookie>` in many places. I think we should copy that design, but we can merge this PR first and do a follow-up. wdyt @davidpdrsn?

davidpdrsn:
> The cookie crate's [changelog](https://github.com/SergioBenitez/cookie-rs/blob/master/CHANGELOG.md#version-018) mentions that it now uses `impl Into<Cookie>` in many places. I think we should copy that design, but we can merge this PR first and do a follow-up. wdyt @davidpdrsn?

Good idea! I've filed https://github.com/tokio-rs/axum/issues/2347

PR #2346, opened 2023-11-24
Remove PATCH versions from all dependencies

Exotic0015:
## Motivation

The [SemVer](https://semver.org/) specification states that PATCH versions should only contain backward-compatible bug fixes. 

## Solution

Removing explicit PATCH version dependencies from cargo.toml files allows dependency maintainers to push bug fixes without the need for axum to update them manually.

### DIFF
--- a/axum-core/Cargo.toml
+++ b/axum-core/Cargo.toml
@@ -18,32 +18,32 @@ tracing = ["dep:tracing"]
 __private_docs = ["dep:tower-http"]
 
 [dependencies]
-async-trait = "0.1.67"
+async-trait = "0.1"
 bytes = "1.0"
 futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
-http = "1.0.0"
-http-body = "1.0.0"
-http-body-util = "0.1.0"
-mime = "0.3.16"
-pin-project-lite = "0.2.7"
-sync_wrapper = "0.1.1"
+http = "1.0"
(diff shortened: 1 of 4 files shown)


paolobarbolini:
In `cargo` it works a little bit differently. Writing `0.3.16` for example is the exact same as writing `^0.3.16`, which is the same as `>=0.3.16, <0.4.0` (see https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility).

Specifying the patch version is indeed useful as it allows specifying the minimum supported patch version of that dependency. `cargo` automatically picks the latest version of that dependency as allowed by the semver range when building a project for the first time (when `Cargo.lock` is missing or when adding a new dependency to `Cargo.toml`). If the dependency later releases further patch versions you can run the process again with `cargo update`

jplatte:
> Removing explicit PATCH version dependencies from cargo.toml files allows dependency maintainers to push bug fixes without the need for axum to update them manually.

This is not true. Cargo will already pull in newer versions of indirect dependencies with the specifications as-is. `foo = "1"`, `foo = "1.0"` and `foo = "1.0.0"` have exactly the same meaning in Cargo. This is documented in detail [here](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html).

Exotic0015:
I see, thanks for clarifying 😄

Issue #2347, opened 2023-11-24
Use `impl Into<Cookie>`


davidpdrsn:
We updated to the latest version of `cookie` in https://github.com/tokio-rs/axum/pull/2343 and they've changed a bunch of methods to use `impl Into<Cookie>`. [Their changelog has the details](https://github.com/SergioBenitez/cookie-rs/blob/master/CHANGELOG.md).

We should make similar changes in axum-extra. At least [`CookieJar::add`](https://docs.rs/axum-extra/latest/axum_extra/extract/cookie/struct.CookieJar.html#method.add) but probably also others.

jplatte:
Related, we should mention the upgrade in the changelog. I only thought about that after pressing merge on #2343.

PR #2348, opened 2023-11-24
fix: use impl Into<Cookie>

mseele:
closes https://github.com/tokio-rs/axum/issues/2347

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -12,10 +12,12 @@ and this project adheres to [Semantic Versioning].
 - **added:** `Clone` implementation for `ErasedJson` ([#2142])
 - **breaking:** Update to prost 0.12. Used for the `Protobuf` extractor
 - **breaking:** Make `tokio` an optional dependency
+- **breaking**: Functions and methods that previously accepted a `Cookie` now accept any `T: Into<Cookie>` ([#2348])
 
 [#1850]: https://github.com/tokio-rs/axum/pull/1850
 [#2142]: https://github.com/tokio-rs/axum/pull/2142
 [#2310]: https://github.com/tokio-rs/axum/pull/2310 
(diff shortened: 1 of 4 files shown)


axum-extra/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning].
 - **added:** `Clone` implementation for `ErasedJson` ([#2142])
 - **breaking:** Update to prost 0.12. Used for the `Protobuf` extractor
 - **breaking:** Make `tokio` an optional dependency
+- **breaking**: Functions and methods that previously accepted a `Cookie` now accept any `T: Into<Cookie>` ([#2347])
```
jplatte, line 15:
```suggestion
- **breaking**: Functions and methods that previously accepted a `Cookie` now accept any `T: Into<Cookie>` ([#2348])
```

Also needs a link definition below.

jplatte (commented):
Thanks for the PR!

This only modifies `CookieJar`, but I think `PrivateCookieJar` and `SignedCookieJar` have the same methods, right?

Finally a small note, if you start working on a help-wanted issue, I would recommend leaving a comment on the issue first to let others know you're working on it, to avoid duplicate work.

axum-extra/CHANGELOG.md
jplatte: The link definition is still missing.

mseele:
Thanks for pointing out that `PrivateCookieJar` and `SignedCookieJar` should also be refactored, I've updated them.

And sorry for not noting that I'm working on a PR. Just saw the issue and jumped right in. Will to that next time 👍

axum-extra/CHANGELOG.md
mseele: Ah, I see. You'll have that in a minute.

jplatte (approved):

PR #2350, opened 2023-11-24
Reduce hyper dependency

tottoto:
## Motivation

Reduces `hyper` dependency.

## Solution

Depends on `hyper` only when it is used.


jplatte (commented):
You are implicitly creating a `hyper` feature by not using `dep:` syntax for the feature dependency. I don't think that is what we want, right? I also wonder if any of the features that enable `hyper` + one of its features should be made to not enable hyper. Would have to be reviewed one by one.

tottoto:
Fixed the issue, which I forgot to add the `dep:` prefix.

tottoto:
The mix use of `dep:crate` and `crate/feature` seems to need to be waited until Rust 1.71. So changing `hyper` optional dependency does not seem possible at the moment.

jplatte:
Yes, `"dep:hyper"` and `"hyper/foo"` can't be mixed in older versions but you can just write `"dep:hyper", "hyper?/foo"`, which has the same meaning as `"hyper/foo"`.

tottoto:
Thanks. Fixed to use the workaround.

jplatte (approved):

Issue #2352, opened 2023-11-25
Update axum's graceful shutdown example


davidpdrsn:
I should be possible to update axum's [graceful shutdown example](https://github.com/tokio-rs/axum/blob/main/examples/graceful-shutdown/src/main.rs) by taking some inspiration from [hyper's example](https://github.com/hyperium/hyper/blob/master/examples/graceful_shutdown.rs) and [axum's server-with-hyper example](https://github.com/tokio-rs/axum/blob/main/examples/serve-with-hyper/src/main.rs).

davidpdrsn:
I've started looking at this.

PR #2355, opened 2023-11-26
Update graceful shutdown example

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/2352

Was a bit more work than I expected because hyper-util doesn't seem to support graceful shutdown yet.

@seanmonstar is this how you'd recommend doing graceful shutdown?

### DIFF
--- a/examples/graceful-shutdown/src/main.rs
+++ b/examples/graceful-shutdown/src/main.rs
@@ -74,7 +74,7 @@ async fn main() {
         // Clone the watch receiver and move it into the task.
         let close_rx = close_rx.clone();
 
-        // Spawn a task to handle the connection. That way we can multiple connections
+        // Spawn a task to handle the connection. That way we can serve multiple connections
         // concurrently.
         tokio::spawn(async move {
             // Hyper has its own `AsyncRead` and `AsyncWrite` traits and doesn't use tokio.
--- a/examples/graceful-shutdown/src/main.rs
+++ b/examples/graceful-shutdown/src/main.rs
@@ -105,7 +105,8 @@ async fn main() {
 
             loop {
                 tokio::select! {
-                    // Poll the connection to process requests.
+                    // Poll the connection. This completes when the client has closed the
+                    // connection, graceful shutdown has completed, or we encounter a TCP error.
                     result = conn.as_mut() => {
                         if let Err(err) = result {
                             debug!("failed to serve connection: {err:#}");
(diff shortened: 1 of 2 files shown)


examples/graceful-shutdown/src/main.rs
```diff
@@ -4,57 +4,165 @@
...
+
+        debug!("connection {remote_addr} accepted");
+
+        // We don't need to call `poll_ready` because `Router` is always ready.
+        let tower_service = app.clone();
+
+        // Clone the watch receiver and move it into the task.
+        let close_rx = close_rx.clone();
+
+        // Spawn a task to handle the connection. That way we can multiple connections
```
jplatte, line 77:
```suggestion
        // Spawn a task to handle the connection. That way we can serve multiple connections
```

examples/graceful-shutdown/src/main.rs
```diff
@@ -4,57 +4,165 @@
...
+                // `with_upgrades` is required for websockets.
+                .with_upgrades();
+
+            // `graceful_shutdown` requires a pinned connection.
+            let mut conn = std::pin::pin!(conn);
+
+            loop {
+                tokio::select! {
+                    // Poll the connection to process requests.
+                    result = conn.as_mut() => {
```
jplatte, line 109: The `conn` future only resolves when the clieng has closed the connection, there has been an unrecoverable TCP error, or the connection is idle and graceful shutdown has been requested, I guess? Might be worth extending the comment above.

examples/graceful-shutdown/src/main.rs
```diff
@@ -4,57 +4,165 @@
...
+    // We only care about the watch receivers that were moved into the tasks so close the residual
+    // receiver.
+    drop(close_rx);
+
+    // Close the listener to stop accepting new connections.
+    drop(listener);
+
+    // Wait for all tasks to complete.
+    debug!("waiting for {} tasks to finish", close_tx.receiver_count());
+    close_tx.closed().await;
```
jplatte, line 143: Nice, didn't know one could await the closing of all watch channel receivers.

jplatte (approved):

davidpdrsn:
@detro Yeah it leaves a lot to be desired, though I don't think it should deter anyone from updating but that's of course up to you.

I encourage you to help contributing to hyper-util. That way everyone benefits! The issue regarding graceful shutdown is [here](https://github.com/hyperium/hyper/issues/2862). Since wrapping up axum 0.7 I have started contributing myself and will continue to do so.

sunsided:
Also related: hyperium/hyper-util#66.

PR #2357, opened 2023-11-26
Prepare `serve` for potentially supporting graceful shutdown

davidpdrsn:
I agree that graceful shutdown is probably the most common thing missing from `axum::serve` and its not exactly easy with the current state of hyper-util (see https://github.com/tokio-rs/axum/pull/2355).

So I could probably convinced to add graceful shutdown support to `serve` 😅 I think a good API for that would be

```rust
axum::serve(listener, service)
    .graceful_shutdown(shutdown_signal_future())
    .await
```

Doing that requires returning some type we control from `serve` but we can use `IntoFuture` so you can still do `serve(...).await`.

This needs to go in before shipping 0.7 since its a breaking change.


jplatte (approved):
Good idea to prepare for the possibility without blocking the release on actually adding it!

PR #2361, opened 2023-11-27
Add a use of route_service with ServeFile to the static-file-server example

uckelman:
## Motivation

The docs for `Router::route_service` give an example of using `ServeFile`, but if you come to the static file server example first, you won't necessarily be aware of `route_service`.

## Solution

This PR adds an example using `route_service`, to make it easier to spot for people interested in serving static files.


davidpdrsn:
Looks good but you have to run `cargo fmt`

uckelman:
Huh, I'm surprised it wants that all on one line.

davidpdrsn (approved):

PR #2365, opened 2023-11-28
fix: back to forbid `unsafe`

Gumichocopengin8:
## Motivation
Thanks to the great work from #1882 by @davidpdrsn, Axum stopped vendoring `hyper-utils`. We can now switch from `#![deny(unsafe_code)]`  to `#![forbid(unsafe_code)]`.

## Solution

one of axum philosophies is [100% safe code](https://github.com/tokio-rs/axum?tab=readme-ov-file#safety), so just enabled back to `#![forbid(unsafe_code)]`


davidpdrsn (approved):
\o/

PR #2373, opened 2023-11-28
Added utility function to convert body into bytes

ItsEthra:
## Motivation
Sometimes it is useful to convert axum body into bytes, for that one needs to use [collect](https://docs.rs/http-body-util/0.1.0/http_body_util/trait.BodyExt.html#method.collect), so to avoid pulling in `http-body-util` I want to provide this function.

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **added:** Add `axum::body::to_bytes` ([#2373])
 
+[#2373]: https://github.com/tokio-rs/axum/pull/2373
+
 # 0.7.1 (27. November, 2023)
 
 - **fix**: Fix readme.
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- **added:** Add `axum::body::to_bytes` ([#2373])
 
 # 0.7.1 (27. November, 2023)
 
--- a/axum/src/body/mod.rs
+++ b/axum/src/body/mod.rs
@@ -26,6 +26,25 @@ use http_body_util::{BodyExt, Limited};
 /// # Ok(())
 /// # }
 /// ```
+///
+/// You can detect if the limit was hit by checking the source of the error:
+///
+/// ```rust
+/// use axum::body::{to_bytes, Body};
+/// use http_body_util::LengthLimitError;
+///
+/// # #[tokio::main]
+/// # async fn main() {
+/// let body = Body::from(vec![1, 2, 3]);
+/// match to_bytes(body, 1).await {
+///     Ok(_bytes) => panic!("should have hit the limit"),
+///     Err(err) => {
+///         let source = std::error::Error::source(&err).unwrap();
+///         assert!(source.is::<LengthLimitError>());
+///     }
+/// }
+/// # }
+/// ```
 pub async fn to_bytes(body: Body, limit: usize) -> Result<Bytes, axum_core::Error> {
     Limited::new(body, limit)
         .collect()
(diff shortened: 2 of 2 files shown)


axum/src/body/mod.rs
```diff
@@ -8,3 +8,17 @@ pub use bytes::Bytes;
...
+use http_body_util::BodyExt;
+
+/// Converts [`Body`] into [`Bytes`].
+/// ```rust
+/// # use axum::body::{to_bytes, Body};
+/// async fn foo() {
+///     let body = Body::from(vec![1, 2, 3]);
+///     assert_eq!(&to_bytes(body).await.unwrap()[..], &[1, 2, 3]);
+/// }
+/// ```
```
davidpdrsn, line 21:
```suggestion
/// Converts [`Body`] into [`Bytes`].
///
/// # Example
///
/// ```rust
/// use axum::body::{to_bytes, Body};
///
/// # async fn foo() -> Result<(), axum::Error> {
/// let body = Body::from(vec![1, 2, 3]);
/// let bytes = to_bytes(body).await?
/// assert_eq!(&bytes[..], &[1, 2, 3]);
/// # Ok(())
/// # }
/// ```
```

davidpdrsn (changes requested):
Sure but we need to careful about making sure the body has a limit. That was a common source of vulnerabilities with hyper's `to_bytes` since it didn't set any default limit.

What about

```rust
pub async fn to_bytes(body: Body, limit: usize)
```

Then you can do `to_bytes(body, usize::MAX)` if you really don't want a limit.

A limit does get applied by axum when using extractors but if you were to extract the whole `Request` and then use `to_bytes` then no limit would be applied.

ItsEthra:
I see, didn't think about that. How should I implement length limiting in that case? I'm thinking about converting into data frames stream and keep joining them until they exceed length limit.

davidpdrsn:
You can use https://docs.rs/http-body-util/latest/http_body_util/struct.Limited.html. That should handle everything for you.

axum/src/body/mod.rs
ItsEthra: Fixed

axum/src/body/mod.rs
```diff
@@ -8,3 +8,28 @@ pub use bytes::Bytes;
...
+/// use axum::body::{to_bytes, Body};
+///
+/// # async fn foo() -> Result<(), axum_core::Error> {
+/// let body = Body::from(vec![1, 2, 3]);
+/// // Use `usize::MAX` if you don't care about the maximum size.
+/// let bytes = to_bytes(body, usize::MAX).await?;
+/// assert_eq!(&bytes[..], &[1, 2, 3]);
+/// # Ok(())
+/// # }
+/// ```
```
davidpdrsn, line 28: Probably also good to show how to detect if the limit was hit

```suggestion
/// ```
///
/// You can detect if the limit was hit by checking the source of the error:
///
/// ```rust
/// use axum::body::{to_bytes, Body};
/// use http_body_util::LengthLimitError;
///
/// # #[tokio::main]
/// # async fn main() {
/// let body = Body::from(vec![1, 2, 3]);
/// match to_bytes(body, 1).await {
///     Ok(_bytes) => panic!("should have hit the limit"),
///     Err(err) => {
///         let source = std::error::Error::source(&err).unwrap();
///         assert!(source.is::<LengthLimitError>());
///     }
/// }
/// # }
/// ```
```

davidpdrsn (approved):
Awesome!

One last thing: Wanna add a line to the changelog?

ItsEthra:
Will add

axum/CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- **added:** Add `axum::body::to_bytes` ([#2373])
```
davidpdrsn, line 10: You're missing a link to the PR

ItsEthra: Oh right, I though github links it automatically

jplatte: GitHub does I think, but the changelog might be viewed elsewhere too, for example in one's editor.

davidpdrsn (approved):
Thanks!

PR #2374, opened 2023-11-29
Add support for routing shorthand.

exzachlyvv:
## Motivation

The current syntax for defining routes on a Router could be more terse and clear. Calling a "route" method is redundant and the HTTP method type is less obvious (especially if chaining multiple).

```
let appLonghand = Router::new().route("/", get(myHandler));
```

## Solution

We can improve the API by providing "get" etc methods directly on the Router struct

```
let appShorthand = Router::new().get("/", myHandler);
```

### DIFF
--- a/axum/src/docs/routing/route.md
+++ b/axum/src/docs/routing/route.md
@@ -124,6 +124,23 @@ async fn serve_asset(Path(path): Path<String>) {}
 # let _: Router = app;
 ```
 
+# Shorthand
+
+You can call `get`, `post`, `delete`, and any other `method_router`s directly on a Router instance.
+
+```rust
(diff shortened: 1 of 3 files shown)


davidpdrsn:
Thanks for the PR!

I think something like this is fine to add but I think it should go on [`RouterExt` in axum-extra](https://docs.rs/axum-extra/latest/axum_extra/routing/trait.RouterExt.html). That already has methods like `typed_get` which are similar.

I don't to add more methods directly to `Router` and make the API surface even bigger.

davidpdrsn:
Think I'll close this for now. If you're still interested in this then please open a new PR that adds it to axum-extra.

Issue #2356, opened 2023-11-26
Update remaining examples to axum 0.7


davidpdrsn:
This tracks the examples that haven't yet been updated to axum 0.7:

- [ ] [rest-grpc-multiplex](https://github.com/tokio-rs/axum/blob/main/examples/rest-grpc-multiplex/src/main.rs) - Blocked. Requires tonic supporting http 1.0 (https://github.com/hyperium/tonic/issues/1579)
- [x] [tls-graceful-shutdown](https://github.com/tokio-rs/axum/blob/main/examples/tls-graceful-shutdown/src/main.rs) - If we're using axum_server then its blocked but maybe it can be done using hyper directly. Not sure.
- [x] [tls-rustls](https://github.com/tokio-rs/axum/blob/main/examples/tls-rustls/src/main.rs) - Blocked. Requires updating axum_server.
- [x] [reqwest-response](https://github.com/tokio-rs/axum/blob/main/examples/reqwest-response/src/main.rs)
- [x] [http-proxy](https://github.com/tokio-rs/axum/blob/main/examples/http-proxy/src/main.rs) - Good to go, no blockers.
- [x] [reverse-proxy](https://github.com/tokio-rs/axum/blob/main/examples/reverse-proxy/src/main.rs) - No blockers. Can use [hyper-util's legacy client](https://docs.rs/hyper-util/latest/hyper_util/client/legacy/struct.Client.html).
- [x] [low-level-rustls](https://github.com/tokio-rs/axum/blob/main/examples/low-level-rustls/src/main.rs) - Fairly certain this one is good to go.
- [x] [low-level-openssl](https://github.com/tokio-rs/axum/blob/main/examples/low-level-openssl/src/main.rs) - Fairly certain this one is good to go.
- [x] [listen-multiple-addrs](https://github.com/tokio-rs/axum/blob/main/examples/listen-multiple-addrs/src/main.rs) - Good to go, no blockers (claimed by @davidpdrsn)
- [x] [testing](https://github.com/tokio-rs/axum/blob/main/examples/testing/src/main.rs) - No blockers. Can use [hyper-util's legacy client](https://docs.rs/hyper-util/latest/hyper_util/client/legacy/struct.Client.html) (claimed by @davidpdrsn)
[...]

PR #2375, opened 2023-11-29
Update example low-level-rustls to axum 0.7

erebe:
Following #2356, the PR update the example low-level-rustls to support axum 0.7


davidpdrsn (approved):
Thanks!

PR #2376, opened 2023-11-29
Update example low-level-openssl to axum 0.7

erebe:
Following https://github.com/tokio-rs/axum/issues/2356, the PR update the example low-level-openssl to support axum 0.7


davidpdrsn (approved):
Thanks!

Issue #2366, opened 2023-11-28
v0.6 hard-linked in docs


rben01:
- [x] I have looked for existing issues (including closed) about this

## Bug Report

In a handful of places in Axum’s docs, a link to an item is explicitly set to e.g., https://docs.rs/axum/0.6.0/axum/foo/bar.html, using `0.6.0` instead of `latest`.

### Version

0.7.1

### Platform

N/A

### Description

[A search for `0.6` in this repo](https://github.com/search?q=repo%3Atokio-rs%2Faxum+0.6+language%3ARust&type=code&l=Rust) yields links in that should be updated in a handful of files, e.g., https://github.com/tokio-rs/axum/blob/6dd5d0be9dee99f3c6c0b5be5e53599433f7aea3/axum-core/src/extract/mod.rs#L44

I believe these should be changed to use `latest` instead of `0.6` or `0.6.0`.

davidpdrsn:
Thanks for noticing. Do you wanna submit a PR to fix the links?

jplatte:
I don't think we should use `latest`, then the docs would be wrong in the future when we release a breaking version of axum-core and matchign version of axum.

jusexton:
I dont mind simply replacing `0.6.0` with `0.7.1` but what happens when `0.7.2` releases? Will this issue need to be re-opened if we're not going to use `latest`?

jplatte:
I think it's not that bad to link to a slightly outdated version, but also I think it works to use `/0.7/`.

jusexton:
Sounds good @jplatte ! PR is open. I also noticed there were quite a few links making use of `latest`. If you'd like the consistency of using explicit versions everywhere, I wouldn't mind making those changes as well.

jplatte:
Yeah, another PR to replace `latest` links sounds good. What links do you think should always be `latest`?

jusexton:
Doc link in the README referencing `latest` axum.

jplatte:
Right okay that sounds reasonable :)

PR #2379, opened 2023-11-29
Updated doc link versions to use more explicit 0.7 version.

jusexton:
As per conversation on https://github.com/tokio-rs/axum/issues/2366, updated doc links to use more explicit `0.7` version instead of `latest`


jplatte (approved):
💯

PR #2382, opened 2023-11-29
Rework error handling example

davidpdrsn:
This reworks the ye olde [error-handling-and-dependency-injection](https://github.com/tokio-rs/axum/blob/main/examples/error-handling-and-dependency-injection/src/main.rs). I've removed all the dependency injection bits because I think they're distracting and deserve their own example.

Hopefully this new version is easier to follow and can serve as a template for people to build upon.


examples/error-handling/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+                    let method = req.method();
+                    let uri = req.uri();
+
+                    // axum automatically adds this extension.
+                    let matched_path = req
+                        .extensions()
+                        .get::<MatchedPath>()
+                        .map(|matched_path| matched_path.as_str());
+
+                    tracing::debug_span!("request", %method, %uri, matched_path)
```
jplatte, line 73: As-is, the uri (and method) won't be in quotes but the matched path will, as the comment at the start of the file shows correctly. I find this pretty weird. Maybe use `let uri = req.uri().to_string();` above instead, and remove the `%` before `uri` here? For the method, it probably makes sense for it not to be in quotes.

examples/error-handling/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        let (status, message) = match self {
+            AppError::JsonRejection(rejection) => {
+                // This error is caused by bad user input so don't log it
+                (rejection.status(), rejection.body_text())
+            }
+            AppError::TimeError(err) => {
+                // Because `TraceLayer` wraps each request in a span that contains the request
+                // method, uri, etc we don't need to include those details here
+                tracing::error!(%err, "error from time_library");
```
jplatte, line 173: Maybe controversial, but I find the `err=` bit in the output a bit weird. IMHO tracing fields are much more useful for fields whose exact value you'd want to match when searching logs, rather than error messages that are generally just searched for specific keywords.

```suggestion
                tracing::error!("error from time_library: {err}");
```

jplatte (approved):

NVedsted:
Not sure what the chances are of me stumbling upon links to the deleted example just as this merges, haha. I just wanted to highlight these references to the deleted/renamed file:
- https://github.com/tokio-rs/axum/blob/5d388c8d91440bc2865a894a3587971a4e7bc47c/axum/src/docs/error_handling.md?plain=1#L46-L49
- https://github.com/tokio-rs/axum/blob/5d388c8d91440bc2865a894a3587971a4e7bc47c/axum/src/handler/mod.rs#L40-L43

Not familiar with contributing, so let me know if an issue would be preferable. :)

davidpdrsn:
Thanks for noticing! I've filed an issue https://github.com/tokio-rs/axum/issues/2464

PR #2391, opened 2023-12-01
Fix from_stream doc link to `Stream`

meepleek:
## Motivation

The link to `Stream` in the doc comment for `from_stream` incorrect.

## Solution

Fix the link 🙂.


davidpdrsn (approved):
Thanks!

Issue #2363, opened 2023-11-27
A route producing the Unit type `()` should generate a HTTP 204 response


jbethune:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

This is not something super important, but I want to describe a potential semantic improvement on how axum uses HTTP.

Axum currently returns HTTP status 200 for a route that returns the unit type:

```
async fn my_route() -> () { // explicit return type for clarity
    println!("I am not returning anything");
}
```
But a [HTTP 204 status code](https://www.rfc-editor.org/rfc/rfc9110.html#status.204) (=no content) would be more precise, because the unit type is essentially an empty value.

### Motivation

This is mostly useful for routes that get called via ajax. By being more explicit, API breakage can be detected more easily and ajax handling logic may branch on the status code.

Somewhat related: `Option<T>` return types could also benefit from a 200/204 distinction.

### Proposal

My understanding is that this part has to be changed:

https://github.com/tokio-rs/axum/blob/2e963cea9d35cd6b2583670146d25842db1f3618/axum-core/src/response/into_response.rs#L125-L129

This could be as simple as creating the response like above, but before returning, update the status code to 204.

### Alternatives

Just leave things as they are. Which would still be fine.

davidpdrsn:
I suspect that might be difficult because of how we implement `IntoResponse` for tuples but worth a shot! I agree it makes sense.

davidpdrsn:
Gonna close this. See https://github.com/tokio-rs/axum/pull/2393#issuecomment-1871626798 for why.

PR #2393, opened 2023-12-01
Make `()` and `None` return a HTTP 204 response instead of 200

jbethune:
## Motivation

I'm proposing an implementation for #2363: The unit type `()` and `None` should generate a HTTP 204 response.

## Solution

This PR does the following things:

* Change the implementation of `IntoResponse` for `()` to always return a HTTP 204 result
* ~~Add an implementation of `IntoResponse` for `Option<T: IntoResponse>` so that HTTP 204 is returned when the `Option` is `None`~~.
* Add unit tests for the above cases in `axum/src/response/mod.rs`. (Please let me know if they should be somewhere else)
* Adjust some of the existing unit tests to expect HTTP 204 responses. Many of the existing unit tests define routes that produce an empty result and the body of the response is not part of those unit tests.

### DIFF
--- a/axum-core/src/response/into_response.rs
+++ b/axum-core/src/response/into_response.rs
@@ -134,18 +134,6 @@ impl IntoResponse for Infallible {
     }
 }
 
-impl<T> IntoResponse for Option<T>
-where
-    T: IntoResponse,
-{
-    fn into_response(self) -> Response {
-        match self {
-            Some(v) => v.into_response(),
-            None => StatusCode::NO_CONTENT.into_response(),
-        }
-    }
-}
-
 impl<T, E> IntoResponse for Result<T, E>
 where
     T: IntoResponse,
--- a/axum-core/src/response/into_response.rs
+++ b/axum-core/src/response/into_response.rs
@@ -116,7 +116,7 @@ pub trait IntoResponse {
 
 impl IntoResponse for StatusCode {
     fn into_response(self) -> Response {
-        let mut res = ().into_response();
+        let mut res = Body::empty().into_response();
         *res.status_mut() = self;
         res
     }
@@ -124,13 +124,25 @@ impl IntoResponse for StatusCode {
 
 impl IntoResponse for () {
     fn into_response(self) -> Response {
-        Body::empty().into_response()
+        StatusCode::NO_CONTENT.into_response()
     }
 }
 
 impl IntoResponse for Infallible {
     fn into_response(self) -> Response {
         match self {}
+    }
+}
+
+impl<T> IntoResponse for Option<T>
+where
+    T: IntoResponse,
+{
+    fn into_response(self) -> Response {
+        match self {
+            Some(v) => v.into_response(),
+            None => StatusCode::NO_CONTENT.into_response(),
+        }
     }
 }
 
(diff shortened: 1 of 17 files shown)


jbethune:
There are also tests in the examples folder. This one fails (line 75):

https://github.com/tokio-rs/axum/blob/801a78a4cf387694636a98df4367443258e815a2/examples/handle-head-request/src/main.rs#L66-L81

Hm. The return value from the handler is

```
 ([("x-some-header", "header from GET")], "body from GET").into_response()
```
Which now produces a HTTP 204 response. I'm not sure yet why that is. I'll keep looking.

jplatte:
I think `Option` should be discussed separately. I seem to remember that I've wanted `None` to result in a 404 status at some point.

jbethune:
> I think `Option` should be discussed separately. I seem to remember that I've wanted `None` to result in a 404 status at some point.

That's a good point. I can easily remove the `Option` parts of this PR.

jbethune:
> About `()`, I'm wondering if it could be considered a breaking change. Seeing how many tests we needed to update I think it's likely users will have to update some of their tests as well. It also makes me think if changing it is even worth the churn.

To be fair, the tests that this PR updates are testing the core behaviour of axum. And the test behaviour is typically: I make a request and I expect a valid response. And most tests only care about the HTTP headers and not about the body. And it's very convenient to write a test that just returns `()` by using a semicolon. This is why there are so many tests that return `()` in the axum testsuite.

But projects that use axum as a library will have a lot more tests that expect an actual body in the response. The routes that return a `()` are called for their side-effects. And it is a good idea to be explicit about this by returning a 204 response. This also makes misuse and breakage in an API more visible to clients.

I've taken my time to update all of the tests in this PR. It was a lot of tests and I looked at each one individually to make sure I'm not overlooking something. But most tests in this project are fairly easy to read through and understand. It was really not that much work to fix them. It will probably be less of an issue for downstream projects. Or do you have a particular project in mind that would be heavily affected by this change?

davidpdrsn:
> Or do you have a particular project in mind that would be heavily affected by this change?

I'm not worried that updating the tests will be hard. I think you're right that it'll be easy. My question is more about whether it's worth the churn and whether this change really matters. I'm still not sure about that and not sure how to decide 🤔

davidpdrsn:
I'm still not sure I think this is a great idea so I think I'll be a bit conservative and close this for now. Thanks for the suggestion though!

oxalica:
I also encountered this.

> explicit about this by returning a 204 response.

If we do not want to change the `()` implementation, could we have a `pub struct NoContent;` under `axum::response` for this purpose? It's more expressive to have `async fn delete_user() -> Result<NoContent, MyJsonError>` than `-> Result<StatusCode/*what code? success or error?*/, MyJsonError>`

jplatte:
I like that idea! Want to send a PR?

lsalkeld:
I'd be interest in extending this further, but I'm not sure if it's within the interest of anyone maintaining Axum? For example:

`pub struct Ok<T: IntoResponse>(pub T);`
`pub struct Created<T: IntoResponse>(pub T);`
`pub struct Accepted;`

I'm not sure if it'd be better to have these be unit structs, which can be used similar to `(StatusCode, impl IntoResponse)`, or if they should *all* have a `T: IntoResponse` field rather than enforcing that certain status codes don't have any content (the exception being 204 No Content, of course).

PR #2395, opened 2023-12-01
Update example reverse-proxy to axum 0.7

markussilvan:
## Motivation

Following [#2356](https://github.com/tokio-rs/axum/issues/2356), the PR updates the example reverse-proxy to support axum 0.7.

## Solution

Use legacy _Client_ from hyper::util.


davidpdrsn (approved):
Thanks!

PR #2400, opened 2023-12-02
Gracefully handle accept errors in `serve`

davidpdrsn:
As pointed out in https://github.com/tokio-rs/axum/discussions/2388 `serve` doesn't handle errors when accepting a new connection.

This copies the approach that hyper 0.14 used https://github.com/hyperium/hyper/blob/v0.14.27/src/server/tcp.rs#L226-L248.

### DIFF
--- a/axum/src/serve.rs
+++ b/axum/src/serve.rs
@@ -8,6 +8,7 @@ use std::{
     net::SocketAddr,
     pin::Pin,
     task::{Context, Poll},
+    time::Duration,
 };
 
 use axum_core::{body::Body, extract::Request, response::Response};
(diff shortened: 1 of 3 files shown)


axum/src/serve.rs
```diff
@@ -147,7 +148,31 @@ where
...
+                        // > A possible scenario is that the process has hit the max open files
+                        // > allowed, and so trying to accept a new connection will fail with
+                        // > `EMFILE`. In some cases, it's preferable to just wait for some time, if
+                        // > the application will likely close some files (or connections), and try
+                        // > to accept the connection again. If this option is `true`, the error
+                        // > will be logged at the `error` level, since it is still a big deal,
+                        // > and then the listener will sleep for 1 second.
+                        //
+                        // hyper allowed customizing this but axum does not.
+                        error!("accept error: {}", e);
```
jplatte, line 171:
```suggestion
                        error!("accept error: {e}");
```

jplatte (approved):

PR #2411, opened 2023-12-07
Body: Implement `From<()>`

Turbo87:
## Motivation

In crates.io I was implementing a function with `request: Request<impl Into<Body>>` and my plan was to use it with `Request::get(...).body(()).unwrap()`. Unfortunately there was no built-in conversion from `()` to `Body` implemented though, so I had to use `Request::get(...).body("").unwrap()` instead. While it is the same amount of characters, the former is using `http_body_util::Empty` and the latter `http_body_util::Full`, which is assume is slightly less efficient?

## Solution

The solution is a simple `impl From<()> for Body` block :)

### DIFF
--- a/axum-core/src/body.rs
+++ b/axum-core/src/body.rs
@@ -84,6 +84,12 @@ impl Default for Body {
     }
 }
 
+impl From<()> for Body {
+    fn from(_: ()) -> Self {
+        Self::empty()
+    }
+}
+
 macro_rules! body_from_impl {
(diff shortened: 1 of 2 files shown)


Turbo87:
> Wanna add a note to the changelog as well?

done :)

davidpdrsn (approved):
Thanks!

I'll deal with CI separately because its unrelated to this change.

PR #2437, opened 2023-12-18
Fix comments on reqwest-response example for axum 0.7

gerardcl:
## Motivation

Fix the comments updated while enabling back the example reqwest-response for axum 0.7 (see #2356)

## Solution

Update comments with current context


davidpdrsn (approved):
Thanks!

PR #2439, opened 2023-12-19
make doc more rigorous

takkuumi:
## Motivation

It serves as a reminder that limiting lock scope is a good practice.


jplatte (approved):
I like it!

PR #2443, opened 2023-12-23
Update Cargo.toml readme field to point to workspace file

trchopan:
## Motivation

Unify the README.md file by using `readme` field [doc](https://doc.rust-lang.org/cargo/reference/manifest.html#the-readme-field)

## Solution

- Remove the symlinked README.md file in root workspace folder.
- Move the README.md from package axum to root workspace folder.
- Change the readme field to point to that file.


davidpdrsn:
But why? It works fine the way it is now.

trchopan:
> But why? It works fine the way it is now.

It works fine. I just suggest a way for avoid symlink in git. Not so important. If you think it is unnecessary I'm happy to close it.

davidpdrsn:
I think I'd rather leave it the way it was. If it aint broke don't fix it I guess.

PR #2451, opened 2023-12-26
Updated docs regarding constraints of Handler arguments

### LINKED ISSUE #2450: Improve the documentation in debugging_handler_type_errors.md regarding what implements Handler

## Feature Request

In the documentation about what functions implement the `Handler` trait, the docs only say that `FromRequest` is required. But looking at the source code, every argument is also given a trait bound of `Send`. So my feature request is to update the documentation to say that arguments of functions need to implement both `FromRequest + Send`. 

https://github.com/tokio-rs/axum/blob/3fda093806d43d64dd70cda0274cd3d73d29b6c7/axum/src/docs/debugging_handler_type_errors.md?plain=1#L7

### Motivation

The motivation for this is to reduce confusion in cases where handler are closures created in generic functions where the argument is generic, like `Json<T>`. In this case, a constraint of `T: FromRequest<...>` is not enough to ensure that any closures with `Json<T>` as an argument actually implement `Handler`

### Proposal

Replace
```
 - Take no more than 16 arguments that all implement [`FromRequest`]. 
```
with
```
 - Take no more than 16 arguments that all implement [`FromRequest`] and [`Send`]. 
```

For people that don't need this piece of information, this small add-on doesn't clutter the documentation, and for those that do, this could save a lot of confusion since `#[debug_handler]` can't be applied on closures and so the developer is left guessing why `Handler` is not implemented.

hydrogen602:
Fixes: #2450

Fixed documentation regarding the trait bounds of arguments in functions that implement `Handler`. The existing docs don't mention `Send` or `FromRequestParts`. Both of these have been added. 

## Motivation

This fixes #2450 as the documentation no longer accurately described the trait bounds of arguments in functions that implement `Handler`. Now it does again.

## Solution

Modified docs.

Feedback regarding the wording would be appreciated.

### DIFF
--- a/axum/src/docs/debugging_handler_type_errors.md
+++ b/axum/src/docs/debugging_handler_type_errors.md
@@ -4,7 +4,9 @@ For a function to be used as a handler it must implement the [`Handler`] trait.
 axum provides blanket implementations for functions that:
 
 - Are `async fn`s.
-- Take no more than 16 arguments that all implement [`FromRequest`].
+- Take no more than 16 arguments that all implement `Send`.
+  - All except the last argument implement [`FromRequestParts`].
+  - The last argument implements [`FromRequest`].
 - Returns something that implements [`IntoResponse`].
 - If a closure is used it must implement `Clone + Send` and be
 `'static`.


davidpdrsn (approved):
Thanks!

PR #2454, opened 2023-12-28
Update lib.rs doc : naming

DjebbZ:
## Motivation

I'm a Rust newbie, and reading `services` seems wrong, since they're called `handlers` right in the next paragraph. `handlers` makes more sense IMO, but I haven't read the rest of the docs yet.

What do you think?

## Solution

`services` -> `handlers`

### DIFF
--- a/axum/src/lib.rs
+++ b/axum/src/lib.rs
@@ -64,7 +64,7 @@
 //!
 //! # Routing
 //!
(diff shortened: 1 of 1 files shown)


davidpdrsn:
Technically the handlers are turned into services internally. You can also route to services with `Router::route_service` so I don't think this is wrong.

PR #2456, opened 2023-12-28
Implement `Clone` for `RouterIntoService`

### LINKED ISSUE #2455: RouterIntoService not Clone?

### Version

* axum 0.7

### Description

I'm trying to turn a Router into a Service so I can use it in `root_router.nest_service(...)`.

* Router has a [`.into_service()`](https://docs.rs/axum/latest/axum/routing/struct.Router.html#method.into_service) method
* [`Router::nest_service`](https://docs.rs/axum/latest/axum/routing/struct.Router.html#method.nest_service) requires the service to implement `Clone`
* [Router](https://docs.rs/axum/latest/axum/struct.Router.html) does implement `Clone`.
* [RouterIntoService](https://docs.rs/axum/latest/axum/routing/struct.RouterIntoService.html) does *not* implement `Clone`. 

All of that together means that, while I can write code that does

```rust
let sub_router = submod::create_router();

let router = Router::new()
  .nest("/sub", sub_router);
```

I can not write code that does

```rust
let sub_router_service = submod::create_router().into_service();

let router = Router::new()
  .nest_service("/sub", sub_router_service);
```

This seems like it *might* be an oversight? It looks to me like [RouterIntoService](https://docs.rs/axum/latest/axum/routing/struct.RouterIntoService.html) should trivially implement `Clone`? (It only has a Router in it.)

---

Why do I want to use `.nest_service` instead of `.nest` in the first place? It's because I have a layer on my subrouter, and when I use `.nest` that layer doesn't get called for non-matching routes.

bryanburgers:
Fixes #2455

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -12,12 +12,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **change:** Update tokio-tungstenite to 0.21 ([#2435])
 - **added:** Enable `tracing` feature by default ([#2460])
 - **added:** Support graceful shutdown on `serve` ([#2398])
+- **added:** `RouterIntoService` implements `Clone` ([#2456])
 
 [#2411]: https://github.com/tokio-rs/axum/pull/2411
 [#2433]: https://github.com/tokio-rs/axum/pull/2433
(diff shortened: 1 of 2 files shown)


davidpdrsn (commented):
Looks good! Wanna update the changelog as well?

davidpdrsn (approved):
Thanks!

PR #2457, opened 2023-12-28
Updating `tls-rustls` example

FemLolStudio:
Following #2356, updated the [tls-rustls](https://github.com/tokio-rs/axum/blob/main/examples/tls-rustls/src/main.rs) for the latest [axum-server](https://github.com/programatik29/axum-server) version, which supports [hyper](https://github.com/hyperium/hyper) 1.0


davidpdrsn (approved):
Oh awesome that axum-server has been updated! Thanks!

PR #2465, opened 2023-12-30
Add an example integration test for SSE

chenhunghan:
## Motivation

The SSE example was missing a test, would be nice to have one.

## Solution

This PR adds an example for integration test.

- Adding the integration test requiring some minor refactoring of the example, this PR moves the create of app into a separate `app()`.
-  Add dev-dependencies: `reqwest`/`reqwest-eventsource`/`eventsource-stream` that were used in tests.

### DIFF
--- a/examples/sse/src/main.rs
+++ b/examples/sse/src/main.rs
@@ -84,7 +84,7 @@ mod tests {
         async fn spawn_app(host: impl Into<String>) -> String {
             let host = host.into();
             // Bind to localhost at the port 0, which will let the OS assign an available port to us
-            let listener = TcpListener::bind(format!("{}:0", _host)).await.unwrap();
+            let listener = TcpListener::bind(format!("{}:0", host)).await.unwrap();
             // Retrieve the port assigned to us by the OS
             let port = listener.local_addr().unwrap().port();
             tokio::spawn(async {
@@ -92,7 +92,7 @@ mod tests {
                 axum::serve(listener, app).await.unwrap();
             });
             // Returns address (e.g. http://127.0.0.1{random_port})
-            format!("http://{}:{}", _host, port)
+            format!("http://{}:{}", host, port)
         }
         let listening_url = spawn_app("127.0.0.1").await;
 
--- a/examples/sse/src/main.rs
+++ b/examples/sse/src/main.rs
@@ -98,7 +98,7 @@ mod tests {
             let listener = TcpListener::bind(format!("{}:0", _host)).await.unwrap();
             // Retrieve the port assigned to us by the OS
             let port = listener.local_addr().unwrap().port();
-            let _ = tokio::spawn(async move {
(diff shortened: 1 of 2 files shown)


examples/sse/src/main.rs
```diff
@@ -44,16 +43,35 @@ async fn main() {
...
+}
+
+#[derive(Deserialize)]
+struct Repeat {
+    take_n: Option<usize>,
+}
+
 async fn sse_handler(
     TypedHeader(user_agent): TypedHeader<headers::UserAgent>,
+    repeat: Query<Repeat>,
```
davidpdrsn, line 63: Adding this specifically for testing feels wrong. I think the test should be able to just take a certain number of events itself. Perhaps using https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.take

chenhunghan: It's using `.take` at line 74, maybe the question is how do you recommend to pass n to `.take()` in the test?

chenhunghan: Like this? https://github.com/tokio-rs/axum/pull/2465/commits/9cc61f1318fc65b880b2bcc682cab98e2e6eb838

examples/sse/src/main.rs
```diff
@@ -63,3 +70,59 @@ async fn sse_handler(
...
+    use eventsource_stream::Eventsource;
+    use tokio::net::TcpListener;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn integration_test() {
+        // A helper function that spawns our application in the background
+        async fn spawn_app(host: impl Into<String>) -> String {
+            let _host = host.into();
```
davidpdrsn, line 85: Why the `_`?

```suggestion
            let host = host.into();
```

examples/sse/src/main.rs
```diff
@@ -63,3 +70,59 @@ async fn sse_handler(
...
+            // Bind to localhost at the port 0, which will let the OS assign an available port to us
+            let listener = TcpListener::bind(format!("{}:0", _host)).await.unwrap();
+            // Retrieve the port assigned to us by the OS
+            let port = listener.local_addr().unwrap().port();
+            tokio::spawn(async {
+                let app = app();
+                axum::serve(listener, app).await.unwrap();
```
davidpdrsn, line 92:
```suggestion
                axum::serve(listener, app()).await.unwrap();
```

davidpdrsn (commented):
A new nits but otherwise looks good!

examples/sse/src/main.rs
davidpdrsn: Yes!

examples/sse/src/main.rs
chenhunghan: Good catch

davidpdrsn (approved):
Thanks!

Issue #2287, opened 2023-10-24
`IntoResponse` implementation for `(StatusCode, R) where R: IntoResponse` overrides error status codes and might be a footgun


reivilibre:
- [x] I have looked for existing issues (including closed) about this

### Version

```
├── axum v0.6.20
│   ├── axum-core v0.3.4
│   ├── axum v0.6.20 (*)
```
(but I checked the source code for the latest version on docs.rs to notice this)

### Description

I'm looking at the code in `async fn hello()` below.

```rust
pub async fn serve(bind: SocketAddr) -> eyre::Result<()> {
    let router = Router::new()
        .route("/", get(hello));

    info!("Listening on {bind:?}");
    axum::Server::try_bind(&bind)
        .context("could not bind listen address")?
        .serve(router.into_make_service_with_connect_info::<SocketAddr>())
        .await?;

    Ok(())
}

async fn hello() -> impl IntoResponse {
    let mut map = HashMap::new();
    map.insert((1, 2), 3);
    (StatusCode::IM_A_TEAPOT, Json(map))
}
```

The general idea is: we're emitting a JSON payload with a custom status code.
But along the lines, something has gone wrong and the JSON serialisation fails (in this case: because the map has keys which are not strings and are not coerced into strings by `serde_json`).

Had I just used `Json(map)` as the return value, then I'd get a 500 Internal Server Error response — very reasonable!

However, instead I am getting the originally desired status code but with a text error message:

```
$ curl -i http://127.0.0.1:8080
HTTP/1.1 418 I'm a teapot
content-type: text/plain; charset=utf-8
content-length: 20
date: Tue, 24 Oct 2023 21:35:49 GMT

key must be a string
```

I find this a little bit unsettling and like it might turn out to be a bit of a footgun.
It's frankly not very likely with emitting Json (but not impossible) but when writing a custom `IntoResponse` to render a template or something like that, it seems quite reasonable to anticipate errors!

On the other hand, I don't know how/if you can fix this with the current API surface. But it felt like it was worth documenting.

(I should note this is entirely academic only and hasn't caused me any trouble in a real project or anything, so of course don't waste too much time on me!)

davidpdrsn:
Hm yeah that is a bit of a footgun for sure.

I think we could

1. Leave things as is but document it and provide a type that doesn't override 5xx status codes.
[...]

PR #2468, opened 2023-12-30
Fix `IntoResponse` for tuples overriding error response codes

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/2287

I think [3rd option](https://github.com/tokio-rs/axum/issues/2287#issuecomment-1778703040) is the way to go. If we went with [option 2](https://github.com/tokio-rs/axum/issues/2287#issuecomment-1778295170) then things like

```rust
(StatusCode::SEE_OTHER, [("location", "...")], StatusCode::INTERNAL_SERVER_ERROR).into_response()
```

would have to skip the `IntoResponseParts` in the middle as well as `StatusCode::SEE_OTHER`, otherwise we'd get a 500 with a `location` header.

However we probably shouldn't skip the `IntoResponseParts` if the first element isn't a status code

```rust
(Extension(error_that_happened), StatusCode::INTERNAL_SERVER_ERROR).into_response()
```

That feels somewhat inconsistent and combined with the fact that it might break middleware that "catch errors" and use tuples to construct new responses I think adding a specific extension to signal "IntoResponse failed" is nicer.

So this PR does just that. It adds `IntoResponseFailed` which if present in response extensions makes other response transformations not run. Thus the 500 status code of

```
(StatusCode::CREATED, Json::<NotJsonCompatible>(...))
```

is maintained.

## TODO

- [ ] Audit our `IntoResponse` and `IntoResponseParts` impls and add `IntoResponseFailed` where necessary.
- [ ] Make sure you can still write middleware that catch and transforms errors.
- [ ] Provide `OverrideAllStatusCodes(StatusCode)` newtype to opt-out and always set the status code.

### DIFF
--- a/axum-core/src/response/into_response.rs
+++ b/axum-core/src/response/into_response.rs
@@ -1,4 +1,4 @@
-use super::{IntoResponseParts, Response, ResponseParts};
+use super::{IntoResponseFailed, IntoResponseParts, Response, ResponseParts};
 use crate::{body::Body, BoxError};
 use bytes::{buf::Chain, Buf, Bytes, BytesMut};
 use http::{
@@ -328,7 +328,9 @@ where
 {
     fn into_response(self) -> Response {
         let mut res = self.1.into_response();
-        *res.status_mut() = self.0;
+        if res.extensions().get::<IntoResponseFailed>().is_none() {
+            *res.status_mut() = self.0;
+        }
         res
     }
 }
@@ -404,18 +406,16 @@ macro_rules! impl_into_response {
                 let ($($ty),*, res) = self;
 
                 let res = res.into_response();
-                let parts = ResponseParts { res };
-
-                $(
-                    let parts = match $ty.into_response_parts(parts) {
+                if res.extensions().get::<IntoResponseFailed>().is_none() {
+                    let parts = ResponseParts { res };
+                    let parts = match ($($ty,)*).into_response_parts(parts) {
                         Ok(parts) => parts,
-                        Err(err) => {
-                            return err.into_response();
-                        }
+                        Err(err) => return err.into_response(),
                     };
-                )*
-
-                parts.res
+                    parts.res
+                } else {
+                    res
+                }
             }
         }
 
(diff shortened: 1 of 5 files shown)


jplatte (commented):
Looks decent, but I'm wondering if we should ship this in 0.7.x. It seems pretty likely that some people are relying on `IntoResponseParts` from a tuple running even if the final tuple element produces an error (e.g. headers used for distributed tracing).

Also, re. `OverrideAllStatusCodes`, could `ForceStatusCode` be a better name? It's a bit shorter.

davidpdrsn:
> Looks decent, but I'm wondering if we should ship this in 0.7.x. It seems pretty likely that some people are relying on `IntoResponseParts` from a tuple running even if the final tuple element produces an error (e.g. headers used for distributed tracing).

Yeah we probably should wait for 0.8.

> Also, re. `OverrideAllStatusCodes`, could `ForceStatusCode` be a better name? It's a bit shorter.

👍

jplatte:
Do you think we should ship 0.8 soon / start planning for what should be part of it? The `async-trait` removal is another big breaking change, plus potentially requiring `Sync` for inner services (or something along those lines, I have some other ideas in that area). Those changes seem large enough to warrant another breaking release, though there might be other things we should wait for (e.g. another matchit release since the routing syntax is changing).

davidpdrsn:
Yeah I think we should ship sooner rather than later. We'll probably get some hype from being the first major framework to adopt AFIT 😅 although coordinating with matchit would be good.

yanns:
@jplatte I've fixed the merge conflicts from `main` and fixed the compilation errors.
Do you think it makes sense to spend time on this? Is it the right direction?

jplatte:
I still think this is probably the right direction, but am pretty swamped with life right now. The associated issue is part of the 0.9 milestone, I think that should help me find this again pretty quickly once I go from my current reactive only-look-at-notifications mode to something more proactive where I revisit the open tasks.

jplatte:
I should be able to take a look soon. Could you rebase, given CI failed previously and there are some merge conflicts? (I doubt they will be very complex)

jplatte:
I'll close this ancient PR and create a new one with a rebased version of this myself.

PR #2475, opened 2023-12-30
Add a separate trait for optional extractors

### LINKED ISSUE #2298: Remove or change `impl FromRequest* for Option<impl FromRequest*>`?
The possibility of using `Option<Extractor>` as an extractor has lead to a bunch of confusion because people have expected or wanted it to have different semantics for specific inner extractors. We should consider removing or altering the two implementations.

One option: We could have separate traits `OptionalFromRequest` and `OptionalFromRequestParts` and implement them only for a handful of types where it actually makes sense to discard the error. We could also implement different behavior, such as making `Option<Path<_>>` equivalent to the current `OptionalPath` from axum-extra. If you feel like this is worth exploring, I can open an issue.

This idea originally came up in https://github.com/tokio-rs/axum/discussions/2281#discussioncomment-7434310.

jplatte:
## Motivation

Closes #2298.

## Solution

- [x] Add `OptionalFromRequest`, `OptionalFromRequestParts`
  - Q: Should the trait items have different names from the ones in the non-optional traits?
- [x] Implement for relevant types
- [x] Deprecate `OptionalFoo` extractors
- [x] Documentation
- [x] Changelog


jplatte:
@davidpdrsn Have you had time to take a look? The main changes are done, and I don't want to do the remaining docs / changelog writing and cleanup if it's not yet clear whether this will be merged.

jplatte:
@mladedav @yanns both of you have been active with contributions recently. Would you be up for reviewing this? For context, see the linked issue.

axum-extra/src/extract/query.rs
```diff
@@ -152,9 +174,11 @@ impl std::error::Error for QueryRejection {
 ///
 /// [example]: https://github.com/tokio-rs/axum/blob/main/examples/query-params-with-empty-strings/src/main.rs
 #[cfg_attr(docsrs, doc(cfg(feature = "query")))]
+#[deprecated = "Use Option<Path<_>> instead"]
```
mladedav, line 177: Query instead of Path

mladedav:
I will read it more thoroughly tomorrow or over the weekend, but it makes sense.

I want to think on other ways to make the option work but on a first sight this may be the best solution.

I just noticed one typo.

mladedav (commented):
One thing that bugs me is that the optional variants use the same rejections so for example if someone uses the optional typed header there still is valid representation in the rejection. for the missing case.

That said the rejections are non exhaustive so they cannot handle explicitly everything so they may omit this one as well. It might just be confusing.

axum/src/extract/matched_path.rs
```diff
@@ -81,6 +82,21 @@ where
     }
 }
 
+#[async_trait]
+impl<S> OptionalFromRequestParts<S> for MatchedPath
+where
+    S: Send + Sync,
+{
+    type Rejection = MatchedPathRejection;
```
mladedav, line 90: I think this should be `Infallible` as missing is the only failure case.

And just to be sure, it seems to me like the only way `MatchedPath` would be missing is if it is accessed inside a fallback handler? If that's so, I think it would be better not to add this impl and have users just use `Result` if they want to guard against using it there.

examples/todos/src/main.rs
```diff
@@ -82,13 +82,11 @@ pub struct Pagination {
 }
 
 async fn todos_index(
-    pagination: Option<Query<Pagination>>,
+    Query(pagination): Query<Pagination>,
```
mladedav, line 85: This should not change I think.

mladedav (commented):
The code looks good to me.

The drawback I see here is having two distinct traits so it might be harder for users to discover how to properly use `Option`s. That said in hindsight, it might be better to force users to use `Result` instead of `Option` if they really mean to just discard any errors.

I am still unsure about whether it is better to use the same rejections for optional and normal extractors if the missing case is already handled in the normal rejection. I think we could use some kind of enum (probably very similar to `Option`) for extractors that have optional variants. But each extractor would have to define `IntoResponse` for its missing case.

I also wonder if we couldn't get away with just one `FromRequestParts` trait where the extracting method would return an enum with three cases like `MyResult<T, T::Missing, T::Rejection>`. Then we could maybe create the extractors with correct semantics without the need for multiple traits.

But if that's not possible or having the two traits is a better tradeoff than whatever issues that potential solution has, the implementation as is looks ok. I think we should just go over the implementations of `FromRequest` and `FromRequstParts` and add a few implementations for the optional conterparts.

examples/todos/src/main.rs
jplatte: Ah, this was intentional but I guess it should really be a separate PR with its own description.

jplatte:
Damn, I didn't realize I failed to respond to most of the feedback here since asking for it. I definitely think having more specific rejection types would be better. I think it's something I skipped initially because it was just easier that way for the first prototype.

I'll see when I can get back to this.

axum-macros/tests/typed_path/pass/option_result.rs
```diff
@@ -1,5 +1,5 @@
-use axum_extra::routing::{TypedPath, RouterExt};
 use axum::{extract::rejection::PathRejection, http::StatusCode};
+use axum_extra::routing::{RouterExt, TypedPath};
```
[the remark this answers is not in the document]
jplatte: Handled separately in https://github.com/tokio-rs/axum/pull/3049.

jplatte:
Suggestions for the docs would be very much welcome 😅

jplatte:
Gentle ping regarding re-review, @mladedav

mladedav (approved):
Thanks for the ping and sorry for the wait.

PR #2473, opened 2023-12-30
Require `Sync` for services

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/1950

This will allow us to use `Arc` inside `Router` to make it cheaper to clone.

## TODO

- [x] changelog

Issue #2449, opened 2023-12-26
Performance regression (requests per second) from 0.6.20 to 0.7.2 when using many routes


kapteinstein:
- [x] I have looked for existing issues (including closed) about this

### Version

```
├── axum v0.7.2
│   ├── axum-core v0.4.1
```

### Platform

```
Darwin 23.2.0 Darwin Kernel Version 23.2.0: Wed Nov 15 21:54:10 PST 2023; root:xnu-10002.61.3~2/RELEASE_X86_64 x86_64
```

### Description

The throughput (requests per second) seems to be reduced if I have many routes specified after upgrading to axum 0.7.2. This is not an issue with axum 0.6.20. I have made a small example/pseudo code to reproduce the unintended behaviour:

```
use axum::{
    routing::get,
    Router,
};

async fn hello() -> String {
    String::from("hello")
}

#[tokio::main]
async fn main() {
    // build our application with a single route
    let app = Router::new()
        .route("/", get(|| async { "Hello, World!" }))
        .route("/0", get(hello))
        .route("/1", get(hello))
        .route("/2", get(hello))
        .route("/3", get(hello))
        .route("/4", get(hello))
        .route("/5", get(hello))
        .route("/6", get(hello))
        // ... skipped the middle lines here on GitHub, 
        // but the pattern is followed to reproduce the unintended behaviour (1000 routes)
        .route("/996", get(hello))
        .route("/997", get(hello))
        .route("/998", get(hello))
        .route("/999", get(hello));

    // I used this for axum 0.7.2
    let listener = tokio::net::TcpListener::bind("[::1]:3444").await.unwrap();
    axum::serve(listener, app).await.unwrap();

    // I used this for axum 0.6.20
    axum::Server::bind(&"[::1]:3444".parse().unwrap())
        .serve(app.into_make_service())
        .await
        .unwrap();
```

I used `oha` to measure the throughput: `oha --ipv6 --no-tui -z 10s http://localhost:3444` 

for axum 0.7.2:
```
  Success rate:	1.0000
  Total:	10.1465 secs
  Slowest:	0.0312 secs
  Fastest:	0.0006 secs
  Average:	0.0030 secs
  Requests/sec:	13334.5141
```

for axum 0.6.20:
```
  Success rate:	1.0000
  Total:	10.1471 secs
  Slowest:	0.0302 secs
  Fastest:	0.0001 secs
  Average:	0.0009 secs
  Requests/sec:	48525.3179
```

[...]

PR #2476, opened 2023-12-30
Internally `Arc` `Router`

davidpdrsn:
Based on https://github.com/tokio-rs/axum/pull/2473.

With `Router` being `Sync` we can now store it as an `Arc` internally which should help a lot with cloning it.

This might impact https://github.com/tokio-rs/axum/issues/2449.

## TODO

- [ ] Changelog
- [ ] Benchmark it and see if there is a difference

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -122,12 +122,23 @@ where
         }
     }
 
-    fn update_router<F, S2>(self, f: F) -> Router<S2>
+    fn map_inner<F, S2>(self, f: F) -> Router<S2>
     where
         F: FnOnce(RouterInner<S>) -> RouterInner<S2>,
     {
         Router {
             inner: Arc::new(f(self.into_inner())),
+        }
+    }
+
+    fn tap_inner_mut<F>(self, f: F) -> Self
+    where
+        F: FnOnce(&mut RouterInner<S>),
+    {
+        let mut inner = self.into_inner();
+        f(&mut inner);
+        Router {
+            inner: Arc::new(inner),
         }
     }
 
@@ -146,9 +157,8 @@ where
     #[doc = include_str!("../docs/routing/route.md")]
     #[track_caller]
     pub fn route(self, path: &str, method_router: MethodRouter<S>) -> Self {
-        self.update_router(|mut this| {
+        self.tap_inner_mut(|this| {
             panic_on_err!(this.path_router.route(path, method_router));
-            this
         })
     }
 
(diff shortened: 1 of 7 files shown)


axum/src/routing/mod.rs
```diff
@@ -111,22 +113,47 @@ where
...
+            inner: Arc::new(RouterInner {
+                path_router: Default::default(),
+                fallback_router: PathRouter::new_fallback(),
+                default_fallback: true,
+                catch_all_fallback: Fallback::Default(Route::new(NotFound)),
+            }),
+        }
+    }
+
+    fn update_router<F, S2>(self, f: F) -> Router<S2>
```
jplatte, line 125: I would call this `map_inner`. Maybe we should also have a `tap_inner[_mut]` for an `FnOnce(&mut RouterInner<S>)` that's implemented in terms of this fn, would simplify a few call sites.

jplatte (commented):
Nice, do we have any benchmarks to check the before/after for?

davidpdrsn:
> Nice, do we have any benchmarks to check the before/after for?

Yeah we have some benchmarks [in this repo](https://github.com/tokio-rs/axum/blob/main/axum/benches/benches.rs). I'm gonna try and run them. Also curious if it impacts https://github.com/tokio-rs/axum/issues/2449

jplatte (approved):
Code looks good, curious about the benchmark results!

davidpdrsn:
Tested with `cargo bench -- routing` which runs a benchmark of a router with 1000 routes.

On `main`:

```
Running "routing" benchmark
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:60849/foo/bar/baz for 10 second(s)
    Latencies:
      0.19ms   0.06ms   0.09ms   2.07ms
    Requests:
      Total: 515851  Req/Sec: 51581.42
    Transfer:
      Total: 36.90 MB Transfer Rate: 3.69 MB/Sec
```

On this branch

```
Running "routing" benchmark
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:60837/foo/bar/baz for 10 second(s)
    Latencies:
      0.07ms   0.02ms   0.01ms   0.77ms
    Requests:
      Total: 1366847 Req/Sec: 136678.29
    Transfer:
      Total: 97.76 MB Transfer Rate: 9.78 MB/Sec
```

So went from 5.1581 rps to 13.6678. Thats pretty nuts 😅

jplatte:
Your dots in those numbers are in a weird position but yay 😄

davidpdrsn:
> Your dots in those numbers are in a weird position but yay 😄

lol sorry about that 😂

> Maybe we should look into running some benches in CI (measuring instruction counts for reproducibility) to catch obvious regressions in perf?

Never tried something like that but worth a shot!

jplatte:
Superseded by #2483, right?

davidpdrsn:
Yes!

Not sure if we still require `Sync` for services or if the mutex approach is fine.

PR #2478, opened 2023-12-31
Remove duplication in serving with and without graceful shutdown

davidpdrsn:
I guess serving without graceful shutdown is the same as serving with shutdown signal that never completes. Does result in the non-graceful case doing a bit more work but since you normally just spawn one server its probably fine.


jplatte (approved):

jplatte:
Closing in favor of https://github.com/tokio-rs/axum/pull/2803.

Issue #2474, opened 2023-12-30
Add `serve_uds` to axum-extra


davidpdrsn:
`axum::serve` intentionally only supports `TcpListener` because it greatly simplifies the API. However that means serving over a unix domain socket requires [a lot of boilerplate](https://github.com/tokio-rs/axum/blob/main/examples/unix-domain-socket/src/main.rs).

We should add a dedicated `serve_uds` function to axum-extra that makes that easier. I should be possible to more or less copy `axum::serve` and use Tokio's `UnixListener` instead.

vi:
Will it also make serving connections over arbitrary `AsyncRead`+`AsyncWrite` pairs easier?

davidpdrsn:
No. hyper 0.14's server used to that but it [leads to nasty trait bounds](https://docs.rs/hyper/0.14.28/hyper/server/struct.Builder.html#method.serve) which I really want to avoid.

fundon:
A trait for a listener already exists: https://docs.rs/tokio-util/latest/tokio_util/net/trait.Listener.html

davidpdrsn:
Oh neat! That might be worth trying to use.

jplatte:
So I guess #2479 will work for UDS and we don't actually want to add a separate `serve_uds`?

davidpdrsn:
Yes!

Turbo87:
@jplatte with https://github.com/tokio-rs/axum/pull/2941 merged, is this issue still relevant?

jplatte:
No, that PR should have closed this issue.

PR #2479, opened 2024-01-01
Make `serve` generic over the listener and IO types

davidpdrsn:
Based on https://github.com/tokio-rs/axum/pull/2478

Fixes https://github.com/tokio-rs/axum/issues/2474

It's actually not as bad as I had feared so I think we can do this for 0.8.

### DIFF
--- a/axum/src/serve.rs
+++ b/axum/src/serve.rs
@@ -25,7 +25,7 @@ use tokio::{
 use tower::util::ServiceExt;
 use tower_service::Service;
 
-/// TODO
+/// Types that can listen for connections.
 pub trait Listener: Send + 'static {
     /// The listener's IO type.
     type Io: AsyncRead + AsyncWrite + Unpin + Send + 'static;
--- a/axum/src/serve.rs
+++ b/axum/src/serve.rs
@@ -6,7 +6,6 @@ use std::{
     future::{poll_fn, Future, IntoFuture},
     io,
     marker::PhantomData,
-    net::SocketAddr,
     sync::Arc,
     time::Duration,
 };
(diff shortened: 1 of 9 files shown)


axum/src/serve.rs
```diff
@@ -19,12 +18,59 @@ use hyper_util::{
...
+
+    /// The listener's address type.
+    type Addr: Send;
+
+    /// Accept a new incoming connection to this listener
+    fn accept(&mut self) -> impl Future<Output = io::Result<(Self::Io, Self::Addr)>> + Send;
+
+    /// Returns the local address that this listener is bound to.
+    fn local_addr(&self) -> io::Result<Self::Addr>;
+}
```
davidpdrsn, line 41: We could also use tokio-util's [`Listener`](https://docs.rs/tokio-util/latest/tokio_util/net/trait.Listener.html) trait however I don't think tokio-util should be in our public API. By writing our own we can also use `async fn accept` rather than `fn poll_accept`.

examples/unix-domain-socket/src/main.rs
```diff
@@ -118,22 +92,14 @@ mod unix {
         peer_cred: UCred,
     }
 
-    impl connect_info::Connected<&UnixStream> for UdsConnectInfo {
-        fn connect_info(target: &UnixStream) -> Self {
-            let peer_addr = target.peer_addr().unwrap();
-            let peer_cred = target.peer_cred().unwrap();
-
+    impl connect_info::Connected<IncomingStream<'_, UnixListener>> for UdsConnectInfo {
```
jplatte, line 95: Is there some reason something like this can't be part of axum? I think we could implement at least `impl Connected<unix::SocketAddr> for unix::SocketAddr`, `impl Connected<IncomingStream<'_, UnixListener>> for unix::SocketAddr` and `impl Connected<IncomingStream<'_, UnixListener>> for UCreds`, right?

davidpdrsn: You cannot do `impl Connected<unix::SocketAddr> for unix::SocketAddr` because `unix::SocketAddr` is not `Clone`.

But we could have some `UdsConnectInfo` in axum that uses `Arc<unix::SocketAddr>` (similarly to [what tonic has](https://docs.rs/tonic/latest/tonic/transport/server/struct.UdsConnectInfo.html)) but I've never actually used unix streams myself so I'm not sure what info that type should contain. Do you have an opinion about that?

jplatte: Hm, no. I've used UDS w/ axum once, but didn't care about connection info then.

davidpdrsn: In that case I think we should just leave it as is. We can always add it if someone requests it and has a use case.

jplatte (approved):

davidpdrsn:
> Is any help needed to bring this change over the finish line?

Nope. Its a breaking change and don't wanna start merging breaking PRs just yet. It is done otherwise.

lojewalo:
> Is any help needed to bring this change over the finish line?

I'm not involved with the project, but seeing as this is a breaking change, it's slotted for 0.8 as far as I can tell, so not sure how soon it will be until this is merged.

As a workaround for myself, I cloned the project and merged this PR specifically into the latest version. If it's viable to use for your situation, you can patch it in via the below snippet. I'm using this in prod with no issues, but YMMV obviously.

```toml
# in your Cargo.toml
[patch.crates-io]
axum = { git = "https://github.com/lojewalo/axum.git", branch = "uds-v0.7.3" }
```

kamulos:
I came here via the issue #2494. If this makes it easier to accept tls connections, could you also provide an example how to do it without the `axum-server` crate?

I am at the moment a bit frustrated about the slow maintenance of the `axum-server` crate and would prefer to avoid depending on it.

elichai:
Any updates on where this stands with the maintainers? do you want to experiment with different designs or can this be merged as-is? 
This also solves for me a case of listening over VSOCK with something a little bit different than the regular TCP syscall-flow

jplatte:
This is a PR by the primary maintainer, approved by the co-maintainer. It's going to land, the question is just when. Which I unfortunately can't give you any details about, you're just going to have to be patient (or use this branch via [`[patch]`](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section)).

xtexChooser:
This PR is added to 0.8 milestone. Will it land when 0.8 publishes?

jplatte:
Yes.

jplatte:
Closing in favor of #2941.

PR #2480, opened 2024-01-02
chore: update feature flag description

lvillis:
## Motivation

Tracing is enabled by default in axum 0.7.3

## Solution

Update lib.rs


davidpdrsn (approved):
Thanks!

PR #2481, opened 2024-01-02
Make Router Sync without breaking the API

jplatte:
## Motivation

https://github.com/tokio-rs/axum/pull/2476 is currently based on https://github.com/tokio-rs/axum/pull/2473, which is a breaking change. Thus, we can't ship it in a patch release for 0.7 to recover performance of existing apps that have upgraded to axum 0.7.

## Solution

Use `Mutex`es in a few places that only ever actually need to be locked for cloning, to make `Router` `Sync` and enable #2476 to be merged as a non-breaking change. For axum 0.8, we can then revert this PR as part of #2473.

### DIFF
--- a/axum/src/boxed.rs
+++ b/axum/src/boxed.rs
@@ -1,4 +1,4 @@
-use std::{convert::Infallible, fmt};
+use std::{convert::Infallible, fmt, sync::Mutex};
 
 use crate::extract::Request;
 use tower::Service;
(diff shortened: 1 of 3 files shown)


davidpdrsn (approved):
Nice!

Probably worth benchmarking and comparing against https://github.com/tokio-rs/axum/pull/2473 but I'm pretty sure we just end up cloning the one `Route` we actually call so the short lock shouldn't matter.

davidpdrsn:
I'm gonna redo https://github.com/tokio-rs/axum/pull/2476 on top of these changes.

PR #2483, opened 2024-01-02
Internally `Arc` `Router`, without breaking changes

davidpdrsn:
Applies https://github.com/tokio-rs/axum/pull/2476 on top of https://github.com/tokio-rs/axum/pull/2481, thus without breaking changes.

I also added a wrapper around mutex that, in test mode, tracks the number of times its been locked, so we can write a test against it.

I ran our benchmarks and saw no difference between https://github.com/tokio-rs/axum/pull/2476 and the lock 💪

### DIFF
--- a/axum/src/util.rs
+++ b/axum/src/util.rs
@@ -1,6 +1,8 @@
 use pin_project_lite::pin_project;
 use std::{ops::Deref, sync::Arc};
 
+pub(crate) use self::mutex::*;
+
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub(crate) struct PercentDecodedStr(Arc<str>);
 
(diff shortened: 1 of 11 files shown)


axum/src/util.rs
```diff
@@ -59,3 +61,65 @@ fn test_try_downcast() {
     assert_eq!(try_downcast::<i32, _>(5_u32), Err(5_u32));
     assert_eq!(try_downcast::<i32, _>(5_i32), Ok(5_i32));
 }
+
+// `AxumMutex` is a wrapper around `std::sync::Mutex` which, in test mode, tracks the number of
+// times its been locked on the current task. That way we can write a test to ensure we don't
```
jplatte, line 66:
```suggestion
// times it's been locked on the current task. That way we can write a test to ensure we don't
```

axum/src/util.rs
```diff
@@ -59,3 +61,65 @@ fn test_try_downcast() {
...
+// accidentally introduce more locking.
+//
+// When not in test mode its a literal type alias for `std::sync::Mutex`.
```
jplatte, line 69:
```suggestion
// When not in test mode, it is just a type alias for `std::sync::Mutex`.
```

jplatte (approved):

Issue #2489, opened 2024-01-03
Implement `TypedPath` for `WithRejection<TypedPath, _>`


gpluscb:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

Allow customisation of `TypedPath` rejections using `WithRejection` as an alternative to the rejection customisation of the `TypedPath` macro.

### Motivation

Rejections for types implementing `TypedPath` can be customised already when using the `TypedPath` macro. However, that requires the desired rejection type is available where the struct implementing `TypedPath` is located.
This might not always be the case, especially when paths are defined in a different place than the rest of the server. For example, I have an API server crate and an API client crate in a workspace that share some common types in a third crate in the same workspace. To make sure I don't typo urls in the client crate, and to make the interaction between the crates more robust in general, I'd like to put `TypedPath` structs in the common crate, such that both the client and server crate can make use of them.
I do not have access to the server error type, which I would like to use as the rejection type, in this common crate.
Usually I'd work around this by just using `WithRejection` in the handlers in the server crate, but this doesn't work in this case because `WithRejection` does not implement `TypedPath`.

### Proposal

`WithRejection<E, R>` should implement `TypedPath` where `E: TypedPath`. I believe this shouldn't be breaking. There might be subtleties that I'm not familiar with that make this harder.

### Alternatives

For the example given above, it is possible to move the rejection type to the common crate and to utilise the custom rejection capabilities of the macro. However, this would pull in types dependencies into the common crate and therefore the API client that are not needed there.
Similarly, a custom rejection type could be added, but this also has the issue of doing work that should be done in the server crate (deciding what rejection responses should look like) in the common crate.
Another workaround could be to implement `TypedPath` and `Display` by hand in the common crate and then implement `FromRequest` in the server crate, but I think orphan rules could rule that option out. Either way it would be very cumbersome and error-prone compared to the macro and `WithRejection`.

davidpdrsn:
Sure! I've submitted https://github.com/tokio-rs/axum/pull/2491

PR #2491, opened 2024-01-04
Implement `TypedPath` for `WithRejection<TypedPath, _>`

davidpdrsn:
Fixes https://github.com/tokio-rs/axum/issues/2489

### DIFF
--- a/axum-extra/src/extract/with_rejection.rs
+++ b/axum-extra/src/extract/with_rejection.rs
@@ -2,7 +2,7 @@ use axum::async_trait;
 use axum::extract::{FromRequest, FromRequestParts, Request};
 use axum::response::IntoResponse;
 use http::request::Parts;
-use std::fmt::Debug;
+use std::fmt::{Debug, Display};
 use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut};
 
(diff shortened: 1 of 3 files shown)


axum-extra/src/extract/with_rejection.rs
```diff
@@ -137,6 +137,27 @@ where
     }
 }
 
+#[cfg(feature = "typed-routing")]
+const _: () = {
```
jplatte, line 141: Why do it this way? I find the `const _` trick a bit weird.

davidpdrsn: Just so you don't need to put `#[cfg(feature = "typed-routing")]` on the imports as well (unless you want warnings) or use fully qualified paths.

jplatte: I think I'd prefer either of those solutions 😅

cc https://github.com/rust-lang/rust/issues/65516 ;)

davidpdrsn: I hope that doesn't land 😅 I think its pretty useful to be able to define stuff everywhere though I do see its a nightmare from and IDE point of view.

But I've fixed it in c1b27565976b0

jplatte (approved):

PR #2498, opened 2024-01-09
fix link to serde::Deserialize

yanns:
https://github.com/tokio-rs/axum/pull/2431/files#r1445838251

### DIFF
--- a/axum-extra/src/extract/json_deserializer.rs
+++ b/axum-extra/src/extract/json_deserializer.rs
@@ -10,7 +10,7 @@ use std::marker::PhantomData;
 
 /// JSON Extractor for zero-copy deserialization.
 ///
-/// Deserialize request bodies into some type that implements [`serde::Deserialize<'de>`].
+/// Deserialize request bodies into some type that implements [`serde::Deserialize<'de>`](`serde::Deserialize`).
 /// Parsing JSON is delayed until [`deserialize`](JsonDeserializer::deserialize) is called.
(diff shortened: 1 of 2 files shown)


axum-extra/src/extract/json_deserializer.rs
```diff
@@ -10,7 +10,7 @@ use std::marker::PhantomData;
 
 /// JSON Extractor for zero-copy deserialization.
 ///
-/// Deserialize request bodies into some type that implements [`serde::Deserialize<'de>`].
+/// Deserialize request bodies into some type that implements [`serde::Deserialize<'de>`](`serde::Deserialize`).
```
jplatte, line 13: This is not what I suggested, can you confirm that what I wrote also works?

```suggestion
/// Deserialize request bodies into some type that implements [`serde::Deserialize<'de>`][serde::Deserialize].
```

I would prefer it because round-bracket links are more commonly used for regular HTTP links (not touched by rustdoc), and the extra backticks seem pointless.

yanns: yes this is also working

jplatte (approved):
Thanks!

jplatte:
Mind adding an entry to the changelog`

yanns:
> Mind adding an entry to the changelog?

https://github.com/tokio-rs/axum/pull/2498/commits/2b8104ded8be31a5208c0dfd8b19aa4dc0404c15

PR #2503, opened 2024-01-10
feat(typed_header_rejection): add is_missing utility to the reason

joshuachp:
## Motivation

It's hard, when creating your own extractor, to check if a `TypedHeader` is missing or there was an actual error. The `TypedHeaderRejectionReason` doesn't/can't implement `PartialEq` since it include the [headers_core](https://docs.rs/headers-core/0.3.0/headers_core/index.html)::[Error](https://docs.rs/headers-core/0.3.0/headers_core/struct.Error.html#) variant, so you need to add a match to check.

```rust
match typed_header {
   Ok(_) => todo!(),
   Err(err) if matches!(err.reason(), TypedHeaderRejectionReason::Missing)  => todo!(),
   Err(_) => todo!()
}
```

## Solution

Add a utility function to check if the `TypedHeaderRejection` is missing. Example:

```rust
match typed_header {
   Ok(_) => todo!(),
   Err(err) if err.is_missing() => todo!(),
   Err(err) => todo!()
}
```

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -9,8 +9,10 @@ and this project adheres to [Semantic Versioning].
 
 - **added:** Implement `TypedPath` for `WithRejection<TypedPath, _>`
 - **fixed:** Documentation link to `serde::Deserialize` in `JsonDeserializer` extractor ([#2498])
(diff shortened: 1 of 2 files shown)


davidpdrsn (commented):
Looks good! Wanna update the changelog as well?

joshuachp:
Updated the CHANGELOG

jplatte (approved):

PR #2507, opened 2024-01-11
Add Scheme extractor 

### LINKED ISSUE #2504: Add Scheme extractor

## Feature Request

I'd like an easy way to see the scheme / protocol section of the request, which can be found in various headers and the URI itself.

### Motivation
Similar to #826, extracting of the scheme of a URI is desirable.

### Proposal

```rust
use axum::extract::Scheme;

async fn foo(Scheme(scheme): Scheme) -> ...
```

### Alternatives
Manually extracting the scheme via the parts of a URI / analyzing various headers, which is a tedious process.

If this is something that should be part of `axum`, I can submit a Pull Request with the implementation I already have (adapted from the existing extractor for hosts), and follow it up with some tests.

bengsparks:
## Motivation
Closes #2504.

## Solution
Detect the scheme in the following order:

1. From `proto` within the `Forwarded` header
2. Reading from `X-Forwarded-Proto` 
3. Extracting from the URI

### DIFF
--- a/axum-extra/src/extract/host.rs
+++ b/axum-extra/src/extract/host.rs
@@ -139,7 +139,7 @@ mod tests {
         assert_eq!(value, "192.0.2.60");
 
         // is case insensitive
-        let headers = header_map(&[(FORWARDED, "HOST=192.0.2.60;proto=http;by=203.0.113.43")]);
+        let headers = header_map(&[(FORWARDED, "host=192.0.2.60;proto=http;by=203.0.113.43")]);
         let value = parse_forwarded(&headers).unwrap();
         assert_eq!(value, "192.0.2.60");
 
--- a/axum-extra/src/extract/host.rs
+++ b/axum-extra/src/extract/host.rs
@@ -139,7 +139,7 @@ mod tests {
         assert_eq!(value, "192.0.2.60");
 
         // is case insensitive
-        let headers = header_map(&[(FORWARDED, "host=192.0.2.60;proto=http;by=203.0.113.43")]);
+        let headers = header_map(&[(FORWARDED, "HOST=192.0.2.60;proto=http;by=203.0.113.43")]);
         let value = parse_forwarded(&headers).unwrap();
         assert_eq!(value, "192.0.2.60");
 
--- a/axum-extra/src/extract/mod.rs
+++ b/axum-extra/src/extract/mod.rs
@@ -56,3 +56,10 @@ pub use crate::json_lines::JsonLines;
 #[cfg(feature = "typed-header")]
 #[doc(no_inline)]
 pub use crate::typed_header::TypedHeader;
+
+#[cfg(feature = "scheme")]
+pub mod scheme;
+
+#[cfg(feature = "scheme")]
+#[doc(no_inline)]
+pub use self::scheme::{SchemeMissing, Scheme};
--- a/axum-extra/src/extract/mod.rs
+++ b/axum-extra/src/extract/mod.rs
@@ -21,6 +21,9 @@ mod query;
 #[cfg(feature = "multipart")]
 pub mod multipart;
 
+#[cfg(feature = "scheme")]
+mod scheme;
+
 pub use self::{
     cached::Cached, host::Host, optional_path::OptionalPath, with_rejection::WithRejection,
 };
@@ -43,6 +46,10 @@ pub use self::query::{OptionalQuery, OptionalQueryRejection, Query, QueryRejecti
 #[cfg(feature = "multipart")]
 pub use self::multipart::Multipart;
 
+#[cfg(feature = "scheme")]
+#[doc(no_inline)]
+pub use self::scheme::{Scheme, SchemeMissing};
+
 #[cfg(feature = "json-deserializer")]
 pub use self::json_deserializer::{
     JsonDataError, JsonDeserializer, JsonDeserializerRejection, JsonSyntaxError,
@@ -56,10 +63,3 @@ pub use crate::json_lines::JsonLines;
 #[cfg(feature = "typed-header")]
 #[doc(no_inline)]
 pub use crate::typed_header::TypedHeader;
-
-#[cfg(feature = "scheme")]
-pub mod scheme;
-
-#[cfg(feature = "scheme")]
-#[doc(no_inline)]
-pub use self::scheme::{Scheme, SchemeMissing};
--- a/axum-extra/src/extract/mod.rs
+++ b/axum-extra/src/extract/mod.rs
@@ -62,4 +62,4 @@ pub mod scheme;
 
 #[cfg(feature = "scheme")]
 #[doc(no_inline)]
-pub use self::scheme::{SchemeMissing, Scheme};
+pub use self::scheme::{Scheme, SchemeMissing};
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -33,6 +33,7 @@ json-lines = [
 ]
 multipart = ["dep:multer", "dep:fastrand"]
 protobuf = ["dep:prost"]
+scheme = []
 query = ["dep:serde_html_form"]
 tracing = ["axum-core/tracing", "axum/tracing"]
 typed-header = ["dep:headers"]
--- a/axum-extra/src/extract/scheme.rs
+++ b/axum-extra/src/extract/scheme.rs
@@ -1,0 +1,152 @@
+//! Extractor that parses the scheme of a request.
+//! See [`Scheme`] for more details.
+
+use axum::{
+    extract::FromRequestParts,
+    response::{IntoResponse, Response},
+};
+use http::{
+    header::{HeaderMap, FORWARDED},
+    request::Parts,
+};
+const X_FORWARDED_PROTO_HEADER_KEY: &str = "X-Forwarded-Proto";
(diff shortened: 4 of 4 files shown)


bengsparks:
The test for extracting the scheme from the URI is missing as I haven't groked how to pass a URI with a scheme to the test server.

bengsparks:
@davidpdrsn Can this be merged or do I need to add more tests?

Wiezzel:
May I suggest additionally supporting `X-Forwarded-Scheme` header? (It's mentioned in a comment, but only `X-Forwarded-Proto` is supported).

Wiezzel:
It also seems to me that the extraction from `parts` is not really working. I've added a debug print statement there and the `parts.uri` seems to contain **only URI path**.

bengsparks:
Thanks for your comments @Wiezzel.
I will follow-up sometime during the weekend.

Wiezzel:
@bengsparks Please check this minimal example:

`Cargo.toml`
```toml
[package]
name = "scheme-extractor"
version = "0.1.0"
edition = "2021"

[dependencies]
axum = "0.7.3"
tokio = { version = "1.36.0", features = ["full"] }

[patch.crates-io]
axum = { git = "https://github.com/bengsparks/axum.git" }
```

`main.rs`
```rust
async fn test(scheme: axum::extract::Scheme) -> String {
    scheme.0
}

#[tokio::main]
async fn main() {
    let app = axum::Router::new()
        .route("/", axum::routing::get(test));

    let listener = tokio::net::TcpListener::bind("0.0.0.0:8000").await.unwrap();
}
```

Then run:
```
$ cargo run &
$ curl -i  http://localhost:8000
HTTP/1.1 400 Bad Request
content-length: 26
date: Mon, 04 Mar 2024 12:52:01 GMT

No scheme found in request
```

jplatte:
I think we've made a point to *not* handle any conventional, or even standard headers for "original" request data from proxies in `ConnectInfo`, because it can be spoofed at least in some cases. I think the same reasoning applies here.

bengsparks:
@jplatte Could you concretise what this means for the PR? Does this mean `X-Forwarded-Proto` should be ignored? In this case, the `Host` extractor would also have to be updated.

Wiezzel:
I could suggest a fallback to HTTP in case the scheme cannot be extract neither from headers, nor from path. This sounds like a reasonable default to me, but it could be optional in case some application needs stricter check.

jplatte:
> @jplatte Could you concretise what this means for the PR? Does this mean `X-Forwarded-Proto` should be ignored? In this case, the `Host` extractor would also have to be updated, as it reads from `X-Forwarded-Host`.

Yes, that is what I meant. I didn't know we even had a `Host` extractor 😅
I'll try to find some time to figure out the background on how that extractor was added, and also to find those discussions I remember about `ConnectInfo`.

bengsparks:
@jplatte I will put this PR on pause then until the matter is resolved; Having a `Host` extractor was half the reason I considered writing this PR to go along with it. There is also the question of what to do if the scheme cannot be found in the URI, despite being provided, which is the case in @Wiezzel's example.

sclu1034:
In my tests, the `Request` never contained a scheme value. Regardless of whether I used the `Request` or `OriginalUri` extractors.

> I think we've made a point to _not_ handle any conventional, or even standard headers for "original" request data from proxies in `ConnectInfo`, because it can be spoofed at least in some cases. I think the same reasoning applies here.

If you wanted to walk that path to the end, you'd have to remove the `Host` extractor entirely, and prevent stuff like `req.uri().host()`, as that's from the `Host` header. The protocol would at least be possible, as the server should know whether it currently speaks with or without TSL.

But as a user, the extractors came across to me as an implementation of common practices. Which is why I would expect them to check those extra headers. Because if they didn't, they'd boil down to a one-line like `req.uri().host()`, and I don't need an extractor for that.

I would not expect them to be subject to security-based mitigations.
Instead, the way I'd expect a security mitigation like this would be either some kind of middleware that I can deliberately add to have "spoofable" headers be stripped, or the reverse, where they are stripped/ignored by default unless some kind of .reverse_proxy(true)` is set on a `Router` or for _axum_ as a whole.

bengsparks:
To reiterate: 
The scheme is resolved through the following, in order:
- `Forwarded` header
- `X-Forwarded-Proto` header
- Request URI

The final bullet point requires that `axum` is served with the `http2` feature flag.
I've tested this manually with `cURL` and `xh`, but the `TestClient` that is used in the tests doesn't have the API to force HTTP/2 to be used. 
How can such a test be included as part of this PR?

sclu1034:
> requires that `axum` is served with the `http2` feature flag

Why is that? Regardless of the HTTP version, something in the stack has to initiate the TLS connection, and therefore knows what scheme is in use.
If it's a limitation in libraries not exposing this, it might be worthwhile to create an upstream feature request for that.

bengsparks:
Ah, I think I've mixed something up. 
I've managed to send requests through to an `axum` instance served with TLS (using `axum_server`), without enabling any feature flags.

bengsparks:
I accidentally closed this Pull Request when updating my fork of axum.
However, I'd like further instruction / advice on what should be done to make this mergeable or if this is even wanted.
If not, I'll simply reclose the PR, no hard feelings :).

mladedav:
It seems you've accidentally pushed the already merged branch removing the `allow`s instead of the scheme extractor one.

We've moved the `Host` extractor to axum-extra and the last consensus was to put this there too. I think (but can't really check) that your changes were already adding it there so I don't know if there's anything blocking it.

Sorry for the delays.

bengsparks:
@mladedav thanks for the quick reply. I've restored the state of the branch

axum-extra/src/extract/host.rs
```diff
@@ -139,7 +139,7 @@ mod tests {
         assert_eq!(value, "192.0.2.60");
 
         // is case insensitive
-        let headers = header_map(&[(FORWARDED, "host=192.0.2.60;proto=http;by=203.0.113.43")]);
+        let headers = header_map(&[(FORWARDED, "HOST=192.0.2.60;proto=http;by=203.0.113.43")]);
```
yanns, line 142: This change does not seem part of this PR right?

bengsparks: Correct, it's more of a bonus, as before the case-insensitivity was not being checked in the Host extractor.
I can still remove it, if you like.

yanns: That's nice! Maybe you could extract this as a separated PR?

bengsparks: It could be done as a follow-up PR, yes

bengsparks:
The `Host` extractor, since opening of this PR, has been moved into `axum-extra`, and to my understanding, it was precisely because of this issue.
I have also since adjusted this PR to match this.

Additionally, adding a separate method to indicate "danger" goes against the spirit of extractors, whose intended usage is documented as "reading via destructuring" i.e. `Path(user_id): Path<u32>`, `Json(payload): Json<Value>`, `Extension(state): Extension<State>`, etc.

yanns:
That's a good point. I'd like that we make users aware of the risks of using those extractors somehow, but I don't have better idea for now. Having them in axum-extra does not feel good enough. It's still very easy to import them and use them without being conscious of the possible security implications.

I'll wait for the point of view of @jplatte here.

jplatte:
@yanns yeah I agree, it would be nice to do more than just exclude these from the main crate. But also I think it wouldn't be great to hold up this PR again because of that.

Let's merge it but make an issue to revisit both `Host` and `Scheme` before v0.8.0?

axum-extra/src/extract/mod.rs
```diff
@@ -56,3 +56,10 @@ pub use crate::json_lines::JsonLines;
 #[cfg(feature = "typed-header")]
 #[doc(no_inline)]
 pub use crate::typed_header::TypedHeader;
+
+#[cfg(feature = "scheme")]
+pub mod scheme;
+
+#[cfg(feature = "scheme")]
+#[doc(no_inline)]
+pub use self::scheme::{Scheme, SchemeMissing};
```
jplatte, line 65: Could you please move these declarations up to the groups of module declarations and `self` re-exports before the `json_lines` re-export? Also, the module should not be public.

yanns:
> @yanns yeah I agree, it would be nice to do more than just exclude these from the main crate. But also I think it wouldn't be great to hold up this PR again because of that.
> 
> Let's merge it but make an issue to revisit both `Host` and `Scheme` before v0.8.0?

Agree with this plan!

axum-extra/src/extract/mod.rs
bengsparks: PR is updated; I hope I've done this correctly :)

jplatte:
Oh, CI failing is my fault.. I'll look into it 🫠

axum-extra/src/extract/host.rs
jplatte:
```suggestion
        let headers = header_map(&[(FORWARDED, "host=192.0.2.60;proto=http;by=203.0.113.43")]);
```

yanns:
I've added an issue about spoofable extractors: https://github.com/tokio-rs/axum/issues/2998

jplatte (approved):

PR #2509, opened 2024-01-11
Re-export `BodyDataStream` from axum

davidpdrsn:
Happened to notice that `BodyDataStream` wasn't re-exported by axum. So you had to depend on axum-core to name that type.


jplatte (approved):
Looking at this type in the docs, I noticed that the method that produces it (`Body::into_data_stream`) has broken intra-doc links to the `Stream` trait. Can be fixed in a separate PR of course.

PR #2517, opened 2024-01-15
Improve documentation of `axum_extra::extract::Query`

AlphaKeks:
This commit documents a potential footgun with using `Option<Vec<T>>` in query
parameters. It also refers to `axum_extra` in the `axum::extract::Query` docs
for handling multiple values in query parameters.

## Motivation

Some people might be inclined to use `Option<Vec<T>>` for an optional list of
query parameters, because they are used to using `Option<T>` for other, non-list
types. This however has a weird quirk to it, that it handles 0, 2, or more
values correctly, but not 1. `Option<Vec<T>>` is also kind of redundant, and an
empty `Vec<T>` with `#[serde(default)]` should be preferred, as mentioned in the
documentation of `serde_html_form`.

## Solution

Add extra notes and examples about parsing multiple values for query parameters
to `axum_extra::extract::Query` and `axum::extract::Query`.

### DIFF
--- a/axum-extra/src/extract/query.rs
+++ b/axum-extra/src/extract/query.rs
@@ -55,7 +55,7 @@ use std::fmt;
 /// While `Option<T>` will handle empty parameters (e.g. `param=`), beware when using this with a
 /// `Vec<T>`. If your list is optional, use `Vec<T>` in combination with `#[serde(default)]`
 /// instead of `Option<Vec<T>>`. `Option<Vec<T>>` will handle 0, 2, or more arguments, but not one
-/// argument. For details, see [the documentation of `serde_html_form`].
+/// argument.
 ///
 /// # Example
 ///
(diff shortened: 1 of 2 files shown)


axum-extra/src/extract/query.rs
```diff
@@ -51,6 +51,35 @@ use std::fmt;
 /// example.
 ///
 /// [example]: https://github.com/tokio-rs/axum/blob/main/examples/query-params-with-empty-strings/src/main.rs
+///
+/// While `Option<T>` will handle empty parameters (e.g. `param=`), beware when using this with a
+/// `Vec<T>`. If your list is optional, use `Vec<T>` in combination with `#[serde(default)]`
+/// instead of `Option<Vec<T>>`. `Option<Vec<T>>` will handle 0, 2, or more arguments, but not one
+/// argument. For details, see [the documentation of `serde_html_form`].
```
jplatte, line 58: What extra details are you talking about here? 😅

jplatte (approved):
Thanks! One oddity, otherwise looks good to me.

axum-extra/src/extract/query.rs
AlphaKeks: The crate level documentation, which originally mentioned this detail. If you find it unnecessary though I can remove it again :slightly_smiling_face:

jplatte: Yeah it does mention this, but doesn't go into more detail than what you wrote here. I think it's better to remove it to avoid confusion.

PR #2518, opened 2024-01-15
Grammar: Fix "it's" vs "its" in several places

DominoTree:
Fix misuse of "its" and "it's" in various locations

## Motivation

While reading docs on docs.rs as well as browsing the source tree, I ran into a few uses of "its" and "it's" that were mixed up, and it bugged me.

## Solution

Searched the source tree for "its" and "it's" and fixed their uses where necessary


jplatte (approved):
Thanks!

PR #2523, opened 2024-01-16
add minijinja templating example

Deffendor:
A small working sample on how to use MiniJinja for templating html pages with Axum.

It creates three routes, each with an own template and all of them use the same layout as base.


jplatte:
@davidpdrsn I know you're a bit skeptical about new examples that are similar to others, but I think dynamic templating like MiniJinja / Tera is different enough from static templating like askama that another example is probably worthwhile. Also MiniJinja might be the most stable Rust templating system there is - it's post-1.0 and I'm a big fan of it API wise :)

examples/templates-minijinja/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+    env.add_template("about", include_str!("../templates/about.jinja"))
+        .unwrap();
+    let env = Arc::new(env);
+
+    // define routes
+    let app = Router::new()
+        .route("/", get(handler_home))
+        .route("/content", get(handler_content))
+        .route("/about", get(handler_about))
+        .layer(Extension(env)); // add state as extension
```
jplatte, line 33: I think it would be nicer to use `with_state` and the `State` extractor instead of `Extension` for this.

examples/templates-minijinja/templates/about.jinja
```diff
@@ -0,0 +1,6 @@
+{% extends "layout" %}
+{% block title %}{{ super() }} | {{ title }} {% endblock %}
+{% block body %}
+<h1>{{ title }}</h1>
+<p>{{ about_text }}</p>
+{% endblock %}
```
jplatte, line 6: Missing final newline in almost all of the jinja files.

jplatte (approved):
Looks good overall, just a few small nitpicks. In any case this also needs David's feedback.

Deffendor:
Thanks for the feedback. I applied the changes but wasn't sure if passing the lifetime of Environment to AppState is the right way to do it. Let me know if that needs to be changed.

If @davidpdrsn is skeptical about new examples and does not want this one to be here I can also move it to the MiniJinja repo. Let me know in that case.

examples/templates-minijinja/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+//! Exposes three pages all sharing the same layout with a minimal nav menu.
+
+use axum::extract::State;
+use axum::http::StatusCode;
+use axum::{response::Html, routing::get, Router};
+use minijinja::{context, Environment};
+use std::sync::Arc;
+
+struct AppState<'a> {
+    env: Environment<'a>,
```
jplatte, line 16:
```suggestion
struct AppState {
    env: Environment<'static>,
```

Since non-static in state is pretty uncommon (might even be completely unsupported, can't remember right now).

Deffendor:
Ok thanks, changed it to static, looks better now.

davidpdrsn (approved):
Thanks!

PR #2524, opened 2024-01-16
Fix fallback handlers on nested routers returning 404

### LINKED ISSUE #2252: Fallback handlers on nested routers returning 404
I'm trying to do an upgrade from axum-0.6.11 to axum-0.6.20 and I noticed that the behavior we were relying on in 0.6.11, where a nested handler could designate a fallback, seems to not work in 0.6.20.

I've got a test case in this repository:

https://github.com/maxburke/axum-test

With axum-0.6.11, `curl -v http://localhost:3000/nest/` returns 200 + "Hello World", but with axum-0.6.20, the same curl command returns 404 + empty response.

dalton-oliveira:
## Motivation
The issue #2252 ia about a breaking change  on `fallback` action of a nested `Router`. Previously, if the nested `Router` with a `fallback` is matched,  it should call it's `fallback` when the route is not defined. Example:

https://github.com/dalton-oliveira/axum/blob/1313119b73663ea1f6e461197195ad89248b6e83/axum/src/routing/tests/nest.rs#L387-L397
  
This breaking change was introduced once it started to use [matchit](https://github.com/ibraheemdev/matchit) for routing to the fallback handler. It uses a special [catch-all-parameter](https://github.com/ibraheemdev/matchit?tab=readme-ov-file#catch-all-parameters)  `/*__private__axum_fallback` which does't match _empty_  path which would be the _root_ of the nested router. There's no option for matching `zero-or-more` pattern. There's an open issue [Support optional last param](https://github.com/ibraheemdev/matchit/issues/25) that, once solved, I think would be sufficient for this scenario and thus a better solution too.

## Solution
This changes `Router::nest` to also nest the root  for the `fallback` `PathRouter` 

Fixes #2252

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- **fixed:** Fallback handlers on nested routers returning 404
 
 # 0.7.4 (13. January, 2024)
 
--- a/axum/src/routing/path_router.rs
+++ b/axum/src/routing/path_router.rs
@@ -180,16 +180,27 @@ where
                 .expect("no path for route id. This is a bug in axum. Please file an issue");
 
             let path = path_for_nested_route(prefix, inner_path);
-
+            let route_nested_root =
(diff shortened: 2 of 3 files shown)


dalton-oliveira:
Closing as it wasn't completely fixing the issue

Issue #2511, opened 2024-01-13
oauth example doesn't validate CSRF token


davidpdrsn:
The CSRF token generated [here](https://github.com/tokio-rs/axum/blob/main/examples/oauth/src/main.rs#L146) needs to be validated for login attemps. The oauth example currently doesn't do that.

Thank you to Nicolas Ettlin for reporting this issue

loganbnielsen:
I can try doing this if that's alright

loganbnielsen:
I can give this a go if that's alright

davidpdrsn:
@loganbnielsen thank you!

loganbnielsen:
Made PR #2534

theodorebje:
Hi! I noticed this issue appears to have been fixed by #2534 (which still has the `S-waiting-for-review` label even though it's merged). Should this issue be closed?

mladedav:
You're right, thanks for the ping.

PR #2534, opened 2024-01-25
Implement CSRF token verification for OAuth example

loganbnielsen:
Addresses #2511


examples/oauth/src/main.rs
```diff
@@ -142,19 +142,37 @@ async fn index(user: Option<User>) -> impl IntoResponse {
...
         .authorize_url(CsrfToken::new_random)
         .add_scope(Scope::new("identify".to_string()))
         .url();
 
-    // Redirect to Discord's oauth service
-    Redirect::to(auth_url.as_ref())
+    // Create session to store csrf_token
+    let mut session = Session::new();
+    session
+        .insert("csrf_token", &csrf_token)
```
iggyzuk, line 157: Since "csrf_token" is used in two placed it might be good to define a constant value. `static CSRF_TOKEN: &str = "csrf_token";` similar to COOKIE_NAME. I suppose "user" also needs one.

iggyzuk (approved):
Looks good and it works, I used it in my project.

examples/oauth/src/main.rs
```diff
@@ -142,19 +142,37 @@ async fn index(user: Option<User>) -> impl IntoResponse {
...
+
+    // Store the session in MemoryStore and retrieve the session cookie
+    let cookie = store
+        .store_session(session)
+        .await
+        .context("failed to store CSRF token session")?
+        .context("unexpected error retrieving CSRF cookie value")?;
+
+    // Attach the session cookie to the response header
+    let cookie = format!("{COOKIE_NAME}={cookie}; SameSite=Lax; Path=/");
```
emmanuelgautier, line 168: The cookie security can be improved setting `HttpOnly`, `Secure`, and `SameSite` to `Strict`.

iggyzuk: In my case setting SameSite to Strict breaks something "Application error: unexpected error getting cookie name". It happens when the login is authorized, so `csrf_token_validation_workflow` is unable to find cookie `COOKIE_NAME`. Note: I'm using Google OAuth2, not Discord.

emmanuelgautier: My bad, you are right. We can not use Strict here since the cookie is sent after a redirect. First comment updated.

loganbnielsen:
Noticed that this has been open for almost a year. Lmk if I can do anything to make this mergeable

emmanuelgautier (approved):

examples/oauth/src/main.rs
mladedav: @loganbnielsen can you add the mentioned options? I get that we don't need to make the whole thing production-ready, but if these settings are more secure let's add them to the example.

mladedav (changes requested):
Thanks for pinging us, let's finish this.

loganbnielsen:
updated

examples/oauth/src/main.rs
```diff
@@ -142,19 +142,37 @@ async fn index(user: Option<User>) -> impl IntoResponse {
...
-    // Redirect to Discord's oauth service
-    Redirect::to(auth_url.as_ref())
+    // Create session to store csrf_token
+    let mut session = Session::new();
+    session
+        .insert("csrf_token", &csrf_token)
+        .context("failed in inserting CSRF token into session")?;
+
+    // Store the session in MemoryStore and retrieve the session cookie
+    let cookie = store
```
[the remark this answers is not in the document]
mladedav: Yes it is. Sorry, I completely overlooked it somehow.

mladedav:
Thanks. I'll just try to run it tomorrow when I get to a PC and if there aren't any issues, I'll merge it.

loganbnielsen:
Sounds good, I sometimes hit a race condition when I run it where the token read attempt occurs before the write has happened so you might need to try a second time if you get that error (I think exponential back off for second would fix this but feels out of scope for a minimal example)

examples/oauth/src/main.rs
emmanuelgautier: All good for the `HttpOnly` and `Secure` flags. This comment can  be resolved.

mladedav (approved):
It works for me so I'll merge this. Thanks again for fixing the example.

PR #2535, opened 2024-01-25
Added MongoDB example

tr1sm0s1n:
Hello! ^^

## Motivation

Recently, I've been trying to create an API service in `Rust` to communicate with **MongoDB** using `axum`. Adding an example would be nice for future developers.

## Solution

- Added a MongoDB example to `examples/mongodb`.
- Used [the official MongoDB Rust Driver](https://github.com/mongodb/mongo-rust-driver) for the communication.
- Added routes to handle CRUD operations.

### DIFF
--- a/examples/mongodb/src/main.rs
+++ b/examples/mongodb/src/main.rs
@@ -7,11 +7,14 @@
 use axum::{
     extract::{Path, State},
     http::StatusCode,
-    response::IntoResponse,
     routing::{delete, get, post, put},
     Json, Router,
 };
-use mongodb::{bson::doc, Client, Collection};
+use mongodb::{
+    bson::doc,
+    results::{DeleteResult, InsertOneResult, UpdateResult},
+    Client, Collection,
+};
 use serde::{Deserialize, Serialize};
 use tower_http::trace::TraceLayer;
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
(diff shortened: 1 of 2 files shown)


examples/mongodb/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+    let result = db.delete_one(doc! { "_id": id }, None).await.unwrap();
+    println!("{:?}", result);
+
+    (StatusCode::OK, Json(result))
+}
+
+// defining Member type
+#[derive(Debug, Deserialize, Serialize)]
+struct Member {
+    _id: u32,
```
davidpdrsn, line 119: Why the `_`?

```suggestion
    id: u32,
```

examples/mongodb/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+        .with_state(collection)
+}
+
+// handler to create a new member
+async fn create_member(
+    State(db): State<Collection<Member>>,
+    Json(input): Json<Member>,
+) -> impl IntoResponse {
+    let result = db.insert_one(input, None).await.unwrap();
+    println!("{:?}", result);
```
davidpdrsn, line 71: Not sure we need all this print. Same goes for the ones below.

```suggestion
```

examples/mongodb/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+        .route("/delete/:id", delete(delete_member))
+        .layer(TraceLayer::new_for_http())
+        .with_state(collection)
+}
+
+// handler to create a new member
+async fn create_member(
+    State(db): State<Collection<Member>>,
+    Json(input): Json<Member>,
+) -> impl IntoResponse {
```
davidpdrsn, line 69: We generally don't recommend returning `impl IntoResponse`, see https://docs.rs/axum/latest/axum/response/index.html#regarding-impl-intoresponse.

So I think instead we should write the full type.

Would also be good to not `unwrap` inside the handler and instead return a `Result`.

examples/mongodb/src/main.rs
Ptrskay3: I believe this is because MongoDB uses the `_id` field as primary key. 
Maybe it could be a `#[serde(rename = "_id")] id: u32` to adhere to Rust's convention?

tr1sm0s1n: Fixed in [3cd9ded](https://github.com/tokio-rs/axum/pull/2535/commits/3cd9dede2d24a2d9ebc25636a17068afd42baf1a).

examples/mongodb/src/main.rs
tr1sm0s1n: Fixed in [3cd9ded](https://github.com/tokio-rs/axum/pull/2535/commits/3cd9dede2d24a2d9ebc25636a17068afd42baf1a).

examples/mongodb/src/main.rs
tr1sm0s1n: Fixed in [3cd9ded](https://github.com/tokio-rs/axum/pull/2535/commits/3cd9dede2d24a2d9ebc25636a17068afd42baf1a).

jplatte (approved):
Sorry for the long long delay on this. I haven't used MongoDB but David initally reviewed this, I guess there can't be much wrong with merging this.

PR #2542, opened 2024-01-26
Fix typos in testing example

rdbo:
This pull request fixes a typo in the file `examples/testing/src/main.rs`, where `ConnectInfo` is sometimes referred to as `connect into`. For example, the route name currently is `/requires-connect-into`. This pull request simply changes that.

## Motivation

I was just browsing the source and found this typo, so I decided to fix it.

## Solution

Not needed.


jplatte (approved):
Thanks!

PR #2567, opened 2024-02-05
examples: Add wildcard-route

Enselic:
I tried to find examples of wildcard routes but didn't find any with

    cd examples ; git grep 'route(".*\*'

so add such an example.


Enselic:
I did found that documentation actually, but as I'm completely new to axum I found it to be incomplete. In particular, it does not demonstrate how to access the variable path in a handler.

But you are probably right that it is better to improve the docs than to add this as an example. PR for that: https://github.com/tokio-rs/axum/pull/2569

PR #2568, opened 2024-02-05
Run cargo clippy --fix to fix CI

Enselic:
I ran this command:

    cargo +beta clippy --fix --workspace --all-targets --all-features -- -D warnings

which fixes a clippy problem that CI currently complains about: https://github.com/tokio-rs/axum/pull/2567

In the case of `NotSendSync` its unused field is important so that it stays not Send and not Sync. So allow it to be unused.


jplatte (approved):
Thanks!

PR #2569, opened 2024-02-06
route.md: Add wildcard example code

Enselic:
To developers completely new to axum, it helps a lot to give example code on to get access to the variable path in a handler.

Rendered:

image

Even if the example is not perfect, my hope is that it will be better than nothing. It certainly would have helped me at least.

But feel free to close this PR if the example is too sloppy.

### DIFF
--- a/axum/src/docs/routing/route.md
+++ b/axum/src/docs/routing/route.md
@@ -56,7 +56,22 @@ Note that `/*key` doesn't match empty segments. Thus:
 - `/*key` doesn't match `/` but does match `/a`, `/a/`, etc.
 - `/x/*key` doesn't match `/x` or `/x/` but does match `/x/a`, `/x/a/`, etc.
 
-Wildcard captures can also be extracted using [`Path`](crate::extract::Path).
+Wildcard captures can also be extracted using [`Path`](crate::extract::Path):
+
+```rust
+use axum::{
+    Router,
+    routing::get,
+    extract::Path,
+};
+
+let app = Router::new().route("/*key", get(handler));
+
+async fn handler(Path(path): Path<String>) -> String {
+    path
+}
+```
+
 Note that the leading slash is not included, i.e. for the route `/foo/*rest` and
 the path `/foo/bar/baz` the value of `rest` will be `bar/baz`.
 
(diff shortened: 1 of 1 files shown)


axum/src/docs/routing/route.md
```diff
@@ -60,6 +60,16 @@ Wildcard captures can also be extracted using [`Path`](crate::extract::Path).
...
+Example:
+
+```rust
+let app = Router::new().route("/*key", get(handler));
+
+async fn handler(path: axum::extract::Path<String>) -> String {
+    path.0
+}
+```
+
```
davidpdrsn, line 72: Lets combine it with the paragraph above

```suggestion
Wildcard captures can also be extracted using [`Path`](crate::extract::Path):

```rust
use axum::{
    Router,
    routing::get,
    extract::Path,
};

let app = Router::new().route("/*key", get(handler));

async fn handler(Path(path): Path<String>) -> String {
    path
}
```

Note that the leading slash is not included, i.e. for the route `/foo/*rest` and
the path `/foo/bar/baz` the value of `rest` will be `bar/baz`.
```

**Comment by Enselic** [comment]
Thanks for the review, I think I fixed your comment now.

**Comment by Enselic** [comment]
I'd like to keep my PR outbox clean, so let's close this for now.

**Comment by jplatte** [comment]
Hey, sorry for the long delay. Other than failing CI (which might fix itself with the next run, let's see), this looks good now!

**Review by jplatte — APPROVED** [review]

File: axum/src/docs/routing/route.md [hunk 2]
```diff
@@ -56,7 +56,22 @@ Note that `/*key` doesn't match empty segments. Thus:
...
+Wildcard captures can also be extracted using [`Path`](crate::extract::Path):
+
+```rust
+use axum::{
+    Router,
+    routing::get,
+    extract::Path,
+};
+
+let app = Router::new().route("/*key", get(handler));
```
jplatte, line 68: CI failure is actually legit, the compiler can't infer the type of `app` here (specifically the generic state type of the router).

```suggestion
let app: Router = Router::new().route("/*key", get(handler));
```

Issue #2550, opened 2024-01-29
Document deriving `FromRequestParts` limited to specific state types


Kinrany:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

edit(jplatte): Apparently it is possible to explicitly set a state type for a derived `FromRequest` / `FromRequestParts` via `#[from_request(state(Type))]`. This is not documented.

Previous issue description:

### Motivation

`FromRequest` and `FromRequestParts` are two great derive macros that follow the general pattern of trait `X` being derivable for type `T` when all the fields of `T` are already `X`.

For completeness, one would expect this to also work in the case of `FromRequestParts<S>` where `S` is a concrete type: if all the fields of `T` are `FromRequestParts<S>`, it's reasonable to expect to be able to derive `FromRequestParts<S>` for `T` itself.

### Proposal

Change the derive macro for `FromRequestParts` (and possibly `FromRequest`) to also work for types that can only implement the trait for a certain state type.

It is likely the easiest to initially support a single state type only, with an intentionally long attribute that specifies the type:

```rust
impl FromRequestParts<Foo> for Bar {...}

#[derive(FromRequestParts)]
#[axum(from_parts_single_state = Foo)]
struct Baz {
   bar: Bar,
}
```

### Alternatives

Without the macro support, the only alternative is to implement the traits manually for types that rely on state.

mladedav:
Is this what you mean or am I misunderstanding somehow?

```
use axum::async_trait;
use axum::http::request::Parts;
use axum::extract::FromRequestParts;

struct Foo;
struct Bar;

#[async_trait]
impl FromRequestParts<Foo> for Bar {
    type Rejection = ();

    async fn from_request_parts(_parts: &mut Parts, _state: &Foo) -> Result<Self, Self::Rejection> {
        todo!()
    }
}

#[derive(FromRequestParts)]
#[from_request(state(Foo))]
struct Baz {
   bar: Bar,
}

fn main() {}
```

jplatte:
It is not mentioned in the documentation though.

jplatte:
Docs were unnoticed in #2581.

Kinrany:
Thank you!

PR #2581, opened 2024-02-08
Document usage of concrete state in `FromRequest` macro (#2550)

mladedav:
## Motivation

The feature was not documented, #2550 

## Solution

Added a documentation with example to the macro. `FromRequestParts` links back to `FromRequest` for information about its options so nothing is needed there.

### DIFF
--- a/axum-macros/src/lib.rs
+++ b/axum-macros/src/lib.rs
@@ -233,6 +233,55 @@ use from_request::Trait::{FromRequest, FromRequestParts};
 /// }
 /// ```
 ///
+/// ## Concrete state
+///
+/// If the extraction can be done only for a concrete state, that type can be specified with
+/// `#[from_request(state(YourState))]`:
+///
+/// ```
+/// use axum::extract::{FromRequest, FromRequestParts};
+///
+/// #[derive(Clone)]
+/// struct CustomState;
+///
+/// struct MyInnerType;
+///
+/// #[axum::async_trait]
+/// impl FromRequestParts<CustomState> for MyInnerType {
+///     // ...
+///     # type Rejection = ();
+///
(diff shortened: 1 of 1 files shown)


jplatte (commented):
I think for other macros, we also infer a concrete state type from `State<Foo>` usage. Have you checked whether a `State<Foo>` field works to infer a concrete state type without `#[from_request(state)]`? If yes, that should also be documented IMO.

jplatte:
Thanks for working on this!

jplatte:
CI failure is unrelated to the change (caused by new Rust release). Best fix would be a separate PR to apply the clippy suggestion. I can do it tomorrow if nobody beats me to it.

mladedav:
Yes, `State<Foo>` works out of the box. I initially wanted to use that in the example but to my surprise it worked even without the annotation so I had to do it the more complicated way. I've added it to the docs.

> I can do it tomorrow if nobody beats me to it.

Someone's already done it, it just awaits approval #2568

axum-macros/src/lib.rs
```diff
@@ -233,6 +233,55 @@ use from_request::Trait::{FromRequest, FromRequestParts};
...
+/// struct CustomState;
+///
+/// struct MyInnerType;
+///
+/// #[axum::async_trait]
+/// impl FromRequestParts<CustomState> for MyInnerType {
+///     // ...
+///     # type Rejection = ();
+///
+///     # async fn from_request_parts(
```
davidpdrsn, line 254: Otherwise we'll get an empty line in the docs

```suggestion
///     # type Rejection = ();
///     #
///     # async fn from_request_parts(
```

davidpdrsn (approved):
Thanks!

PR #2584, opened 2024-02-09
Add rejection tracing to all extractors (#2526)

### LINKED ISSUE #2526: Rejection tracing not implemented for axum_extra::extract::Query
**Description:**
Currently, the rejection tracing feature in axum (version 0.7.4) works well for all `axum::extract::*` types when using the `env_filter` feature. However, it seems that rejection tracing is not implemented for `axum_extra::extract::Query`.

**Steps to Reproduce:**
1. Setup `tracing_subscriber` to enable rejection tracing:
 ```rust
    let subscriber = tracing_subscriber::fmt()
        .with_env_filter(
            tracing_subscriber::EnvFilter::from_default_env()
                .add_directive("axum::rejection=trace".parse().unwrap()),
        )
        .finish();

    tracing::subscriber::set_global_default(subscriber)
        .expect("Failed to set up tracing");
```
2. Use `axum_extra::extract::Query` in your handler.
3. Observe that rejections from `axum_extra::extract::Query` are not traced to the terminal.

**Expected Behavior:**
Rejections from `axum_extra::extract::Query` should be traced to the terminal when using the rejection tracing feature in axum.

Thank you for your attention to this issue.

mladedav:
## Motivation

Rejection tracing was missing at some points.
Closes #2526

## Solution

This should add the tracing eveyrwhere, I searched by `into_response` implementations.

I've also changed the places where it already was to call `body_text` only once as it makes an unnecessary copy of the string.

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning].
 
 # Unreleased
 
-- None.
+- **added:** New `tracing` feature which enables logging rejections from
+  built-in extractor with the `axum::rejection=trace` target ([#2596])
 
 # 0.9.2 (13. January, 2024)
 
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -12,7 +12,7 @@ repository = "https://github.com/tokio-rs/axum"
 version = "0.9.2"
 
 [features]
-default = []
+default = ["tracing"]
 
 async-read-body = ["dep:tokio-util", "tokio-util?/io", "dep:tokio"]
 cookie = ["dep:cookie"]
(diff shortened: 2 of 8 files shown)


davidpdrsn (commented):
Thanks! Do you wanna update the changelog as well?

mladedav:
Sure. When I added it to the changelog, I realized I'm using the same macro that's used in `axum` itself, so the target is `axum::rejection` which could be potentially confusing since it originates in `axum-extra`.

Do you think it'd be better to change the macro to also (optinally?) take the target as an argument and use `axum-extra::rejection` instead or is this ok?

mladedav:
When I looked at #2527, I realized that this would not work as currently implemented because the macro uses `cfg(features = ...)` but since it is expanded in `axum-extra` which does not define the `tracing` feature at all, this would never actually log anything.

I'll fix this hopefully during today.

mladedav:
I've added `tracing` feature to `axum-extra`, same as is in `axum`. However, it's on by default in axum but I wasn't sure here so it's off.

I ultimately went this way because `tracing` wasn't a dependency so it felt wrong just adding it without it being behind a feature and once this needed feature anyway, there was no reason not to reuse the macro from `axum-core`.

jplatte:
Removing a feature even Uhr it does nothing is a breaking change. I agree that we should eventually remove it unless we want to actually make it do something, but please don't outright remove it for now.

jplatte (approved):

PR #2585, opened 2024-02-11
docs: circumvent oddities with SSE and compression

BenJeau:
## Motivation

I was working with SSEs for my personal project and wasn't able to figure out why sometimes I received some events, only if the events were big. Thought it was something with Chrome, Firefox, or Postman. Was able to view the full response with curl but not within the rest.

After trying different things, I saw https://github.com/tokio-rs/axum/discussions/2034#discussioncomment-6120929 and it fixed my issue.

## Solution

The solution is to either remove the compression layer itself or to add an exception for SSEs. I've added a note on that in the docs related to SSEs

### DIFF
--- a/axum/src/response/sse.rs
+++ b/axum/src/response/sse.rs
@@ -32,10 +32,11 @@
 //! latest events. If you're using that layer, you can do the following to not
 //! compress SSE:
 //!
-//! ```ignore
-//! CompressionLayer::new()
-//!     .compress_when(DefaultPredicate::new()
-//!     .and(NotForContentType::new("text/event-stream"))
+//! ```
+//! # use tower_http::compression::{predicate::{NotForContentType, DefaultPredicate}, Predicate, CompressionLayer};
+//! let predicate = DefaultPredicate::new().and(NotForContentType::new("text/event-stream"));
+//!
+//! let layer = CompressionLayer::new().compress_when(predicate);
 //! ```
 
 use crate::{
--- a/axum/src/response/sse.rs
+++ b/axum/src/response/sse.rs
@@ -24,6 +24,19 @@
 //! }
 //! # let _: Router = app;
 //! ```
+//!
+//! ## Compression
+//!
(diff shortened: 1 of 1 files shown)


davidpdrsn (commented):
It should be pretty straight forward to fix in tower-http. Would you be up for giving it a shot? There is an issue for it https://github.com/tower-rs/tower-http/issues/420. Should be able to update `DefaultPredicate` to include event streams.

BenJeau:
That seems reasonable as adding it to the `DefaultPredicate`, opened a PR on `tower-http`'s side https://github.com/tower-rs/tower-http/pull/465

BenJeau:
With the code merged from `tower-http`'s side, I still think there should be a mention about compression in the Axum's SSE docs - maybe not specifically about `tower-http` - in case someone uses compression but not with the `DefaultPredicate`.

@davidpdrsn what do you think? If you think otherwise, feel free to close this PR

stevenengler:
Just wanted to mention that if you still want compression on your SSE streams rather than disabling it altogether, you can write your own `futures_util::stream::Stream` that wraps the `axum::body::BodyDataStream` and compresses and flushes SSE stream events. This avoids the buffering issues with tower's compression middleware. I wrote a small example at #2728 in the hope that someone passing through might find it helpful.

paolobarbolini:
I guess since https://github.com/tower-rs/tower-http/pull/465 was merged this is not necessary anymore?

> Just wanted to mention that if you still want compression on your SSE streams rather than disabling it altogether, you can write your own `futures_util::stream::Stream` that wraps the `axum::body::BodyDataStream` and compresses and flushes SSE stream events.

This would still be cool to have, but it should be an issue in tower-http

jplatte:
I think it's probably not that important to have a note about this sort of problem in the docs now that the one well-known instance of it is fixed. Also seems like the author was not interested in updating it to explain to the underlying problem after the tower-http instance was fixed.

PR #2586, opened 2024-02-11
Fix layers being cloned for each request

### LINKED ISSUE #2582: Example in `MethodRouter::layer` documentation doesn't work
[The docs for `MethodRouter::layer`](https://docs.rs/axum/latest/axum/routing/method_routing/struct.MethodRouter.html#method.layer) contain this example:

```rust
use tower::limit::ConcurrencyLimitLayer;

async fn handler() {}

    "/",
    // All requests to `GET /` will be sent through `ConcurrencyLimitLayer`
    get(handler).layer(ConcurrencyLimitLayer::new(64)),
);
```

This compiles, but no concurrency limit is enforced. This is because each request to `GET /` is sent through a different `ConcurrencyLimit` instance, each with its own semaphore.

Full runnable example (test with `hey -n 50 http://localhost:3000/api/test`):

```rust
use std::time::Duration;
use tower::limit::ConcurrencyLimitLayer;

async fn handler() {
    eprintln!("/api/test in");
    tokio::time::sleep(Duration::from_secs(1)).await;
    eprintln!("/api/test out");
}

#[tokio::main]
async fn main() {
        "/api/test",
        get(handler).layer(ConcurrencyLimitLayer::new(5)),
    );

    // run our app with hyper, listening globally on port 3000
}
```

davidpdrsn:
I would expect https://github.com/tokio-rs/axum/issues/2582 to work since we no longer clone the whole router for each request (https://github.com/tokio-rs/axum/pull/2483).

Turns out we were missing a `.with_state(())` to eagerly apply layers like we used to do for [`.into_ make_service()`](https://github.com/tokio-rs/axum/blob/main/axum/src/routing/mod.rs#L468). This might also have a performance impact.

edit(jplatte): Closes #2582.

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -7,7 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- **fixed:** Fixed layers being cloned when calling `axum::serve` directly with
+  a `Router` or `MethodRouter` ([#2586])
+
+[#2586]: https://github.com/tokio-rs/axum/pull/2586
 
 # 0.7.4 (13. January, 2024)
 
--- a/axum/src/routing/method_routing.rs
+++ b/axum/src/routing/method_routing.rs
@@ -1239,7 +1239,7 @@ const _: () = {
         }
 
         fn call(&mut self, _req: IncomingStream<'_>) -> Self::Future {
-            std::future::ready(Ok(self.clone()))
+            std::future::ready(Ok(self.clone().with_state(())))
         }
     }
 };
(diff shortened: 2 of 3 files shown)


dayvejones:
Hello! Will this PR allow the `RateLimitLayer `(Tower) to be used?

https://github.com/tokio-rs/axum/discussions/2520

jplatte (approved):
This makes me think that there's a performance benefit to be had by having a separate `RouterService` again (where inner routes are known to not require state be cloned into them anymore).

jplatte (approved):
Okay I was uncertain because of the comments in #2634 but actually David wrote that this fixes #2582 so assume it is tested. And even if it wasn't fixing anything, we have enough tests to tell that it doesn't break anything obvious.

PR #2623, opened 2024-02-28
add example about compression

yanns:
## Motivation

When I wanted to decompress request bodies, I had some difficulties finding the right way to do it.

## Solution

Having an example could help others finding a way to use compression and decompression quickly.


dayvejones:
Thanks! This is a useful example. Can you show how much overhead is generated by working with and without compression?

yanns:
> Can you show how much overhead is generated by working with and without compression?

Do you mean showing the difference in bytes? in cpu resource? in memory?

yanns:
I could also add some tests. Other examples do not have tests in general, so I don't know if it's something we should do or not.

yanns:
> Yes, of course, this is not a necessity. Just my wondering - is this feature useful or expensive. Any tests.

I don't have any scientific tests. And the results depend a lot on the kind of data you're compressing.
I use compression from a certain threshold so that small payloads are sent without, and large payloads are compressed.

examples/compression/README.md
```diff
@@ -0,0 +1,32 @@
+# compression
+
+This example shows how to:
+- automatically decompress request bodies when necessary
+- compress response bodies based on the `accept` header.
+
+## Running
+
+```
+cargo run
```
mladedav, line 10:
```suggestion
cargo run -p example-compression
```

examples/compression/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "compression"
```
mladedav, line 2: All the examples are named `example-*` because they are part of the same workspace as the normal project. I don't think it should be named `compression` as that could confuse people into thinking this is something else than an example.

```suggestion
name = "example-compression"
```

examples/compression/src/main.rs
```diff
@@ -0,0 +1,286 @@
...
+        ServiceBuilder::new()
+            .layer(RequestDecompressionLayer::new())
+            .layer(CompressionLayer::new()),
+    )
+}
+
+async fn root(Json(value): Json<Value>) -> Json<Value> {
+    Json(value)
+}
+
```
mladedav, line 37: Can you add a visual delimiter here like `// ===== TESTS =====` so that it is clearer for the reader that anything bellow is superfluous for the example they were looking for?

mladedav (approved):
This looks good to me

examples/compression/Cargo.toml
yanns: https://github.com/tokio-rs/axum/pull/2623/commits/46a6276247e2f779cc16d98304657b0a40d8d322

examples/compression/src/main.rs
yanns: https://github.com/tokio-rs/axum/pull/2623/commits/8efe6f1b73c4c8562fd4772c02410849c8044076

examples/compression/README.md
yanns: https://github.com/tokio-rs/axum/pull/2623/commits/46a6276247e2f779cc16d98304657b0a40d8d322

mladedav (approved):

examples/compression/README.md
```diff
@@ -0,0 +1,32 @@
...
+curl -v -g 'http://localhost:3000/' \
+    -H "Content-Type: application/json" \
+    -H "Content-Encoding: gzip" \
+    --compressed \
+    --data-binary @data/products.json.gz
+```
+
+(Notice the `Content-Encoding: gzip` in the request, and `content-encoding: gzip` in the response.)
+
+## Sending non compressed requests
```
jplatte, line 25:
```suggestion
## Sending uncompressed requests
```

examples/compression/src/main.rs
```diff
@@ -0,0 +1,288 @@
...
+use tower::ServiceBuilder;
+use tower_http::{compression::CompressionLayer, decompression::RequestDecompressionLayer};
+use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
+
+#[tokio::main]
+async fn main() {
+    tracing_subscriber::registry()
+        .with(
+            tracing_subscriber::EnvFilter::try_from_default_env()
+                .unwrap_or_else(|_| "compression=trace".into()),
```
jplatte, line 12:
```suggestion
                .unwrap_or_else(|_| "example-compression=trace".into()),
```

examples/compression/src/main.rs
jplatte: Do we have a marker like that elsewhere? Seems a bit weird to me honestly.

mladedav: I don't think so, but we also don't have tests in examples so there really never was a need.

 But personally, when I skimmed through the code, I suddenly found myself between the `#[test]` decorated functions and it wasn't as easy for me to visually find where the code ends and the tests begin since.

But I don't mind if you don't like this and you'd rather just have it as it was.

yanns: I could also move the tests to a separated `test.rs`. WDYT?

jplatte: That sounds like a good idea! 👍

yanns: https://github.com/tokio-rs/axum/pull/2623/commits/179fff01852679bafc7324aad242161fb1dbe651

PR #2633, opened 2024-03-05
Move Html from axum-extra to axum

### LINKED ISSUE #2632: Move axum-extra's Html response into axum
In https://github.com/tokio-rs/axum/pull/1921, David added an `Html` response type that has a different bound on the inner type than `axum`'s `Html` response type. It was meant to be replace the latter for axum 0.7.0 originally, but was forgotten for that release.

We should move it with the next breaking-change release.

creativcoder:
## Motivation

As per #2632 this PR moves `Html` type and impl from axum-extra to axum crate.

## Solution

Closes #2632


axum/src/response/mod.rs
```diff
@@ -40,15 +37,15 @@ pub struct Html<T>(pub T);
...
-    T: Into<Body>,
+    T: IntoResponse,
 {
     fn into_response(self) -> Response {
         (
             [(
-                header::CONTENT_TYPE,
-                HeaderValue::from_static(mime::TEXT_HTML_UTF_8.as_ref()),
+                http::header::CONTENT_TYPE,
+                http::HeaderValue::from_static(mime::TEXT_HTML_UTF_8.as_ref()),
```
jplatte, line 46: Why add `http::` when those symbols are already imported?

jplatte:
```suggestion
                header::CONTENT_TYPE,
                HeaderValue::from_static(mime::TEXT_HTML_UTF_8.as_ref()),
```

axum/src/response/mod.rs
```diff
@@ -1,8 +1,5 @@
 #![doc = include_str!("../docs/response.md")]
 
-use axum_core::body::Body;
-use http::{header, HeaderValue};
-
 mod redirect;
```
jplatte, line 3:
```suggestion
use http::{header, HeaderValue};

mod redirect;
```

jplatte (approved):

Issue #2637, opened 2024-03-10
Update serde_html_form to 0.2.4 in axum extra


LMH01:
- [X] I have looked for existing issues (including closed) about this

### Motivation

I'm currently writing an api endpoint where I require that `Option<Vec<T>>` can properly be parsed in a queries. Currently there is an issue that it can not be correctly parsed.
The fix is already implemented in `serde_html_form 0.2.4`: https://github.com/jplatte/serde_html_form/issues/11

### Proposal

We should update the version of `serde_html_form` to `2.4.0`.

### Alternatives

-

jplatte:
This is not necessary: https://github.com/tokio-rs/axum/pull/2638#issuecomment-1987227198

PR #2638, opened 2024-03-10
axum-extra: update serde_html_form to 0.2.4 in

LMH01:
## Motivation

A bug fix was introduced in this version which I require in my project.

See #2637.

## Solution

Update `serde_html_form` to `0.2.4` to include this fix: https://github.com/jplatte/serde_html_form/issues/11

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -59,7 +59,7 @@ headers = { version = "0.4.0", optional = true }
 multer = { version = "3.0.0", optional = true }
 percent-encoding = { version = "2.1", optional = true }
 prost = { version = "0.12", optional = true }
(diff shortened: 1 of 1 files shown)


jplatte:
The existing dependency specification already means `>=0.2.0 <0.3.0`. There is no need to change `axum-extra` to get the new version of `serde_html_form`. If you have a project where `Cargo.lock` locks in older version of it, run `cargo update -p serde_html_form` to get the latest version in the range (or `cargo update` to update all dependencies to the latest semver-compatible versions).

jplatte:
Cargi documentation for dependency specifications: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

LMH01:
Thanks for your comment and sorry to have bothered you.

PR #2643, opened 2024-03-12
Make axum generic over it's Listener type

StratusFearMe21:
## Motivation
I'd like making `axum` applications powered by Unix Sockets to be easier than having to make your own event loop from scratch. This PR allows you to put a `UnixListener` in `axum::serve`, or any other type that you implement `Accept` on.

## Solution
This PR adds the `Accept` trait, which abstracts over the `accept()` function of the Listener type. This PR also adds the `LocalAddr` trait, which get's the associated Listener's Socket type's `SocketAddr` by calling the `local_addr()` function on the type. From there, changes were made all around the codebase to make sure that any type implementing `Accept` could be put into `axum::serve`

### DIFF
--- a/axum-core/src/lib.rs
+++ b/axum-core/src/lib.rs
@@ -1,4 +1,3 @@
-#![cfg_attr(nightly_error_messages, feature(diagnostic_namespace))]
 //! Core types and traits for [`axum`].
 //!
 //! Libraries authors that want to provide [`FromRequest`] or [`IntoResponse`] implementations
--- a/axum-core/src/lib.rs
+++ b/axum-core/src/lib.rs
@@ -1,3 +1,4 @@
+#![cfg_attr(nightly_error_messages, feature(diagnostic_namespace))]
 //! Core types and traits for [`axum`].
 //!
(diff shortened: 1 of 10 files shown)


StratusFearMe21:
The idea is that the new `Accept` trait that this PR introduces is as flexible as possible, so as of right now, it uses `hyper::rt::Read + hyper::rt::Write` as trait bounds. Im totally open to suggestions though about how we can replace these trait bounds with something non-hyper

jplatte:
Did you see the existing issue at #2474 / PR at #2479?

StratusFearMe21:
Huh, somehow I didn't see those

jplatte:
That's unfortunate. I recommend next time you open an issue to discuss any non-trivial changes before making a PR. As for this change, is it okay if I close this as duplicate? Maybe your work on this could still be useful if it leads to any useful feedback on the other PR, if you want to check for any noteworthy differences and leave a review there that would be appreciated.

StratusFearMe21:
Yeah, that's fine

Duplicate of #2474  / #2479

PR #2650, opened 2024-03-14
Fix nightly tests

mladedav:
## Motivation

The `diagnostic_namespace` was stabilized in nightly 1.78 and gives warnings. Some trybuild tests were broken.

## Solution

The `diagnostic_namespace` feature was removed and tests were updated automatically.


jplatte (approved):
Thanks!

jplatte:
You'll have to bump the nightly version in the GitHub actions workflow file.

mladedav:
Fixed.

I thought it pulled latest nightly and that's why I saw breakage in the other PR, turns out there were just actual errors and when I tried to run the tests locally I must have ran them from the parent directory so the rust-toolchain file was ignored.

tisonkun:
Now with an earlier nightly toolchain I got:

```
error[E0658]: `#[diagnostic]` attribute name space is experimental
   --> /Users/tison/.cargo/registry/src/index.crates.io-6f17d22bba15001f/axum-0.7.5/src/handler/mod.rs:130:5
    |
130 |     diagnostic::on_unimplemented(
    |     ^^^^^^^^^^
    |
    = note: see issue #111996 <https://github.com/rust-lang/rust/issues/111996> for more information
    = help: add `#![feature(diagnostic_namespace)]` to the crate attributes to enable
```

```
[toolchain]
channel = "nightly-2023-12-19"
```

Is it possible to switch the feature gate by different nightly toolchain? Or how can I work around this compile error.

tisonkun:
I use `axum = { version = "=0.7.4" }` to pin on an earlier axum version for now.

jplatte:
> Is it possible to switch the feature gate by different nightly toolchain? Or how can I work around this compile error.

I think it's possible but I'd rather not. Instead, we can have an explicit feature to turn that attribute on. I'm thjnking something like `msrv1.80` (if that's the right version, with a `-` instead of `.` if necessary).

For now, you can pin the version. Also I would recommend checking a lockfile into git when you are using nightly.

Issue #2521, opened 2024-01-16
Set tcp nodelay on sockets


Clueliss:
- [x] I have looked for existing issues (including closed) about this

### Motivation
Currently `axum` does not set tcp-nodelay on accepted connections.
As demonstrated in this issue: https://github.com/hyperium/hyper/issues/3269 this default behaviour can heavily degrade performance in some use cases.

### Proposal
Add some way to set `TCP_NODELAY` on the accepted connections (via [`TcpStream::set_nodelay`](https://docs.rs/tokio/latest/tokio/net/struct.TcpStream.html#method.set_nodelay)).

### Variants
1. Make it the default behaviour for `axum::serve` by just setting it on each accepted connection
2. Add a function to `axum::serve::Serve<>` (like `graceful_shutdown`) of which the usage could look something like this:
```rust
axum::serve(...).set_nodelay(true).await
```

### Some Notes/Questions
As far as I understood you want to keep this part of `axum` relatively simple and prefer to refer to `hyper-util` for more complex tasks. However, this would be a relatively small change (I can create a PR if you want) so it doesn't add much code complexity and can give huge performance boosts.

So I guess questions:
1. Do you want this at all?
2. If you do: What would you like for an interface?

nylonicious:
#2479 is going to let you set settings for the accepted connections before handing them over. However it's a breaking change so unfortunately you will need to wait for `0.8`.

x1957:
I've also encountered this issue, it's very important！thanks

jplatte:
A `tcp_nodelay` method on `Serve` would be appreciated. Then for 0.8 we can see if it makes sense to remove again because of #2479.

PR #2653, opened 2024-03-15
Feature: `Serve::tcp_nodelay`

Clueliss:
Fixes: #2521

## Motivation
See #2521 and linked issue in hyper

## Solution
Adds the `tcp_nodelay` method to `Serve` (and `WithGracefulShutdown`) to enable the user to instruct the server to set `TCP_NODELAY` on incoming connections.

Regarding tests: I wasn't really sure what you would like to have there (since `with_graceful_shutdown` also doesn't seem to have tests or I didn't find them, not sure), so pointers are appreciated.

### DIFF
--- a/axum/src/serve.rs
+++ b/axum/src/serve.rs
@@ -102,6 +102,7 @@ where
     Serve {
         tcp_listener,
         make_service,
+        tcp_nodelay: None,
         _marker: PhantomData,
     }
 }
@@ -112,6 +113,7 @@ where
 pub struct Serve<M, S> {
     tcp_listener: TcpListener,
     make_service: M,
+    tcp_nodelay: Option<bool>,
     _marker: PhantomData<S>,
 }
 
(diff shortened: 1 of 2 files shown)


mladedav:
Hi, this looks good to me. Can you just add an entry to the changelog please?

axum/src/serve.rs
```diff
@@ -240,9 +278,42 @@ pub struct WithGracefulShutdown<M, S, F> {
...
     make_service: M,
     signal: F,
+    tcp_nodelay: Option<bool>,
     _marker: PhantomData<S>,
 }
 
+impl<M, S, F> WithGracefulShutdown<M, S, F> {
+    /// Instructs the server to set the value of the `TCP_NODELAY` option on every accepted connection.
+    ///
+    /// See also [`TcpStream::set_nodelay`]
```
jplatte, line 288:
```suggestion
    /// See also [`TcpStream::set_nodelay`].
```

jplatte (commented):
Is `TCP_NODELAY` ever on by default, such that `.tcp_nodelay(false)` would make sense to have? I was originally thinking that the method would be parameter-less.

Clueliss:
> Is `TCP_NODELAY` ever on by default, such that `.tcp_nodelay(false)` would make sense to have? I was originally thinking that the method would be parameter-less.

The idea was that there is maybe some configuration or OS where it could be enabled by default that I'm just not aware of. Just wanted to make sure that we don't mess with that default value and users can explicitly disable it.

Again, I personally couldn't give you an example of such an OS or config.
I just had https://github.com/hyperium/hyper/issues/3269#issuecomment-1887653439 in mind where it was mentioned that MacOS probably has a slightly different implementation.

("[..] On Mac OS there was no difference in performance between `/stream` and `/regular` and this probably has to do with a slight difference in the implementation details or settings tuning of Nagle's algorithm across the two OSes. [..]" - @huntharo)

The comment doesn't imply that its not enabled tho.
I can make it parameter-less if you prefer

jplatte:
Hey, I'll try to get to this again soon. Your reasoning for having the bool parameter makes sense, let's keep that.

jplatte (approved):
Looks good, could you rebase on top of / merge main? Should fix CI and merge conflicts.

PR #2654, opened 2024-03-16
Add multipart/form-data response builders to axum-extra

### LINKED ISSUE #2624: Add a native way to create and respond to requests with multipart forms

### Motivation

Currently, as far as I can tell, Axum does not provide a native way to respond to requests with multipart forms. 

An endpoint might return a multipart form for a variety of reasons, but the most common being the need to include multiple pieces of data (a json, a file, plaintext) in a single request.

For example, I would like to send a json and a file in a single response.

### Proposal
I am not comfortable with the Axum codebase, so I am unsure of the most idiomatic ways do this, but there are a few options that may work:

Implement `IntoResponse` for the `Multipart` extractor, and add methods like `add_field()` to the existing impls for multipart.

It may also make more sense to create a separate MultipartBuilder and implement `IntoResponse` for that type, and add builder methods, [Reqwest's implementation](https://github.com/tokio-rs/axum/blob/main/examples/reqwest-response/src/main.rs) could be used as a reference.

I don't think there are any significant drawbacks, it will marginally increase the complexity of exposed formdata APIs, but I believe the benefits outweigh the drawbacks.

### Alternatives
As far as I can tell, alternatives involve:
- Having users wrap `reqwest`'s implementation and manually interop
  - This adds a level of complexity, for no real advantages from an interface perspective.
- Having users entirely create multipart form builders from scratch.

I am willing to attempt to implement this, however I am not confident in my abilities to do so.

zleyyij:
Closes #2624 (mine)
This pull request provides builders that enable `multipart/form-data` forms to be returned from handlers. 

## Motivation
`multipart/form-data` forms provide a way to include multiple types of data in a single response as key value pairs. You could include an image and a JSON, or any other combination of data. 

As an example, this could be used for a pastebin, so that a single request to an axum backend could return metadata about a paste in a json, and the paste itself. The paste wouldn't need to conform to UTF-8, and could be sent separately from the json.

## Solution
I added a `MultipartForm` struct that implements `IntoResponse`, and implementers can add parts to that form by defining `Part`s. I did not implement `IntoResponse` on the `Multipart` extractor, as advised in the original issue.

One dependency was added if the `multipart` feature is enabled. `fastrand` is used to generate boundaries, mirroring Reqwest's implementation.

Thank you for your time :)


zleyyij:
I may need some help fixing those CI errors, I am unsure how to address them

zleyyij:
Ok, I managed to fix all of those CI errors, thank you for your patience

axum-extra/src/multipart_builder.rs
```diff
@@ -0,0 +1,337 @@
...
+#[derive(Debug)]
+pub enum TransferEncoding {
+    /// If not specified, encoding defaults to UTF-8
+    Default,
+    /// If transferring raw binary data that is not guaranteed to be valid UTF-8.
+    Binary,
+}
+
+/// Create multipart forms to be used in API responses.
+/// This struct implements [IntoResponse], and so it can be returned from a handler like normal.
```
mladedav, line 17:
```suggestion
/// This struct implements [IntoResponse], and so it can be returned from a handler.
```

axum-extra/src/multipart_builder.rs
```diff
@@ -0,0 +1,337 @@
...
+    /// ```rust
+    /// use axum_extra::multipart_builder::{MultipartForm, Part};
+    ///
+    /// let mut form = MultipartForm::new();
+    /// form
+    ///     .part(Part::text("foo", "abc"))
+    ///     .part(Part::text("other_field_name", "def"))
+    ///     .part(Part::file("file", "file.txt", vec![0x68, 0x68, 0x20, 0x6d, 0x6f, 0x6d]));
+    /// ```
+    pub fn part(&mut self, part: Part) -> &mut Self {
```
mladedav, line 56: I think this should be named so it's clearer what it does, e.g. `add_part` or `push_part`.

Personally, I'm not also sure this API is that useful,wouldn't it be better to just have something like `parts_mut` which would return a mutable reference to the parts vector and then the user can just push there directly, but also remove stuff, clear it, or any other operation they might need to take.

On the other hand they can just build the vector beforehand and then call `from_parts` just before returning the response.

axum-extra/src/multipart_builder.rs
```diff
@@ -0,0 +1,337 @@
...
+    ///
+    /// ```rust
+    /// use axum_extra::multipart_builder::{MultipartForm, Part};
+    ///
+    /// // create a form with a single part that has a field with a name of "foo",
+    /// // and a value of "abc"
+    /// let parts: Vec<Part> = vec![Part::text("foo", "abc")];
+    /// let form = MultipartForm::with_parts(parts);
+    /// ```
+    pub fn text(name: &str, contents: &str) -> Self {
```
mladedav, line 136: Generally, it makes sense to require `String` if that's what you'll need.

Currently, if someone had a large string content, instead of moving it into this constructor (assuming the caller doesn't need it anymore), they would pass only slice here and this method would clone it for no reason.

axum-extra/src/multipart_builder.rs
```diff
@@ -0,0 +1,337 @@
...
+        Self {
+            name: name.to_owned(),
+            filename: None,
+            mime_type: "text/plain".to_owned(),
```
mladedav, line 140: We already depend on `mime` so I'd suggest using that to construct the string. It could avoid some typo bugs. It would be great if it would be possible to use it also for the multi part content type. I don't know if it's possible with the dynamic boundary though.

axum-extra/src/multipart_builder.rs
```diff
@@ -0,0 +1,337 @@
...
+
+use axum::response::{IntoResponse, Response};
+use http::{header, HeaderMap};
+use std::time::{SystemTime, UNIX_EPOCH};
+
+/// The `Content-Transfer-Encoding` setting for a part.
+#[derive(Debug)]
+pub enum TransferEncoding {
+    /// If not specified, encoding defaults to UTF-8
+    Default,
```
mladedav, line 11: Defaults by us choosing it? You're setting it explicitly in the `Part::text` for example.

So I think this should be named `TextUtf8` or something se more descriptive.

If you want to signal somewhere you're using just a default that might change one day, derive `Default` and use that.

axum-extra/src/multipart_builder.rs
```diff
@@ -0,0 +1,337 @@
...
+/// Generate a single portion of the boundary by performing an xor
+/// shift on the current time
+fn gen_boundary_portion() -> u64 {
+    // *very* rudimentary xor-shift used
+    // the current time is used as a seed because it doesn't really matter what the data is, it should
+    // just be vaguely random
+    let mut x = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_millis() as u64;
```
mladedav, line 264: Is this also copied from reqwest?

So we take system time, do 3 xors and shifts and repeat. I would be very surprised if this took a millisecond, I would even expect all of the four "pseudorandom" numbers to be the same one most of the time. Then again it's not really that probable the data would contain the same random `u64` in hex four times next to each other separated by a hyphen. So it serves its purpose.

If you want to keep this, just use the last result as the next seed I guess. But I don't think that using `rand` (or `fastrand`) would hurt performance here much. This seems like a case of premature optimization.

axum-extra/src/multipart_builder.rs
```diff
@@ -0,0 +1,337 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{MultipartForm, Part};
+    use axum::{body::Body, http};
+    use axum::{routing::get, Router};
+    use http::{Request, Response};
+    use http_body_util::BodyExt;
+    // for `collect`
```
mladedav, line 278: You don't need to write why you import something. Personally, I would recommend against it because it's just a matter of time before someone needs to change something here and suddenly the comments are wrong.

axum-extra/src/lib.rs
```diff
@@ -83,6 +83,9 @@ pub mod routing;
 #[cfg(feature = "json-lines")]
 pub mod json_lines;
 
+#[cfg(feature = "multipart")]
+pub mod multipart_builder;
```
mladedav, line 87: Why "builder"?

axum-extra/src/multipart_builder.rs
```diff
@@ -0,0 +1,337 @@
...
+impl IntoResponse for MultipartForm {
+    fn into_response(self) -> Response {
+        // see RFC2388 for details
+        let boundary = generate_boundary();
+        let mut headers = HeaderMap::new();
+        headers.insert(
+            header::CONTENT_TYPE,
+            format!("multipart/form-data; boundary={}", boundary)
+                .parse()
+                .unwrap(),
```
mladedav, line 71: I would caution against unwraps. I think this should not happen, but either write a comment about why it's totally ok or just bite the bullet and return an explicit 500 here on error.

mladedav (dismissed):
Thanks, for the PR I think the functionality looks good. I have not personally worked with multipart on this level so I don't know much of the technical details but I did comment on some of the other stuff.

axum-core/Cargo.toml
```diff
@@ -34,6 +34,7 @@ tower-service = "0.3"
 # optional dependencies
 tower-http = { version = "0.5.0", optional = true, features = ["limit"] }
 tracing = { version = "0.1.37", default-features = false, optional = true }
+true = { version = "0.1.0", optional = true }
```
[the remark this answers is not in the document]
zleyyij: I have no clue, I don't remember ever adding that.

axum-extra/src/multipart_builder.rs
zleyyij: You're right, it did feel kind of odd, I was trying to create something similar to  Reqwest's implementation (`.part()`) chained calls, but with `.with_parts()` it feels redundant, so I removed it. I can add `parts_mut` if you feel it would prudent.

axum-extra/src/multipart_builder.rs
zleyyij: No, they use their own implementation of an xor shift
https://github.com/seanmonstar/reqwest/blob/7e7011a657e2f97d3b52dda7bcd8af4353c06b8a/src/util.rs#L27

This impl is definitely a bit jank.

I used this here because frankly I tried to use `fastrand` first, but couldn't figure out how to make it work without CI errors. I am not experienced at all working with `cargo.toml` at this scale. I'll dig into that some more and try to figure it out, thanks for the feedback

axum-extra/src/lib.rs
zleyyij: There's already a `Multipart` extractor, so I wanted to clarify between that and the functionality for creating forms.

axum-extra/src/multipart_builder.rs
zleyyij: I figured this out, the impl uses fastrand now.

axum-extra/src/multipart_builder.rs
zleyyij: Right now all of the `Part::*` constructors return Self directly, but `Part::raw_part` relies on a mime type passed in as an argument. 

I know that it's generally considered bad practice to add third party crates to the public api without good cause, but the only other option I can see here is having the function internally parse the provided string into a valid mime type, which should return a result, making this portion of the interface a little bit clunkier.

Thoughts? I could include mime in the public interface, or change the interface to return a result, making it less ergonomic.

zleyyij:
I have commit ready to fix some of those minor issues, I just need to figure out how to handle mime type stuff from a public api level

zleyyij:
I don't know how to do that, any advice would be appreciated

zleyyij:
I gave up and just made any mime parsing happen internally, and the associated functions return an error.

I ran as much CI as I could with `act` locally, hopefully it passes

zleyyij:
Is there anything more this PR needs?

jplatte:
No, it's just waiting on my review. I've started looking at it now.

axum-extra/src/multipart_builder.rs
```diff
@@ -0,0 +1,330 @@
...
+#[derive(Debug)]
+pub struct MultipartForm {
+    parts: Vec<Part>,
+}
+
+impl MultipartForm {
+    /// Construct a new empty multipart form with no parts.
+    pub fn new() -> Self {
+        MultipartForm { parts: Vec::new() }
+    }
```
jplatte, line 28: What use is this if there's no method to add a part?

axum-extra/src/multipart_builder.rs
```diff
@@ -0,0 +1,330 @@
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use axum_extra::multipart_builder::{MultipartForm, Part};
+    ///
+    /// let parts: Vec<Part> = vec![Part::text("foo".to_string(), "abc"), Part::text("bar".to_string(), "def")];
+    /// let form = MultipartForm::with_parts(parts);
+    /// ```
+    pub fn with_parts(parts: Vec<Part>) -> Self {
```
jplatte, line 40: Maybe there should be an `impl FromIterator<Part>`? Then you could do

```rust
MultipartForm::from_iter([
    first_part,
    second_part,
    third_part,
)]
```

or

```rust
some_iterator
    .map(|item| /* turn item into Part */)
    .collect()
```

axum-extra/src/multipart_builder.rs
```diff
@@ -0,0 +1,330 @@
...
+// - (optionally) a `Content-Type` header, which if not set, defaults to `text/plain`.
+// If the field contains a file, then the file should be identified with that file's MIME type (eg: `image/gif`).
+// If the `MIME` type is not known or specified, then the MIME type should be set to `application/octet-stream`.
+// - If the part does not conform to the default encoding, then the `Content-Transfer-Encoding` header may be supplied.
+// Valid settings for that header are: "base64", "quoted-printable", "8bit", "7bit", and "binary".
+/// A single part of a multipart form as defined by
+/// <https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4>
+/// and RFC2388.
+#[derive(Debug)]
+pub struct Part {
```
jplatte, line 100: It's hard to tell where the regular comment ends and the doc comment starts. Maybe move the regular comment inside the definition and separate it from the first field's docs by an empty line?

axum-extra/src/multipart_builder.rs
```diff
@@ -0,0 +1,330 @@
...
+        serialized_part += &format!("Content-Type: {}\r\n", self.mime_type);
+        // if an encoding was set, add that
+        // determine what encoding to label the body of the field with
+        let encoding: Option<&str> = match self.encoding {
+            TransferEncoding::TextUTF8 => None,
+            TransferEncoding::Binary => Some("binary"),
+        };
+        if let Some(encoding) = encoding {
+            serialized_part += &format!("Content-Transfer-Encoding: {}\r\n", encoding);
+        }
```
jplatte, line 227: Are you sure that this is a thing? I googled and found `Content-Encoding` and `Transfer-Encoding` docs, but the only mention of `Content-Transfer-Encoding` was in relation to some old RFC that has been replaced since (AFAIU).

axum-extra/src/lib.rs
zleyyij: That's definitely an improvement IMO, I felt multipart_builder was verbose but didn't know what else to do

axum-extra/src/multipart_builder.rs
zleyyij: The ability to add a part was removed in a previous iteration, so this is stale, good catch :)

axum-extra/src/multipart_builder.rs
zleyyij: So I actually implemented most of this using RFC 2388 (whoops), which has been superceded by RFC 7578, which deprecates that field. I updated the comments and code to follow RFC 5758

zleyyij:
Those changes have been applied, thank you for your feedback :)

axum-extra/src/response/multiple.rs
```diff
@@ -0,0 +1,292 @@
...
+
+impl FromIterator<Part> for MultipartForm {
+    fn from_iter<T: IntoIterator<Item = Part>>(iter: T) -> Self {
+        Self {
+            parts: iter.into_iter().collect(),
+        }
+    }
+}
+
+/// A boundary is defined as a user defined (arbitrary) value that does not occur in any of the data.
```
jplatte, line 207:
```suggestion
/// A boundary is defined as a user defined (arbitrary) value that does not occur in any of the data.
///
```

axum-extra/src/response/multiple.rs
```diff
@@ -0,0 +1,292 @@
...
+            filename: Some(file_name.to_owned()),
+            // If the `MIME` type is not known or specified, then the MIME type should be set to `application/octet-stream`.
+            // See RFC2388 section 3 for specifics.
+            mime_type: mime::APPLICATION_OCTET_STREAM,
+            contents,
+        }
+    }
+
+    /// Create a new part with more fine-grained control over the semantics of that part. The caller
+    /// is assumed to have set a valid MIME type.
```
jplatte, line 141:
```suggestion
    /// Create a new part with more fine-grained control over the semantics of that part.
    ///
    /// The caller is assumed to have set a valid MIME type.
```

axum-extra/src/response/multiple.rs
```diff
@@ -0,0 +1,292 @@
...
+            name,
+            filename: None,
+            mime_type: mime::TEXT_PLAIN_UTF_8,
+            contents: contents.as_bytes().to_vec(),
+        }
+    }
+
+    /// Create a new part containing a generic file, with a `Content-Type` of `application/octet-stream`
+    /// using the provided file name, field name, and contents. If the MIME type of the file is known, consider
+    /// using `Part::raw_part`.
```
jplatte, line 117:
```suggestion
    /// Create a new part containing a generic file, with a `Content-Type` of `application/octet-stream`
    /// using the provided file name, field name, and contents.
    ///
    /// If the MIME type of the file is known, consider using `Part::raw_part`.
```

axum-extra/src/response/multiple.rs
```diff
@@ -0,0 +1,292 @@
...
+    filename: Option<String>,
+    /// The `Content-Type` header. While not strictly required, it is always set here
+    mime_type: Mime,
+    /// The content/body of the part
+    contents: Vec<u8>,
+}
+
+impl Part {
+    /// Create a new part with `Content-Type` of `text/plain` with the supplied name and contents.
+    /// This form will not have a defined file name.
```
jplatte, line 94:
```suggestion
    /// Create a new part with `Content-Type` of `text/plain` with the supplied name and contents.
    ///
    /// This form will not have a defined file name.
```

jplatte (approved):
I don't have the energy to do another review, but I think this has seen enough attention that we should just merge it by now.

There's just some doc comment rewrapping I wanna apply before merging (can likely commit it myself).

jplatte:
Do you want this to be released in another axum-extr 0.9.x patch release? It's not a huge effort but if you don't care, I'll let it sit until I have another reason to make a patch release, or it's time for 0.10.0.

zleyyij:
Feel free to let it sit, this is fairly low priority for me.

Thanks :)

axum-extra/src/response/multiple.rs
```diff
@@ -0,0 +1,296 @@
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use axum_extra::response::multiple::{MultipartForm, Part};
+    ///
+    /// let parts: Vec<Part> = vec![Part::text("foo".to_string(), "abc"), Part::text("bar".to_string(), "def")];
+    /// let form = MultipartForm::with_parts(parts);
+    /// ```
+    #[deprecated]
```
jplatte, line 27: I just found this for another reason... Why did we add a new deprecated item here?? @zleyyij

zleyyij: That definitely shouldn't be there, I'm so sorry. I probably added it during a refactor with the intention of changing that method before deciding it would be better if I didn't.

I'll remove it.

Thank you so much for your patience, I'm so sorry about all the hassle this PR has created

zleyyij: Will be resolved by https://github.com/tokio-rs/axum/pull/2957

PR #2675, opened 2024-03-24
axum: remove unnecessary `Arc::clone`

mladedav:
## Motivation

There was an unnecessary arc clone which is always dropped right afterward.

The `try_unwrap` call returns `Ok` only if the `Arc` has a strong count of exactly 1, but since we have just cloned it we know it will be more than 1. So the `try_unwrap` never works and the more expensive `clone` always happens anyway.

## Solution

I have removed the `Arc` operations.

### DIFF
--- a/axum/src/routing/path_router.rs
+++ b/axum/src/routing/path_router.rs
@@ -111,12 +111,10 @@ where
     }
 
     fn set_node(&mut self, path: &str, id: RouteId) -> Result<(), String> {
-        let mut node = (*self.node).clone();
-        if let Err(err) = node.insert(path, id) {
-            return Err(format!("Invalid route {path:?}: {err}"));
-        }
-        self.node = Arc::new(node);
-        Ok(())
+        let node = Arc::make_mut(&mut self.node);
+
+        node.insert(path, id)
+            .map_err(|err| format!("Invalid route {path:?}: {err}"))
     }
 
     pub(super) fn merge(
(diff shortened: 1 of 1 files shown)


axum/src/routing/path_router.rs
```diff
@@ -111,8 +111,7 @@ where
     }
 
     fn set_node(&mut self, path: &str, id: RouteId) -> Result<(), String> {
-        let mut node =
-            Arc::try_unwrap(Arc::clone(&self.node)).unwrap_or_else(|node| (*node).clone());
+        let mut node = (*self.node).clone();
```
jplatte, line 114: Could you still use fully qualified syntax here? Just so it's clear that this is a 'shallow' clone. I suspect we're not consistent about that overall, but would probably use fully qualified if we were to make things consistent.

jplatte (commented):
Thanks! Just one nitpick.

axum/src/routing/path_router.rs
mladedav: This was actually a deep clone.

However, I took a minute to think about how to actually do what the code probably wanted to do in the first place (which I guess I should have done at the start) and changed it to use `Arc::make_mut`. So as long as the router hasn't been cloned, it'll cheaply update it and otherwise fall back to a clone of `Node`.

jplatte (approved):
Huh, I wouldn't have noticed that this meant to be `make_mut`, but now that you said it it makes perfect sense 😅 

Thanks!

PR #2688, opened 2024-03-29
Upgrade reqwest to 0.12

tisonkun:
New reqwest version uses http 1 and hyper 1.

Reduce some intermediate conversions.


jplatte (approved):
Thanks!

PR #2696, opened 2024-04-02
feat: Update reverse proxy example to support both http and https

afifurrohman-id:
## Motivation

Example for reverse proxy currently only using http, but sometimes we want to proxy https server as well.

## Solution

Update existing example and added it.

### DIFF
--- a/examples/reverse-proxy/src/main.rs
+++ b/examples/reverse-proxy/src/main.rs
@@ -9,6 +9,7 @@
 //! cargo run -p example-reverse-proxy
 //! ```
 
+use axum::http::{header::HOST, StatusCode};
 use axum::{
     body::Body,
     extract::{Request, State},
@@ -16,7 +17,6 @@ use axum::{
     response::{IntoResponse, Response},
     routing, Router,
 };
-use axum::http::{header::HOST, StatusCode};
 use hyper_tls::HttpsConnector;
 use hyper_util::{client::legacy::connect::HttpConnector, rt::TokioExecutor};
 
@@ -51,7 +51,7 @@ async fn handler(State(client): State<Client>, mut req: Request) -> Result<Respo
 
     let mut uri = format!("http://127.0.0.1:3000{}", path_query);
     if path == "/https" {
-        uri = format!("https://example.com")
+        uri = String::from("https://example.com");
     }
 
     *req.uri_mut() = Uri::try_from(uri).unwrap();
--- a/examples/reverse-proxy/src/main.rs
+++ b/examples/reverse-proxy/src/main.rs
@@ -1,5 +1,7 @@
-//! Reverse proxy listening in "localhost:4000" will proxy all requests to "localhost:3000"
+//! Reverse proxy listening in "localhost:4000" will proxy all `GET` requests to "localhost:3000" except for path /https is example.com
 //! endpoint.
+//!
+//! On unix like OS: make sure `ca-certificates` is installed.
 //!
 //! Run with
 //!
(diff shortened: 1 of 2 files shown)


yanns (approved):

examples/reverse-proxy/src/main.rs
```diff
@@ -45,10 +49,16 @@ async fn handler(State(client): State<Client>, mut req: Request) -> Result<Respo
...
-    let uri = format!("http://127.0.0.1:3000{}", path_query);
+    let mut uri = format!("http://127.0.0.1:3000{}", path_query);
+    if path == "/https" {
+        uri = String::from("https://example.com");
+    }
 
     *req.uri_mut() = Uri::try_from(uri).unwrap();
 
+    //? Remove incorrect header host, hyper will add automatically for you.
+    req.headers_mut().remove(HOST).unwrap();
```
jplatte, line 60: Why the question mark in front here? Personally I don't know if this is the right behavior here, @yanns can you comment on it since you approved (in addition to @afifurrohman-id)?

Even if we keep it I don't think there's any reason to `.unwrap()` there though.

yanns: When proxing a HTTP request, the initial HTTP request contains as `HOST` the address of the proxy (`locahost:4000`). If we use that value against another server, that server will refuse the HTTP request as it expects `localhost:3000` instead.

jplatte: So this was previously broken? Also, should the header value be added as some (de-facto) standard proxied-host header?

yanns: > So this was previously broken? 
I guess yes. I have not checked. I've built several proxy and this is one of the usual issues.

> Also, should the header value be added as some (de-facto) standard proxied-host header?

If we want to go in the direction of "good behaving" proxy, yes they are several headers that should be set:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

mladedav: I just tried the original example and it works.

The `host` header would be an issue if the two servers were running externally visible on the same port as virtual servers with nginx or something else in front of them that would forward the requests based on the `host` header.

`axum`, `hyper`, or anything else in the stack does not check the header, because it can't even know what to check it for. The request just arrives either on `localhost:4000` or `localhost:3000` and then it is processed.

Personally, I think that handling of this and the addition of an external dependency on `example.com` is a bit sad and it would be better to just add an axum tls server (based on any of the other examples) and call that instead.

But even with `example.com`, the header change is not really needed as `curl -v https://example.com -H 'host: localhost:3000'` returns a 404 -- it's not the same page you get if you let curl use the correct `host` (and it might be returned by a reverse proxy and not the actual server), but it is a response to an `https` request so it does show that this works.

Generally, I think the examples should be as simple as possible to showcase the thing they mean to (here that's relaying a request, potentially with a change to the protocol), but details such as extension headers can and should be omitted for brevity.

afifurrohman-id: > Why the question mark in front here?

I just for clarify why we remove `HOST` header, as @yanns say.

afifurrohman-id: > If we want to go in the direction of "good behaving" proxy, yes they are several headers that should be set:
> 
> * https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host
> * https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
> * https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

I also think like so, we need add it as well, like some other reverse proxy, it will also make more great example.

afifurrohman-id: > Personally, I think that handling of this and the addition of an external dependency on `example.com` is a bit sad and it would be better to just add an axum tls server (based on any of the other examples) and call that instead.

Yes, first because it think to simple as possible use an `example.com` can add more precisely how we can interact with external service, but will add some external network request as well.

Turbo87:
from looking at the diff of this PR, it seems like it does quite a bit more than just adding support for HTTPS. it also adds a new special route, it adds various forwarded headers, it prints request errors to the console, etc.

while some of this might be useful in a real-world reverse proxy, I'm not sure if it adds to the understanding of how to build a basic reverse proxy with axum.

IMHO the existing example is simple enough to understand the basic concept, and we shouldn't pretend that the proposed example code is a production-ready reverse proxy implementation either. I'd vote for keeping it as-is and potentially documenting that this is an abbreviated example where many features of real-world reverse proxies are missing.

Issue #1146, opened 2022-07-07
Make `TestClient` public


rouge8:
### Motivation

I want to write tests for my axum app. `TestClient` makes that easy to do so.

### Proposal

Make the existing `TestClient` public. 

### Alternatives

Use [axum-test-helper](https://github.com/cloudwalk/axum-test-helper) or vendor the `TestClient` myself.

davidpdrsn:
It's just a thin wrapper around reqwest so I recommend you use that instead.

You can also have a look at the testing example https://github.com/tokio-rs/axum/blob/main/examples/testing/src/main.rs

Eric-Arellano:
Hi, thanks for Axum! I heard about it on Rustacean Station and am really enjoying it.

Thoughts on adding an explicit section to the main page of the docs on testing, and linking to this example? It took some digging through GitHub to find the recommended approach. I'm happy to submit a patch if you'd like.

davidpdrsn:
Sure. Wanna make a PR?

PR #2699, opened 2024-04-06
feat: expose test client under testing feature

tisonkun:
## Motivation

There is a crate [axum-test-helper](https://github.com/cloudwalk/axum-test-helper) for the same purpose and it seems widely used. Since the crate is less actively maintained, I suppose we can implement the same on the upstream so keep it updated.

## Solution

```rust
#[cfg(any(test, feature = "testing"))]
pub mod test_helpers;
```

... and export all inner methods for test clients.

@jplatte

### DIFF
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -32,6 +32,7 @@ matched-path = []
 multipart = ["dep:multer"]
 original-uri = []
 query = ["dep:serde_urlencoded"]
+testing = ["dep:reqwest"]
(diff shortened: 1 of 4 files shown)


tisonkun:
Some compile errors happen. I may spend more time investigating, while the "expected outer doc comment" seems a false-positive.

Anyway, I'd like to hear from the maintainer whether this is a good way to go.

jplatte:
This was previously rejected in https://github.com/tokio-rs/axum/issues/1146, sorry.

PR #2705, opened 2024-04-10
fix server leaking IO resources

nicolaiunrein:
## Motivation

TcpStreams are not dropped when the future returned from `serve` is dropped. This is the case with and without graceful shutdown. If there are open connections for long running requests these will possibly hang on to the TcpStream forever making it imposibble to start a new server or listen on this port until the process exits and the OS cleans up the port.

## Solution

I propose adding a drop signal and using `tokio::select` on this and the body of `task::spawn`

# Example
```rust
use std::sync::Arc;

#[tokio::main]
async fn main() {
    let notify = Arc::new(tokio::sync::Notify::new());
    let cancel = notify.clone();
        "/",
        get(move || async move {
            notify.notify_waiters();
            loop {
                tokio::time::sleep(std::time::Duration::from_secs(1)).await;
            }
        }),
    );

    println!("starting server");

    axum::serve(listener, app)
        .with_graceful_shutdown(async move { cancel.notified().await })
        .await
        .unwrap();

    println!("server is shut down");

    // but port 3000 is still blocked

    loop {
        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
    }
}

```

### DIFF
--- a/axum/src/serve.rs
+++ b/axum/src/serve.rs
@@ -22,10 +22,27 @@ use hyper_util::server::conn::auto::Builder;
 use pin_project_lite::pin_project;
 use tokio::{
     net::{TcpListener, TcpStream},
-    sync::watch,
+    sync::{watch, Notify},
 };
 use tower::util::{Oneshot, ServiceExt};
 use tower_service::Service;
+
+#[derive(Clone, Default)]
+struct DropNotify(Arc<Notify>);
+
+impl std::ops::Deref for DropNotify {
+    type Target = Notify;
+
+    fn deref(&self) -> &Self::Target {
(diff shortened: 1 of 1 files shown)


nicolaiunrein:
I just realised that the implementation and example is wrong for the graceful case. But in both cases I think it would very nice if there was a way to shutdown the server forcefully dropping all open connections. In our project we use a lot of sse streams in child processes spawned on windows. Letting the OS close the connections does not work in our case because windows seems to not close connections if the parent process is still alive. I can think of other situations where it is desirable to close all open connections. And the current implementation is at least very surprising! I will update my PR tomorrow.

PR #2720, opened 2024-04-22
Add ErrorKind::DeserializeError to specialize ErrorKind::Message (`extract::path::ErrorKind`)

vegardgs-ksat:
## Motivation

Expose the `key` and `value` of the `extract::Path` extrator for when the serde deserialization fails. This allow custom extractors wrapping the `Path` extractor to specialize their responses with more context about the failed operation. Currently, all failed deserializations end up as part of the `extract::path::ErrorKind::Message` variant.

I encountered this situation when attempting to create my own extrator, using it to deserialize `uuid::Uuid` resources, since I require specializing the response for these error conditions.

```rust
//! Manual implementation wrapping `axum::extract::Path` extrator.

use axum::extract::FromRequestParts;
use axum::extract::{path::ErrorKind, rejection::PathRejection};
use axum::http::{request::Parts, StatusCode};
use serde::de::DeserializeOwned;

pub struct Path<T>(pub T);

#[axum::async_trait]
impl<T, S> FromRequestParts<S> for Path<T>
where
    T: DeserializeOwned + Send,
    S: Send + Sync,
{
    type Rejection = (StatusCode, axum::Json<serde_json::Value>);

        match axum::extract::Path::<T>::from_request_parts(parts, state).await {
            Ok(value) => Ok(Self(value.0)),
            Err(rejection) => match &rejection {
                PathRejection::FailedToDeserializePathParams(f) => match f.kind() {
                    ErrorKind::ParseErrorAtKey { key, value, .. } => Err((
                        StatusCode::BAD_REQUEST,
                        axum::Json(serde_json::json!({
                            "code": "MALFORMED_VALUE",
                            "message": format!("Value '{value}' could not be parsed"),
                            "target": key
                        })),
                    )),
                    
                    // Nominal UUID deserialization issuess occur in ErrorKind::Message
                    
                    _ => Err((
                        StatusCode::BAD_REQUEST,
                        axum::Json(serde_json::json!({
                            "code": "MALFORMED_VALUE",
                            "message": f.body_text(),
                            "target": "",
                        })),
                    )),
                },
                PathRejection::MissingPathParams(..) | _ => {
                    Err((
                        axum::Json(serde_json::json!({
                            "message": "An internal server error occurred",
                        })),
                    ))
                }
            },
        }
    }
}

// This is the endpoint handler using the custom Path extractor.
async fn find_resource(Path(resource_id): Path<uuid::Uuid>) -> String {
    todo!()
}
```

## Solution

This commit introduces another `extract::path::ErrorKind` variant that captures the serde error nominally captured through the `serde::de::Error` trait impl on `PathDeserializeError`. We augment the deserialization error with the captured (key, value), allowing `extract::Path`, and wrapping extractors, to gain programmatic access to the key name, and attempted deserialized value.

This allows me to expand the handling in the wrapping Path extractor.
```rust
ErrorKind::DeserializeError { key, value, message } => Err((StatusCode::BAD_REQUEST, axum::Json(serde_json::json!({
  "code": "MALFORMED_VALUE",
  "message": format!("Value '{value}' could not be parsed"),
  "target": key
})),
```

### DIFF
--- a/axum/src/extract/path/de.rs
+++ b/axum/src/extract/path/de.rs
@@ -94,7 +94,21 @@ impl<'de> Deserializer<'de> for PathDeserializer<'de> {
                 .got(self.url_params.len())
                 .expected(1));
         }
-        visitor.visit_borrowed_str(&self.url_params[0].1)
+        let key = &self.url_params[0].0;
+        let value = &self.url_params[0].1;
+        visitor
+            .visit_borrowed_str(value)
+            .map_err(|e: PathDeserializationError| {
+                if let ErrorKind::Message(message) = &e.kind {
+                    PathDeserializationError::new(ErrorKind::DeserializeError {
+                        key: key.to_string(),
+                        value: value.as_str().to_owned(),
+                        message: message.to_owned(),
+                    })
+                } else {
+                    e
+                }
+            })
     }
 
     fn deserialize_unit<V>(self, visitor: V) -> Result<V::Value, Self::Error>
@@ -362,7 +376,19 @@ impl<'de> Deserializer<'de> for ValueDeserializer<'de> {
     where
         V: Visitor<'de>,
     {
-        visitor.visit_borrowed_str(self.value)
+        visitor
+            .visit_borrowed_str(self.value)
+            .map_err(|e: PathDeserializationError| {
+                if let (ErrorKind::Message(message), Some(key)) = (&e.kind, self.key.as_ref()) {
+                    PathDeserializationError::new(ErrorKind::DeserializeError {
+                        key: key.key().to_owned(),
+                        value: self.value.as_str().to_owned(),
+                        message: message.to_owned(),
+                    })
+                } else {
+                    e
+                }
+            })
     }
 
     fn deserialize_bytes<V>(self, visitor: V) -> Result<V::Value, Self::Error>
@@ -606,6 +632,15 @@ impl<'de> SeqAccess<'de> for SeqDeserializer<'de> {
 enum KeyOrIdx<'de> {
     Key(&'de str),
     Idx { idx: usize, key: &'de str },
+}
+
+impl<'de> KeyOrIdx<'de> {
+    fn key(&self) -> &'de str {
+        match &self {
+            Self::Key(key) => key,
+            Self::Idx { key, .. } => key,
+        }
+    }
 }
 
 #[cfg(test)]
@@ -923,6 +958,19 @@ mod tests {
             Vec<Vec<String>>,
             ErrorKind::UnsupportedType {
                 name: "alloc::vec::Vec<alloc::string::String>",
+            }
+        );
+    }
+
+    #[test]
+    fn test_deserialize_key_value() {
+        test_parse_error!(
+            vec![("id", "123123-123-123123")],
+            uuid::Uuid,
+            ErrorKind::DeserializeError {
+                key: "id".to_owned(),
+                value: "123123-123-123123".to_owned(),
+                message: "UUID parsing failed: invalid group count: expected 5, found 3".to_owned(),
             }
         );
     }
--- a/axum/src/extract/path/mod.rs
+++ b/axum/src/extract/path/mod.rs
@@ -303,6 +303,16 @@ pub enum ErrorKind {
         name: &'static str,
     },
 
+    /// Failed to deserialize the value with a custom deserialization error.
+    DeserializeError {
+        /// The key at which the invalid value was located.
+        key: String,
+        /// The value that failed to deserialize.
+        value: String,
+        /// The deserializaation failure message.
+        message: String,
+    },
+
     /// Catch-all variant for errors that don't fit any other variant.
     Message(String),
 }
@@ -345,6 +355,11 @@ impl fmt::Display for ErrorKind {
                 f,
                 "Cannot parse value at index {index} with value `{value:?}` to a `{expected_type}`"
             ),
+            ErrorKind::DeserializeError {
+                key,
+                value,
+                message,
+            } => write!(f, "Cannot parse `{key}` with value `{value:?}`: {message}"),
         }
     }
 }
@@ -369,6 +384,7 @@ impl FailedToDeserializePathParams {
     pub fn body_text(&self) -> String {
         match self.0.kind {
             ErrorKind::Message(_)
+            | ErrorKind::DeserializeError { .. }
             | ErrorKind::InvalidUtf8InPathParam { .. }
             | ErrorKind::ParseError { .. }
             | ErrorKind::ParseErrorAtIndex { .. }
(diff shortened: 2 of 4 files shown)


axum/src/extract/path/mod.rs
```diff
@@ -347,6 +357,11 @@ impl fmt::Display for ErrorKind {
                 f,
                 "Cannot parse value at index {index} with value `{value:?}` to a `{expected_type}`"
             ),
+            ErrorKind::DeserializeError {
+                key,
+                value,
+                message,
+            } => write!(f, "Cannot parse `{key}` with value `{value:?}`: {message}"),
```
mladedav, line 364: Do not use debug for `value`, it would add redundant quotes. You can also that in the new tests.

axum/src/extract/path/de.rs
```diff
@@ -362,7 +375,18 @@ impl<'de> Deserializer<'de> for ValueDeserializer<'de> {
     where
         V: Visitor<'de>,
     {
-        visitor.visit_borrowed_str(self.value)
+        visitor
+            .visit_borrowed_str(self.value)
+            .map_err(|e: Self::Error| match e.kind {
+                ErrorKind::Message(message) => {
+                    PathDeserializationError::new(ErrorKind::DeserializeError {
+                        key: self.key.map(|k| k.key().to_owned()).unwrap_or_default(),
```
mladedav, line 383: When can this be `None`? Does using an empty string as a fallback make sense here? Can we maybe use another variant, even the catch all `Message` as a fallback instead?

axum/src/extract/path/de.rs
```diff
@@ -94,7 +94,20 @@ impl<'de> Deserializer<'de> for PathDeserializer<'de> {
                 .got(self.url_params.len())
                 .expected(1));
         }
-        visitor.visit_borrowed_str(&self.url_params[0].1)
+        let key = &self.url_params[0].0;
+        let value = &self.url_params[0].1;
+        visitor
+            .visit_borrowed_str(value)
+            .map_err(|e: Self::Error| match e.kind {
```
mladedav, line 101: Nit, I'm not sure you need to specify the type at all, but if you want to do that, I'd rather use `ErrorKind` for consistency's sake since you're using that on the next lines.

axum/src/extract/path/de.rs
```diff
@@ -94,7 +94,20 @@ impl<'de> Deserializer<'de> for PathDeserializer<'de> {
...
+                ErrorKind::Message(message) => {
+                    PathDeserializationError::new(ErrorKind::DeserializeError {
+                        key: key.to_string(),
+                        value: value.as_str().to_owned(),
+                        message,
+                    })
+                }
+                kind => PathDeserializationError::new(kind),
```
mladedav, line 109: This is very nitty so take it or leave it, but if you really care just about one variant, you can use `if let` you'll have less indented code (maybe not in this case because the match is on the same line as the closure argument) and in the else branch you can directly return `e` instead of constructing the error again from a kind.

mladedav (changes requested):
This looks good, just a few minor points.

axum/src/extract/path/mod.rs
vegardgs-ksat: I was just mirroring the practice established by the other formats for the other variants: https://github.com/tokio-rs/axum/blob/e3bb7083c886247f4e6931e149ef6067e6b82e1b/axum/src/extract/path/mod.rs#L330-L349

I was a bit perplexed about this as well, and thought it did not make sense to use debug format for `value`.

Should I change it do non-Debug for this new variant only, or change the other variants as well?

axum/src/extract/path/de.rs
vegardgs-ksat: From what I understand of the code, it is only `None` when deserializing maps. This is also evident from this comment: https://github.com/tokio-rs/axum/blob/e3bb7083c886247f4e6931e149ef6067e6b82e1b/axum/src/extract/path/de.rs#L450-L451

I can match on the existence of the key, and keep it to `ErrorKind::Message` if key is ever `None`.

axum/src/extract/path/de.rs
vegardgs-ksat: The compiler barfs at me without annotation:
```
error[E0282]: type annotations needed
  --> axum/src/extract/path/de.rs:99:52
   |
99 |         visitor.visit_borrowed_str(value).map_err(|e| match e.kind {
   |                                                    ^        - type must be known at this point
   |
help: consider giving this closure parameter an explicit type
   |
99 |         visitor.visit_borrowed_str(value).map_err(|e: /* Type */| match e.kind {
   |                                                     ++++++++++++

For more information about this error, try `rustc --explain E0282`.
error: could not compile `axum` (lib) due to 1 previous error
```

I will change it to `ErrorKind`.

axum/src/extract/path/mod.rs
vegardgs-ksat: I added an additional commit that changes the format for all variants.

mladedav: Thanks. Sorry I didn't notice you just wrote it consistently with the rest, otherwise I wouldn't bring it up here.

mladedav:
LGTM, @jplatte can you take a look when you have some free time?

mladedav (approved):

vegardgs-ksat:
Rebased against latest main for an up-to-date PR.

vegardgs-ksat:
@mladedav anything holding up this PR?

jplatte (approved):
Yeah, I was holding it up by not reviewing, sorry. I did so now, and the change looks good to me.

One question, and a request though:
- Is `ErrorKind::Message` still constructed anywhere? If not, would make sense to remove it, I wouldn't want to release this in a patch of 0.7 anyways.
- Could you please add a changelog entry?

vegardgs-ksat:
>     * Is `ErrorKind::Message` still constructed anywhere? If not, would make sense to remove it, I wouldn't want to release this in a patch of 0.7 anyways.

It is indirectly constructed through the `PathDeserializationError::custom()` method, [here](https://github.com/tokio-rs/axum/blob/280d16a61059f57230819a79b15aa12a263e8cca/axum/src/extract/path/de.rs#L282) and [here](https://github.com/tokio-rs/axum/blob/280d16a61059f57230819a79b15aa12a263e8cca/axum/src/extract/path/de.rs#L251)

>     * Could you please add a changelog entry?

Done! Added it under the "# Unreleased" section

vegardgs-ksat:
I have a feeling that the failing CI job is unrelated to this PR.

mladedav:
Yeah, it seems unrelated. I don't know the public api check but when I get some time, I'll try to figure out what's happening and then we'll merge this. Other branches seem to have also failed, it's possible one of our dependencies we're exposing started exposing another dependency we didn't mean to expose.

Sorry the whole thing takes so long.

mladedav:
Thanks for sticking with this and pushing it through the finish line!

PR #2729, opened 2024-05-03
feature: percent-decode before routing

### LINKED ISSUE #2678: Routing percent-encoded paths

### Motivation

URIs can be percent-encoded but that should not change what resource is accessed.

More specifically, [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3) says:

>    URIs that differ in the replacement of an unreserved character with
   its corresponding percent-encoded US-ASCII octet are equivalent: they
   identify the same resource. 
   
There are also some characters that are sometimes encoded and sometimes not in the real world, e.g. hyper will probably have handling of `{`, `}`, and `"` in paths [configurable](https://github.com/hyperium/hyper/issues/3594).

### Proposal

We can unescape unreserved characters inside the path, which can be decoded at any time. E.g. requests `/axum` and /%61xum` can be interpreted as the same one. We can also normalize reserved characters by percent-encoding them, if they are in the path (with some exceptions that already have special meaning like `%` , `?`, `#`, and so on; we can however encode `{`, `"`, ` `, etc. to normalize these before routing).

We can also encode special characters when registering a route such as internally turning current `.route("/100%",..)` into `.route("/100%25", ...)` (`%25` is percent encoding for `%`). Special case of this is that a user can currently write `.route("/what?",...)` which can never match any request.

With these two changes users can use special characters in route like `r#"/"qoutes"/etc"#` and match it with both encdoed and not encoded variants.

### Alternatives

Using a middleware before `axum::Router` that decodes (or rather otherwise normalizes) the percent-encoding. Not all percent-decoded paths are valid paths in URI so it would most likely have to percent-decode unreserved characters and percent-encode reserved characters to normalize.

This can be combined with encoding special characters when they are to be registered to `matchit` (e.g. braces, percent, question mark,...) either by providing something like `route_encoded` or by having the user encode any needed characters themselves.

I mention this primarily in case we would want to support people who want to have control over percent decoding themselves, but I think this can be also just built inside `axum` itself.

mladedav:
Note that this PR is built on top of #2645 so it also contains its changes. The only relevant change is the last commit.

Closes #2678 

## Motivation

Matching was happening before percent decoding. Therefore, `/a` and `/%61` were two different paths although according to the spec they shouldn't be.

We were also unable to construct routes with special characters that would have to be percent encoded by clients such as `?` or users would have to specify their paths already percent encoded when calling `route`.

## Solution

The paths are percent decoded before passing them to `matchit`. One exception is the forward slash `/`, which must not be decoded as that could change the matching behavior. This is achieved by encoding the `%` in `%2f` which would otherwise be decoded to `/` and interpreted as a path separator by `matchit::Router`. This way, the `matchit::Router` still sees `%2f` which gets decoded later on when inserting matched captures to request extensions.

For example with a routes `/api/{resource}` and `/api/car/{name}`, making a call such as `/api/car%2flada` should match the first route, not the second one.

One drawback of this is that if a user makes a request such as `/%252f_is_slash`, which would be captured by route `/{capture}`, the value of the captured path should be `%2f_is_slash` but because slashes are decoded after everything else, this will actually produce `/_is_slash`. This should be very rare scenario though.

Other characters work as intended, e.g. request to `/%2561_a` would result in a capture of `%61_a`.


2ndDerivative:
Is this blocked by anything that needs work still?

PR #2731, opened 2024-05-06
Add Extension<NonCloneType> debug_handler ui test

### LINKED ISSUE #2294: Add special handling of `Extension` and `Clone` for `debug_handler`

jplatte:
## Motivation

#2294 indicates that the error message for `T: Clone` constraint on `Extension<T>`'s `FromRequestParts` impl being unmet is bad even with `#[debug_handler]`. I decided to check and found that's false.

## Solution

Add a test to make sure we don't regress.

Closes #2294.
Supersedes #2301.

### DIFF
--- a/axum-macros/tests/debug_handler/fail/extension_not_clone.rs
+++ b/axum-macros/tests/debug_handler/fail/extension_not_clone.rs
@@ -1,0 +1,9 @@
+use axum::extract::Extension;
+use axum_macros::debug_handler;
+
+struct NonCloneType;
+
+#[debug_handler]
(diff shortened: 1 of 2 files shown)


yanns (approved):

jplatte:
Note to self: co-authered syntax is wrong, need to fix.

davidpdrsn (approved):

PR #2733, opened 2024-05-10
docs: remove axum-sessions from ecosystem (it moved to tower-sessions)

arcstur:
tower-sessions already is in the ecosystem


jplatte:
Sorry, seems like I missed this! You're indeed right, I previously merged #2862 which changed `axum-sessions` to `tower-sessions` and in the process duplicated the entry.

jplatte (approved):

Issue #2655, opened 2024-03-16
Remove tracing feature from axum-core


jplatte:
It doesn't actually do anything, but removing it is a breaking change.

jplatte:
No longer relevant after #2739.

PR #2739, opened 2024-05-13
axum-core: correctly use features for tracing macro

mladedav:
## Motivation

An alternative to #2655.

Currently, `axum-core` defines a `tracing` feature which is never used, because it was intended only for expansion inside a macro, but the `#[cfg(feature = "tracing")]` attribute is expanded and is evaluated based on whether the consuming crate (in this case either `axum` or `axum-extra`) has the feature enabled.

## Solution

Based on the feature, we can either expand the macro to a `tracing::event!` call or we can just expand to nothing. This way, the feature works as intended.

One difference between this and removing the feature is that with this change, if either `tracing` or `tracing-extra` enables the `tracing` feature, it will transitively enable the same feature in `axum-core` so extractors from both crates will have their rejections logged.

If we instead remove the feature from `axum-core`, each crate's `tracing` feature will be independent which may result in some extractors being logged and others not. So although it gives more control to the user, I personally think the behavior may be surprising.

Also, removing the feature is a breaking change while this is not.

### DIFF
--- a/axum-core/src/macros.rs
+++ b/axum-core/src/macros.rs
@@ -1,4 +1,5 @@
 /// Private API.
+#[cfg(feature = "tracing")]
 #[doc(hidden)]
 #[macro_export]
 macro_rules! __log_rejection {
@@ -7,7 +8,6 @@ macro_rules! __log_rejection {
         body_text = $body_text:expr,
         status = $status:expr,
     ) => {
-        #[cfg(feature = "tracing")]
         {
             tracing::event!(
                 target: "axum::rejection",
@@ -19,6 +19,17 @@ macro_rules! __log_rejection {
             );
         }
     };
+}
+
+#[cfg(not(feature = "tracing"))]
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __log_rejection {
+    (
+        rejection_type = $ty:ident,
+        body_text = $body_text:expr,
+        status = $status:expr,
+    ) => {};
 }
 
 /// Private API.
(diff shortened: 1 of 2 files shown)


axum-core/src/macros.rs
```diff
@@ -7,7 +8,6 @@ macro_rules! __log_rejection {
         body_text = $body_text:expr,
         status = $status:expr,
     ) => {
-        #[cfg(feature = "tracing")]
         {
             tracing::event!(
                 target: "axum::rejection",
```
jplatte, line 13: Hmm, I get where you're coming from re. confusion but actually I find it also pretty weird that right now, we use this tracing target for both `axum` and `axum-extra` extractors... I'm wondering if it wouldn't be better to just copy-paste the macro into each crate.

davidpdrsn:
> Currently, `tracing-core` defines a tracing feature which is never used

Just making sure I understand, you mean

> Currently, ~~`tracing-core`~~ `axum-core` defines a tracing feature which is never used

right?

axum-core/src/macros.rs
jplatte: Okay, let's merge this then.

mladedav:
Right, I meant `axum-core`. I've also changed the PR name and description accordingly.

The code is fixed but CI is still failing due to unrelated issues. I'll send a separate PR for that.

jplatte:
Note that the warnings in the cargo-public-api-crates job are not what makes it fail - we likely need to pin an older nightly to fix it as https://github.com/davidpdrsn/cargo-public-api-crates has no fix commit, or even an open issue or PR for this:

```
Error: missing field `expr` at line 1 column 128550
```

jplatte (approved):

PR #2740, opened 2024-05-15
example: Microsoft Entra integration using azure_identity

duvet86:
## Motivation
I couldn't find a good example of integration between axum and Microsoft Entra using azure_identity crate.

## Solution
I hope this will help other developers integrating axum with azure identity.

I am not sure I am using the app state correctly with AuthorizationCodeFlow struct since it does not support the clone trait.

### DIFF
--- a/examples/oauth-azure-identity/Cargo.toml
+++ b/examples/oauth-azure-identity/Cargo.toml
@@ -4,25 +4,26 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


jplatte:
Hi, sorry but this is definitely not a common enough thing people would need when using axum, so I wouldn't want to have it in this repo and maintenain it.

However, if you put it into its own repository, I'd be happy to accept a PR that adds a link to it to our `ECOSYSTEM.md` file.

PR #2742, opened 2024-05-19
Derive PartialEq for Json extractor

jkylling:
## Motivation

It should be possible to test handlers using
```rust
assert_eq!((StatusCode::OK, Json(MyStruct)), my_handler().await);
```
Currently, this can only be achieved with
```rust
let (status_code, Json(response)) = my_handler().await;
assert_eq!(StatusCode::OK, status_code);
assert_eq!(MyStruct, response);
```

### DIFF
--- a/axum/src/json.rs
+++ b/axum/src/json.rs
@@ -87,7 +87,7 @@ use serde::{de::DeserializeOwned, Serialize};
 /// let app = Router::new().route("/users/:id", get(get_user));
 /// # let _: Router = app;
 /// ```
-#[derive(Debug, Clone, Copy, Default)]
+#[derive(Debug, Clone, Copy, Default, PartialEq)]
 #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
 #[must_use]
 pub struct Json<T>(pub T);


jplatte:
If we do this, we should do it for all the extractors that just consist of a single generic field, including those in `axum-extra`.

I think your second test code snippet is actually better though, since it will print a simpler / more specific error message if the assert fails.
Is this sort of test code your only motivation for wanting to add this trait impl?

jkylling:
> If we do this, we should do it for all the extractors that just consist of a single generic field, including those in `axum-extra`.
> 
> I think your second test code snippet is actually better though, since it will print a simpler / more specific error message if the assert fails. Is this sort of test code your only motivation for wanting to add this trait impl?

Yes, test code is the only motivation. It's only an ergonomics improvement. I was writing a test for an endpoint returning a string first, and it was very natural to write `assert_eq!((StatusCode::OK, "Hello world".to_string()), my_handler().await);`. For a `Json` response, my initial approach before stopping to think was to do
```rust
let response = my_handler().await;
assert_eq!(StatusCode::OK, response.0);
assert_eq!(MyStruct, response.1.0);
```
which felt a bit ugly. Though, it's probably a good rule of thumb to pattern match on assignments from single field tuple structs.

jplatte:
Given the discussion above, I'll close. Feel free to open an issue or discussion if you have more to say on this.

PR #2769, opened 2024-06-05
Adding docs on using tokio task_local to share state

Gaelik-git:
I stumbled upon this [discussion](https://github.com/tokio-rs/axum/discussions/2631) for using tokio task_local to share state between handlers and middleware.

Is it okay to add some docs, would you prefer to have some test or example to show the behaviour ?

### DIFF
--- a/axum/src/lib.rs
+++ b/axum/src/lib.rs
@@ -295,7 +295,7 @@
 //!
 //! ## Using [tokio's `task_local` macro](https://docs.rs/tokio/1/tokio/macro.task_local.html):
 //!
-//! This allows to share state with IntoResponse implementations.
+//! This allows to share state with `IntoResponse` implementations.
 //!
 //! ```rust,no_run
 //! use axum::{
@@ -352,7 +352,6 @@
 //! let app: Router = Router::new()
 //!     .route("/", get(handler))
 //!     .route_layer(middleware::from_fn(auth));
-//!
 //! ```
 //!
 //! # Building integrations for axum
--- a/axum/src/lib.rs
+++ b/axum/src/lib.rs
@@ -293,6 +293,68 @@
 //! The downside to this approach is that it's a little more verbose than using
 //! [`State`] or extensions.
 //!
+//! ## Using [tokio's `task_local` macro](https://docs.rs/tokio/1/tokio/macro.task_local.html):
+//!
+//! This allows to share state with IntoResponse implementations.
+//!
+//! ```rust,no_run
+//! use axum::{
+//!     extract::Request,
+//!     http::{header, StatusCode},
+//!     middleware::{self, Next},
+//!     response::{IntoResponse, Response},
+//!     routing::get,
+//!     Router,
+//! };
+//! use tokio::task_local;
+//!
+//! #[derive(Clone)]
+//! struct CurrentUser {
+//!     name: String,
+//! }
+//! task_local! {
+//!     pub static USER: CurrentUser;
+//! }
+//!
+//! async fn auth(req: Request, next: Next) -> Result<Response, StatusCode> {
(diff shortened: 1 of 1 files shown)


axum/src/docs/middleware.md
```diff
@@ -535,6 +535,60 @@ let app = Router::new()
 automatically moved to response extensions. You need to manually do that for the
 extensions you need.
 
+State can also be passed from middleware to handlers using [tokio task_local]:
```
jplatte, line 538:
```suggestion
State can also be passed from middleware to handlers using
[tokio's `task_local` macro](https://docs.rs/tokio/1/tokio/macro.task_local.html):
```

axum/src/docs/middleware.md
```diff
@@ -535,6 +535,60 @@ let app = Router::new()
...
+    http::StatusCode,
+    middleware::{self, Next},
+    response::{IntoResponse, Response},
+    routing::get,
+    Router,
+};
+use tokio::task_local;
+
+#[derive(Clone)]
+struct CurrentUser {/* ... */}
```
jplatte, line 552:
```suggestion
struct CurrentUser {
    // ...
}
```

axum/src/docs/middleware.md
```diff
@@ -535,6 +535,60 @@ let app = Router::new()
...
+    let auth_header = req
+        .headers()
+        .get(http::header::AUTHORIZATION)
+        .and_then(|header| header.to_str().ok());
+
+    let auth_header = if let Some(auth_header) = auth_header {
+        auth_header
+    } else {
+        return Err(StatusCode::UNAUTHORIZED);
+    };
```
jplatte, line 568:
```suggestion
    let auth_header = req
        .headers()
        .get(http::header::AUTHORIZATION)
        .and_then(|header| header.to_str().ok())
        .ok_or(StatusCode::UNAUTHORIZED)?;
```

jplatte (commented):
Thanks for the PR! I think it makes sense to showcase this in the docs, but in addition to the small improvement suggestions below, there are two more general things:

- I think this documentation belongs in the main documentation, in the "Sharing state with handlers" section
- To me, the real important aspect about this is that unlike other state sharing mechanisms, this one actually also works to share state with `IntoResponse` impls; I think the docs should call this out

jplatte:
Hey @Gaelik-git, are you still interested in getting back to this?

Gaelik-git:
Yep sorry for the late response, I moved and modified the example for it to include the IntoReponse implementation
But my example does not make much sense I think, devs will still get the point I think but if you have other sugestion do not hesitate

axum/src/lib.rs
```diff
@@ -293,6 +293,68 @@
 //! The downside to this approach is that it's a little more verbose than using
 //! [`State`] or extensions.
 //!
+//! ## Using [tokio's `task_local` macro](https://docs.rs/tokio/1/tokio/macro.task_local.html):
+//!
+//! This allows to share state with IntoResponse implementations.
```
jplatte, line 298:
```suggestion
//! This allows to share state with `IntoResponse` implementations.
```

axum/src/lib.rs
```diff
@@ -293,6 +293,68 @@
...
+//! }
+//!
+//! async fn handler() -> UserResponse {
+//!     UserResponse
+//! }
+//!
+//! let app: Router = Router::new()
+//!     .route("/", get(handler))
+//!     .route_layer(middleware::from_fn(auth));
+//!
```
jplatte, line 355: The codeblock will look better without an empty line at the end.

```suggestion
```

jplatte (approved):
Thanks!

Gaelik-git:
glad I could contribute

PR #2777, opened 2024-06-09
Reduce extra allocation when constructing statically known body

cynecx:
## Motivation

Constructing empty bodies or bodies which are fully "available" (`String`, `Vec<u8>`, ...) go through an extra indirection (One additional `Box`) which isn't that efficient (but simplifies code). We can reduce this single allocation quite easily without much code complexity and effort, so I hope this won't add additional burden to the maintenance of the code. However, this will ultimately result in a layout size increase (On 64-bit archs this will go from 16 bytes to 40 bytes).

## Solution

Add some specialization, through an enum, for all the types that we can statically embed in the body.

### DIFF
--- a/axum-core/src/body.rs
+++ b/axum-core/src/body.rs
@@ -36,7 +36,7 @@ where
 
 /// The body type used in axum requests and responses.
 #[derive(Debug)]
-pub struct Body(BoxBody);
+pub struct Body(Kind);
 
 impl Body {
     /// Create a new `Body` that wraps another [`http_body::Body`].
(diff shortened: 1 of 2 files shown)


cynecx:
Very gentle ping :)

cynecx:
Another gentle ping.

jplatte:
Hey, sorry but I don't want to accept micro-optimizations that have potential downsides without prior discussion. axum is currently running in relatively low-maintenance mode and I want to avoid breaking or regressing anything.

PR #2778, opened 2024-06-09
docs: remove unnecessary link targets

mladedav:
## Motivation

CI is emitting warnings due to unnecessary links in docs due to `#[warn(rustdoc::redundant_explicit_links)]` which is on by default.

## Solution

Remove the unnecessary links. Each changed file is currently included from only a single place so this is ok.

## Alternatives

Suppress this warning.

### DIFF
--- a/axum/src/docs/extract.md
+++ b/axum/src/docs/extract.md
@@ -19,8 +19,7 @@ Types and traits for extracting data from requests.
 
 A handler function is an async function that takes any number of
 "extractors" as arguments. An extractor is a type that implements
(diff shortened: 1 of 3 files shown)


mladedav:
This will not actually fix the CI as noted [here](https://github.com/tokio-rs/axum/pull/2739#issuecomment-2156781770). But we might want to get rid of these warnings nonetheless either by doing this or suppressing the lint.

jplatte (approved):
Nice, but I can't merge because of the CI issue 😉

PR #2779, opened 2024-06-09
CI: pin nightly version for `cargo public-api-crates`

mladedav:
## Motivation

The CI is failing due to changes in `nightly` which `cargo-public-api-crates` is not ready for.

## Solution

Pin the `nightly` version to a prior version that is known to work until the tool is updated to handle current `nightly`.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -57,7 +57,9 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     # Pinned version due to failing `cargo-public-api-crates`.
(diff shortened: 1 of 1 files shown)


jplatte (approved):
Thanks for figuring this out! Looks like the relevant jobs are all passing now.

mladedav:
Finally :confetti_ball: 

I didn't look too closely into what the tool actually does, but hopefully when the 2024 edition lands, we can  switch to the private/public dependencies feature.

jplatte:
Well, likely not unfortunately: https://github.com/rust-lang/rust/issues/44663#issuecomment-2148249421

PR #2781, opened 2024-06-10
Allow querying of HTTP status codes for query and form rejections

htrefil:
## Motivation

`axum_extra::{query::QueryRejection, form::FormRejection}` are missing a method that would allow users to query the status code without converting it into a response via `IntoResponse` first.

This is useful when one wants to customize rejection responses.

## Solution

Add `pub fn status(&self) -> http::StatusCode` methods to both rejections.

### DIFF
--- a/axum-extra/src/extract/form.rs
+++ b/axum-extra/src/extract/form.rs
@@ -77,13 +77,24 @@ pub enum FormRejection {
     FailedToDeserializeForm(Error),
 }
 
+impl FormRejection {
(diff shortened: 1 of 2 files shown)


Turbo87 (approved):
This appears reasonable to me and matches the `status()` fn on the rejections in the `axum` package, which is automatically created by the macros in `axum-core`.

I think it would make sense to refactor these rejections to use the `axum-core` macros instead, but that is a larger refactoring and can be done in a follow-up pull request.

PR #2786, opened 2024-06-13
feat: allow newline runs in data

drbh:
This PR allows runs of `\n` chars to be included in an event stream data. Subsequently this will stop `\n` from being stripped from a message and includes it in the payload. 

This is helpful in a use case where the server needs to send text with multiple lines in a single event. 

I believe this still aligns with the SSE specification since I cannot find specifics around removing/including newline runs. Please let me know what you think of this change and if anything should be updated.

Thanks


ttys3:
`\n` is not allowed in the data. 

https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation

Issue #2768, opened 2024-06-04
Add a File type to axum-extra


joeydewaal:
- [x] I have looked for existing issues (including closed) about this

## Feature Request
Would it be useful to add a response type for a file in Axum? This response type would set headers like Content-Type, Content-Disposition and Content-Length. This could be a `File` type just like the `Html` and other types but one that is more generic. I would suppose that this `File` type would live in the `axum-extra` crate.

### Motivation

I think this would make Axum more user-friendly. Currently the way to respond with a file is through:

-  The `Html`,  `JavaScript`, `Css` and `Wasm` types.
    - This could be too specific for a user.
- Through the `Body` type and setting the right headers.
    - This is good but could be made easier.

### Proposal

I was thinking of something like this:
```rust
pub struct File {
    file_name: String,
    mime_type: Mime,
    source: Body,
}


impl<F, M, B> From<(F, M, B)> for File
where
    F: Into<String>,
    M: Into<Mime>,
    B: Into<Body>,
{
    fn from((file, mime, source): (F, M, B)) -> Self {
        Self {
            file_name: file.into(),
            mime_type: mime.into(),
            source: source.into(),
        }
    }
}

impl IntoResponse for File {
    fn into_response(self) -> Response {
        let mut headers = HeaderMap::new();

        headers.append(
            header::CONTENT_TYPE,
            self.mime_type.to_string().parse().unwrap(),
        );
        headers.append(
            header::CONTENT_DISPOSITION,
            format!("attachment; filename=\"{}\"", self.file_name)
                .parse()
                .unwrap(),
        );

        if let Some(content_length) = self.source.size_hint().exact() {
            headers.append(header::CONTENT_LENGTH, content_length.into());
        }
        (headers, self.source).into_response()
    }
}

// example 1
async fn file_handler1() -> impl IntoResponse {
    let bytes = [0; 4096].to_vec();
    File::from(("main.rs", mime::TEXT_PLAIN_UTF_8, bytes))
}

// example 2
async fn file_handler2() -> impl IntoResponse {
    use tokio::fs::File as TokioFile;

    let stream = TokioFile::open("./main.rs")
        .map_ok(|file| FramedRead::new(file, BytesCodec::new()))
        .try_flatten_stream();
    File::from(("main.rs", mime::TEXT_PLAIN_UTF_8, Body::from_stream(stream)))
}
```
[...]

PR #2789, opened 2024-06-17
Add an Attachment type to axum-extra

joeydewaal:
Refs #2768

## Motivation

When trying to send an attachment from Axum it includes,
* Setting a `Content-Disposition` header.
* Using the provided `Html`, `Css`, `JavaScript` or `Wasm` types or setting a `Content-Type` header.

 I think this could be made a bit easier.

## Solution

Add an `Attachment` response type to axum-extra. This type adds a `Content-Disposition` header and optionally a `Content-Type` header to the response.

example:
```rust
async fn cargo_toml() -> Result<Attachment<String>, (StatusCode, String)> {
    let file_contents = tokio::fs::read_to_string("Cargo.toml").await
        .map_err(|err| (StatusCode::NOT_FOUND, format!("File not found: {err}")))?;

    Ok(Attachment::new(file_contents)
        .filename("Cargo.toml")
        .content_type("text/x-toml"))
}
```

### DIFF
--- a/axum-extra/src/response/attachment.rs
+++ b/axum-extra/src/response/attachment.rs
@@ -5,7 +5,7 @@ use tracing::trace;
 /// A file attachment response type
 ///
 /// This type will set the `Content-Disposition` header for this response. In response a webbrowser
-/// will download the contents localy.
+/// will download the contents locally.
 ///
 /// Use the `filename` and `content_type` methods to set the filename or content-type of the
 /// attachment. If these values are not set they will not be sent.
@@ -34,7 +34,7 @@ use tracing::trace;
 /// use the [`Attachment`] type in a tuple.
 ///
 /// # Note
-/// If the content length is known and this header is manualy set to a different length hyper
+/// If the content length is known and this header is manually set to a different length hyper
 /// panics.
 ///
 /// ```rust
--- a/axum-extra/src/response/attachment.rs
+++ b/axum-extra/src/response/attachment.rs
@@ -66,7 +66,7 @@ impl<T: IntoResponse> Attachment<T> {
 
     /// Sets the content-type of the [`Attachment`]
     pub fn content_type<H: TryInto<HeaderValue>>(mut self, value: H) -> Self {
-        if let Some(content_type) = value.try_into().ok() {
+        if let Ok(content_type) = value.try_into() {
             self.content_type = Some(content_type);
         } else {
             trace!("Attachment content-type contains invalid characters");
--- a/axum-extra/src/response/attachment.rs
+++ b/axum-extra/src/response/attachment.rs
@@ -91,8 +91,7 @@ where
             bytes.extend_from_slice(filename.as_bytes());
             bytes.push(b'\"');
 
-            HeaderValue::from_bytes(&bytes)
-                .expect("This was a HeaderValue so this can not fail")
+            HeaderValue::from_bytes(&bytes).expect("This was a HeaderValue so this can not fail")
         } else {
             HeaderValue::from_static("attachment")
         };
--- a/axum-extra/src/response/attachment.rs
+++ b/axum-extra/src/response/attachment.rs
@@ -1,0 +1,112 @@
+use axum::response::IntoResponse;
+use http::{header, HeaderMap, HeaderValue};
+
+/// A file attachment response type
+///
+/// This type will set the `Content-Disposition` header for this response. In response a webbrowser
+/// will download the contents localy.
+///
+/// Use the `filename` and `content_type` methods to set the filename or content-type of the
+/// attachment. If these values are not set they will not be sent.
+///
+///
+/// # Example
+///
+/// ```rust
+///  use axum::{http::StatusCode, routing::get, Router};
+///  use axum_extra::response::Attachment;
+///
+///  async fn cargo_toml() -> Result<Attachment<String>, (StatusCode, String)> {
+///      let file_contents = tokio::fs::read_to_string("Cargo.toml")
+///          .await
+///          .map_err(|err| (StatusCode::NOT_FOUND, format!("File not found: {err}")))?;
+///      Ok(Attachment::new(file_contents)
+///          .filename("Cargo.toml")
+///          .content_type("text/x-toml"))
+///  }
+///
+///  let app = Router::new().route("/Cargo.toml", get(cargo_toml));
+///  let _: Router = app;
+/// ```
+///
+/// Hyper will set the `Content-Length` header if it knows the length. To manually set this header
+/// use the [`Attachment`] type in a tuple.
+///
+/// # Note
+/// If the content length is known and this header is manualy set to a different length hyper
+/// panics.
+///
+/// ```rust
+/// async fn with_content_length() -> impl IntoResponse {
+///     (
+///         [(header::CONTENT_LENGTH, 3)],
+///         Attachment::new([0, 0, 0])
+///             .filename("Cargo.toml")
+///             .content_type("text/x-toml"),
+///     )
+/// }
+/// ```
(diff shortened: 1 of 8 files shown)


axum-extra/src/response/attachment.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    /// Sets the filename of the [`Attachment`]
+    ///
+    /// This updates the `Content-Disposition` header to add a filename.
+    pub fn filename<H: TryInto<HeaderValue>>(mut self, value: H) -> Self {
+        if let Some(filename) = value.try_into().ok() {
+            self.filename = Some(filename);
+        } else {
+            trace!("Attachment filename contains invalid characters");
+        }
```
jplatte, line 75: If calling this fn multiple times, I think it makes more sense if the last call fails, that the field is `None`. Wdyt?

```suggestion
        self.filename = if let Ok(filename) = value.try_into() {
            Some(filename)
        } else {
            trace!("Attachment filename contains invalid characters");
            None
        };
```

axum-extra/src/response/attachment.rs
```diff
@@ -0,0 +1,119 @@
...
+            let content_disposition = HeaderValue::from_bytes(&bytes)
+                .expect("This was a HeaderValue so this can not fail");
+
+            headers.append(header::CONTENT_DISPOSITION, content_disposition);
+        } else {
+            headers.append(
+                header::CONTENT_DISPOSITION,
+                HeaderValue::from_static("attachment"),
+            );
+        }
```
jplatte, line 115: Can be simplified a bit:

```suggestion
        let content_disposition = if let Some(filename) = self.filename {
            let mut bytes = b"attachment; filename=\"".to_vec();
            bytes.extend_from_slice(filename.as_bytes());
            bytes.push(b'\"');

            HeaderValue::from_bytes(&bytes)
                .expect("This was a HeaderValue so this can not fail")
        } else {
            HeaderValue::from_static("attachment")
        };

        headers.append(header::CONTENT_DISPOSITION, content_disposition)
```

axum-extra/src/response/mod.rs
```diff
@@ -3,13 +3,21 @@
 #[cfg(feature = "erased-json")]
 mod erased_json;
 
+#[cfg(feature = "attachment")]
+mod attachment;
+
+
```
jplatte, line 9: One newline too extra, wondering why I'm not seeing rustfmt complain in CI.

jplatte (commented):
Thanks for the PR! A few small suggestions below.

jplatte:
CI failure seems unrelated, if you have an idea on how to fix it, a separate PR would be appreciated. Otherwise I'll try to find some time to fix it soon.

axum-extra/src/response/attachment.rs
joeydewaal: Yes, that seems fair. That makes this function more predictable.

axum-extra/src/response/attachment.rs
```diff
@@ -0,0 +1,119 @@
...
+/// ```rust
+/// async fn with_content_length() -> impl IntoResponse {
+///     (
+///         [(header::CONTENT_LENGTH, 3)],
+///         Attachment::new([0, 0, 0])
+///             .filename("Cargo.toml")
+///             .content_type("text/x-toml"),
+///     )
+/// }
+/// ```
```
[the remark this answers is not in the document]
joeydewaal: My docs are too specific. I'll change it to a note that says something like: When using this type with hyper, hyper will set a `Content-Type` if it is know.

joeydewaal:
Thanks for your review and changes, I changed the code. I don't know why CI is failing if I have some time I'll try and fix it.

jplatte:
I looked and the rustdoc failure is really just 2 spaces of indentation missing. Since I can't approve / merge my own PRs, do you mind just adding that to this? It's such a trivial change that I don't mind it being rolled up into another PR.

joeydewaal:
Ofcourse!

jplatte:
Okay, looks like there's more to it. Sorry, I'll go fix it myself soon-ish (hope David will be around soon-ish to merge).

joeydewaal:
Almost there, it's complaining about `MakeErasedRouter` is never constructed. Is it okay is I add an `#[allow(dead_code)]` to that type?

jplatte:
Sure, I suspect that there's more that should be done eventually, but to unblock CI that seems like the best solution.

axum-extra/src/routing/typed.rs
```diff
@@ -84,13 +84,8 @@ use serde::Serialize;
...
 ///
 /// - A `TypedPath` implementation.
-/// - A [`FromRequest`] implementation compatible with [`RouterExt::typed_get`],
-/// [`RouterExt::typed_post`], etc. This implementation uses [`Path`] and thus your struct must
-/// also implement [`serde::Deserialize`], unless it's a unit struct.
-/// - A [`Display`] implementation that interpolates the captures. This can be used to, among other
-/// things, create links to known paths and have them verified statically. Note that the
-/// [`Display`] implementation for each field must return something that's compatible with its
-/// [`Deserialize`] implementation.
+/// - A [`FromRequest`] implementation compatible with [`RouterExt::typed_get`], [`RouterExt::typed_post`], etc. This implementation uses [`Path`] and thus your struct must also implement [`serde::Deserialize`], unless it's a unit struct.
```
jplatte, line 87: Please could you re-wrap these long lines? There just have to be extra spaces in front to not trigger the rustdoc warning / error.

```suggestion
/// - A [`FromRequest`] implementation compatible with [`RouterExt::typed_get`],
///   [`RouterExt::typed_post`], etc. This implementation uses [`Path`] and thus your struct must
///   also implement [`serde::Deserialize`], unless it's a unit struct.
```

axum-extra/src/response/attachment.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /// Sets the filename of the [`Attachment`].
+    ///
+    /// This updates the `Content-Disposition` header to add a filename.
+    pub fn filename<H: TryInto<HeaderValue>>(mut self, value: H) -> Self {
+        self.filename = if let Ok(filename) = value.try_into() {
+            Some(filename)
+        } else {
+            trace!("Attachment filename contains invalid characters");
```
jplatte, line 61: This seems like it should never be hit, so it should be prominent in logs.

```suggestion
            error!("Attachment filename contains invalid characters");
```

axum-extra/src/response/attachment.rs
```diff
@@ -0,0 +1,103 @@
...
+        };
+        self
+    }
+
+    /// Sets the content-type of the [`Attachment`]
+    pub fn content_type<H: TryInto<HeaderValue>>(mut self, value: H) -> Self {
+        if let Ok(content_type) = value.try_into() {
+            self.content_type = Some(content_type);
+        } else {
+            trace!("Attachment content-type contains invalid characters");
```
jplatte, line 72: Same here.

axum-extra/src/response/attachment.rs
```diff
@@ -0,0 +1,103 @@
...
+///          .content_type("text/x-toml"))
+///  }
+///
+///  let app = Router::new().route("/Cargo.toml", get(cargo_toml));
+///  let _: Router = app;
+/// ```
+///
+/// # Note
+///
+/// When using this type in Axum with hyper, hyper will set the `Content-Length` if it is known.
```
jplatte, line 35:
```suggestion
/// If you use axum with hyper, hyper will set the `Content-Length` if it is known.
```

jplatte (approved):
Thank you!

PR #2790, opened 2024-06-18
add `has_routes` function and docu to `Router`

cbeck88:
## Motivation

The motivation is to be able to avoid panics in generic code that wants to call `route_layer`.

For example, you might have a generic function that takes a `Router` configured by an application (`/api/v1/my_route`, `/api/v1/my_route2`,...) and adds a bunch of additional configuration that is common to your applications, like an http metrics layer using `route_layer`, followed by infrastructure or diagnostic routes (`/health`, `/ready`, ...) that you do not need metrics for.

`route_layer` currently panics if you call it on `Router::new()`. But in some cases that is legitimate, for example, you may have an application that performs important periodic tasks, and needs to have `/health` and `/ready` to exist in your k8s infrastructure, but doesn't actualy have any HTTP API to expose. So then it would be very logical to pass `Router::new()` to your generic function.

## Solution

With this non-breaking change, the helper function would be able to call `is_empty()` first on the argument, and skip the (panicking) call to `route_layer` if the router is empty. This is a bit nicer than the workaround of adding dummy routes to the `Router::new()` just to avoid a `Router` panic.

Edit: After discussion, changed to a `has_routes` function instead of `is_empty` for a more clear semantic

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **change:** Avoid cloning `Arc` during deserialization of `Path`
 - **added:** `axum::serve::Serve::tcp_nodelay` and `axum::serve::WithGracefulShutdown::tcp_nodelay` ([#2653])
-- **added:** `Router::has_routes` function added ([#2790])
+- **added:** `Router::has_routes` function ([#2790])
 
 [#2653]: https://github.com/tokio-rs/axum/pull/2653
 [#2790]: https://github.com/tokio-rs/axum/pull/2790
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -9,8 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **change:** Avoid cloning `Arc` during deserialization of `Path`
 - **added:** `axum::serve::Serve::tcp_nodelay` and `axum::serve::WithGracefulShutdown::tcp_nodelay` ([#2653])
+- **added:** `Router::has_routes` function added ([#2790])
 
 [#2653]: https://github.com/tokio-rs/axum/pull/2653
+[#2790]: https://github.com/tokio-rs/axum/pull/2790
 
 # 0.7.5 (24. March, 2024)
 
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -307,6 +307,12 @@ where
         })
     }
 
+    /// True if the router currently has at least one route added.
+    #[track_caller]
+    pub fn has_routes(&self) -> bool {
+        self.inner.path_router.has_routes()
+    }
+
     #[track_caller]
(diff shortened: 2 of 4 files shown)


cbeck88:
CI failures seem not related to my code, i guess i can try rebasing

jplatte:
Yes, a rebase should fix it.

However, I'm a bit skeptical about this addition. Your use case is certainly valid, but maybe it would make more sense to have a method like `route_layer_allow_noop` or whatever. Actually we were also considering a separate method on `Router` that you would call to disable a non-critical check in the context of the matchit upgrade, maybe the same pattern would make sense here.

cc @mladedav I'm curious whether you have an opinion on this

mladedav:
I'm not really sure about this. I think both `route_layer_no_panic` and `is_empty` seem reasonable to me. The latter may also help in a few other scenarios I guess if there are similar panics elsewhere. But I think either is fine.

I assume we don't want just outright allow `route_layer` on an empty `Router` because a lot of people starting with axum would try that and be surprised their middlewares are not applied.

I'm not a fan of `strict`/`lenient` config inside the router, I see the one in matchit context as a temporary solution so that people who upgrade are alerted that their captures won't work and they need to migrate, but I assumed we'd remove that in next semi-major release and just allow the paths in question.

cbeck88:
One thing that occurs to me looking at this again is, a router can hold more things than just routes so maybe it's potentially confusing to define 'is_empty' this way.

For example, if you have 'Router::new().layer(...)' this function would say it's empty, and that's the most helpful thing for the use case scenario I mentioned, but for some other user in another scenario they might expect it to say it's not empty since it has some content now.

Happy to rework this as you like, lmk what you think. Thanks

jplatte:
One more option: `Router::has_routes` as the inverse of `is_empty`. Maybe a bit more natural?

```rust
if router.has_routes() {
    router = router.route_layer(layer);
}
```

axum/src/routing/mod.rs
```diff
@@ -307,6 +307,12 @@ where
         })
     }
 
+    /// True if the router currently has at least one route added.
+    #[track_caller]
```
jplatte, line 311: This attribute only makes sense on functions that can panic.


```suggestion
```

axum/src/routing/path_router.rs
```diff
@@ -290,6 +290,11 @@ where
         }
     }
 
+    #[track_caller]
```
jplatte, line 293:
```suggestion
```

axum/CHANGELOG.md
```diff
@@ -9,8 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **change:** Avoid cloning `Arc` during deserialization of `Path`
 - **added:** `axum::serve::Serve::tcp_nodelay` and `axum::serve::WithGracefulShutdown::tcp_nodelay` ([#2653])
+- **added:** `Router::has_routes` function added ([#2790])
```
jplatte, line 12:
```suggestion
- **added:** `Router::has_routes` function ([#2790])
```

jplatte (approved):

cbeck88:
thank you!

PR #2792, opened 2024-06-19
Update TLS examples to use better HTTP->HTTPS redirect

jamesqh:
I tried to implement the redirect from the TLS examples in my project, and it didn't work - I like to use port 6789 when I'm testing things locally and this caused big problems. So I thought I'd try fixing it and filing a PR.

tl;dr: The redirect showcased in the TLS examples is very fragile, I am proposing a more robust alternative.

## Motivation:

The HTTP->HTTPS redirect showcased in these examples changes the port with a simple find-and-replace over the entire authority, including the hostname. This works well in the median case but:

* Custom HTTPS ports with default HTTP ports fail to redirect properly, as the HTTP port is not present in the string: `http://example.com` -> `https://example.com` (not `https://example.com:4443` or whatever port you're using.) This is the one that bit me!

* IP addresses can be mangled: `http://80.80.80.80` -> `https://443.443.443.443`, `http://[::80]` -> `https://[::443]`

* As, I think worst of all, can hostnames, like this real and very cool website: `http://io808.com` -> `https://io4438.com` !

## Solution:

I've written an alternative based on parsing into an Authority; this type's `.port()` should guarantee with a `None` return that the authority string is a bare hostname/IP, or with `Some(port)` that the string ends with `:{port}` and so we can safely strip that suffix. Adding new port only conditionally would be easy, but browsers tend to be smart about eliding unnecessary `:443`s, so it seemed an overcomplication?

I also changed the args, since this redirect solution doesn't actually need to know the HTTP port, and since clippy's suggestion to use an `&str` seemed reasonable.

I'm not very expert at Rust, so hopefully my solution doesn't have any glaring problems! It may be better to simply use the `url` crate, as was proposed (and rejected) previously, but it was more fun for me to write this.

### DIFF
--- a/examples/tls-graceful-shutdown/src/main.rs
+++ b/examples/tls-graceful-shutdown/src/main.rs
@@ -6,7 +6,7 @@
 
 use axum::{
     handler::HandlerWithoutStateExt,
-    http::{StatusCode, Uri},
+    http::{uri::Authority, StatusCode, Uri},
     response::Redirect,
     routing::get,
     BoxError, Router,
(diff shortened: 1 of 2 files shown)


Turbo87 (approved):
while I'm not a huge fan of the `unwrap()` calls, this at least looks less error prone as before. IMHO this seems good to merge.

Issue #2796, opened 2024-06-21
Remove `heck` from axum-macros


vsuryamurthy:
Not super important, but I did notice that `heck` is not used anywhere withing axum-macros. If I am correct, it should be removed probably?

jplatte:
I can't find any use of it either, want to make a PR?

vsuryamurthy:
Sure can!

PR #2798, opened 2024-06-21
Remove unused dependency heck from axum-macros

vsuryamurthy:
Resolves https://github.com/tokio-rs/axum/issues/2796


jplatte (approved):
Thank you!

PR #2801, opened 2024-06-22
Update extract doc with optional extractors present in axum-extra

vsuryamurthy:
This PR involves updating the documentation of extract in order to point the user to more optional extractors available in `axum-extra`.

### DIFF
--- a/axum/src/docs/extract.md
+++ b/axum/src/docs/extract.md
@@ -280,7 +280,7 @@ let app = Router::new().route("/users", post(create_user));
 # let _: Router = app;
 ```
 
-An alternate option is to make use of the optional extractors in [axum-extra] that return `None`
+Another option is to make use of the optional extractors in [axum-extra] that return `None`
 if there are no query parameters in the request or return `Some(T)` if deserialization was successful.
 
 # Customizing extractor responses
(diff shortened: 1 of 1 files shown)


axum/src/docs/extract.md
```diff
@@ -280,6 +280,9 @@ let app = Router::new().route("/users", post(create_user));
 # let _: Router = app;
 ```
 
+An alternate option is to make use of the optional extractors in [axum-extra] that return `None`
```
jplatte, line 283:
```suggestion
Another option is to make use of the optional extractors in [axum-extra] that return `None`
```

axum/src/docs/extract.md
```diff
@@ -280,6 +280,9 @@ let app = Router::new().route("/users", post(create_user));
...
+if there are no query parameters in the request or return `Some(T)` if deserialization was successful.
```
jplatte, line 284:
```suggestion
only for some rejections from their "base" extractor and reject the request otherwise.
```

jplatte (commented):
The new doc sentence seems to specify `OptionalQuery` behavior specifically, I've left some wording suggestions but feel free to make different suggestions yourself.

jplatte (approved):
Sorry for the delay, updated text looks good!

PR #2808, opened 2024-06-26
Fix a few typos in docs and comments

cpick:
## Motivation

These are mostly small issues where one word is accidentally used in place of another.

## Solution

Removed unnecessary words, replaced incorrect words, fixed link.


jplatte (approved):
Thank you!

PR #2812, opened 2024-06-29
[Reqwest example]: Include headers into response

Valerioageno:
## Topic 

Example fix 

## Motivation

While working on a reverse proxy with reqwest I think I noticed that the headers are not actually passed from the request to the response (I'm new to axum so I might be wrong).

Shouldn't be like this?

## Solution

We update the `response_builder`'s `HeadersMap`

### DIFF
--- a/examples/reqwest-response/src/main.rs
+++ b/examples/reqwest-response/src/main.rs
@@ -62,17 +62,12 @@ async fn proxy_via_reqwest(State(client): State<Client>) -> Response {
 
     let mut response_builder = Response::builder().status(reqwest_response.status().as_u16());
 
-    {
-        let headers = response_builder.headers_mut().unwrap();
+    let headers = response_builder.headers_mut().unwrap();
 
-        reqwest_response
-            .headers()
-            .into_iter()
-            .for_each(|(name, value)| {
-                let name = HeaderName::from_bytes(name.as_ref()).unwrap();
-                let value = HeaderValue::from_bytes(value.as_ref()).unwrap();
-                headers.insert(name, value);
-            });
+    for (name, value) in reqwest_response.headers() {
+        let name = HeaderName::from_bytes(name.as_ref()).unwrap();
+        let value = HeaderValue::from_bytes(value.as_ref()).unwrap();
+        headers.insert(name, value);
     }
 
     response_builder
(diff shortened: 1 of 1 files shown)


examples/reqwest-response/src/main.rs
```diff
@@ -60,15 +60,20 @@ async fn proxy_via_reqwest(State(client): State<Client>) -> Response {
...
-        let value = HeaderValue::from_bytes(value.as_ref()).unwrap();
-        (name, value)
-    }));
+    {
+        let headers = response_builder.headers_mut().unwrap();
+
+        reqwest_response
+            .headers()
+            .into_iter()
+            .for_each(|(name, value)| {
```
mladedav, line 71: Nit, if you're going to iterate like this, it's more idiomatic to use a simple `for`. The docs for `for_each` itself mention it too.

Valerioageno: Nice catch. Thanks. Updated

mladedav (approved):
Once #2843 lands the CI will hopefully pass. LGTM

Valerioageno:
Hey @mladedav, thanks for your approval. I think I need one more since the rebase

mladedav (approved):
I'm not really an approver, so although I can click approve, it doesn't really have any value other than informational that _I_ think it's ok and that _someone_ has seen it. But merging will have to wait until an actual maintainer gets to it.

jplatte (commented):
I also think this was broken before, but instead of using a completely different approach (that I think handles mutliple header values for the same header name differently), could you just add `*response_builder.headers_mut().unwrap() = headers;` to after the `extend` call that was already there, and keep the `HeaderMap::with_capacity` creation and such?

Valerioageno:
I haven't thought about it! Thanks for spotting. Actually cleaner solution while keeping the `HeaderMap::with_capacity` initialization.

examples/reqwest-response/src/main.rs
```diff
@@ -60,16 +60,19 @@ async fn proxy_via_reqwest(State(client): State<Client>) -> Response {
         }
     };
 
-    let response_builder = Response::builder().status(reqwest_response.status().as_u16());
+    let mut response_builder = Response::builder().status(reqwest_response.status().as_u16());
 
     // Here the mapping of headers is required due to reqwest and axum differ on the http crate versions
     let mut headers = HeaderMap::with_capacity(reqwest_response.headers().len());
+
```
jplatte, line 67:
```suggestion
```

jplatte (approved):
Thanks!

PR #2825, opened 2024-07-09
chore: fix gRPC multiplex example

### LINKED ISSUE #2325: Update rest-grpc-multiplex example to hyper 1.0
- [ x] I have looked for existing issues (including closed) about this

### Motivation
Hyper recently moved to 1.0 and the `hyper::Server` struct is no longer available.

### Proposal
Migrate the code in `examples/rest-grpc-multiplex`

### Alternatives
One can look into `examples/hyper-1-0` and figure it out, but it is not as easy.

### LINKED ISSUE #2736: Update gRPC multixplex example for axum 0.7 and hyper 1

## Feature Request
Update example `rest-grpc-multiplex`

### Motivation

Currently, the example code in https://github.com/tokio-rs/axum/tree/main/examples/rest-grpc-multiplex is commented out
and marked as TODO. 

The code in the example is not really straight forward for beginners and can be a good starting point for learning exercises.

### Proposal

Update the code to be compatible with axum 0.7 and hyper 1.

abhiaagarwal:
## Motivation

Updates the gRPC example with the new release of tonic that uses axum 0.7 and hyper 1. Closes #2736.

## Solution

Owing to `tower::steer`, I was able to remove the entire multiplex service in `multiplex_service.rs` file. Additionally, tonic thankfully gives us a method to covert its server into axum routes, so we avoid all type inference issues (since they are the same types!)

It should be possible to merge the two routers together directly, but I haven't investigated how to make routing possible. I'm not also sure what will happen to the middleware associated with each router. I chose to keep them separate.


spencerbart:
Using TraceLayer::new_for_grpc() doesn't work when multiplexing this way but works when just using tonic without multiplexing.

```Rust
let layer = tower::ServiceBuilder::new()
    .layer(TraceLayer::new_for_grpc())
    .into_inner();
```

Here's an example

https://github.com/spencerbart/cards

maxnachlinger:
> Scratch all of that. I just needed to put `.layer(TraceLayer::new_for_grpc())` after I call `.into_router()` on tonic's Server::builder().

@spencerbart trivial point, but you can add your TraceLayer before calling `into_router()` too :) 
```rs
    Server::builder()
        .layer(TraceLayer::new_for_grpc()) // you can add your layer here too
        .add_service(your_awesome_service)
        .into_router()
```

spencerbart:
> > Scratch all of that. I just needed to put `.layer(TraceLayer::new_for_grpc())` after I call `.into_router()` on tonic's Server::builder().
> 
> @spencerbart trivial point, but you can add your TraceLayer before calling `into_router()` too :)
> 
> ```rust
>     Server::builder()
>         .layer(TraceLayer::new_for_grpc()) // <--
>         .add_service(your_awesome_service)
>         .into_router()
> ```
@maxnachlinger That's what I had before and it didn't work.

For some reason calling `.layer(TraceLayer::new_for_grpc())` before `.into_router()` doesn't log the traces. You have to create that layer after you call `.into_router()`.

maxnachlinger:
@spencerbart hah, sorry for the churn man :)

abhiaagarwal:
Yep, I agree that this is very much a minimal POC, and the actual interaction between the grpc and rest server and its associated middleware isn't really that great...

In an ideal world, we should be able to merge the routers together and route based on Content-type, which would needs its own middleware. I think it's possible but I haven't investigated it.

brocaar:
I have been playing with this, but I have not (yet) solved how to work with layers using this example. As you already commented, when using `Server::builder().layer(...).into_router()`, it looks like the layer is not in the returned routes. On the other hand when using `Server::builder().into_router().layer(...)`, I get type errors (e.g. when using the `GrpcWebLayer`, https://docs.rs/tonic-web/latest/tonic_web/).

daheige:
> Yep, I agree that this is very much a minimal POC, and the actual interaction between the grpc and rest server and its associated middleware isn't really that great...
> 
> In an ideal world, we should be able to merge the routers together and route based on Content-type, which would needs its own middleware. I think it's possible but I haven't investigated it.

You're absolutely right. Thanks for the pull that allowed me to have the grpc service and http service run on the same port in tonic 0.12.1 release. Here is the crates package I used：
```toml
tonic = "0.12.1"
prost = "0.13.1"
tokio = {version = "1.38.0",features = ["full"]}
# note: Must be the same as the tonic version
tonic-reflection = "0.12.1"
# note: Must be the same as the tonic version
tonic-build = "0.12.1"
```

examples/rest-grpc-multiplex/Cargo.toml
```diff
@@ -5,16 +5,17 @@ edition = "2021"
 publish = false
 
 [dependencies]
-axum = { path = "../../axum" }
+# needs to match tonic
+axum = { version = "0.7.5" }
```
jplatte, line 9: I wouldn't want examples within this repo to pull in a crates.io release of axum.

I'm considering moving examples to a separate repo anyways though, see #2942 for discussion.

jplatte (commented):
Hey, sorry for letting this linger for so long!

I know next to nothing about GRPC, so it's hard for me to review this. @mladedav how about you?

daheige:
> Hey, sorry for letting this linger for so long!
> 
> I know next to nothing about GRPC, so it's hard for me to review this. @mladedav how about you?

I used the solution mentioned above to merge multiple services into one service. This solution is indeed feasible. See the project link below for details.
https://github.com/daheige/rs-rpc/blob/main/src/multiplex_service.rs#L112

examples/rest-grpc-multiplex/src/main.rs
```diff
@@ -4,84 +4,90 @@
...
-//     // combine them into one service
-//     let service = MultiplexService::new(rest, grpc);
+    // combine them into one service
+    let service = Steer::new(vec![rest, grpc], |req: &Request, _services: &[_]| {
+        if req
+            .headers()
+            .get(CONTENT_TYPE)
+            .map(|content_type| content_type.as_bytes())
+            .filter(|content_type| content_type.starts_with(b"application/grpc"))
+            .is_some()
```
mladedav, line 81: Nit, you could do just `.and_then(|content_type| content_type.as_bytes().starts_with(b"...").then_some(1)).unwrap_or(0)` without an explicit condition instead.

(Although now I see you have just copied this from the deleted file)

mladedav (commented):
I did use tonic before, but not like this so I still want to take a closer look to some of the parts used here, but I'd say this looks good.

I want to also double check if we can't make it work with a different versions of Axum between us and tonic. If we can't I'd vote to grant an exception here to depend on crates.io especially if we want to move it soon afterwards anyway.

abhiaagarwal:
> I did use tonic before, but not like this so I still want to take a closer look to some of the parts used here, but I'd say this looks good.
> 
> I want to also double check if we can't make it work with a different versions of Axum between us and tonic. If we can't I'd vote to grant an exception here to depend on crates.io especially if we want to move it soon afterwards anyway.

I could probably do a cargo patch to resolve this. Lemme give it a shot

mladedav:
If you mean doing `[patch.crates-io]` inside `Cargo.toml`, then I don't think we want to do that since once we make breaking changes, `tonic` in the example will fail to compile.

I see two options to reconcile different `axum` versions in the exmaple
- You can map the response type of `tonic`'s router to our response type.
```
    let grpc = Router::new().nest_service(
        "/",
        grpc.into_service()
            .map_response(|response| response.map(Body::new)),
    );
```
Then you can even just merge the two routers with `let router = rest.merge(grpc);` and you don't need `Steer` and `Sharded` at all. This works only if the two services don't use the same routes though. I think that's actually better than having completely different behavior based on whether `content-type` starts with `application/grpc`, but otherwise this seems like another instance of "I want to route based on something other than path/HTTP method". This compiles but does not run both routers define a route for `/`.

- You can create a layer between `Steer` and the two different routers so that it is happy with the types.
```
#[derive(Clone)]
enum MultiplexedService {
    Axum(axum::Router),
    Tonic(tonic::service::AxumRouter),
}

impl Service<Request<Body>> for MultiplexedService {
...
    fn call(&mut self, req: Request<Body>) -> Self::Future {
        match self {
            MultiplexedService::Axum(router) => Box::pin(router.call(req)),
            MultiplexedService::Tonic(router) => {
                let future = router.call(req);
                Box::pin(async {
                    let response = future.await;
                    response.map(|response| response.map(Body::new))
                })
            }
        }
    }
}
```

Also, it seems some of the methods from `tonic` you're using are already deprecated, could you check that please?

Turbo87:
thanks for your work, @abhiaagarwal!

in https://github.com/tokio-rs/axum/pull/3105 it was decided to remove this example though and potentially keep it in a different and/or dedicated repository. since the example was removed I guess we can close this pull request.

abhiaagarwal:
@Turbo87 all good!

Issue #2826, opened 2024-07-09
`FromRef` triggers `clippy::clone_of_ref_ptr`


AlphaKeks:
- [x] I have looked for existing issues (including closed) about this

### Version

0.7.5

### Platform

Linux 6.6.34 #1-NixOS SMP PREEMPT_DYNAMIC Sun Jun 16 11:47:49 UTC 2024 x86_64 GNU/Linux

### Description

When deriving [`FromRef`](https://docs.rs/axum/0.7.5/axum/extract/trait.FromRef.html), clippy will trigger [`clippy::clone_on_ref_ptr`](https://rust-lang.github.io/rust-clippy/master/#/clone_on_ref_ptr). This lint is allow-by-default, which is probably why nobody reported it yet.

You can reproduce the issue by adding `axum@0.7.5` as a dependency with the `macros` feature enabled, and writing the following code:
```rust
use std::sync::Arc;

use axum::extract::FromRef;

#[derive(Clone, FromRef)]
pub struct MyState {
	substate: Arc<MySubstate>,
}

pub struct MySubstate {}
```

And then running clippy with the offending lint set to `warn` (or higher):
```sh
$ cargo clippy -- -Wclippy::clone_on_ref_ptr

warning: using `.clone()` on a ref-counted pointer
 --> src/lib.rs:7:12
  |
7 |     substate: Arc<MySubstate>,
  |               ^^^ help: try: `Arc::<MySubstate>::clone(&substate: Arc)`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_ref_ptr
  = note: requested on the command line with `-W clippy::clone-on-ref-ptr`
```

I would assume adding a `clippy::clone_on_ref_ptr` to the `#[allow(…)]` attribute on [this line](https://docs.rs/axum-macros/0.4.1/src/axum_macros/from_ref.rs.html#57) would fix the problem. I can open a PR if that's a fine solution.

jplatte:
> I would assume adding a `clippy::clone_on_ref_ptr` to the `#[allow(…)]` attribute on [this line](https://docs.rs/axum-macros/0.4.1/src/axum_macros/from_ref.rs.html#57) would fix the problem. I can open a PR if that's a fine solution.

Sounds good!

PR #2827, opened 2024-07-09
Fix clippy warning on `FromRef` derive macro

AlphaKeks:
## Motivation

Fixes a clippy warning caused by the `FromRef` derive macro.

See #2826.

## Solution

Add the lint to the `#[allow(…)]` attribute above the generated function that triggers it.

### DIFF
--- a/axum-macros/src/from_ref.rs
+++ b/axum-macros/src/from_ref.rs
@@ -54,7 +54,7 @@ fn expand_field(state: &Ident, idx: usize, field: &Field) -> TokenStream {
     };
 
     quote_spanned! {span=>
-        #[allow(clippy::clone_on_copy)]
+        #[allow(clippy::clone_on_copy, clippy::clone_on_ref_ptr)]
         impl ::axum::extract::FromRef<#state> for #field_ty {
             fn from_ref(state: &#state) -> Self {
                 #body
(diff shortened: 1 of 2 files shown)


AlphaKeks:
The test is failing because of [the latest serde release](https://github.com/serde-rs/serde/releases/tag/v1.0.204) where they leverage the new `#[diagnostic::on_unimplemented]` feature to give better error messages.

jplatte:
Would you mind running `TRYBUILD=overwrite cargo test` from `axum-macros` to fix this?

AlphaKeks:
It seems like CI is running a very old version of nightly (from ~4 months ago). Running the tests on a newer release (from the past few days) results in _a lot_ more failing tests, not only in `axum-macros`. I could fix just this particular case by Switching to the exact same release that CI is using worked, but I think it's noteworthy that these tests and nightly CI should be updated at some point.

jplatte:
Yeah, I tried to bump nightly in #2805, I guess locally the trybuild tests generated different output from CI because of lockfile differences. Also David isn't currently around much to approve my PRs.

If you care, I would be happy to merge a PR that bumps nightly, and passes CI unlike mine 😄

jplatte (approved):
💯

PR #2841, opened 2024-07-23
Upgrade tokio-tungstenite to 0.23

hatchan:
## Motivation

The latest version of tokio-tungstenite updates both tungstenite and rustls.

## Solution

Upgrade tokio-tungstenite to 0.23.

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -10,11 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **change:** Avoid cloning `Arc` during deserialization of `Path`
 - **added:** `axum::serve::Serve::tcp_nodelay` and `axum::serve::WithGracefulShutdown::tcp_nodelay` ([#2653])
 - **added:** `Router::has_routes` function ([#2790])
-- **change:** Update tokio-tungstenite to 0.23 ([TODO])
+- **change:** Update tokio-tungstenite to 0.23 ([#2841])
 
 [#2653]: https://github.com/tokio-rs/axum/pull/2653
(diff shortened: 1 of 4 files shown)


hatchan:
I've also replaced `clippy::mismatched_target_os` with `unexpected_cfgs` as that was blocking the CI. `clippy::mismatched_target_os` seems to have been removed from Rust 1.81 and this is their recommendation. 

```
error: lint `clippy::mismatched_target_os` has been removed: this lint has been replaced by `unexpected_cfgs`
  --> axum-macros/src/lib.rs:18:5
   |
18 |     clippy::mismatched_target_os,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D renamed-and-removed-lints` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(renamed_and_removed_lints)]`
```

Let me know if this should be in a different PR, and/or if it requires a changelog entry.

hatchan:
I've also solved/ignored some other clippies, see previous 2 commits.

jplatte:
Only things that matter for users of the library should be mentioned in the changelog, this is not the case for clippy fixes usually (unless public API is changes in response to a clippy lint). Having those fixes as a separate PR would be nice though.

hatchan:
> Only things that matter for users of the library should be mentioned in the changelog, this is not the case for clippy fixes usually (unless public API is changes in response to a clippy lint). Having those fixes as a separate PR would be nice though.

I can separate it in a different PR, but it should work now :partying_face:

hatchan:
Once #2843 is merged I will rebase this PR on `main` and mark this PR as ready for review.

jplatte (approved):

PR #2849, opened 2024-07-26
Macro for JsonDeserializer extractor to avoid boilerplate code

manuelgdlvh:
Macro for the JsonDeserializer extractor to avoid boilerplate code in most situations where the handler returns a Response or its derivatives, and returns early in case of failure.

## Motivation

In most situations, the controller layer is responsible for managing the result of the service layer and returning a specific response (not an expected result). Given this scenario, using the JsonDeserializer extractor, we should handle both when the payload is correct and when it is not. 
This macro returns the expected payload, and if not, it returns early a Response structure, avoiding duplicate code.

## Solution

A macro that returns the payload if everything went well, otherwise it returns the error as a Response structure that implements IntoResponse.

### DIFF
--- a/axum-extra/src/extract/json_deserializer.rs
+++ b/axum-extra/src/extract/json_deserializer.rs
@@ -209,6 +209,16 @@ fn json_content_type(headers: &HeaderMap) -> bool {
     is_json_content_type
 }
 
+#[macro_export]
+macro_rules! deserialize_or_bail {
+      ($payload:expr) => {
+        match $payload.deserialize() {
+            Ok(data) => data,
+            Err(e) => return e.into_response(),
+        }
+    };
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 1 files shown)


jplatte:
I don't think this makes a lot of sense. Have you considered changing your handler to return <code><a href="https://docs.rs/axum/latest/axum/response/type.Result.html">axum::response::Result</a>&lt;JsonDeserializer&lt;T&gt;&gt;</code>, call `.deserialize()?` and wrapping your other return values in `Ok()`?

manuelgdlvh:
Hi @jplatte , 
i made this small change to avoid exactly that, Ok(_) and Result<_, _> in all handlers considering that handlers must to manage all resulting of below layers.
Something like bail! macro (more or less) in anyhow crate does to keep less boilerplate!

jplatte:
Well, using `?` is usually much easier than macros for early returns. It's not really expected that many handlers return exactly a `Response` so I don't think this macro is useful for many users of axum.

If you want, you can use the [Discussions](https://github.com/tokio-rs/axum/discussions) section to discuss the merit of this kind of macro or your style of writing handlers more generally, but I'm going to close this PR because I'm fairly certain it doesn't make sense for axum to include this macro.

Issue #2767, opened 2024-06-04
Is the `reqwest-response` example outdated?


frederikhors:
Is the example here outdated now that `http1` porting is done, right?

https://github.com/tokio-rs/axum/blob/main/examples/reqwest-response/src/main.rs

jplatte:
Hm, it's already upgraded to reqwest 0.12, but it looks more complicated than necessary. I'm not sure actually. If you find a way to cut down on conversion boilerplate (might be simple, or not), a PR would be appreciated.

frederikhors:
I'll PR as soon I understand how. We're trying to find a way here: https://users.rust-lang.org/t/a-proxy-with-axum-0-7-and-reqwest-0-12-based-on-http-1/112489/2.

frederikhors:
I found two ways:

1. This is one way, from https://users.rust-lang.org/t/a-proxy-with-axum-0-7-and-reqwest-0-12-based-on-http-1/112489/5:

    ```rust
    use http_body_util::BodyExt;

    let http_client = reqwest::Client::new();

    let (headers, body) = req.into_parts();

    let req = reqwest::Request::try_from(axum::extract::Request::from_parts(
        headers,
        body.collect().await.unwrap().to_bytes(),
    ))
    .unwrap();

    let res = http_client.execute(req).await.unwrap();
    ```

1. This is another way, from https://stackoverflow.com/a/78577792:

    ```rust
    use sync_wrapper::SyncStream;

    let http_client = reqwest::Client::new();

    let req = req.map(|body| reqwest::Body::wrap_stream(SyncStream::new(body.into_data_stream())));

    let req = reqwest::Request::try_from(req).unwrap();

    let res = http_client.execute(req).await.unwrap();
    ```

I think I'll go with SyncStream.

jplatte:
In case you're waiting for a reply, yes that sounds like a good plan :)

The first snippet would buffer the entire request body in memory, which can be a big problem for large requests.

PaulDotSH:
> In case you're waiting for a reply, yes that sounds like a good plan :)
> 
> The first snippet would buffer the entire request body in memory, which can be a big problem for large requests.

I would be open to create a PR if the issue owner doesn't have the time to

PR #2852, opened 2024-07-29
Fixes: #2767 Update reqwest-response example

redforks:
## Motivation

Update reqwest-response example, as explained by <https://github.com/tokio-rs/axum/issues/2767#issuecomment-2150559926>

## Solution

Based on <https://github.com/tokio-rs/axum/issues/2767#issuecomment-2150559926>

### DIFF
--- a/examples/reqwest-response/Cargo.toml
+++ b/examples/reqwest-response/Cargo.toml
@@ -7,7 +7,6 @@ publish = false
 [dependencies]
 axum = { path = "../../axum" }
 reqwest = { version = "0.12", features = ["stream"] }
-sync_wrapper = { version = "1.0.1", features = ["futures"] }
 tokio = { version = "1.0", features = ["full"] }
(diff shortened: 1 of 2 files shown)


jplatte (commented):
I think there was some sort of confusion in #2767. This example is currently showing how to stream a response from `reqwest` to a client, it's *not* showing how to actually proxy a request. The function name `proxy_via_reqwest` is misleading, but there is a separate example for proxying and I don't think it makes sense to show the same here.

Could you update this PR to only remove the response conversion boilerplate, and leave the request sending as-is? If you want, you can also rename the function to `stream_reqwest_response`.

redforks:
Reconstruct code:

1. revert the change that use SyncStream
2. Remove two unnecessary type conversion: status and headers

jplatte (approved):
Thanks, that looks good!

PR #2857, opened 2024-07-30
Avoid reallocation for String body extraction

yhx-12243:
The conversion from `Bytes` to `String` does not need an extra allocation if we use `Bytes: Into<Vec<u8>>` and `String::from_utf8`.

### DIFF
--- a/axum-core/src/extract/request_parts.rs
+++ b/axum-core/src/extract/request_parts.rs
@@ -106,9 +106,7 @@ where
                 }
             })?;
 
-        let string = std::str::from_utf8(&bytes)
-            .map_err(InvalidUtf8::from_err)?
-            .to_owned();
+        let string = String::from_utf8(bytes.into()).map_err(InvalidUtf8::from_err)?;
 
         Ok(string)
     }
--- a/axum-core/Cargo.toml
+++ b/axum-core/Cargo.toml
@@ -19,7 +19,7 @@ __private_docs = ["dep:tower-http"]
 
 [dependencies]
 async-trait = "0.1.67"
-bytes = "1.0"
+bytes = "1.2"
 futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
 http = "1.0.0"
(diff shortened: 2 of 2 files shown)


mladedav (approved):

jplatte:
Looks like this requires us to raise our minimum version of `bytes`, can you bump it to 1.6.1 in `axum-core/Cargo.toml`?

axum-core/src/extract/request_parts.rs
```diff
@@ -106,9 +106,7 @@ where
                 }
             })?;
 
-        let string = std::str::from_utf8(&bytes)
-            .map_err(InvalidUtf8::from_err)?
-            .to_owned();
+        let string = String::from_utf8(bytes.into()).map_err(InvalidUtf8::from_err)?;
```
FSMaxB, line 109: Won't the `bytes.into()` still reallocate the data? Unless there is some special code in the vtable of whatever representation of `Bytes` that underlies this that checks the reference count and avoids a copy if it is 1?

FSMaxB: Is that what this does? https://docs.rs/bytes/1.6.1/src/bytes/bytes.rs.html#1108-1128

yhx-12243:
> Looks like this requires us to raise our minimum version of `bytes`, can you bump it to 1.6.1 in `axum-core/Cargo.toml`?

Sorry I see, it requires a minimum version of 1.2 of bytes.

axum-core/src/extract/request_parts.rs
mladedav: From what I remember when I looked at `bytes`, you have two ways `Bytes` may be represented. Either with an owned `Vec` or if there are at least two instances referencing the same allocation (even if disjunct), it gets promoted to an `Arc`-backed storage.

But I think the bytes will already be cloned when the request body is collected so I think it should prevent one clone from happening.

jplatte (approved):

PR #2865, opened 2024-08-07
Avoid some extra clones/allocations

novacrazy:
## Motivation

I noticed when writing a `Layer` implementation that it requires `Clone`, and after running a small hello world noticed it cloned _a lot_. After looking at the code I saw that some of the clones were pointless, just to get a mutable reference that then cloned again internally.

## Solution

This adds a couple specialized routines that help avoid a few clones, and cleans up some code elsewhere to avoid a couple more allocations.

~~The most contentious change might be to `MethodRouter::call_with_state`, but a `match` might be more efficient, and the new macro should more easily help catch issues the comment about destructuring was after.~~ Edit: Partially reverted, overlooked something simple.

I'm still looking for places where clones/allocations can be avoided.

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -656,14 +656,14 @@ where
         }
     }
 
-    fn call_with_state(&mut self, req: Request, state: S) -> RouteFuture<E> {
+    fn call_with_state(self, req: Request, state: S) -> RouteFuture<E> {
         match self {
             Fallback::Default(route) | Fallback::Service(route) => {
-                RouteFuture::from_future(route.oneshot_inner(req))
+                RouteFuture::from_future(route.oneshot_inner_owned(req))
             }
             Fallback::BoxedHandler(handler) => {
-                let mut route = handler.clone().into_route(state);
-                RouteFuture::from_future(route.oneshot_inner(req))
+                let route = handler.into_route(state);
+                RouteFuture::from_future(route.oneshot_inner_owned(req))
             }
         }
     }
--- a/axum/src/routing/path_router.rs
+++ b/axum/src/routing/path_router.rs
@@ -331,9 +331,11 @@ where
             }
         }
 
-        let path = req.uri().path().to_owned();
+        // split apart to get multiple mutable references to parts,
+        // this avoids the need to clone the URI/path.
+        let (mut parts, body) = req.into_parts();
 
-        match self.node.at(&path) {
+        match self.node.at(parts.uri.path()) {
             Ok(match_) => {
                 let id = *match_.value;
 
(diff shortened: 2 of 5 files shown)


pickfire:
@novacrazy I am interested how did you get to figure out where allocations occur. Did you overwrite the global allocator to make it tracks allocation?

novacrazy:
@pickfire I had a hunch it was inefficient somewhere, so I put a `println!` in my layer/service `Clone` implementation to confirm. Then I just read through Axum’s code, starting at the Router’s own Service call until I saw all the cloning going on. Nothing special.

When I have some more time I’d like to figure out how to remove even more clones. 1-2 should be possible.

jplatte (commented):
Hey, thanks for the PR! I'm not intimately familiar with the code pieces you're changing here. Would you mind splitting this PR into smaller ones? It seems like there's a bunch of independent changes here that would be easier to discuss separately.

Bmwx12013 (changes requested):
Chancha

yanns (approved):
This is looking good to me, but I'm no expert here.

yanns:
@novacrazy could you check the conflicts with `main`?

mladedav (approved):

yanns:
@novacrazy I've created https://github.com/tokio-rs/axum/pull/2968 that would merge your changes with `main`. I hope this can help you updating to `main`.

axum/src/routing/path_router.rs
```diff
@@ -342,22 +344,24 @@ where
...
                     .expect("no route for id. This is a bug in axum. Please file an issue");
 
+                let req = Request::from_parts(parts, body);
+
                 match endpoint {
                     Endpoint::MethodRouter(method_router) => {
                         Ok(method_router.call_with_state(req, state))
                     }
-                    Endpoint::Route(route) => Ok(route.clone().call(req)),
+                    Endpoint::Route(route) => Ok(route.clone().call_owned(req)),
```
yanns, line 364: This change is not reflected in any tests. Do you think we could test this?

axum/src/routing/route.rs
```diff
@@ -42,13 +42,27 @@ impl<E> Route<E> {
         )))
     }
 
+    /// Variant of [`Route::call`] that takes ownership of the route to avoid cloning.
+    pub(crate) fn call_owned(self, req: Request<Body>) -> RouteFuture<E> {
+        let req = req.map(Body::new);
+        RouteFuture::from_future(self.oneshot_inner_owned(req))
```
yanns, line 48: When comparing with `main`, it would now be `self.oneshot_inner_owned(req).not_top_level()`.

I'm not sure about the `.not_top_level()`. And there are no tests checking if this is necessary or not.
Do you think we could test this?

axum/src/routing/mod.rs
```diff
@@ -656,14 +656,14 @@ where
...
         match self {
             Fallback::Default(route) | Fallback::Service(route) => {
-                RouteFuture::from_future(route.oneshot_inner(req))
+                RouteFuture::from_future(route.oneshot_inner_owned(req))
             }
             Fallback::BoxedHandler(handler) => {
-                let mut route = handler.clone().into_route(state);
-                RouteFuture::from_future(route.oneshot_inner(req))
+                let route = handler.into_route(state);
+                RouteFuture::from_future(route.oneshot_inner_owned(req))
```
yanns, line 666: There are no tests covering this change. Do you think we could test this?

jplatte:
Given latest comments, I want to emphasize once again that this would be much easier to get merged as several smaller PRs, however small the diff may be.

novacrazy:
If @yanns wants to take over with #2968 that's fine with me. I ended up just writing my own zero-cost web framework to replace Axum in my stack.

jplatte:
Okay, I'll close this in favor of the new PR then.

yanns:
I've extracted one part that is covered by tests in https://github.com/tokio-rs/axum/pull/2969.
I'll tackle the other parts in other PRs, trying to find way to add test coverage.

axum/src/routing/route.rs
yanns: This is a note for me.
The current `.not_top_level()` is covered by `axum/src/routing/tests/mod.rs:1086`

PR #2875, opened 2024-08-15
Made DefaultBodyLimit const friendly

StNicolay:
Derived Copy for DefaultBodyLimit
Made DefaultBodyLimit::max and DefaultBodyLimit::disabled const

## Motivation

Making DefaultBodyLimit more const friendly will reduce the code duplication for projects trying to add this layer to multiple endpoints but not to the entire route or router by removing the unnecessary clone calls

## Solution

By deriving Copy and making those methods const we allow the following code to be used:

```rust
const LIMIT: DefaultBodyLimit = DefaultBodyLimit::disable();

Router::new()
    .route(
        "/files",
        get(file::download)
            .post(file::upload.layer(LIMIT))
            .delete(file::delete)
            .patch(file::modify.layer(LIMIT)),
    )
```

instead of

```rust
let limit = DefaultBodyLimit::disable();


Router::new()
    .route(
        "/files",
        get(file::download)
            .post(file::upload.layer(limit.clone()))
            .delete(file::delete)
            .patch(file::modify.layer(limit.clone())),
    )
```

### DIFF
--- a/axum-core/src/extract/default_body_limit.rs
+++ b/axum-core/src/extract/default_body_limit.rs
@@ -72,7 +72,7 @@ use tower_layer::Layer;
 /// [`RequestBodyLimit`]: tower_http::limit::RequestBodyLimit
 /// [`RequestExt::with_limited_body`]: crate::RequestExt::with_limited_body
 /// [`RequestExt::into_limited_body`]: crate::RequestExt::into_limited_body
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 #[must_use]
 pub struct DefaultBodyLimit {
     kind: DefaultBodyLimitKind,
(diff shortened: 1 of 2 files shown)


axum-core/src/extract/default_body_limit.rs
```diff
@@ -72,7 +72,7 @@ use tower_layer::Layer;
 /// [`RequestBodyLimit`]: tower_http::limit::RequestBodyLimit
 /// [`RequestExt::with_limited_body`]: crate::RequestExt::with_limited_body
 /// [`RequestExt::into_limited_body`]: crate::RequestExt::into_limited_body
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
```
jplatte, line 75: What is `Copy` needed for, in your example? I'm a little bit hesitant about adding it.

jplatte (commented):
Hey, thanks for the PR!

Can you add a changelog entry for it?

axum-core/src/extract/default_body_limit.rs
StNicolay: It allows not using clone when DefaultBodyLimit was constructed during runtime (by reading the config file for example). I understand, that removing it would be a breaking change, but I consider it worth it, since such changes to this layer are unlikely

jplatte: Well, I agree that changes that would make it not `Copy` are somewhat unlikely, but we tend to be conservative about that kind of thing. Anyways, in this specific case I'm fine merging as-is.

jplatte (approved):

PR #2880, opened 2024-08-18
Upgrade tower and other dependencies

elichai:
## Motivation

Upgrade dependencies to reduce duplicate dependencies in binaries depending on `axum` together with other projects using `tower`

## Solution

Upgrade everything to latest version,
I did not upgrade `matchit` to 0.8 as this is a breaking change for routing matching so I'm not sure how to move forward with this, the changes are: https://github.com/ibraheemdev/matchit/releases/tag/v0.8.0

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -49,7 +49,7 @@ http-body-util = "0.1.0"
 mime = "0.3"
 pin-project-lite = "0.2"
 serde = "1.0"
-tower = { version = "0.4", default-features = false, features = ["util"] }
+tower = { version = "0.5", default-features = false, features = ["util"] }
 tower-layer = "0.3"
 tower-service = "0.3"
 
@@ -60,7 +60,7 @@ form_urlencoded = { version = "1.1.0", optional = true }
 headers = { version = "0.4.0", optional = true }
 multer = { version = "3.0.0", optional = true }
 percent-encoding = { version = "2.1", optional = true }
-prost = { version = "0.12", optional = true }
+prost = { version = "0.13", optional = true }
 serde_html_form = { version = "0.2.0", optional = true }
 serde_json = { version = "1.0.71", optional = true }
 serde_path_to_error = { version = "0.1.8", optional = true }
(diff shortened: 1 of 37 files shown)


elichai:
This should fix the failed CI: https://github.com/tower-rs/tower/pull/788

martinetd:
Thanks, I also had been looking at the same after noticing duplicates in our dependencies and found this update.

Just commenting on failed checks:
- msrv, it looks like prost 0.13 requires rust 1.70 (while axum's msrv is 1.66); 1.70 looks old enough to me to just update requirements but I have no opinion here, I don't otherwise depend on prost so would be happy to keep it lower for a while longer if there is no security implication (didn't check deeply, doesn't seem to be any)
- duplicates ('bans licenses sources' check)
  - sync_wrapper: tower 0.5 requires 0.1.2 so we need to stick with the old version until they in turn upgrade
  - tower 0.4 is *still* pulled from reqwest 0.12.4 -> hyper-util v0.1.8 -> tower v0.4.13; so I guess we need to start with updating hyper-util?... that's apparently holding on another msrv problem as well: https://github.com/hyperium/hyper-util/pull/144


dependency tracking in rust is so much fun.

jplatte:
tower and tower-http are upgraded now. It would make a lot of sense to do the rest of the updates from this PR, but I would prefer separate PRs for examples vs. non-example code.

W.r.t. `cargo-deny`, it's not a big deal to add more entries to the allowlist for duplicate dependencies. Really I don't find error-on-duplicate-deps to be such a useful thing in general.

jplatte:
Oh and thanks for the effort, and sorry for letting this PR (and others) linger for so long.

martinetd:
Thank you!

> W.r.t. cargo-deny, it's not a big deal to add more entries to the allowlist for duplicate dependencies. Really I don't find error-on-duplicate-deps to be such a useful thing in general.

Good to know; I'm more used to the world of distro packaging where depending on two versions of a .so would generally just blow up but if you think it's not a problem here I'll trust your judgement on it.
It looks like trying to upgrade everything in lockstep would be difficult here anyway and might as well start somewhere

jplatte:
I'll close this since most of the upgrades have been merged in separate PRs. There's still the `low-level-rustls` example that needs its encryption dependencies upgraded, if anybody involved here wants to make a PR for that, it would be appreciated :)

PR #2881, opened 2024-08-18
feat: local_addr for Serve and WithGracefulShutdown

### LINKED ISSUE #2879: add means to find the socket address for Serve and WithGracefulShutdown
- [*] I have looked for existing issues (including closed) about this

## Feature Request

I would like to get the socket address of `Serve` and `WithGracefulShutdown`.

### Motivation

I am the maintainer of Axum Test, and the initial motivation behind this is to help with testing servers that have been gracefully shutdown. The way Axum Test works is by having a `TestServer`, and one passes in a `Router` or a `IntoMakeService`, and the `TestServer` then sets that up to be served.

This is relevant, as it means a developer cannot call `Serve::with_graceful_shutdown` before passing their application into a `TestServer`.

I can extend my code to allow that. However I need to know what address the `Serve` is running on, in order to then make requests to it. Ideally I'd like to be able to just get it from `Serve` directly.

### Proposal

I'd propose to add the methods `Serve::local_addr` and `WithGracefulShutdown::local_addr`. This is pretty simple to do, and the code would be something like:

```rust
#[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
impl<M, S> Serve<M, S> {
    pub fn local_addr(&self) -> io::Result<std::net::SocketAddr> {
        self.tcp_listener.local_addr()
    }
}

impl<M, S, F> WithGracefulShutdown<M, S, F> {
    pub fn local_addr(&self) -> io::Result<std::net::SocketAddr> {
        self.tcp_listener.local_addr()
    }
}
```

I would be happy to make a PR for this myself if there is interest.

### Alternatives

Could offer a means to expose the `TcpListener`.

Another alternative could be to move the `with_graceful_shutdown` to be attached earlier, before calling `serve`. This would make it separate to what the service is listening to, and could also solve my issue (although this would be a large and breaking API change).

JosephLenton:
Adds methods for exposing the address the Axum servers are bound to.

 * add `Serve::local_addr`, which returns the address it is bound to.
 * add `WithGracefulShutdown::local_addr`, which returns the address it is bound to.

Closes #2879

## Motivation

I would like to be able to know the address of a server, as it makes testing code easier to write.

## Solution

I have added methods to expose the `SocketAddr` on `Serve` and `WithGracefulShutdown`.

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -11,10 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **added:** `axum::serve::Serve::tcp_nodelay` and `axum::serve::WithGracefulShutdown::tcp_nodelay` ([#2653])
 - **added:** `Router::has_routes` function ([#2790])
 - **change:** Update tokio-tungstenite to 0.23 ([#2841])
+- **added:** `Serve::local_addr` and `WithGracefulShutdown::local_addr` functions ([#2881])
 
 [#2653]: https://github.com/tokio-rs/axum/pull/2653
 [#2790]: https://github.com/tokio-rs/axum/pull/2790
 [#2841]: https://github.com/tokio-rs/axum/pull/2841
+[#2881]: https://github.com/tokio-rs/axum/pull/2881
 
 # 0.7.5 (24. March, 2024)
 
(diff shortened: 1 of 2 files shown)


jplatte:
Could you please add a change log entry though?

JosephLenton:
@jplatte that is added!

PR #2884, opened 2024-08-21
Use `env!("CARGO_CRATE_NAME")` in the example to simplify the tracing setup code

hongweipeng:
…etup code

## Motivation
The example code is full of tracing registration codes, and the log level is hard-coded. I think it can be replaced with the `module_path!` macro.

## Solution
Simplify the code.

### DIFF
--- a/examples/error-handling/src/main.rs
+++ b/examples/error-handling/src/main.rs
@@ -45,8 +45,9 @@ use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 async fn main() {
     tracing_subscriber::registry()
         .with(
-            tracing_subscriber::EnvFilter::try_from_default_env()
-                .unwrap_or_else(|_| "example_error_handling=debug,tower_http=debug".into()),
+            tracing_subscriber::EnvFilter::try_from_default_env().unwrap_or_else(|_| {
+                concat!(env!("CARGO_CRATE_NAME"), "=debug,tower_http=debug").into()
+            }),
         )
         .with(tracing_subscriber::fmt::layer())
         .init();
--- a/examples/error-handling/src/main.rs
+++ b/examples/error-handling/src/main.rs
@@ -46,7 +46,7 @@ async fn main() {
     tracing_subscriber::registry()
         .with(
             tracing_subscriber::EnvFilter::try_from_default_env().unwrap_or_else(|_| {
-                concat!(env!("CARGO_CRATE_NAME"), "=debug,tower_http=debug").into()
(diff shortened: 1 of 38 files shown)


jplatte:
Interesting idea. Two ideas:

- Use `concat!` instead of `format!`
- Use the crate name instead of the module path so this still applies the log level to the entire crate's logs if moved into a submodule

What do you think? I think together this would look like `concat!(env!("CARGO_CRATE_NAME"), "=debug"))`.

ttys3:
`CARGO_CRATE_NAME` is better.

and for `concat` , I do not think it is necessary. keep using `format!` is much clear.

ttys3:
I'm afraid that `concat!(env!("CARGO_CRATE_NAME"), "=debug,tower_http=debug")` is much less clear compared to:

`format!("{}=debug,tower_http=debug", env!("CARGO_CRATE_NAME"))`

I do not believe there will be performance issues for our usage scenario.

examples/error-handling/src/main.rs
```diff
@@ -45,8 +45,9 @@ use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 async fn main() {
     tracing_subscriber::registry()
         .with(
-            tracing_subscriber::EnvFilter::try_from_default_env()
-                .unwrap_or_else(|_| "example_error_handling=debug,tower_http=debug".into()),
+            tracing_subscriber::EnvFilter::try_from_default_env().unwrap_or_else(|_| {
+                concat!(env!("CARGO_CRATE_NAME"), "=debug,tower_http=debug").into()
```
ttys3, line 49: I'm afraid that `concat!(env!("CARGO_CRATE_NAME"), "=debug,tower_http=debug")` is much less clear compared to:

`format!("{}=debug,tower_http=debug", env!("CARGO_CRATE_NAME"))`

I do not believe there will be performance issues for our usage scenario.

ttys3 (changes requested):

examples/error-handling/src/main.rs
jplatte: Yeah, it's a bit ugly. Feel free to change it back.

mladedav: Personally, I'd say having
```rust
let crate_name = env!("CARGO_CRATE_NAME");
format!("{crate_name}=debug,tower_http=debug")
```
would be a more readable alternative, but feel free to ignore this, either is fine.

jplatte (approved):
Thanks!

PR #2894, opened 2024-08-28
Add support for WebSockets over HTTP/2

SabrinaJewson:
## Motivation

WebSockets over HTTP/2 are currently not supported. It is defined in [RFC 8441](https://datatracker.ietf.org/doc/html/rfc8441) (websockets over HTTP/3 are defined in [RFC 9220](https://datatracker.ietf.org/doc/rfc9220/) and it is exactly the same as HTTP/2).

## Solution

Implement the changes necessary to support them, which just requires looking at the `:protocol` pseudo-header in the websocket extractor and advertising our support by enabling `SETTINGS_ENABLE_CONNECT_PROTOCOL` in HTTP/2.

Additionally, I needed to fix the handling of responses to CONNECT requests: Axum was incorrectly adding the `Content-Length` header to them (see also https://github.com/hyperium/hyper/pull/3748), which was causing things to break.


jplatte (commented):
This looks like a very high-quality PR, but I really don't know enough about websockets _or_ HTTP/2 to review it.

@mladedav are you able to review this?

Also @SabrinaJewson would you be willing to maintain the websocket stuff going forward (be pinged on issues and PRs to help)?

SabrinaJewson:
I’m happy to be pinged!

axum/src/extract/ws.rs
```diff
@@ -122,17 +122,21 @@ use tokio_tungstenite::{
 
 /// Extractor for establishing WebSocket connections.
 ///
-/// Note: This extractor requires the request method to be `GET` so it should
-/// always be used with [`get`](crate::routing::get). Requests with other methods will be
-/// rejected.
+/// For HTTP/1.1 requests, this extractor requires the request method to be `GET`;
+/// in later versions, `CONNECT` is used instead. Thus it should either be used
+/// with [`any`](crate::routing::any), or placed behind
+/// [`on`](crate::routing::on)`(`[`MethodFilter`]`::GET.or(`[`MethodFilter`]`::POST), ...)`.
```
yanns, line 128: I guess we're using `POST` here because we don't support `CONNECT` directly?
Do you think that we should add `CONNECT` as well?

axum/src/routing/method_routing.rs
```diff
@@ -1034,13 +1034,11 @@ where
                     match $svc {
                         MethodEndpoint::None => {}
                         MethodEndpoint::Route(route) => {
-                            return RouteFuture::from_future(route.clone().oneshot_inner($req))
-                                .strip_body($method == Method::HEAD);
+                            return route.clone().oneshot_inner($req);
```
yanns, line 1037: I see that this is the same as https://github.com/tokio-rs/axum/pull/2897.
Do you think we should first merge https://github.com/tokio-rs/axum/pull/2897 to reduce the scope of this PR?

yanns (commented):
I'm also no expert in ws and http2.
The example is running great! :100:

axum/src/extract/ws.rs
SabrinaJewson: No, that’s just a typo on my part…

axum/src/routing/method_routing.rs
SabrinaJewson: Both PRs benefit from this change, so it can be merged in either order!

axum/src/extract/ws.rs
SabrinaJewson: Although I do now reälize that `MethodFilter::CONNECT` doesn’t exist. I pushed a commit to consistently use `any` instead.

yanns (approved):

jplatte (approved):
Could you add a changelog entry, please?

SabrinaJewson:
Done and done.

jplatte (approved):

jplatte:
Let me know if you want this backported (though maybe a follow-up PR adding `MethodFilter::CONNECT` would be good to have in first).

ekzhang:
Thanks for this really awesome new feature!

---
Just FYI — this change resulted in some breakage to a downstream project https://sshx.io in a hard to detect way after upgrading to axum 0.8.0. What happened was that I had a route:

```rust
Router::new().route("/s/{name}", get(socket::get_session_ws))
```

And when I upgraded from axum 0.6 to axum 0.8 (along with a dozen other dependencies), I enabled the http/2 feature and used `axum::serve` as a replacement for `hyper`. But I didn't realize that websockets over http/2 support was added, and that we needed to change the `get()` to `any()` since it wasn't marked as a breaking change. As a result browsers in prod behind a TLS load balancer were being advertised that WebSocket over HTTP/2 was supported, but the backend web server still only had WebSocket over HTTP/1.1 due to the `get()` extractor.

---

Just sharing this in case anyone else finds it helpful, spent a couple hours debugging. But thanks for the great library of course!

Issue #2659, opened 2024-03-17
Consider changing the behavior of / route when nesting


jplatte:
Currently, given a router

```rust
let inner = Router::new().route("/", method_router);
```

and an outer router

- `Router::new().nest("/foo", inner)` => only the route `/foo` exists
- `Router::new().nest("/foo/", inner)` => only the route `/foo/` exists

It seems like people (me included) generally prefer the first form, but oftentimes expect different behavior (see #2267). We should (re)consider what behavior(s) are reasonable for the upcoming release. Also worth checking: What happens with `nest_service`?

jplatte:
cc @yanns @mladedav @ibraheemdev do any of you have ideas / opinions on this?

mladedav:
I've also sometimes guessed wrong how the nesting with root handlers works. One option would be to simply concatenate, but there should be rules to disallow things like  this:
```rust
Router::new().nest("/foo", Router::new().route("bar", todo)); // matches `/foobar` ??
```

What I imagine could work is that axum would require:
- paths in `nest`
  - must start with a slash
  - must not end with a slash
- paths in `route`
  - must be empty or start with a slash

For example:
```rust
Router::new().nest("", Router::new().route("/foo", todo)); // error - nest must start with a slash
Router::new().nest("/", Router::new().route("/foo", todo)); // error - nest must not ends with a slash
Router::new().nest("/foo", Router::new().route("", todo)); // matches `/foo`
Router::new().nest("/foo", Router::new().route("/", todo)); //matches `/foo/`
Router::new().nest("/foo", Router::new().route("/bar", todo)); //matches `/foo/bar`
Router::new().nest("/foo", Router::new().route("/bar/", todo)); //matches `/foo/bar/`
Router::new().nest("/foo/", Router::new().route("/bar", todo)); // error - nest ends with a slash
Router::new().nest("/foo", Router::new().route("bar", todo)); // error - route doesn't start with a slash
```

Currently, axum only requires all paths to start with a slash as far as I know.

This would allow setting routes with or without trailing slashes and would even prevent users from building the same router multiple ways (`nest` with or without a trailing slash is identical except for handling the root as far as I know).

The thing I dislike about this the most is that
```
let router = Router::new().route("/", root).route("", empty);
```
would behave differently based on whether it would be nested or not. On top level the two routes are equivalent (so we should maybe even reject starting the server due to conflicting routes) but in a nested router they are different.

[...]

Issue #2251, opened 2023-10-04
Add a method-not-allowed fallback on the Router level


ThibSrb:
- [x] I have looked for existing issues (including closed) about this

### Motivation

It's a common mistake to expect `Router::fallback` to be triggered on method-not-allowed.
Furthermore, as shown by [#1005](https://github.com/tokio-rs/axum/issues/1005), it is pretty easy to misinterpret the current behavior for a bug.
While returning HTTP 405 is normal in this case, it would be helpful to have a simple baked-in solution to handle it,  in a way similar to fallback.

### Proposal

Add a `Router::method_not_allowed_fallback` method that allows defining a fallback for this specific situation

#### Usage

```rust
use std::net::SocketAddr;

use axum::{routing::get, Router, response::IntoResponse};
use hyper::Server;

async fn hello_world() -> impl IntoResponse {
    "Hello, world!\n"
}

async fn default_fallback() -> impl IntoResponse {
    "Default fallback\n"
}

async fn handle_405() -> impl IntoResponse {
    "Method not allowed fallback"
}

#[tokio::main]
async fn main() -> Result<(), hyper::Error> {
    let router = Router::new()
        .route("/", get(hello_world))
        .fallback(default_fallback)
        .method_not_allowed_fallback(handle_405);

    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));

    Server::bind(&addr).serve(router.into_make_service()).await
}
```

### Alternatives

An included middleware such as `MethodNotAllowedLayer(fallback: Handler)`.

#### Usage

```rust
use std::net::SocketAddr;

use axum::{response::IntoResponse, routing::get, Router, middleware::MethodNotAllowedLayer};
use hyper::Server;

async fn hello_world() -> impl IntoResponse {
    "Hello, world!\n"
}

async fn default_fallback() -> impl IntoResponse {
    "Default fallback\n"
}

async fn handle_405() -> impl IntoResponse {
    "Method not allowed fallback"
}

#[tokio::main]
async fn main() -> Result<(), hyper::Error> {
    let router = Router::new()
        .route("/", get(hello_world))
        .layer(MethodNotAllowedLayer(handle_405))
        .fallback(default_fallback);

    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));

    Server::bind(&addr).serve(router.into_make_service()).await
}
```

Anyway, I think it would be great to mention this topic in the `Router::fallback` documentation.

davidpdrsn:
[...]

PR #2896, opened 2024-09-01
Clarify some subtleties of routing

SabrinaJewson:
## Motivation

I was confused when using Axum of the following things:

- whether `nest("/foo")` will match `/foo` and/or `/foo/`;
- whether a `fallback` is called when the method is not allowed.

## Solution

Document the behaviours.

---

Also relevant: #2659, #2251

### DIFF
--- a/axum/src/docs/routing/nest.md
+++ b/axum/src/docs/routing/nest.md
@@ -82,6 +82,11 @@ let app = Router::new()
 # let _: Router = app;
 ```
 
+Additionally, while the wildcard route `/foo/*rest` will not match the
+paths `/foo` or `/foo/`, a nested router at `/foo` will match the path `/foo`
+(but not `/foo/`), and a nested router at `/foo/` will match the path `/foo/`
+(but not `/foo`).
+
 # Fallbacks
 
 If a nested router doesn't have its own fallback then it will inherit the
--- a/axum/src/docs/routing/fallback.md
+++ b/axum/src/docs/routing/fallback.md
@@ -23,7 +23,11 @@ async fn fallback(uri: Uri) -> (StatusCode, String) {
 
 Fallbacks only apply to routes that aren't matched by anything in the
 router. If a handler is matched by a request but returns 404 the
-fallback is not called.
+fallback is not called. Note that this applies to [`MethodRouter`]s too: if the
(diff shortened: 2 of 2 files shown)


axum/src/docs/routing/nest.md
```diff
@@ -82,6 +82,11 @@ let app = Router::new()
 # let _: Router = app;
 ```
 
+Additionally, while the wildcard route `/foo/*rest` will not match the
+paths `/foo` or `/foo/`, a nested router at `/foo` will match the path `/foo`
```
mladedav, line 86: Are you sure about the wildcard? `/foo/` should match and I'm pretty sure it does at least in older versions. `rest` will match an empty string but that's allowed.

SabrinaJewson: The docs [explicit say this doesn’t happen](https://docs.rs/axum/latest/axum/struct.Router.html#wildcards):

 >Note that `/*key` doesn't match empty segments. Thus:
> 
> - `/*key` doesn't match `/` but does match `/a`, `/a/`, etc.
> - `/x/*key` doesn't match `/x` or `/x/` but does match `/x/a`, `/x/a/`, etc.

mladedav: Right, my bad then. I must have misremembered.

jplatte (commented):
Thanks for the PR, docs improvements are super welcome.
I'm not convinced it's completely correct though, see review comment below.

Also, sorry for the lag.

axum/src/docs/routing/nest.md
```diff
@@ -82,6 +82,11 @@ let app = Router::new()
...
+(but not `/foo/`), and a nested router at `/foo/` will match the path `/foo/`
```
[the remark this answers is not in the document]
SabrinaJewson: Didn’t you open https://github.com/tokio-rs/axum/issues/2659 which explains this behaviour?
ts
nots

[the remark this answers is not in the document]
jplatte: I completely forgot 🫠 

Guess I am still of the same opinion, this is super confusing and should be changed. Makes it even more important to have clear docs!

jplatte (approved):

PR #2904, opened 2024-09-09
Fix warnings and errors from beta Rust

bengsparks:
## Motivation

The CI cannot pass due to the marking of two new warnings / errors, namely pattern exhaustiveness (by `rustc`) and long doc comments (by `clippy`).

## Solution

1. Pattern exhaustiveness
```rs
error: unreachable pattern
   --> axum/src/middleware/map_response.rs:281:25
    |
281 |                         Err(err) => match err {}
    |                         ^^^^^^^^ matches no values because `std::convert::Infallible` is uninhabited
```
Because `unreachable pattern` is a useful compiler warning, and within all of `axum`, there are only 4 occurrences, it is simplest to just mark the match arms with `#[allow]`.
There does not seem to be a way to disable more specific instances of unreachable patterns.

Because the examples are separate from the `axum` lib, it's simplest to just use the same `#[allow]` in each example's `main.rs` at a module level.

NOTE: It seems that this warning has irked enough people for an exception for [empty match arms to be made and nominated for backporting to beta](https://github.com/rust-lang/rust/pull/129103), so hopefully these commits can be reverted soon.

2. `clippy::too-long-first-doc-paragraph`

```rs
error: first doc comment paragraph is too long
   --> axum/src/extract/ws.rs:785:5
    |
785 | /     /// Indicates that an endpoint is terminating the connection because it has received a type of
786 | |     /// data it cannot accept (e.g., an endpoint that understands only text data MAY send this if
787 | |     /// it receives a binary message).
    | |____^
```

This lint is likely here to stay, so this requires actual changes, which were made in accordance with [the fix recommended by the lint's author(s)](https://rust-lang.github.io/rust-clippy/master/index.html#/clippy::too-long-first-doc-paragraph).

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -783,8 +783,9 @@ pub mod close_code {
     pub const PROTOCOL: u16 = 1002;
 
     /// Indicates that an endpoint is terminating the connection because it has received a type of
-    /// data it cannot accept (e.g., an endpoint that understands only text data MAY send this if
-    /// it receives a binary message).
+    /// data that it cannot accept.  
+    ///
+    /// For example, an endpoint MAY send this if it understands only text data, but receives a binary message.
     pub const UNSUPPORTED: u16 = 1003;
 
     /// Indicates that no status code was included in a closing frame.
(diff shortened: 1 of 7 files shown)


axum/src/extract/ws.rs
```diff
@@ -783,8 +783,9 @@ pub mod close_code {
     pub const PROTOCOL: u16 = 1002;
 
     /// Indicates that an endpoint is terminating the connection because it has received a type of
-    /// data it cannot accept (e.g., an endpoint that understands only text data MAY send this if
-    /// it receives a binary message).
+    /// data that it cannot accept.
```
jplatte, line 786: Double spaces at the end of the line will add a line break (in addition to the new paragraph). I'm actually a bit surprised that rustfmt doesn't change these.

```suggestion
    /// data that it cannot accept.
```

jplatte (approved):
Thanks!

PR #2912, opened 2024-09-14
Add request id example

conways-glider:
Add a request-id example using tower, tower-http, and tracing.

## Motivation

This is something I've had to do a few times now in axum that I had to figure out through a combination of axum docs and tower-http docs. Having a concrete example shows a way of doing this in a way that axum supports without having to dig through 2 docs.rs pages (or 3 if you're looking in tower as well). 

Figuring out the middleware, the order, and how to write the `TraceLayer` to avoid extra clones has prevented issues for other engineers I've interacted with as well, so the goal is also to clarify an easy way to achieve request ids.

## Solution

- Add an example of how to add randomized request ids at `examples/request-id`.
- Utilize `tower` and `tower-http` crates to handle request-id middleware.
- Use `tracing` and `tracing-subscriber` crates to show how to add these request ids to each requests span.

### DIFF
--- a/examples/request-id/src/main.rs
+++ b/examples/request-id/src/main.rs
@@ -87,5 +87,6 @@ async fn main() {
 }
 
 async fn handler() -> Html<&'static str> {
+    info!("Hello world!");
     Html("<h1>Hello, World!</h1>")
 }
--- a/examples/request-id/src/main.rs
+++ b/examples/request-id/src/main.rs
@@ -5,6 +5,7 @@
 //! ```
 
 use axum::{
+    extract::MatchedPath,
     http::{HeaderName, Request},
     response::Html,
     routing::get,
@@ -15,7 +16,7 @@ use tower_http::{
     request_id::{MakeRequestUuid, PropagateRequestIdLayer, SetRequestIdLayer},
     trace::TraceLayer,
 };
-use tracing::info_span;
+use tracing::{info, info_span};
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
 const REQUEST_ID_HEADER: &str = "x-request-id";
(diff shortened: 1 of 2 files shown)


yanns (approved):

examples/request-id/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+                let request_id = request.headers().get(REQUEST_ID_HEADER);
+
+                match request_id {
+                    Some(request_id) => info_span!(
+                        "http_request",
+                        matched_path,
+                        method = ?request.method(),
+                        request_id = ?request_id,
+                    ),
+                    None => info_span!(
```
mladedav, line 67: This should never be actually hit, right? If so, I think it would be better to either place a comment here saying as much or using `tracing::error!` or similar to signal that this would be a bug.

Right now it seems to me like this expects to sometimes not have the request ID.

examples/request-id/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+            MakeRequestUuid,
+        ))
+        .layer(
+            TraceLayer::new_for_http().make_span_with(|request: &Request<_>| {
+                // Log the matched route's path (with placeholders not filled in).
+                // Use request.uri() or OriginalUri if you want the real path.
+                let matched_path = request
+                    .extensions()
+                    .get::<MatchedPath>()
+                    .map(MatchedPath::as_str);
```
mladedav, line 55: Do we need this for the example? I feel like this diverts attention from the main focus which are the IDs.

examples/request-id/src/main.rs
conways-glider: Good call! I've updated `None` case to log an error.

examples/request-id/src/main.rs
conways-glider: I included it to match the other logging example, but I think you're right. 

I removed all extra fields from the span!

mladedav (approved):
Thanks

PR #2935, opened 2024-09-24
Upgrade private dependencies

### LINKED ISSUE #2928: Update Tungstenite to version 0.24.0
- [ ] I have looked for existing issues (including closed) about this

### Version
0.7.6

### Platform
Windows

### Description

Update the version of tungstinite to version 0.24. It's fixes an issue where it doesn't correctly flush the stream and allows for tungstenite to be used with `arti_client`.

jplatte:
## Motivation

Some dependencies are outdated.

## Solution

Upgrade them...

Closes #2928.

### DIFF
--- a/examples/http-proxy/Cargo.toml
+++ b/examples/http-proxy/Cargo.toml
@@ -9,6 +9,6 @@ axum = { path = "../../axum" }
 hyper = { version = "1", features = ["full"] }
 hyper-util = "0.1.1"
 tokio = { version = "1.0", features = ["full"] }
-tower = { version = "0.4", features = ["make"] }
+tower = { version = "0.4", features = ["make", "util"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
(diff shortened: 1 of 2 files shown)


examples/http-proxy/Cargo.toml
```diff
@@ -9,6 +9,6 @@ axum = { path = "../../axum" }
 hyper = { version = "1", features = ["full"] }
 hyper-util = "0.1.1"
 tokio = { version = "1.0", features = ["full"] }
-tower = { version = "0.4", features = ["make"] }
+tower = { version = "0.4", features = ["make", "util"] }
```
mladedav, line 12: This seems like an accidental change. The version did not change and I don't expect the feature to do anything on its own.

It's just an example though so I guess it doesn't hurt anything, feel free to ignore.

mladedav (approved):

examples/http-proxy/Cargo.toml
jplatte: Yeah I don't know exactly what happened here but it's definitely required, and should have been included before AFAICT.

PR #2936, opened 2024-09-24
Upgrade most example dependencies

jplatte:
## Motivation

Outdated deps need upgrading.

## Solution

Bump all the versions in `Cargo.toml` that don't need code changes.


jplatte:
@programatik29 I backed out the upgrades of `rustls-pemfile` and `tokio-rustls` in `examples/low-level-rustls` because those required some looking into that I don't currently have time for.

I figured you would have all the context already, right? Maybe you can make a PR to upgrade those?

mladedav (approved):
👍

PR #2941, opened 2024-09-27
Make `serve` generic over the listener and IO types

jplatte:
Supersedes #2479.

### DIFF
--- a/axum/src/serve.rs
+++ b/axum/src/serve.rs
@@ -1,12 +1,12 @@
 //! Serve services.
 
 use std::{
+    any::TypeId,
     convert::Infallible,
     fmt::Debug,
     future::{poll_fn, Future, IntoFuture},
     io,
     marker::PhantomData,
-    net::SocketAddr,
     sync::Arc,
     time::Duration,
 };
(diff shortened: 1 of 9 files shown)


daheige (approved):
LGTM!

axum/src/serve.rs
```diff
@@ -18,12 +18,59 @@ use hyper_util::rt::{TokioExecutor, TokioIo};
...
+    /// Returns the local address that this listener is bound to.
+    fn local_addr(&self) -> io::Result<Self::Addr>;
+}
+
+impl Listener for TcpListener {
+    type Io = TcpStream;
+    type Addr = std::net::SocketAddr;
+
+    #[inline]
+    async fn accept(&mut self) -> io::Result<(Self::Io, Self::Addr)> {
```
mladedav, line 48: Currently, we ignore these errors. Or more precisely, we ignore "connection errors" and log and sleep on all other errors.

I think we could just have a return type of `(Self::Io, Self::Addr)` here and let the underlying method do whatever it wants to handle transient accept errors, log them, sleep, panic, or anything else.

Or we could let it return a `Result` but then I think those errors should be fatal and we'd just propagate them. And in that case I'd also consider whether to have an associated type for the error type instead of forcing `std::io::Error`.

mladedav (commented):
This looks good to me.

The underlying theme of the comments seems to be what do we want this to eventually look like and whether to move some functionality in `serve` into the `Listener` instances so I guess that's the main thing I'd want your opinions on.

jplatte:
@mladedav I think I've addressed all of your concern with the three new commits I've added. Let me know what you think!

mladedav (approved):
I think this is great. Thanks for finishing it.

PR #2947, opened 2024-09-28
Remove mutexes around boxed services

jplatte:
They're no longer needed with services being required to implement Sync. (#2473)


daheige (approved):
LGTM.
There is really no need for mutex locks here, and I agree with this change.

mladedav (approved):
👍

PR #2948, opened 2024-09-29
MSRV in READMEs

mladedav:
## Motivation

Old MSRV was in all READMEs.

## Solution

Update to our actual MSRV.

One  instance of pre-1.66 conditional compilation was also in the code so I got rid of it since we do not support lower versions anymore.

### DIFF
--- a/axum-core/README.md
+++ b/axum-core/README.md
@@ -14,7 +14,7 @@ This crate uses `#![forbid(unsafe_code)]` to ensure everything is implemented in
 
 ## Minimum supported Rust version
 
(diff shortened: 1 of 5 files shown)


mladedav:
I didn't remember the readme before approving the MSRV bump PR so I'm fixing it here.

jplatte (approved):
Thanks for catching!

mladedav:
Do I just click squash and merge or do I have to also do something else so that the commit message contains the PR number or anything like that?

jplatte:
The PR number is included by default, you don't need to do anything extra.

PR #2950, opened 2024-09-30
Reference the requirements for from_fn_with_state

keirlawson:
## Motivation

I was unable to find this information when navigating the docs myself.

## Solution

Provide a link.


jplatte (approved):
Thanks! I've removed the redundant link definition, since rustdoc will figure it out without that.

PR #2956, opened 2024-10-04
Move the Host extractor to axum-extra

jplatte:
## Motivation

There's not one obvious way to implement it (should it handle proxy headers at all, which ones, why, ...). We have more opinionated stuff in `axum-extra`.

## Solution

Move `Host` from `axum` to `axum-extra`.


yanns (approved):

mladedav (approved):
Let's merge this now. I said elsewhere that we could wait for #2242 to move the fixed version, but I'll just recreate the PR after moving this.

jplatte:
Moved the changelog entry to the beginning of the changes as part of the merge, as I think breaking changes generally should be first. We should move the websocket one up too (separate PR).

PR #2957, opened 2024-10-04
axum-extra: Remove stray deprecation in multipart builder

zleyyij:
## Motivation
In https://github.com/tokio-rs/axum/pull/2654, I added support for creating multi part forms, but a stray `#[deprecated` was left there. This PR removes that.


jplatte (approved):
Thanks!

PR #2958, opened 2024-10-04
Release axum 0.8.0-alpha.1 and related crates

jplatte:
Let's get people to try the matchit upgrade and native `async fn` in traits! 🎉


mladedav (approved):
🎉

Issue #2937, opened 2024-09-24
Consider adding `axum::json!`


carllerche:
## Feature Request

Consider adding an `axum::json!` macro that matches the `serde_json` macro syntax and avoids the stutter:

```rust
async fn hello_world() -> impl IntoResponse {
    axum::json!({ "hello": "world" })
}
```

### Motivation

Returning JSON is common. This would avoid adding an extra dependency in the end user's Cargo.toml and avoid some stutter (`Json(json!(...))`) without exposing the `serde_json` dependency.

SabrinaJewson:
Given that this function returns `impl IntoResponse` anyway, might it be worth using an approach like [`typed-json`](https://docs.rs/typed-json) to avoid the heap allocation?

jplatte:
That's exactly the sort of thing I was planning to write myself! Really nice that it already exists. I think that crate can be used directly by wrapping the generated type in `axum::Json` (and documenting that people are free to write either `-> Json<impl Serialize>` or `-> impl IntoResponse`). Would you be interested in making a PR for it?

PR #2962, opened 2024-10-06
Add `axum_extra::json!`

SabrinaJewson:
## Motivation

Closes: #2937

## Solution

Add `axum_extra::json!`, a macro that forwards to `typed_json::json!` to produce an ad-hoc JSON response with zero allocations.

I chose to put this in `axum-extra` rather than core Axum for two reasons:
1. It adds an extra dependency on `typed-json`. It’s a lightweight dependency and not exposed in the public API, but a dependency nonetheless.
2. It allows for `json!` to return an `ErasedJson`. Using `Json<impl Serialize>` isn’t ideal because if the `json!` macro borrows data, the type can’t be returned from a function. Using `Response` directly isn’t ideal because we lose `Clone`.

I had to modify a couple lines in the `Cargo.toml` of `axum-extra` in order to get the tests to compile on my machine – just enabling some feature flags.

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning].
 
 - **added:** Add `json!` for easy construction of JSON responses ([#2962]).
 
+[#2962]: https://github.com/tokio-rs/axum/pull/2962
+
 # 0.10.0
 
 # alpha.1
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning].
 
 # Unreleased
 
-- **added:** Add `json!` for easy construction of JSON responses ([#2962]).
+- **added:** Add `json!` for easy construction of JSON responses ([#2962])
 
 [#2962]: https://github.com/tokio-rs/axum/pull/2962
 
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog],
 and this project adheres to [Semantic Versioning].
 
+# Unreleased
+
+- **added:** Add `json!` for easy construction of JSON responses ([#2962]).
+
 # 0.10.0
 
 # alpha.1
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -104,4 +104,8 @@ allowed = [
     "tokio",
     "tower_layer",
     "tower_service",
+    # typed-json does not appear in the public API, but does appear in `#[doc(hidden)]` re-exports
+    # accessed by macros that are not a part of the public API.
+    # See also: https://github.com/davidpdrsn/cargo-public-api-crates/issues/6
+    "typed-json",
 ]
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -104,8 +104,4 @@ allowed = [
     "tokio",
     "tower_layer",
     "tower_service",
-    # typed-json does not appear in the public API, but does appear in `#[doc(hidden)]` re-exports
-    # accessed by macros that are not a part of the public API.
-    # See also: https://github.com/davidpdrsn/cargo-public-api-crates/issues/6
-    "typed-json",
 ]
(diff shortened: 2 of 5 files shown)


axum-extra/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog],
 and this project adheres to [Semantic Versioning].
 
+# Unreleased
+
+- **added:** Add `json!` for easy construction of JSON responses ([#2962]).
+
```
jplatte, line 11:
```suggestion
- **added:** Add `json!` for easy construction of JSON responses ([#2962]).

[#2962]: https://github.com/tokio-rs/axum/pull/2962

```

axum-extra/src/response/erased_json.rs
```diff
@@ -72,3 +81,61 @@ impl IntoResponse for ErasedJson {
...
+        $crate::response::ErasedJson::new(
+            $crate::response::__private_erased_json::typed_json::json!($($t)*)
+        )
+    }
+}
+
+/// Not public API. Re-exported as `crate::response::__private_erased_json`.
+pub mod private {
+    pub use typed_json;
+}
```
jplatte, line 141: Maybe it helps do mark one or both of these as `#[doc(hidden)]` too? 😅

axum-extra/Cargo.toml
```diff
@@ -102,4 +104,8 @@ allowed = [
     "tokio",
     "tower_layer",
     "tower_service",
+    # typed-json does not appear in the public API, but does appear in `#[doc(hidden)]` re-exports
+    # accessed by macros that are not a part of the public API.
+    # See also: https://github.com/davidpdrsn/cargo-public-api-crates/issues/6
+    "typed-json",
```
jplatte, line 110: AFAIK, `cargo-public-api-crates` operates on rustdoc JSON, so I wonder why this is necessary.

axum-extra/src/response/erased_json.rs
```diff
@@ -72,3 +81,61 @@ impl IntoResponse for ErasedJson {
...
+/// Any variable or expression implementing [`Serialize`]
+/// can be interpolated as a value in the literal.
+/// If the [`Serialize`] implementation decides to fail,
+/// or if a map with non-string keys is used,
+/// a 500 response will be issued
+/// whose body is the error message in UTF-8.
+///
+/// Internally,
+/// this function uses the [`typed_json::json!`] macro
+/// so that no allocations are performed during serialization.
```
jplatte, line 97: `ErasedJson::new` *does* allocate, so this wording needs to be tweaked a bit. Maybe we could state that it allocates a lot less than `Json(serde_json::json!(...))`?

axum-extra/Cargo.toml
SabrinaJewson: Huh. I might be wrong about this. I need to do some more investigation.

axum-extra/CHANGELOG.md
SabrinaJewson: Thanks – somehow I thought it would link automatically 🤣

jplatte: Maybe it actually does within GitHub, but changelogs are also read locally from time to time 😄

axum-extra/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog],
 and this project adheres to [Semantic Versioning].
 
+# Unreleased
+
+- **added:** Add `json!` for easy construction of JSON responses ([#2962]).
```
jplatte, line 10:
```suggestion
- **added:** Add `json!` for easy construction of JSON responses ([#2962])
```

axum-extra/src/response/erased_json.rs
```diff
@@ -12,6 +12,15 @@ use serde::Serialize;
 /// This allows returning a borrowing type from a handler, or returning different response
 /// types as JSON from different branches inside a handler.
 ///
+/// Like [`axum::Json`],
+/// if the [`Serialize`] implementation decides to fail
```
jplatte, line 16:
```suggestion
/// if the [`Serialize`] implementation fails
```

axum-extra/src/response/erased_json.rs
```diff
@@ -72,3 +81,65 @@ impl IntoResponse for ErasedJson {
         }
     }
 }
+
+/// Construct an [`ErasedJson`] response from a JSON literal.
+///
+/// A `Content-Type: application/json` header is automatically added.
+/// Any variable or expression implementing [`Serialize`]
+/// can be interpolated as a value in the literal.
+/// If the [`Serialize`] implementation decides to fail,
```
jplatte, line 90:
```suggestion
/// If the [`Serialize`] implementation fails,
```

jplatte (approved):
Thanks!

PR #2969, opened 2024-10-09
avoid one state clone

yanns:
This is an extraction of a part of https://github.com/tokio-rs/axum/pull/2865

### DIFF
--- a/axum/src/routing/method_routing.rs
+++ b/axum/src/routing/method_routing.rs
@@ -1101,26 +1101,23 @@ where
         macro_rules! call {
             (
                 $req:expr,
-                $method:expr,
                 $method_variant:ident,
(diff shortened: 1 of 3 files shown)


jplatte (approved):
I wonder whether services being required to impl `Sync` now, changes what we can and should do here. But for a start, this is definitely an improvement :)

Maybe use git's `--author` flag to assign authorship to the original author? Your own git identity will then be recorded as `Committer` (as always).

yanns:
> Maybe use git's `--author` flag to assign authorship to the original author? Your own git identity will then be recorded as `Committer` (as always).

Good idea! It's done.

jplatte:
If you rebase-merge now (it's already approved), the author will be kept.

PR #2970, opened 2024-10-09
Avoid clone in fallback

yanns:
The first commit creates test coverage on state clone in fallback.
The second commit removes one clone.

This is an extraction of a part of https://github.com/tokio-rs/axum/pull/2865

### DIFF
--- a/axum/src/test_helpers/counting_cloneable_state.rs
+++ b/axum/src/test_helpers/counting_cloneable_state.rs
@@ -1,0 +1,52 @@
+use std::sync::{
+    atomic::{AtomicBool, AtomicUsize, Ordering},
+    Arc,
+};
+
+pub(crate) struct CountingCloneableState {
+    state: Arc<InnerState>,
+}
+
+struct InnerState {
+    setup_done: AtomicBool,
+    count: AtomicUsize,
+}
+
+impl CountingCloneableState {
+    pub(crate) fn new() -> Self {
+        let inner_state = InnerState {
+            setup_done: AtomicBool::new(false),
+            count: AtomicUsize::new(0),
+        };
+        CountingCloneableState {
+            state: Arc::new(inner_state),
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


axum/src/test_helpers/counting_cloneable_state.rs
```diff
@@ -0,0 +1,59 @@
+use std::sync::{
+    atomic::{AtomicBool, AtomicUsize, Ordering},
+    Arc, Mutex,
+};
+
+pub(crate) struct CountingCloneableState {
+    state: Arc<Mutex<InnerState>>,
+}
```
jplatte, line 8: Nice 👍🏼

jplatte (approved):

axum/src/test_helpers/counting_cloneable_state.rs
yanns: actually, we don't need a `Mutex` as we're using atomic values: https://github.com/tokio-rs/axum/pull/2970/commits/d9ef8e02c7087d5bd52605ffc23dc008ccfaaea5

yanns:
@jplatte do we prefer "squash and merge"?
Or I could make a better git log with only 2 commits, and use "rebase and merge"?

jplatte:
I'd say it doesn't matter much here. I implied in the other PR that squash doesn't preserve authorship, but actually I think GitHub will add `Co-authored-by` lines for separate authors in squash-merge commits.

Generally I default to squash-merge in axum bc. that's what David pretty much always used, making some exceptions for larger PRs with well-separated commits.

yanns:
OK thanks. I'll keep the history here to show that the 2nd commit decreases the number of clones.

jplatte:
One nitpick about the commit messages, please capitalize them in the future (unless there's some prefix like `macros: `, though I haven't seen that used much in this repo anyways).

PR #2971, opened 2024-10-10
avoid cloning the uri/path

yanns:
extract of https://github.com/tokio-rs/axum/pull/2865

### DIFF
--- a/axum/src/routing/path_router.rs
+++ b/axum/src/routing/path_router.rs
@@ -367,9 +367,9 @@ where
             }
         }
 
-        let path = req.uri().path().to_owned();
+        let (mut parts, body) = req.into_parts();
 
-        match self.node.at(&path) {
+        match self.node.at(parts.uri.path()) {
             Ok(match_) => {
                 let id = *match_.value;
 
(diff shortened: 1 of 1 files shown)


mladedav (approved):
Looks good.

yanns:
> I wonder whether we should run any benchmarking on this. In theory this could regress things as the request parts are moved around more (`into_parts`, `from_parts`).

I don't think this could regress things.

`let (mut parts, body) = req.into_parts();` is just a destructure of the Request:

```rust
    #[inline]
    pub fn into_parts(self) -> (Parts, T) {
        (self.head, self.body)
    }
```

and `Request::from_parts(parts, body)` is in the other direction:
```rust
    #[inline]
    pub fn from_parts(parts: Parts, body: T) -> Request<T> {
        Request { head: parts, body }
    }
```

Both should be no-op.

jplatte:
Both will involve moving things around on the stack, which is not a no-op. In many cases such moves are optimized out when compiling with optimizations, but it's not guaranteed.

yanns:
I don't have the expertise for this.
I was assuming that destructuring a tuple for example would be no-op.
And I'm assuming that avoiding duplicating the uri path is worth it to avoid allocations.

jplatte:
Okay, I'll look into it a bit later, if time permits. I think we have some existing benchmarks that I can run, probably need to add some more to distinguish short / long paths.

jplatte:
Ran benchmarks locally, the most pronounced result was an improvement for the only test that has a >8 byte long path.

Mixed results for other benchmarks, likely just noise.

jplatte (approved):

PR #2972, opened 2024-10-10
state always need to implement Clone

yanns:
Fix https://github.com/tokio-rs/axum/discussions/2821

### DIFF
--- a/axum/src/extract/state.rs
+++ b/axum/src/extract/state.rs
@@ -21,9 +21,6 @@ use std::{
 /// //
 /// // here you can put configuration, database connection pools, or whatever
 /// // state you need
-/// //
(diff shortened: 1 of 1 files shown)


jplatte (commented):
Honestly, looking at what's left: Wouldn't it maybe make more sense to just outright delete this section of the docs? The explanation for the `Clone` bound isn't really much of an explanation, is it?

And for other bounds, you also don't usually have any special docs to justify them. I think this section was only useful to have back when there were actually some circumstances when you didn't need to impl `Clone` for the toplevel state type.

Does that make sense to you?

yanns:
Yes I agree

jplatte (approved):

PR #2973, opened 2024-10-10
Avoid clone in layer

yanns:
Part of https://github.com/tokio-rs/axum/pull/2865

The first commit introduces a test to cover the state cloning
The second commit removes one clone

### DIFF
--- a/axum/src/routing/tests/mod.rs
+++ b/axum/src/routing/tests/mod.rs
@@ -942,7 +942,7 @@ async fn state_isnt_cloned_too_much_in_layer() {
 
     client.get("/").await;
 
-    assert_eq!(state.count(), 4);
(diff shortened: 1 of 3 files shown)


axum/src/routing/tests/mod.rs
```diff
@@ -924,6 +925,26 @@ async fn state_isnt_cloned_too_much() {
     assert_eq!(state.count(), 3);
 }
 
+#[crate::test]
+async fn state_isnt_cloned_too_much_in_layer() {
```
yanns, line 929: I'm not sure if it's the right place to put this test

jplatte: Fine by me.

jplatte (approved):

Issue #2964, opened 2024-10-07
Expose Websocket close frame when closing from server side


deep-gaurav:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

If I understand correctly, currently we cannot set close reason and code when closing websocket from server side, axum is already exposing `CloseCode` and `CloseFrame` for close msg from client side, but we cannot pass it when closing from server.

### Motivation

For me i want to reject some websocket connection but websocket spec doest let js get http status code and msg when connect to websocket and failing. Recommended method to do so is to accept connection and immediately close with reason according to https://stackoverflow.com/a/50685387. 
Currently I cannot do this. close method does not take any parameters.

### Proposal

To maintain backward compatibility we can expose another method, `close_with_msg(msg:CloseFrame)` which will pass it to tungstenite.

### Alternatives

maybe we can do `socket.send(axum::extract::ws::Message::Close(()))` before calling `socket.close()` but this seems weird to be, I dont know if sending close frame is also closing connection or not.

jplatte:
cc @SabrinaJewson

SabrinaJewson:
You can implement this yourself by sending a close message to the WebSocket.

This issue also caused me to discover that Axum’s current `.close()` method is incorrect, as it takes ownership of the WebSocket; in reälity, only the write half is closed and for true graceful shutdown one should read more messages from the read half.

I think I’ll make a PR to remove `.close()` (it’s redundant since you can send close messages) and update the documentation.

deep-gaurav:
Yeah I did discover that just sending CloseFrame message manually closes the connection. but sending a CloseMessage doesnt take ownership of socket so it's not apparent that it'll also shutdown the websocket connection. Could be just lack of knowledge on my part.

SabrinaJewson:
> but sending a CloseMessage doesnt take ownership of socket so it's not apparent that it'll also shutdown the websocket connection

So this is actually a good thing. The way the WebSocket closing protocol is supposed to work is that:
- Peer A sends the close frame to peer B, does not close the connection, and does not send further frames
- Peer B responds with a close frame, and does not send further frames
- Peer A receives any remaining frames from peer B as well as the close frame
- The connection is finally closed

If close methods take ownership, this protocol is impossible to implement correctly as peer A would not receive the remaining frames from peer B.

PR #2974, opened 2024-10-10
Remove `.close()`, and document clean WebSocket closing

SabrinaJewson:
## Motivation

The current `.close()` method is mostly useless, because it does not allow sending a close frame and takes ownership of the WebSocket, which even forbids implementing the normal [WebSocket close handshake](https://datatracker.ietf.org/doc/html/rfc6455#section-1.4).

## Solution

Remove `.close()`, and document the behaviour of WebSocket close handshakes.

Closes: #2964

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -623,7 +623,7 @@ pub enum Message {
     /// you may still read messages,
     /// but attempts to send another message will error.
     /// After receiving a close frame,
-    /// Axum will automatically respond with a close frame if necessary
+    /// axum will automatically respond with a close frame if necessary
     /// (you do not have to deal with this yourself).
     /// Since no further messages will be received,
     /// you may either do nothing
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -621,12 +621,12 @@ pub enum Message {
     ///
     /// After sending a close frame,
     /// you may still read messages,
-    /// but any attempts to send a message will error.
+    /// but any attempts to send another message will error.
     /// After receiving a close frame,
-    /// the server will automatically respond with a close frame if necessary
+    /// axum will automatically respond with a close frame if necessary
     /// (you do not have to deal with this yourself).
     /// Since no further messages will be received,
-    /// so you may either do nothing
+    /// you may either do nothing
     /// or explicitly drop the connection.
     Close(Option<CloseFrame<'static>>),
 }
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -507,11 +507,6 @@ impl WebSocket {
             .map_err(Error::new)
     }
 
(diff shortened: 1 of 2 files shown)


axum/src/extract/ws.rs
```diff
@@ -615,6 +610,24 @@ pub enum Message {
...
+    /// by simply dropping the [`WebSocket`].
+    /// However, you may also use the graceful closing protocol, in which
+    /// 1. peer A sends a close frame, and does not send any further messages;
+    /// 2. peer B responds with a close frame, and does not send any further messages;
+    /// 3. peer A processes the remaining messages sent by peer B, before finally
+    /// 4. both peers close the connection.
+    ///
+    /// After sending a close frame,
+    /// you may still read messages,
+    /// but any attempts to send a message will error.
```
jplatte, line 624:
```suggestion
    /// but any attempts to send another message will error.
```

axum/src/extract/ws.rs
```diff
@@ -615,6 +610,24 @@ pub enum Message {
...
+    /// After receiving a close frame,
+    /// the server will automatically respond with a close frame if necessary
```
jplatte, line 626: would it make sense to change this to:

```suggestion
    /// axum will automatically respond with a close frame if necessary
```

?

axum/src/extract/ws.rs
```diff
@@ -615,6 +610,24 @@ pub enum Message {
...
+    /// (you do not have to deal with this yourself).
+    /// Since no further messages will be received,
+    /// so you may either do nothing
+    /// or explicitly drop the connection.
```
jplatte, line 630:
```suggestion
    /// Since no further messages will be received,
    /// you may either do nothing
    /// or explicitly drop the connection.
```

jplatte (commented):
Thanks! Got a few questions / suggestions about the new docs.

axum/src/extract/ws.rs
```diff
@@ -615,6 +610,24 @@ pub enum Message {
...
+    /// 1. peer A sends a close frame, and does not send any further messages;
+    /// 2. peer B responds with a close frame, and does not send any further messages;
+    /// 3. peer A processes the remaining messages sent by peer B, before finally
+    /// 4. both peers close the connection.
+    ///
+    /// After sending a close frame,
+    /// you may still read messages,
+    /// but attempts to send another message will error.
+    /// After receiving a close frame,
+    /// Axum will automatically respond with a close frame if necessary
```
jplatte, line 626:
```suggestion
    /// axum will automatically respond with a close frame if necessary
```

axum is never capitalied :)

SabrinaJewson:
Ah. I capitalied “axum” based on other examples in the repo of it beïng capitalized. They should probably be removed, I guess.

jplatte:
Yeah, it should. In the readme it's consistently lowercase.

PR #2977, opened 2024-10-11
Remove `ServeFuture`

mladedav:
## Motivation

The custom future type did nothing, it just passed through the `poll` method to the inner boxed future. The only thing I could tell is different is that it implemented `Debug` but I'm not convinced that's something that a future type needs anyway.

The commit that added this custom future called this a preparation for graceful shutdown which has been since implemented and does not need this so I don't think this deletion would cause trouble later.

## Solution

I removed the code and let `Serve::into_future` return directly the `Pin<Box<dyn Future<Output = _>>>`.

### DIFF
--- a/axum/src/serve.rs
+++ b/axum/src/serve.rs
@@ -210,7 +210,7 @@ where
     S::Future: Send,
 {
     type Output = io::Result<()>;
-    type IntoFuture = private::ServeFuture;
+    type IntoFuture = futures_util::future::BoxFuture<'static, io::Result<()>>;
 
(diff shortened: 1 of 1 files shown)


jplatte:
This sort of thing is commonly done to allow changing the inner type in a non-breaking release. Is there any advantage to using `Pin<Box<_>>` directly, other than reducing the amount of code a bit?

mladedav:
I don't think there should be any difference, I just noticed the code doesn't actually do anything so I tried removing - it. We can keep it for the flexibility then.

PR #2978, opened 2024-10-11
Add `struct NoContent` as a self-described shortcut

oxalica:
## Motivation 

Refs: #2363, #2393

## Solution

We cannot change the behavior of `<() as IntoResponse>::into_response`, so let's add a newtype struct to explicitly use 204 No Content status.

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+- **added:** Add `NoContent` as a self-described shortcut for `StatusCode::NO_CONTENT` ([#2978])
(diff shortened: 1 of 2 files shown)


jplatte (approved):
Thanks!

axum/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+- **added:** Add `NoContent` as a self-described shortcut for `StatusCode::NO_CONTENT`.
```
jplatte, line 10: Can you add a reference to this PR here (like below) and remove the trailing `.`?

oxalica: Fixed.

PR #2991, opened 2024-10-16
Improve docs regarding state and extensions

### LINKED ISSUE #2917: make it easier to get from State to Extension
[`Router::with_state`](https://docs.rs/axum/latest/axum/routing/struct.Router.html#method.with_state) mentions `Extension` as an alternative to `State`, but only after scrolling past 4 examples.  [`extract::State`](https://docs.rs/axum/latest/axum/extract/struct.State.html) doesn't mention `Extension` at all.

extensions are more modular, as they can be added to various handlers while keeping other handlers pure.

Lachstec:
## Motivation
Issue #2917 mentions that it should be easier to discover the documentation for `Extension` from the docs of `State` and vice versa. Indeed, Extension is very easy to miss as it is only mentioned after a couple of examples.

## Solution
This PR updates the docs by 
- moving the section mentioning extensions up in `Router::with_state`
- Adding a short mention of `Extension` to `extract::State`

Closes #2917

### DIFF
--- a/axum/src/docs/routing/with_state.md
+++ b/axum/src/docs/routing/with_state.md
@@ -1,4 +1,5 @@
-Provide the state for the router.
+Provide the state for the router. State passed to this method is global and will be used
+for all requests this router receives. That means it is not suitable for holding state derived from a request, such as authorization data extracted in a middleware. Use [`Extension`] instead for such data.
 
 ```rust
 use axum::{Router, routing::get, extract::State};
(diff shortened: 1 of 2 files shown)


lolbinarycat:
"state derived from a request" seems like a fairly narrow usecase.  `Extension` is also useful when converting a previously stateless system into a stateful one, since changing the value of `State` also changes the type of the router, which can cause issues.

mladedav (approved):
> "state derived from a request" seems like a fairly narrow usecase. Extension is also useful when converting a previously stateless system into a stateful one, since changing the type of the State parameter also changes the type of the router, which can cause issues.

I agree with how the docs are written now. `Extension` has been historically used for passing around application state, but users should use `State` for that now. It's type safe and more maintainable.

I'd recommend anyone who needs to make their application stateful to use `State` instead of `Extension` even though they will have to make more changes if they're passing around `Router`s. I can't think of any scenarios other than the one in docs where I'd recommend an `Extension`.

lolbinarycat:
well, i genuinely could not get `State` to function properly when i tried.  does every handler need to accept a `State` parameter for it to work?

jplatte:
@lolbinarycat No, that's not required. Please open a new discussion under [Q&A](https://github.com/tokio-rs/axum/discussions/categories/q-a) to get help with that.

PR #2992, opened 2024-10-16
fix(sse): skip sse incompatible chars of `serde_json::RawValue`

### LINKED ISSUE #2981: `Event::json_data` can construct an invalid SSE event by passing a `serde_json::value::RawValue`

## Bug Report

When using `Event::json_data` with a `serde_json::value::RawValue` it is possible to construct an invalid SSE event.
This happens because `RawValue` will also capture newlines between object entries as it only validates that the bytes are valid json.
This means that the string `"{\n\"foo\":\"bar\"}"` when deserialized into a `RawValue` and serialized with `Event::json_data` will not be parsed correctly as internally it gets written to the buffer using `serde_json::to_writer`.

### Reproduction

Looking at the output of `dbg!` of the bytes should be enough to see the issue but I thought I would verify it again using a SSE parsing lib. 

```rs
//! ```cargo
//! [dependencies]
//! axum = "0.7"
//! eventsource-stream = "0.2"
//! serde_json = { version = "1", features = ["raw_value"]}
//! futures = "0.3"
//! ```

use axum::response::{sse::Event, IntoResponse, Sse};
use eventsource_stream::Eventsource;
use futures::TryStreamExt;
use serde_json::value::RawValue;
use std::error::Error;

fn main() -> Result<(), Box<dyn Error>> {
    let data = "{\n\"foo\":\"bar\"}";
    let var_name = serde_json::from_str::<Box<RawValue>>(data)?;
    let valid_event = Event::default().data(var_name.get());
    let invalid_event = Event::default().json_data(&var_name)?;
    let stream = Sse::new(futures::stream::iter([
        Ok::<_, std::convert::Infallible>(valid_event),
        Ok(invalid_event),
    ]))
    .into_response()
    .into_body()
    .into_data_stream()
    .inspect_ok(|b| {
        dbg!(String::from_utf8_lossy(b.as_ref()));
    })
    .eventsource();
    let events = futures::executor::block_on(stream.try_collect::<Vec<_>>())?;
    for e in events {
        assert_eq!(
            serde_json::from_str::<serde_json::Value>(&dbg!(e.data))?,
            serde_json::from_str::<serde_json::Value>(data)?
        );
    }
    Ok(())
}
```

## Solutions

1. Document that all use (also nested) of `serde_json::value::RawValue` could lead to invalid Events (easy)
2. Internally use `serde_json::to_string` instead and than call to `Event::data` which will split the newlines correctly into multiple data fields. (Slow)
3. Continue using `serde_json::to_writer` but wrap buffer in some custom writer which will skip newlines which should be fine as newlines are guaranteed to be escaped in json strings. (should be pretty fast still)

I would be interested in making a PR for 3. if that is the desired solution.

Threated:
As discussed the writer will skip `\n` and also `\r` as I noticed it can also be a valid character between json object keys.

Closes #2981

### DIFF
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -117,7 +117,7 @@ quickcheck = "1.0"
 quickcheck_macros = "1.0"
 reqwest = { version = "0.12", default-features = false, features = ["json", "stream", "multipart"] }
 serde = { version = "1.0", features = ["derive"] }
-serde_json = "1.0"
+serde_json = { version = "1.0", features = ["raw_value"] }
 time = { version = "0.3", features = ["serde-human-readable"] }
 tokio = { package = "tokio", version = "1.25.0", features = ["macros", "rt", "rt-multi-thread", "net", "test-util"] }
 tokio-stream = "0.1"
--- a/axum/src/response/sse.rs
+++ b/axum/src/response/sse.rs
@@ -208,12 +208,29 @@ impl Event {
     where
         T: serde::Serialize,
     {
+        struct IgnoreNewLines<'a>(bytes::buf::Writer<&'a mut BytesMut>);
+        impl std::io::Write for IgnoreNewLines<'_> {
+            fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+                let mut last_split = 0;
+                for delimiter in memchr::memchr2_iter(b'\n', b'\r', buf) {
+                    self.0.write_all(&buf[last_split..delimiter])?;
+                    last_split = delimiter + 1;
+                }
+                self.0.write_all(&buf[last_split..])?;
+                Ok(buf.len())
+            }
+
+            fn flush(&mut self) -> std::io::Result<()> {
+                self.0.flush()
+            }
+        }
         if self.flags.contains(EventFlags::HAS_DATA) {
             panic!("Called `EventBuilder::json_data` multiple times");
         }
 
         self.buffer.extend_from_slice(b"data: ");
-        serde_json::to_writer((&mut self.buffer).writer(), &data).map_err(axum_core::Error::new)?;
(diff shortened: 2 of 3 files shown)


Threated:
CI failure seems to be unrelated

axum/Cargo.toml
```diff
@@ -42,36 +42,36 @@ __private_docs = ["tower/full", "dep:tower-http"]
 
 [dependencies]
 axum-core = { path = "../axum-core", version = "0.5.0-alpha.1" }
+
+# optional dependencies
```
mladedav, line 46: Please remove the changes to this section. It was formatted like this intentionally and the comment especially doesn't make sense now.

axum/src/response/sse.rs
```diff
@@ -213,7 +213,24 @@ impl Event {
         }
 
         self.buffer.extend_from_slice(b"data: ");
-        serde_json::to_writer((&mut self.buffer).writer(), &data).map_err(axum_core::Error::new)?;
+        struct IgnoreNewLines<'a>(bytes::buf::Writer<&'a mut BytesMut>);
```
mladedav, line 216: Move this to the top of the function please. I think it's more readable to first have the helper and then the data operations, right now it's in the middle of the write operations which is a bit confusing.

mladedav (changes requested):
Thanks for fixing this :+1: 

Just two minor nits

axum/Cargo.toml
Threated: Ups I ran cargo sort because the CI failed on cargo sort did not notice it changed all of these.

mladedav (approved):
Looks good, thanks.

I'll merge this once we get the CI working again.

jplatte (commented):
Mind adding a changelog entry?

PR #2993, opened 2024-10-16
Fixes TSRs redirecting to toplevel inside nested Router

erin-desu:
Originally, if you used `route_with_tsr` (let's say `/nyan/`) inside nested router (let's say `/neko`), it got redirected to `/nyan/`, where `/neko/nyan/` was expected.

Fixed using `OriginalUri` instead of plain `Uri`.

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -39,7 +39,7 @@ typed-header = ["dep:headers"]
 typed-routing = ["dep:axum-macros", "dep:percent-encoding", "dep:serde_html_form", "dep:form_urlencoded"]
 
 [dependencies]
-axum = { path = "../axum", version = "0.8.0-alpha.1", default-features = false }
+axum = { path = "../axum", version = "0.8.0-alpha.1", default-features = false, features = ["original-uri"] }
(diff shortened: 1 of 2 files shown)


erin-desu:
CI failure is not my fault :sob:

mladedav (approved):
Looks good to me, we'll take care of the CI.

@jplatte this adds a dependency from `axum-extra` on `axum`'s `original-uri` feature which adds one uri clone and adds a request extension to each request. I think that should be fine but wanted to doublecheck with you.

jplatte:
Hm, I find that a bit awkward. Maybe we should move `route_with_tsr` behind a feature flag (can be on by default)? Doesn't have to be in this PR though.

erin-desu:
> Hm, I find that a bit awkward. Maybe we should move `route_with_tsr` behind a feature flag (can be on by default)? Doesn't have to be in this PR though.

Well, we are getting Axum 0.8 soon but it's already in alpha and idk if that's a change suited for it? If so i can add it in the fork.

jplatte:
Let's not worry about it in this PR. The way it is now, it can be backported to 0.7.

jplatte:
This has been released as part of axum-extra 0.9.5.

PR #2994, opened 2024-10-16
chore: fix new clippy lint

mladedav:
## Motivation

`beta` has a new lint about eliding lifetimes in impls.

## Solution

Elided the lifetimes as suggested by clippy.

I've also checked that nightly doesn't trigger any new lints so we should be ok for 3 months again.

### DIFF
--- a/axum-extra/src/extract/cookie/private.rs
+++ b/axum-extra/src/extract/cookie/private.rs
@@ -289,7 +289,7 @@ struct PrivateCookieJarIter<'a, K> {
     iter: cookie::Iter<'a>,
 }
 
-impl<'a, K> Iterator for PrivateCookieJarIter<'a, K> {
+impl<K> Iterator for PrivateCookieJarIter<'_, K> {
     type Item = Cookie<'static>;
 
     fn next(&mut self) -> Option<Self::Item> {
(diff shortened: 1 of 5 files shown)


jplatte (approved):
Thanks!

mladedav:
There are still some issues with linking an aws sys library in rustls examples. I'm not sure what's the issue, I'll see if I can make it work tomorrow.

mladedav:
What I think has happened is our cache does not take into account the runner and in the last month GitHub switched the `ubuntu-latest` runner from 22.04 to 24.04. Our cache looks at the hash of our dependencies and the compiler so it reused older cache which was originally used for the beta toolchain since it's the same compiler as the current stable toolchain. But the `-sys` library used older libc in the 22.04 image for compilation and some functions then weren't available for linking in the 24.04.

Anyway, that's just my theory. I've deleted the one cache entry and it runs now, so hopefully it's going to be fine going forward.

jplatte:
That makes sense, I heard that they updated `-latest` to 24.04 but rolled it back because it broke so many things. Want to open up a follow-up PR to go to `-24.04` explicitly? Probably not a great idea to use the unversioned one anywhere.

PR #2996, opened 2024-10-18
Explicit CI runner version

mladedav:
## Motivation

[Comment with context](https://github.com/tokio-rs/axum/pull/2994#issuecomment-2419223788)

We are currently using `ubuntu-latest` runners which may change and usually do change every two years. This can lead to issues as in #2994 where caches can be reused in a different version of ubuntu which can then break linking.

## Solution

We set the runner version explicitly. This does not prevent similar issue as in the linked PR, but it will happen in the PR that upgrades the runner instead of random other PR that happened to run actions after the change.

I've also included the `ubuntu-24.04` string in the cache key so that the issue doesn't happen again. This makes migration harder because this should be changed when a runner changes and I have not found a good way to deduplicate this because of [this issue](https://github.com/actions/runner/issues/480). But if both strings are correct, we will never use cache with objects built with older libraries.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v4
     - uses: taiki-e/install-action@protoc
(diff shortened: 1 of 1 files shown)


jplatte (approved):

jplatte:
Small nitpick: Commit messages should start in uppercase (after the optional `prefix:`).

mladedav:
Ok. I know they should normally start with upper case, but it seemed to me it's usually not done after the prefix. For example conventional commits don't do it either in their [examples](https://www.conventionalcommits.org/en/v1.0.0/#examples). But I don't mind either way and I'll try to keep it in mind and use upper case here.

PR #2997, opened 2024-10-19
Upgrade matchit

jplatte:
## Motivation

A few patch versions have been release since 0.8.0, but the anticipated problematic change (https://github.com/ibraheemdev/matchit/pull/61) has not landed yet.

## Solution

Bump the version requirement.


mladedav:
The CI failed because of change in casing of a panic message.

Otherwise looks good, let's do it.

mladedav (commented):
The way you had it before, we wouldn't upgrade to 0.8.5, but with the current change we would.

I just checked that it should be fine with the PR with suffixes (or at least tests seem to pass), but users still won't be able to use the suffixes because we make some checks regarding captures like making sure that the whole path segment is captured.

So I just want to make sure that this is intended.

axum/Cargo.toml
```diff
@@ -48,7 +48,7 @@ http = "1.0.0"
 http-body = "1.0.0"
 http-body-util = "0.1.0"
 itoa = "1.0.5"
-matchit = "=0.8.0"
+matchit = "0.8.4"
```
jplatte, line 51:
```suggestion
matchit = "=0.8.4"
```

jplatte:
Wasn't intended 😅
I just needed to raise the minimum version because our MSRV CI job checks that tests pass with minimum versions of our dependencies.

mladedav (approved):

mladedav:
Thanks, we tried the new version in #3140 but decided along with the `matchit` maintainer to wait for 0.9 instead which seems to be fairly close. See that issue, the PR linked there, and the `matchit` issue linked in our issue for more details.

We use precise version `=0.8.4` here so the fact that it has been released is safe for `axum` users even if it added functionality we wouldn't be able to handle.

Issue #2998, opened 2024-10-19
Spoofable extractors are used with the knowledge of the risks


yanns:
The ticket follows the discussion in https://github.com/tokio-rs/axum/pull/2507#issuecomment-2423925900

Some extractors, like `Host` or `Scheme`, can use the values of some HTTP headers that could be spoofed by malicious users.

We should find a way to make users aware of the risks of using those extractors.

Some ideas:
- using `unsafe`. This is not the idea of `unsafe` and we would be mis-using it. I think that this can be discarded.
- encapsulating the value in a new struct like `SpoofableValue` so that users have to call some function to get the value. The name and the documentation of the function should make the user aware of the risk. Example:
```rust
async fn handler(Host(host): Host) -> String {
  val value = host.spoofable_value();
  value
}
```

bengsparks:
Perhaps something along the lines of:

```rs
/// Wrap spoofable extractor
pub struct Spoofable<E>(pub E);

/// Allow `Spoofable` to be used with spoofable extractors in handlers
impl <S, E> FromRequestParts<S> for Spoofable<E> where E: FromSpoofableRequestParts<S> {

}

/// axum private trait
trait FromSpoofableRequestParts<S>: Sized {
    type Rejection: IntoResponse;

    async fn from_request_parts(
        parts: &mut Parts, 
        state: &S
    ) -> impl Future<Output = Result<Self, Self::Rejection>> + Send;
}

/// Mark `Host` as a spoofable extractor
impl <S> FromSpoofableRequestParts<S> for Host { ... } 

/// Use spoofable extractor
async fn handler(Spoofable(Host(host)): Spoofable<Host>) -> String {
    println("{host}");
}
```

yanns:
I've made one PoC so that we can better imagine how the API would be: https://github.com/tokio-rs/axum/pull/3000

@bengsparks could you also make one for the approach you're suggesting. It seems very interesting!

mladedav:
Is it possible to add on either of the extractors something like `Host::unspoofable_value(&self) -> Option<String>`?

I don't think host can be extracted from anything that cannot be spoofed and scheme could theoretically be extracted from connect info, but the way it is implemented now, it prefers the scheme the client used originally if the server is behind a proxy, i.e. it tries to extract from the proxy headers first which might be what the user is interested in.

[...]

PR #3000, opened 2024-10-20
PoC of introducing SpoofableValue

yanns:
PoC to check which solution to pick for https://github.com/tokio-rs/axum/issues/2998

### DIFF
--- a/axum-extra/src/extract/mod.rs
+++ b/axum-extra/src/extract/mod.rs
@@ -63,3 +63,16 @@ pub use crate::json_lines::JsonLines;
 #[cfg(feature = "typed-header")]
 #[doc(no_inline)]
 pub use crate::typed_header::TypedHeader;
+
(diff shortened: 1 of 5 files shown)


axum-extra/src/extract/mod.rs
```diff
@@ -63,3 +63,16 @@ pub use crate::json_lines::JsonLines;
...
+
+#[derive(Debug, Clone)]
+pub struct SpoofableValue(String);
+
+impl SpoofableValue {
+    pub fn new(value: String) -> Self {
+        Self(value)
+    }
+
+    pub fn spoofable_value(self) -> String {
```
yanns, line 75: Here we could add documentation to make users aware of risks using this method.

davidpdrsn:
Do we still need this given https://github.com/tokio-rs/axum/issues/3442?

yanns:
no, I'm closing this.

PR #3001, opened 2024-10-20
PoC of introducing `FromSpoofableRequestParts` and `Spoofable`

bengsparks:
## Motivation

PoC to check which solution to pick for https://github.com/tokio-rs/axum/issues/2998.
Attempts to come as close to the original idea of extractors as possible, without increasing implementation complexity.

## Solution

Add a new trait `FromSpoofableRequestParts` that is private to `axum-extra`.
This trait is only to be implemented by extractors that read from spoofable portions of HTTP requests.
No extractor should ever implement both `FromSpoofableRequestParts` and `FromRequestsParts`.

A blanket implementation of `FromRequestsParts` is provided by `Spoofable` for all extractors that implement `FromSpoofableRequestParts`, so that reading via destructuring i.e. `Spoofable(Extractor(inner))` can continue to be used in handlers.

```rs
use axum_extra::extract::{Scheme, Spoofable};
use tokio::net::TcpListener;

#[tokio::main]
async fn main() {
    let app = Router::new().route("/", get(f));

}

async fn f(Spoofable(Scheme(scheme)): Spoofable<Scheme>) -> String {
    scheme
}
```

See `examples/spoofable-scheme` for a short demonstration.

### DIFF
--- a/axum-extra/src/extract/mod.rs
+++ b/axum-extra/src/extract/mod.rs
@@ -26,7 +26,7 @@ pub struct Spoofable<E>(pub E);
 impl<S, E> FromRequestParts<S> for Spoofable<E>
 where
     E: spoof::FromSpoofableRequestParts<S>,
-    S: Sync
+    S: Sync,
 {
     type Rejection = E::Rejection;
 
--- a/axum-extra/src/extract/mod.rs
+++ b/axum-extra/src/extract/mod.rs
@@ -6,6 +6,38 @@ mod optional_path;
 pub mod rejection;
 mod with_rejection;
 
+/// Private mod, public trait trick
+mod spoof {
+    pub trait FromSpoofableRequestParts<S>: Sized {
+        type Rejection: axum::response::IntoResponse;
+
+        fn from_request_parts(
+            parts: &mut axum::http::request::Parts,
+            state: &S,
+        ) -> impl std::future::Future<Output = Result<Self, Self::Rejection>> + Send;
+    }
+}
+
+/// Wrap spoofable extractor
+#[derive(Debug)]
+pub struct Spoofable<E>(pub E);
+
+/// Allow `Spoofable` to be used with spoofable extractors in handlers
+impl<S, E> FromRequestParts<S> for Spoofable<E>
+where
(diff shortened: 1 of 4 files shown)


yanns:
Thanks!
How would we force `Host` and `Schema` to use this? We would remove the `FromRequestParts` for them right?

bengsparks:
The traits share the same function signatures and bounds, so simply exchanging the trait name within the `impl` block is sufficient.

jplatte:
How about `Host<WithProxyHeaders>` and `Host<WithoutProxyHeaders>` as an alternative? I find "spoofable" sounds a bit awkward, and while the proxy thing may not sound as dangerous, it would still get people thinking.

yanns:
To finish this PR: could you use this for `Host` and `Schema`?
In the documentation, should we link to https://owasp.org/www-community/pages/attacks/ip_spoofing_via_http_headers? or is it better not to have any link, but only a description? We could re-use some of the content of the page:

> Malicious actors utilize spoofing to inject payloads via HTTP headers, leading to generating inaccurate logs or inject malicious payloads via HTTP headers.
> As remediation, always validate and sanitize the extract value:
> - Implement strict input validation to ensure that client-provided data, including headers, is legitimate.
> - Sanitize input to remove potentially harmful content.

bengsparks:
> To finish this PR: could you use this for `Host` and `Schema`?
Yes; I added an example to the PR that showcases `Spoofable<Scheme>`.
The implementation for `Host` would be similar.

> In the documentation, should we link to https://owasp.org/www-community/pages/attacks/ip_spoofing_via_http_headers? or is it better not to have any link, but only a description? We could re-use some of the content of the page:
> 
> > Malicious actors utilize spoofing to inject payloads via HTTP headers, leading to generating inaccurate logs or inject malicious payloads via HTTP headers.
> > As remediation, always validate and sanitize the extract value:
> > - Implement strict input validation to ensure that client-provided data, including headers, is legitimate.
> > - Sanitize input to remove potentially harmful content.
Those bullet points sound good enough! Maybe add an inline example in the docs that uses `scheme.parse<ExpectedSchemeEnum>` to validate the extracted Scheme

PR #3012, opened 2024-11-01
Fix missing `post` imports in `router` docs example

noxxxxxious:
## Motivation
When going through the router docs, I noticed that some of the examples were using post but were not importing them (as if the import statement was copy pasted from a section that didn't use `post`).

## Solution
Add the `post` import into the examples where it is being used.

### DIFF
--- a/axum/src/docs/routing/route.md
+++ b/axum/src/docs/routing/route.md
@@ -80,7 +80,7 @@ To accept multiple methods for the same route you can add all handlers at the
 same time:
 
 ```rust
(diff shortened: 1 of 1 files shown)


jplatte:
All the examples you modified are tested and work already. None of them use the free-standing `post` function, they all use the `post` method of `MethodRouter`.

noxxxxxious:
Ah! Sorry about that! Thought I had caught something that was overlooked. Please pardon my noobiness.

jplatte:
No worries.

PR #3014, opened 2024-11-02
Remove once_cell from jwt example

Hmikihiro:
## Motivation

I think using `std::sync::LazyLock` can reduce dependency `once_cell`.


yanns (approved):

fr0staman:
`std::sync::LazyLock` MSRV is 1.80.0, which `axum` MSRV with 0.8.0 is 1.75.0.

Maybe in future, but not right now, maybe

ttys3:
since Rust it self growing fast, I think following newest Rust in main branch is good

jplatte:
I think we already don't enforce an MSRV other than stable for examples. The CI failure looks unrelated.

PR #3030, opened 2024-11-16
axum-extra: Add links to features table

Turbo87:
## Motivation

This makes it a bit easier to find the referenced types when looking at the features table.


jplatte (approved):
Thanks!

PR #3031, opened 2024-11-16
fix: avoid setting content-length before middleware (backport)

SabrinaJewson:
Backports #2897.


SabrinaJewson:
CI failure looks unrelated…?

jplatte:
Just cherry-picked the fix for that CI issue (we had it on main too). Should work now..

jplatte (approved):
Thanks!

PR #3033, opened 2024-11-16
axum/routing: Add `content-length` header assertions to `middleware_adding_body` test

Turbo87:
I noticed in our crates.io test suite, that the `content-length` header was no longer set on responses after [the latest update](https://github.com/rust-lang/crates.io/pull/9969) and wondered if https://github.com/tokio-rs/axum/pull/2897 had broken it. I quickly extended the corresponding test to check, and it looks like everything is fine.

> I noticed in our crates.io test suite, that the `content-length` header was no longer set on responses

We are using `router.oneshot(request).await` over there, which seems to have changed behavior due to the PR linked above. The axum test suite is using real `reqwest` calls though, which don't appear to show the same behavior.

### DIFF
--- a/axum/src/routing/tests/mod.rs
+++ b/axum/src/routing/tests/mod.rs
@@ -1091,5 +1091,11 @@ async fn middleware_adding_body() {
 
     let client = TestClient::new(app);
     let res = client.get("/").await;
+
+    let headers = res.headers();
+    let header = headers.get("content-length");
+    assert!(header.is_some());
+    assert_eq!(header.unwrap().to_str().unwrap(), "3");
+
     assert_eq!(res.text().await, "…");
 }


Turbo87:
looks like the change is not just in our (crates.io) test suite, but also when running the server regularly (see https://github.com/tokio-rs/axum/pull/2897#issuecomment-2480808712) 🤔

mladedav (approved):

mladedav:
Can you also check this on the 0.7.x branch (or specifically on the release commit) before the merge to main? That's where the release was created so maybe it's broken there and after merging back to main it works ok.

In either case I'll take a look tomorrow unless someone else figures this out in the meantime.

Turbo87:
> Can you also check this on the 0.7.x branch

seems to pass on that branch as well.

we use quite a few middlewares in the crates.io codebase, so maybe that somehow makes this trigger while it doesn't in the tests here?

mladedav:
I've added a bit of description to the `crates.io`'s [upgrade PR](https://github.com/rust-lang/crates.io/pull/9969) but basically the compression layer (and possibly others) started using chunked transfer instead of content type. More info will be in #3035 

Thanks again for notifying us.

PR #3046, opened 2024-11-23
Add Attachment type in features documentation

joeydewaal:
## Motivation
In #2789 I added an `Attachment` response type but forgot to add this feature to the docs.


mladedav (approved):
Thanks for the follow up.

PR #3047, opened 2024-11-24
Add an encapsulated file stream in axum-extra to make it more conveni…

YanHeDoki:
## Motivation
In use will need to return a file stream, encapsulates a file stream structure to save the return time to add the header

## Solution

In axum-extra add a response structure, the construction needs to pass a file stream, in the response will add the header: “application/octet-stream”, I added the documentation and examples these will be more detailed

```rust
 use axum::{
     http::StatusCode,
     response::{Response, IntoResponse},
     Router,
     routing::get
 };
use tokio::net::TcpListener;
 use axum_extra::response::file_stream::FileStream;
 use tokio::fs::File;
 use tokio_util::io::ReaderStream ;

#[tokio::main]
async fn main() {
    let app = Router::new().route("/", get(file_stream));
}

 async fn file_stream() -> Result<Response, (StatusCode, String)> {    
     let     stream=ReaderStream::new(File::open("file_path").await.map_err(|e| (StatusCode::NOT_FOUND, format!("File not found: 
       {e}")))?);
     let file_stream_resp = FileStream::new(stream)
         .file_name("You want to set the file name");

     Ok(file_stream_resp.into_response())
 }
```
See examples/stream-to-file for a short demonstration  I've added usage to this example.

**I just started using rust, so some of the code may not be well written**


axum-extra/src/response/file_stream.rs
```diff
@@ -0,0 +1,212 @@
...
+        // Validate Response Headers
+        assert_eq!(
+            response.headers().get("content-type").unwrap(),
+            "application/octet-stream"
+        );
+        assert_eq!(
+            response.headers().get("content-disposition").unwrap(),
+            "attachment; filename=\"CHANGELOG.md\""
+        );
+        assert_eq!(response.headers().get("content-length").unwrap(), "8098");
```
yanns, line 209: This will change each time we change the `CHANGELOG.md` file no?
It'd be better to add a specific file only for this test?

examples/stream-to-file/Cargo.toml
```diff
@@ -11,3 +11,5 @@ tokio = { version = "1.0", features = ["full"] }
 tokio-util = { version = "0.7", features = ["io"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+async-stream = "0.3"
+axum-extra = { path = "../../axum-extra", features = ["fileStream"] }
```
yanns, line 15: Please add a line ending.

axum-extra/src/response/file_stream.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// Create a file stream.
+    pub fn new(stream: S) -> Self {
+        Self {
+            stream,
+            file_name: None,
+            content_size: None,
+        }
+    }
+
+    /// Set the file name of the file.
```
yanns, line 64: Should we add a `from_file` function where we read the file, set the name and the content size?

axum-extra/src/response/file_stream.rs
YanHeDoki: Sorry for the writing error here, should have actively gotten the length once

axum-extra/src/response/file_stream.rs
YanHeDoki: Filenames and capitalization are not mandatory and the parameter is a stream, so I didn't do it.

examples/stream-to-file/Cargo.toml
YanHeDoki: I reformatted the code and pushed it, I'm rust beginner not very familiar with these ci checks yet, sorry!

axum-extra/src/response/file_stream.rs
```diff
@@ -0,0 +1,271 @@
...
+    ///     response::{Response, IntoResponse},
+    ///     Router,
+    ///     routing::get
+    /// };
+    /// use axum_extra::response::file_stream::FileStream;
+    /// use std::path::PathBuf;
+    /// use tokio_util::io::ReaderStream;
+    /// use tokio::fs::File;
+    /// async fn file_stream() -> Result<Response, (StatusCode, String)> {
+    ///     Ok(FileStream::<ReaderStream<File>>::from_path(PathBuf::from("test.txt"))
```
yanns, line 85: I'm not sure we need the `Ok(...)`

examples/stream-to-file/src/main.rs
```diff
@@ -84,6 +91,120 @@ async fn show_form() -> Html<&'static str> {
...
+    FileStream::<AsyncReaderStream>::from_path("./CHANGELOG.md".into())
+        .await
+        .map_err(|_| (StatusCode::INTERNAL_SERVER_ERROR, "Failed to open file").into_response())
+        .into_response()
+}
+
+/// If you want to control the returned files in more detail you can implement a Stream
+/// For example, use the try_stream! macro to construct a file stream and set which parts are needed.
+async fn file_download_handler() -> Response {
+    let file_stream = match try_stream("./CHANGELOG.md", 5, 25, 10).await {
```
yanns, line 147: `"./CHANGELOG.md"` will only works if we start the example from some directly, but I guess it's acceptable. We don't want to make the examples too complex.

axum-extra/src/response/file_stream.rs
YanHeDoki: > I'm not sure we need the `Ok(...)`

Yes, it was unnecessary. I removed it.

axum-extra/Cargo.toml
```diff
@@ -15,6 +15,7 @@ version = "0.10.0-alpha.1"
 default = ["tracing", "multipart"]
 
 async-read-body = ["dep:tokio-util", "tokio-util?/io", "dep:tokio"]
+fileStream = ["dep:tokio-util", "tokio-util?/io", "dep:tokio"]
```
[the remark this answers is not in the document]
YanHeDoki: > `error_response`

I changed it to ` kebab-case`

examples/stream-to-file/src/main.rs
YanHeDoki: > `"./CHANGELOG.md"` will only works if we start the example from some directly, but I guess it's acceptable. We don't want to make the examples too complex.

Most of the time “. /CHANGELOG.md” exists, tried creating a file when the use example is started, it's too complicated for the example and may leave the file behind.

yanns: What I mean is if I run `cd examples/stream-to-file && cargo run` then the path `"./CHANGELOG.md"` is not valid.

YanHeDoki: > What I mean is if I run `cd examples/stream-to-file && cargo run` then the path `"./CHANGELOG.md"` is not valid.

Yes, perhaps it's better to create a file before launching the serve

axum-extra/Cargo.toml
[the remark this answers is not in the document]
YanHeDoki: > error_response

But I don't know if error_response is already used, so I didn't change error_response

examples/stream-to-file/src/main.rs
YanHeDoki: > What I mean is if I run `cd examples/stream-to-file && cargo run` then the path `"./CHANGELOG.md"` is not valid.

Now that I've modified the example, I'll create a file to test it first

axum-extra/src/response/file_stream.rs
```diff
@@ -0,0 +1,270 @@
...
+                let file_content = b"Hello, this is the simulated file content!".to_vec();
+                let size = file_content.len() as u64;
+                let reader = Cursor::new(file_content);
+
+                // response file stream
+                let stream = ReaderStream::new(reader);
+                FileStream::new(stream)
+                    .file_name("test")
+                    .content_size(size)
+                    .into_response()
```
yanns, line 187: Should we add more tests:
- one where the filename is not set
- one where the content size is not set
- one where the filename and the content size are not set

YanHeDoki: > Should we add more tests:
> 
> * one where the filename is not set
> * one where the content size is not set
> * one where the filename and the content size are not set

Yeah, I added more tests.

axum-extra/Cargo.toml
```diff
@@ -67,7 +68,7 @@ prost = { version = "0.13", optional = true }
 serde_html_form = { version = "0.2.0", optional = true }
 serde_json = { version = "1.0.71", optional = true }
 serde_path_to_error = { version = "0.1.8", optional = true }
-tokio = { version = "1.19", optional = true }
+tokio = { version = "1.19", optional = true, features = ["fs"] }
```
paolobarbolini, line 71: Shouldn't this feature only be enabled when `file-stream` is enabled?

YanHeDoki:
> Can you add an entry in the https://github.com/tokio-rs/axum/blob/main/axum-extra/CHANGELOG.md ?

Yes, I added it.

axum-extra/Cargo.toml
YanHeDoki: > Shouldn't this feature only be enabled when `file-stream` is enabled?

Yes, I fixed it.

axum-extra/src/response/file_stream.rs
```diff
@@ -0,0 +1,391 @@
...
+{
+    fn into_response(self) -> Response {
+        let mut resp = Response::builder().header(header::CONTENT_TYPE, "application/octet-stream");
+
+        if let Some(file_name) = self.file_name {
+            resp = resp.header(
+                header::CONTENT_DISPOSITION,
+                format!("attachment; filename=\"{}\"", file_name),
+            );
+        };
```
jplatte, line 145:
```suggestion
        }
```

axum-extra/src/response/file_stream.rs
```diff
@@ -0,0 +1,391 @@
...
+
+        if let Some(content_size) = self.content_size {
+            resp = resp.header(header::CONTENT_LENGTH, content_size);
+        };
```
jplatte, line 149:
```suggestion
        }
```

jplatte (commented):
There's a few things I'd like to see improved before we ship this, but maybe it's easier if we merge now and I post a follow-up PR. It's nothing big.

YanHeDoki:
> improved

Okay, I improved it. Can you review it again？

YanHeDoki:
> There's a few things I'd like to see improved before we ship this, but maybe it's easier if we merge now and I post a follow-up PR. It's nothing big.

Okay, I improved it. Can you review it again？

axum-extra/src/response/mod.rs
```diff
@@ -12,6 +12,10 @@ pub mod multiple;
 #[cfg(feature = "error_response")]
 mod error_response;
 
+#[cfg(feature = "file-stream")]
+/// Module for handling file streams.
+pub mod file_stream;
```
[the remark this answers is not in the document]
YanHeDoki: > The main type from this module should also be re-exported from this module.
I improved it, but I'm not very proficient in this part yet, is the modification right?

ttys3:
the usage of this PR feature seems rather limited.

for example, it seems that it does not handle [HTTP range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests)

and it limited the response header. it fixed to `application/octet-stream`  and `attachment; filename=xxx`

and even in download mode, it does not support inline, only support attachment model.

people may want to steam a video file, and need HTTP range requests support.   `application/octet-stream` is too limited.

as a framework, the feature builtin should be more general.

I suggest put this as an external crate, if it is limited to  `application/octet-stream`

what do you think? @yanns @jplatte

yanns:
@ttys3 yes I see your point.
It's a bit sad for @YanHeDoki to receive this feedback after many iterations, but I guess it's the right decision.
@YanHeDoki would you mind putting this code into an external crate and add it to the [ecosystem](https://github.com/tokio-rs/axum/blob/main/ECOSYSTEM.md)?

axum-extra/src/response/file_stream.rs
```diff
@@ -0,0 +1,391 @@
...
+
+        if let Some(content_size) = self.content_size {
+            resp = resp.header(header::CONTENT_LENGTH, content_size);
+        }
+
+        resp.body(body::Body::from_stream(self.stream))
+            .unwrap_or_else(|e| {
+                (
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    format!("build FileStream responsec error:{}", e),
```
paolobarbolini, line 155: Typo.

```suggestion
                    format!("build FileStream response error: {}", e),
```

axum-extra/src/response/file_stream.rs
```diff
@@ -0,0 +1,391 @@
...
+    /// use tokio::fs::File;
+    /// use tokio_util::io::ReaderStream;
+    /// async fn file_stream() -> Response {
+    ///     FileStream::<ReaderStream<File>>::from_path(PathBuf::from("test.txt"))
+    ///     .await
+    ///     .map_err(|e| (StatusCode::NOT_FOUND, format!("File not found: {e}")))
+    ///     .into_response()
+    /// }
+    /// let app = Router::new().route("/FileStreamDownload", get(file_stream));
+    /// # let _: Router = app;
```
paolobarbolini, line 90: I've made the code cleaner by looking at the `Attachment` example. I haven't tried it locally if it builds

```suggestion
    /// use std::path::Path;
    ///
    /// use axum::{
    ///     http::StatusCode,
    ///     response::{Response, IntoResponse},
    ///     Router,
    ///     routing::get
    /// };
    /// use axum_extra::response::file_stream::FileStream;
    /// use tokio::fs::File;
    /// use tokio_util::io::ReaderStream;
    ///
    /// async fn file_stream() -> Result<FileStream<ReaderStream<File>>, (StatusCode, String)> {
    ///     FileStream::from_path(&Path::new("Cargo.toml"))
    ///         .await
    ///         .map_err(|e| (StatusCode::NOT_FOUND, format!("File not found: {e}")))
    /// }
    ///
    /// let app = Router::new().route("/Cargo.toml", get(file_stream));
    /// # let _: Router = app;
```

axum-extra/src/response/file_stream.rs
```diff
@@ -0,0 +1,391 @@
...
+    /// ```
+    pub async fn from_path(path: PathBuf) -> io::Result<FileStream<AsyncReaderStream>> {
+        // open file
+        let file = File::open(&path).await?;
```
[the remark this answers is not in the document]
YanHeDoki: > This should be able to take `&Path`. Any opinions on taking `impl AsRef<Path>` instead?

I've modified it.

YanHeDoki:
I added file streaming support for range headers with the `try_range_response` and `into_range_response` methods, also add documentation and tests and use examples

I just wanted simple support for streaming file downloads, so I didn't make it complicated, and perhaps it should have been added to the `attachment `feature would have been better. maybe continue to improve the functionality

 @yanns @jplatte

axum-extra/src/response/file_stream.rs
```diff
@@ -126,6 +133,165 @@ where
...
+                })?;
+                if n == 0 {
+                    break; // EOF
+                }
+                total_read += n as u64;
+                yield buffer[..n].to_vec();
+
+        }
+    };
+    Ok(stream)
```
paolobarbolini, line 294: It should be possible to replace this entire part with [`AsyncReadExt::take`](https://docs.rs/tokio/1.42.0/tokio/io/trait.AsyncReadExt.html#method.take)

jplatte:
I still think it makes sense to have this as a feature. It's independent of `Attachment` in its purpose, and can be combined with it if wanted.

examples/stream-to-file/Cargo.toml
jplatte, line 1: I think it would be good to split the example change into a separate PR.

axum-extra/src/response/file_stream.rs
```diff
@@ -0,0 +1,638 @@
...
+use http::{header, StatusCode};
+use std::{io, path::Path};
+use tokio::{
+    fs::File,
+    io::{AsyncReadExt, AsyncSeekExt},
+};
+use tokio_util::io::ReaderStream;
+
+/// Alias for `tokio_util::io::ReaderStream<File>`.
+pub type AsyncReaderStream = ReaderStream<File>;
```
jplatte, line 17: Why the type alias? `ReaderStream<File>` is not a very complex type.

axum-extra/src/lib.rs
```diff
@@ -25,8 +25,9 @@
 //! `tracing` | Log rejections from built-in extractors | Yes
 //! `typed-routing` | Enables the [`TypedPath`](crate::routing::TypedPath) routing utilities | No
 //! `typed-header` | Enables the [`TypedHeader`] extractor and response | No
+//! `fileStream` | Enables the [`fileStream`](crate::response::file_stream) response | No
```
jplatte, line 28:
```suggestion
//! `FileStream` | Enables the [`FileStream`](crate::response::FileStream) response | No
```

axum-extra/src/lib.rs
```diff
@@ -25,8 +25,9 @@
...
 //!
-//! [`axum`]: https://crates.io/crates/axum
+//! [`axum`]: <https://crates.io/crates/axum>
```
jplatte, line 30:
```suggestion
//! [`axum`]: https://crates.io/crates/axum
```

axum-extra/src/response/file_stream.rs
```diff
@@ -0,0 +1,638 @@
...
+                let bytes_to_read = std::cmp::min(buffer_size as u64, end - total_read);
+                let n = file.read(&mut buffer[..bytes_to_read as usize]).await.map_err(|e| {
+                    std::io::Error::new(std::io::ErrorKind::Other, e)
+                })?;
+                if n == 0 {
+                    break; // EOF
+                }
+                total_read += n as u64;
+                yield buffer[..n].to_vec();
+
```
jplatte, line 291:
```suggestion
        while total_read < end {
            let bytes_to_read = std::cmp::min(buffer_size as u64, end - total_read);
            let n = file.read(&mut buffer[..bytes_to_read as usize]).await.map_err(|e| {
                std::io::Error::new(std::io::ErrorKind::Other, e)
            })?;
            if n == 0 {
                break; // EOF
            }
            total_read += n as u64;
            yield buffer[..n].to_vec();
```

axum-extra/src/response/file_stream.rs
```diff
@@ -0,0 +1,638 @@
...
+        }
+    };
+    Ok(stream)
```
jplatte, line 294:
```suggestion
    };

    Ok(stream)
```

axum-extra/src/response/file_stream.rs
```diff
@@ -0,0 +1,638 @@
...
+
+        if let Some(content_size) = self.content_size {
+            resp = resp.header(header::CONTENT_LENGTH, content_size);
+        }
+
+        resp.body(body::Body::from_stream(self.stream))
+            .unwrap_or_else(|e| {
+                (
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    format!("build FileStream responsec error: {}", e),
```
jplatte, line 321:
```suggestion
                    format!("build FileStream responsec error: {e}"),
```

axum-extra/src/response/file_stream.rs
YanHeDoki: > AsyncReadExt::take

Yes, it's better to use `AsyncReadExt::take `replacement here, I fixed the

YanHeDoki: > It should be possible to replace this entire part with [`AsyncReadExt::take`](https://docs.rs/tokio/1.42.0/tokio/io/trait.AsyncReadExt.html#method.take) and then going back to `ReaderStream`

Yes, it's better to use AsyncReadExt::take replacement here, I fixed the

axum-extra/src/response/file_stream.rs
YanHeDoki: > Why the type alias? `ReaderStream<File>` is not a very complex type.

It's indeed unnecessary, I just removed it for the sake of a shorter generic, and it's fine to remove it, I removed the

examples/stream-to-file/Cargo.toml
YanHeDoki: > I think it would be good to split the example change into a separate PR.

Are all stream-to-file example changes so far? Is it right to add another PR after this one?

YanHeDoki: > I think it would be good to split the example change into a separate PR.

I've split the example changes

jplatte (approved):

jplatte:
Thank you! I think there's still some improvements to be done, but that can be done separately.

PR #3051, opened 2024-11-26
Disallow nesting at the root of the router

### LINKED ISSUE #2651: Consider disallowing root nesting
Before the 0.8.0 release, I would like to come back to `nest("/", router)` and `nest_service("/", svc)`. I believe I suggested disallowing them (in favor of `merge` and `fallback_service`) before, but if so there must have been some pushback by David since we still allow both.

Any pointers or thoughts appreciated.

jplatte:
## Motivation

It's confusing, there isn't really any nesting going on.

## Solution

Remove support for nesting at the root. `merge` can fully replace `nest("" | "/", ...)`, `fallback_service` can fully replace `nest_service("" | "/", ...)`.

Closes #2651.

### DIFF
--- a/axum/src/extract/matched_path.rs
+++ b/axum/src/extract/matched_path.rs
@@ -324,14 +324,14 @@ mod tests {
             "/{*path}",
             any(|req: Request| {
                 Router::new()
-                    .nest("/", Router::new().route("/foo", get(|| async {})))
+                    .nest("/foo", Router::new().route("/bar", get(|| async {})))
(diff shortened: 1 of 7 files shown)


yanns (approved):
I don't use this feature, so I cannot judge if the change is good for the users.
From the code point of view, all is looking good.

jplatte:
I'm pretty confident that the suggested replacements are better suited for all use cases.

If not, it's backwards-compatible to revert this.

PR #3053, opened 2024-11-28
Fix PrivateCookieJar documentation example comments

SantaClaas:
Update PrivateCookieJar documentation example comments to reflect that key is used for encryption and not signing.

## Motivation
I was looking for a way to encrypt cookies and looked at PrivateCookieJar but got confused that the examples talked about signing. This probably snuck in when copying the documentation from SignedCookieJar as they are similar.

## Solution
Updated comments in example code to reflect what PrivateCookieJar does.


jplatte (approved):
Thanks!

jplatte:
CI errors are unrelated and going to be fixed via https://github.com/tokio-rs/axum/pull/3054.

PR #3071, opened 2024-12-08
refactor: remove raw string hashes

### LINKED ISSUE #3070: `axum/src/extract/path/mod.rs` has some unnecessary `#` symbols

hamirmahal:
Closes #3070

## Motivation

There are some unnecessary `#` symbols we can remove to make `axum/src/extract/path/mod.rs` a bit cleaner and improve code readability and maintainability.

## Solution

We remove the unnecessary `#` symbols.

### DIFF
--- a/axum/src/extract/path/mod.rs
+++ b/axum/src/extract/path/mod.rs
@@ -949,7 +949,7 @@ mod tests {
         let body = res.text().await;
         assert_eq!(
             body,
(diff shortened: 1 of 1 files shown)


hamirmahal (commented):
This is just a small refactor to improve code readability. It shouldn't result in any user-facing changes.

jplatte (approved):
I think these also don't need to be raw string literals in general, right?

hamirmahal:
Ah, I think you're right.

PR #3073, opened 2024-12-10
axum 0.8.0-rc.1

### LINKED ISSUE #3060: Add tcp_nodelay removal to changelog
Note to self.

jplatte:
Closes #3060.

@yanns @mladedav I'd love to do another call for testing with all these changes in. @mladedav do you think you're gonna spend some time on the percent-encoded routing thing (not as in code, but as in getting back to the discussion)? Or should we ship without it?

### DIFF
--- a/axum-core/CHANGELOG.md
+++ b/axum-core/CHANGELOG.md
@@ -7,13 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+# 0.5.0
+
+## rc.1
+
 - **breaking:**: `Option<T>` as an extractor now requires `T` to implement the
   new trait `OptionalFromRequest` (if used as the last extractor) or
   `OptionalFromRequestParts` (other extractors) ([#2475])
 
 [#2475]: https://github.com/tokio-rs/axum/pull/2475
-
-# 0.5.0
 
 ## alpha.1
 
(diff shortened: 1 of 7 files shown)


yanns (commented):
Most of users will look at the changelog from axum, and will miss the ones in axum-core. 
I'd suggest to copy/paste the core change log into axum. WDYT?

yanns:
@jplatte What do you think about copy / pasting the change log? Alternatively, we could provide a link from axum to axum-core change log. But I think that the copy / paste is easier for our users.

jplatte:
Sorry, been pretty busy. I think it's gonna be a bit easier to link to the other changelog, but copy-pasting may be easier. Either way I agree that it would be good to have those entries easier to discover.

jplatte:
I've added a note. For the GH release I'll adjust it to point to the right release page.

yanns (approved):

yanns:
I did the same for https://github.com/tokio-rs/axum/releases/tag/axum-v0.8.0-alpha.1

jplatte:
Huh? But I already had it there? All I see is that the the links to the alpha changelogs seem to be gone now 😕

mladedav:
Sorry, I was (and still am) a bit busier than usual. I'll definitely get back to the percent encoding, but I think we can ship the manor version without it.

jplatte:
I figured as much and moved the percent encoding thing to the 0.9.0 milestone. But maybe that's the wrong place too? I guess it might not even be that clear whether it's going to be breaking.

Anyways, I've added two more things to the 0.8.0 [milestone](https://github.com/tokio-rs/axum/milestone/7) that should probably be done before the release, will spend some time on them over the holidays and probably release around the new year if no new blockers come up.

PR #3076, opened 2024-12-12
chore: remove unused `nodelay` field in `Serve`

shikhar:
see https://github.com/tokio-rs/axum/issues/3060


jplatte (approved):
Not really related to the changelog issue but yeah, this should not be there anymore. Thanks!

PR #3078, opened 2024-12-15
breaking: update tokio-tungstenite to 0.26 and update examples

adryzz:
## Motivation

tokio-tungstenite had a new (breaking) release, 0.25/0.26 that finally allows sending shared data over websockets, and it would be cool if this new breaking change could get in axum 0.8.0 given it's a new major release

## Solution

the websocket extract has been changed, and it now exposes  `Bytes` and `Utf8Bytes` (UTF-8 wrapper around `Bytes`)

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -1145,21 +1145,21 @@ mod tests {
     }
 
     async fn test_echo_app<S: AsyncRead + AsyncWrite + Unpin>(mut socket: WebSocketStream<S>) {
-        let input = tungstenite::Message::Text("foobar".into());
+        let input = tungstenite::Message::Text(tungstenite::Utf8Bytes::from_static("foobar"));
         socket.send(input.clone()).await.unwrap();
         let output = socket.next().await.unwrap().unwrap();
         assert_eq!(input, output);
 
         socket
             .send(tungstenite::Message::Ping(
-                Bytes::from_static(b"ping").into(),
+                Bytes::from_static(b"ping"),
             ))
             .await
             .unwrap();
         let output = socket.next().await.unwrap().unwrap();
         assert_eq!(
             output,
-            tungstenite::Message::Pong(Bytes::from_static(b"ping").into())
+            tungstenite::Message::Pong(Bytes::from_static(b"ping"))
         );
     }
 }
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -114,11 +114,7 @@ use std::{
 use tokio_tungstenite::{
     tungstenite::{
         self as ts,
-        protocol::{
-            self,
-            frame::{Payload, Utf8Payload},
-            WebSocketConfig,
-        },
+        protocol::{self, WebSocketConfig},
     },
     WebSocketStream,
 };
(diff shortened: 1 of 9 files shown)


yanns:
The CI checks are failing. Can you check?

adryzz:
given that we have our own downstream `Message` type that is 1:1 with `tungstenite`'s, should we have our own downstream `Payload`/`Utf8Payload` types too?

jplatte:
Yes, we don't want to have tungstenite as a public dep. However I'm a bit undecided about what to expose exactly - I'll comment later after work.

Also cc @SabrinaJewson

SabrinaJewson:
So, my intial observation is that Tungstenite’s [`Payload`](https://docs.rs/tungstenite/latest/tungstenite/protocol/frame/enum.Payload.html) type is very weirdly-designed. It has three separate variants for `BytesMut`, `Bytes` and `Vec<u8>`. But the thing is:

- the `BytesMut` variant seems pointless, since one can convert `Bytes ↔ BytesMut` with zero cost;
- the only benefit of having the `Vec<u8>` variant is to avoid a small heap allocation if one has a `Vec<u8>` with excess capacity.

To me, neither of these advantage seem worth it to add a whole new `Payload` type to the public API instead of using `Bytes` directly (and I do question why `tungstenite` made this strange design choice). For the UTF-8 case, it seems like this should come from a utility crate, but in the absence of that making our own UTF-8 wrapper of `Bytes` isn’t _too_ bad…? related: https://github.com/tokio-rs/bytes/issues/697

jplatte:
Those were pretty much my thoughts looking at the tungstenite change too 😅

So let's switch to `Bytes` then? And a `Utf8Bytes` type in addition? Pretty sure we even have one in the codebase already, used for path deserialization IIRC. Just not public at the moment.

alexheretic:
The payload variants do have reason to exist. For example, removing `Payload::Vec` variant impacts benched write performance by 10-30%. See https://github.com/snapview/tungstenite-rs/pull/465

axum/src/extract/ws.rs
```diff
@@ -666,8 +670,8 @@ impl Message {
     /// Consume the WebSocket and return it as binary data.
     pub fn into_data(self) -> Vec<u8> {
```
alexheretic, line 671: Return a vec here no longer makes sense, upstream has `fn into_data(self) -> Payload`.

axum/src/extract/ws.rs
```diff
@@ -676,10 +680,12 @@ impl Message {
     /// Attempt to consume the WebSocket message and convert it to a String.
     pub fn into_text(self) -> Result<String, Error> {
```
alexheretic, line 681: Return a `String` here no longer makes sense, upstream has `fn into_text(self) -> Result<Utf8Payload>`.

examples/chat/src/main.rs
```diff
@@ -79,15 +79,15 @@ async fn websocket(stream: WebSocket, state: Arc<AppState>) {
...
+            check_username(&state, &mut username, name.as_str());
 
             // If not empty we want to quit the loop else we want to quit function.
             if !username.is_empty() {
                 break;
             } else {
                 // Only send our client that username is taken.
                 let _ = sender
-                    .send(Message::Text(String::from("Username already taken.")))
+                    .send(Message::Text("Username already taken.".into()))
```
alexheretic, line 90: This example could use a static payload, something like using upstream `Utf8Payload::from_static("Username already taken.")`

examples/websockets-http2/src/main.rs
```diff
@@ -75,7 +75,7 @@ async fn ws_handler(
                 res = ws.recv() => {
                     match res {
                         Some(Ok(ws::Message::Text(s))) => {
-                            let _ = sender.send(s);
+                            let _ = sender.send(s.to_string());
```
alexheretic, line 78: This use case should show the shared bytes reuse, something like using `Utf8Payload::share`.

examples/websockets/src/client.rs
```diff
@@ -65,7 +66,9 @@ async fn spawn_client(who: usize) {
 
     //we can ping the server for start
     sender
-        .send(Message::Ping("Hello, Server!".into()))
+        .send(Message::Ping(Payload::Shared(
+            "Hello, Server!".as_bytes().into(),
```
alexheretic, line 70: This should use static bytes

examples/websockets/src/main.rs
```diff
@@ -101,7 +101,11 @@ async fn ws_handler(
 /// Actual websocket statemachine (one will be spawned per connection)
 async fn handle_socket(mut socket: WebSocket, who: SocketAddr) {
     // send a ping (unsupported by some browsers) just to kick things off and get a response
-    if socket.send(Message::Ping(vec![1, 2, 3])).await.is_ok() {
+    if socket
+        .send(Message::Ping(vec![1, 2, 3].into()))
```
alexheretic, line 105: This should use static bytes

SabrinaJewson:
> For example, removing `Payload::Vec` variant impacts benched write performance by 10-30%. 

Do you know why that’s the case? Is it really just the memcpy due to having excess capacity in the `Vec`? Is it the cost of virtualization?

Looking at the implementation of `Bytes`, it should be possible to avoid the memcpy. Currently, `Bytes` (in its `Vec` representation) stores the base ptr twice, which is unnecessary; if it instead used that field to store capacity, there would be no need for shrinking the allocation. Maybe we could look into making this change in `Bytes` upstream, thus allowing the storage of `Vec<u8>` inside `Bytes` to be truly zero-cost?

The reason I’m somewhat against a `Payload` enum is that it strikes me as a code smell: `Bytes` is _already_ intended to be the type representing “an enum over all the `'static` thread-safe bytes containers”. So the addition of `Payload` means we have two layers of bytes-type-that-can-hold-any-container. If one is to go ahead with the design, I think it should be made very clear in the documentation what the exact limitations of `Bytes` are that it can’t subsume this use-case.

alexheretic:
I do agree, I think payload started as `Bytes`. It is just the performance regressions driving the enum. I (re)investigating if those perf issues can be resolved: https://github.com/snapview/tungstenite-rs/pull/473

alexheretic:
It looks like the perf issues can be resolved. If possible please review that upstream PR and suggest changes so we can get the next breaking tungstenite version as right as possible.

adryzz:
alright, will update this PR as soon as i get home with the newly released version.

what to do about `Utf8Bytes`? i can't seem to find a similiar type in the codebase already

jplatte:
I just checked, looks like I was wrong. I think we can just newtype tungstenite's type then.

axum/src/extract/ws.rs
```diff
@@ -664,32 +781,32 @@ impl Message {
...
             Self::Text(string) => Ok(string),
-            Self::Binary(data) | Self::Ping(data) | Self::Pong(data) => Ok(String::from_utf8(data)
-                .map_err(|err| err.utf8_error())
-                .map_err(Error::new)?),
-            Self::Close(None) => Ok(String::new()),
-            Self::Close(Some(frame)) => Ok(frame.reason.into_owned()),
+            Self::Binary(data) | Self::Ping(data) | Self::Pong(data) => {
+                Ok(Utf8Bytes::try_from(data).map_err(Error::new)?)
+            }
+            Self::Close(None) => Ok(Bytes::new().try_into().map_err(Error::new)?),
```
alexheretic, line 800:
```suggestion
            Self::Close(None) => Ok(Utf8Bytes::from_static("")),
```
Or if you derive `Default` for `Utf8Bytes` (as upstream does):
```suggestion
            Self::Close(None) => Ok(<_>::default()),
```

axum/src/extract/ws.rs
```diff
@@ -553,16 +553,133 @@ impl Sink<Message> for WebSocket {
...
+        f.write_str(self.as_str())
+    }
+}
+
+impl TryFrom<Bytes> for Utf8Bytes {
+    type Error = std::str::Utf8Error;
+
+    #[inline]
+    fn try_from(bytes: Bytes) -> Result<Self, Self::Error> {
+        std::str::from_utf8(&bytes)?;
```
alexheretic, line 614: This check is done by the upstream TryFrom impl

```suggestion
```

examples/chat/src/main.rs
```diff
@@ -109,7 +111,7 @@ async fn websocket(stream: WebSocket, state: Arc<AppState>) {
     let mut send_task = tokio::spawn(async move {
         while let Ok(msg) = rx.recv().await {
             // In any websocket error, break loop.
-            if sender.send(Message::Text(msg)).await.is_err() {
+            if sender.send(Message::Text(msg.into())).await.is_err() {
```
alexheretic, line 114: It might be worth also adding similar methods `Message::text`, `Message::binary` as upstream to make code like this a little nicer. As a mixed user of axum & tungstenite you maybe would expect them to be available in both.

examples/websockets-http2/src/main.rs
alexheretic: This could probably be left as a follow up task, if so maybe just add a TODO here?

axum/src/extract/ws.rs
```diff
@@ -553,16 +553,133 @@ impl Sink<Message> for WebSocket {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// UTF-8 wrapper for [Bytes].
+///
+/// An [Utf8Bytes] is always guaranteed to contain valid UTF-8.
+pub struct Utf8Bytes(ts::Utf8Bytes);
```
jplatte, line 560:
```suggestion
/// UTF-8 wrapper for [Bytes].
///
/// An [Utf8Bytes] is always guaranteed to contain valid UTF-8.
#[derive(Clone, Debug, Default, PartialEq, Eq)]
pub struct Utf8Bytes(ts::Utf8Bytes);
```

examples/websockets-http2/src/main.rs
jplatte: An issue usually works better than a fixme comment for stuff like this. Feel free to open one.

axum/src/extract/ws.rs
```diff
@@ -664,44 +779,60 @@ impl Message {
...
+            Self::Text(ref string) => Ok(string.as_str()),
             Self::Binary(ref data) | Self::Ping(ref data) | Self::Pong(ref data) => {
                 Ok(std::str::from_utf8(data).map_err(Error::new)?)
             }
             Self::Close(None) => Ok(""),
             Self::Close(Some(ref frame)) => Ok(&frame.reason),
         }
     }
+
+    /// Create a new text WebSocket message from a stringable.
```
jplatte, line 816: Not sure how to word this, but "stringable" is not a word in my vocabulary :D

axum/src/extract/ws.rs
```diff
@@ -553,16 +553,131 @@ impl Sink<Message> for WebSocket {
...
+impl From<Utf8Bytes> for Bytes {
+    #[inline]
+    fn from(Utf8Bytes(bytes): Utf8Bytes) -> Self {
+        bytes.into()
+    }
+}
+
+impl<T> PartialEq<T> for Utf8Bytes
+where
+    for<'a> &'a str: PartialEq<T>,
```
jplatte, line 656: Note to self: revisit.

jplatte (approved):
There's a few things I'd like to look at and possibly improve before releasing, but this is definitely a big step in the right direction so I'll merge 👍

Issue #3075, opened 2024-12-12
Move axum-extra's Query type to axum


jplatte:
From https://github.com/tokio-rs/axum/issues/1041#issuecomment-2539584964

> I just came across this alternative `Query` impl and I was wondering: is there is a downside to making this implementation the default in `axum` itself? are there plans to eventually do that?

As the creator of the underlying crate I may be biased a bit. There's probably some cases where it's a tad slower but that should be everything in terms of downsides.

Lachstec:
Was also unaware of it, but I think that the version in extra is really nice and would have been very useful to me in the past. If it is desired to make it the default

jplatte:
Wdyt @yanns @mladedav ?

yanns:
I've never used the `Query` myself, so take my input with this in consideration.
The new query is looking good, and I don't see any reasons not to use it.

Should we add more tests, especially concerning encoding of special characters in the values?

manortec:
I think the Query and Json modules have similar uses and importance. I think Query should be consistent with Json, so their file locations should be in the same place.

SabrinaJewson:
It seems fine to me in principle. I wonder if the inefficiencies could be avoided by lazily parsing the input? That is, by default no `IndexMap` is generated, but if the `Deserialize` implementation calls `deserialize_seq` or `deserialize_option` it will scan ahead and build the map?

jplatte:
IIRC something like that is already the case. I have some benchmarks in the repo and it's barely slower (needs some extra branches only) for cases supported by both crates.

Turbo87:
I'm wondering whether we should address https://github.com/jplatte/serde_html_form/pull/23 first. 

From what I've seen so far the two query extractor seem to behave slightly differently in that regard, and it would be unfortunate to introduce a breaking change twice if we think about aligning the two implementations again anyway.

jplatte:
Oh! I didn't think my crate had different behavior in any case supported by `serde_urlencoded`, but I think you're right that that's different. I don't think that is going to be easy to adress, I don't think your PR is the roght approach but it will take me some time to formulate my thinking in detail.

In that case the best solution is probably to ship 0.8 without this.

PR #3081, opened 2024-12-18
Query/Form: Use `serde_path_to_error` to report fields that failed to parse

Turbo87:
## Motivation

I've started using the `Query` extractor more and more over the past couple of days and noticed that its error messages don't help the user in understanding which field specifically failed to parse.

## Solution

This solves the problem roughly similar to our existing usage of `serde_path_to_error` in the `Json` extractor. The resulting error messages change from e.g.

> Failed to deserialize query string: invalid digit found in string

to

> Failed to deserialize query string: my_broken_field: invalid digit found in string


axum/Cargo.toml
```diff
@@ -68,6 +68,7 @@ tower-service = "0.3"
 # optional dependencies
 axum-macros = { path = "../axum-macros", version = "0.5.0-rc.1", optional = true }
 base64 = { version = "0.22.1", optional = true }
+form_urlencoded = { version = "1.1.0", optional = true }
```
Turbo87, line 71: I would prefer not to add this explicit extra dependency too, but `serde_urlencoded` currently has no way to construct a `Deserializer` without it and we need a `Deserializer` instance to pass into `serde_path_to_error`.

jplatte: Since we're switching to `serde_html_form` anyways, if you have an idea for how to make the `Deserializer` publically constructible in a nice way (haven't looked at that code in a while myself), feel free to make a PR.

jplatte: Great. Let's wait until after https://github.com/tokio-rs/axum/issues/3075 then, and then you can update the PR to use that (I released v0.2.7 of `serde_html_form` with this).

jplatte: With #3075 postponed, I think we can leave things as they are here. Upstreaming the constructor to `serde_urlencoded` wouldn't hurt, but isn't really necessary either.

jplatte (commented):
Mind doing the same thing for `Form` and both `serde_html_form` extractors in axum-extra?

Turbo87:
> doing the same thing for `Form`

it looks like the `Form` extractor already includes the field name (from the tests: `Invalid URL: Cannot parse `res` with value `123123-123-123123`: UUID parsing failed: invalid group count: expected 5, found 3`), though in this case the user isn't aware of the internal name of the field, so I don't know how useful that even is 😅 

> and both `serde_html_form` extractors in axum-extra?

yeah, I'll add it to the other `Query` extractor too

Turbo87:
nevermind, I was looking at the `Path` extractor, not the `Form` extractor... 🤦‍♂️ 

too early in the morning 😂

Turbo87:
> Mind doing the same thing for `Form` and both `serde_html_form` extractors in axum-extra?
> 
> Also, changelog entries

alright, should all be done now :)

jplatte (approved):
Thanks!

Issue #3079, opened 2024-12-18
axum 0.8.0-rc.1: optional query parameters


yanns:
I'm trying out https://github.com/tokio-rs/axum/releases/tag/axum-v0.8.0-rc.1 on some internal services, and I have an issue with the change about optional extractors: https://github.com/tokio-rs/axum/pull/2475

My code:
```
#[derive(Deserialize, Debug)]
struct ConsistencyParams {
    #[serde(deserialize_with = "deserialize_untagged_enum_case_insensitive")]
    consistency: Consistency,
}

#[debug_handler]
async fn expand(
   ...
    consistency: Option<Query<ConsistencyParams>>,
   ...
) -> Response {
```

When I call the HTTP endpoint without the `consistency` query parameter, the request is rejected with:
```
2024-12-18T08:43:16.504317Z TRACE http-request{correlation_id="reference-535743ab-319d-46ae-8241-d4b49d370ee5"}: axum::rejection: rejecting request status=400 body="Failed to deserialize query string: missing field `consistency`" rejection_type="axum::extract::rejection::FailedToDeserializeQueryString"
2024-12-18T08:43:16.504393Z  INFO http-request{correlation_id="reference-535743ab-319d-46ae-8241-d4b49d370ee5"}: reference_service::telemetry::logging::access_logs: POST /{project_key}/reference-expansion 400 http.method="POST" http.status_code=400 http.received_bytes=96 http.sent_bytes=63 ctp.api_endpoint="POST /{project_key}/reference-expansion" duration=0.00019 type="reference-http-access"
thread 'component_tests::eventual_consistency::expanded_obj_are_strong_consistent_on_user_session_by_default' panicked at tests/reference/component_tests/component_tests_helpers.rs:329:9:
assertion `left == right` failed: body: Ok("Failed to deserialize query string: missing field `consistency`")
  left: 400
 right: 200
```

The URL is looking like this:
```
/project-a/reference-expansion?expand=mainProduct&expand=order
```

When I look at the description of https://github.com/tokio-rs/axum/pull/2475, I can read:
```
/// # `Option<Query<T>>` behavior
///
/// If `Query<T>` itself is used as an extractor and there is no query string in
/// the request URL, `T`'s `Deserialize` implementation is called on an empty
/// string instead.
///
/// You can avoid this by using `Option<Query<T>>`, which gives you `None` in
/// the case that there is no query string in the request URL.
///
/// Note that an empty query string is not the same as no query string, that is
[...]

PR #3088, opened 2024-12-19
Remove OptionalFromRequestParts impl for Query

jplatte:
Closes https://github.com/tokio-rs/axum/issues/3079.
cc @Turbo87

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -7,12 +7,8 @@ and this project adheres to [Semantic Versioning].
 
 # Unreleased
 
-- **breaking:** `axum::extract::ws::Message` now uses `Bytes` in place of `Vec<u8>`,
-  and a new `Utf8Bytes` type in place of `String`, for its variants ([#3078])
-- **changed:** Upgraded `tokio-tungstenite` to 0.26 ([#3078])
 - **changed:** Query/Form: Use `serde_path_to_error` to report fields that failed to parse ([#3081])
 
-[#3078]: https://github.com/tokio-rs/axum/pull/3078
 [#3081]: https://github.com/tokio-rs/axum/pull/3081
 
 # 0.10.0
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -7,9 +7,11 @@ and this project adheres to [Semantic Versioning].
 
 # Unreleased
 
+- **breaking:** Remove `OptionalFromRequestParts` impl for `Query` ([#3088])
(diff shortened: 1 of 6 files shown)


axum-extra/src/extract/query.rs
```diff
@@ -216,11 +216,9 @@ impl std::error::Error for QueryRejection {
 ///
 /// [example]: https://github.com/tokio-rs/axum/blob/main/examples/query-params-with-empty-strings/src/main.rs
 #[cfg_attr(docsrs, doc(cfg(feature = "query")))]
-#[deprecated = "Use Option<Query<_>> instead"]
```
Turbo87, line 219: While I don't think people should use `Option<Query<_>>`, should we recommend using `Query<_>` with `Option<_>` fields instead and keep this deprecated?

axum/src/docs/extract.md
```diff
@@ -108,18 +108,10 @@ struct Pagination {
     per_page: usize,
```
Turbo87, line 108: should these be `Option<usize>` now or was is already this way before?

Turbo87 (approved):

axum-extra/src/extract/query.rs
jplatte: Well this extractor was specifically added with the intention of allowing "all query parameters specified, or none". I don't really get why that's a useful thing, but we can always deprecate / remove this later.

mladedav: When you say deprecate, we cannot lint that through the attribute though, right?

axum-extra/CHANGELOG.md
```diff
@@ -7,13 +7,11 @@ and this project adheres to [Semantic Versioning].
 
 # Unreleased
 
-- **breaking:** `axum::extract::ws::Message` now uses `Bytes` in place of `Vec<u8>`,
```
mladedav, line 10: These changelogs disappeared by accident or did they move somewhere?

axum-extra/src/extract/query.rs
jplatte: Huh, why would I not mean the attribute?

jplatte: Note that this is about `OptionalQuery` in axum-extra, not the `OptionalFromRequestParts` impl for `Query` (already removed in this PR).

axum-extra/CHANGELOG.md
jplatte: Moved to `axum` changelog since that's where they belong (first commit).

axum-extra/src/extract/query.rs
mladedav: Sorry, I got confused about what the deprecation was about.

PR #3100, opened 2024-12-25
ci: Use updated fork of cargo-public-api-crates

jplatte:
So we no longer need to pin an old nightly.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -65,22 +65,16 @@ jobs:
         crate: [axum, axum-core, axum-extra, axum-macros]
     steps:
     - uses: actions/checkout@v4
-    # Pinned version due to failing `cargo-public-api-crates`.
-    - uses: dtolnay/rust-toolchain@master
-      with:
-        toolchain: nightly-2024-06-06
+    - uses: dtolnay/rust-toolchain@nightly
     - uses: Swatinem/rust-cache@v2
       with:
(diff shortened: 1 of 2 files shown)


Turbo87 (approved):
seems reasonable 

is there a reason why we're not using https://github.com/obi1kenobi/cargo-semver-checks here?

jplatte:
No reason other than that nobody has put in the effort yet (and I think we'd have to discuss _how_ to integrate it too). But to be clear that would be *in addition*, right?

Turbo87:
I don't have direct experience with either, but from their readmes it sounded like they do roughly the same thing unless I've missed something 😅

jplatte:
They do related but independent jobs, and both use rustdoc's JSON output to do their job: cargo-semver-checks checks whether there's any breaking changes between two versions of a crate (I use it locally prior to publishing a non-semver-breaking crate release to ensure I'm not accidentally shipping something that _does_ break, but it can also be integrated into CI of course). cargo-public-api-crates ensures that your crate does not publically export items from / implements traits from any of its dependencies unless white-listed (such that upgrading that dependency is never a breaking change) - at some point this will be superseded by native Cargo support for marking dependencies as public or private (https://github.com/rust-lang/rust/issues/44663).

Turbo87:
I see. looks like I somehow misunderstood what cargo-public-api-crates does 😅

jplatte:
Well, it's hardly documented. Probably something I should fix in my fork too at some point ^^

PR #3104, opened 2024-12-26
Unignore `Cargo.lock` file

Turbo87:
This makes it much easier to checkout a previous revision of the project and have a working set of dependencies. That helps for example with things like running `git bisect` when finding regressions.

see also https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html


Turbo87:
I would recommend to also use https://github.com/renovatebot/renovate to regularly update the lockfile (and probably also dev-dependencies), but that will need to be installed in the `tokio-rs` GitHub org to be able to use it and idk if any of our axum maintainers have sufficient permissions to do so 🤔

Turbo87:
@jplatte for future PRs: are you watching the repo (i.e. do you get notifications for new PRs anyway?) or would you prefer an explicit ping or review request?

jplatte:
Makes sense to me.

Re. renovate: We could set up dependabot instead, this doesn't require any special permissions. It seems to support quite many modes of operation. How about a weekly PR? I'd prefer every other week but unfortunately it seems like they only support a small fixed number of cadences.

Re. review: I get notified of everything already, no need to ping or request review.

jplatte (approved):

Turbo87:
> Re. renovate: We could set up dependabot instead, this doesn't require any special permissions. It seems to support quite many modes of operation.

I used to recommend dependabot over renovatebot a few years ago, but have changed my mind since then. renovatebot is way more flexible and configurable. the only advantage of dependabot is that it was bought by GitHub and integrated somewhat, but the original creator isn't even working for GitHub anymore...

> How about a weekly PR? I'd prefer every other week but unfortunately it seems like they only support a small fixed number of cadences.

The default cadence for lockfile updates in renovatebot is weekly, which works well in my experience.

In renovatebot the schedule is essentially a cronjob syntax (see e.g. https://docs.renovatebot.com/presets-schedule/#scheduleautomergemonthly) which could be adjusted to run on fixed days 2-3 times per month, if we wanted to go that way.

PR #3105, opened 2024-12-26
Remove broken `rest-grpc-multiplex` example

Turbo87:
This example has not been updated to hyper v1 for a year and requires every developer to have `protoc` installed on their machine to be able to run `cargo build --workspace`. Both of these points combined IMHO shift the tradeoff enough to it not being valuable enough to keep the broken example around.


Turbo87:
> This example has not been updated to hyper v1 for a year

well, I just discovered https://github.com/tokio-rs/axum/pull/2825, so that point is not as strong as I thought. still, the `protoc` requirement is quite annoying as my Rust Rover IDE is constantly complaining about it 😅

jplatte:
I agree with this. The fact that the upgrade PR has been lingering for quite a while is IMO another reason to just remove it. It's probably better to have an example like this in the tonic repo.

jplatte (approved):

PR #3106, opened 2024-12-26
Improve `crate-ci/typos` configuration

Turbo87:
see commit messages 😉

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -257,4 +257,4 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Check the spelling of the files in our repo
-        uses: crate-ci/typos@v1.20.8
+        uses: crate-ci/typos@v1.28.4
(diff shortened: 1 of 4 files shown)


jplatte:
If both are supported (pretty sure they are), I'd prefer using `.typos.toml` over `_typos.toml`.

Turbo87:
I just discovered that we could also use `[workspace.metadata.typos]` in the `Cargo.toml` file. I think that would me my preferred choice, unless you disagree

jplatte (approved):
Not sure what to think of the config in `Cargo.toml`, really. For cargo plugins it makes sense to me, for more or less entirely language-agnostic tools like typos it feels a little weird. But it's nice to have one file less, I suppose.

Turbo87:
> for more or less entirely language-agnostic tools like typos it feels a little weird

huh, I guess I figured with the org name being `crate-ci` it was somewhat Rust-specific, but you're right, that there isn't much in it that is specific to Rust.

> But it's nice to have one file less, I suppose.

yeah, even two files less in this case :D

PR #3108, opened 2024-12-26
clippy: Merge config files

Turbo87:
The content of `.clippy.toml` was unintentionally overriding the contents of `clippy.toml` 😅

Related:

- https://doc.rust-lang.org/clippy/configuration.html

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,1 +1,4 @@
 allow-mixed-uninlined-format-args = false
+disallowed-types = [
+    { path = "tower::util::BoxCloneService", reason = "Use our internal BoxCloneService which is Sync" },
+]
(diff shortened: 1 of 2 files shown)


jplatte:
Nice. But let's use the dot-prefixed filename.

Turbo87:
> let's use the dot-prefixed filename.

done

.clippy.toml
```diff
@@ -1 +1,4 @@
 allow-mixed-uninlined-format-args = false
+disallowed-types = [
+    { path = "tower::util::BoxCloneService", reason = "Use our internal BoxCloneService which is Sync" },
```
jplatte, line 3: This reminds me, there is now https://docs.rs/tower/latest/tower/util/struct.BoxCloneSyncService.html that we may want to replace our own with.

jplatte (approved):

PR #3109, opened 2024-12-26
Replace custom `BoxCloneService` struct with `tower::util::BoxCloneSyncService`

Turbo87:
Resolves https://github.com/tokio-rs/axum/pull/3108#discussion_r1897981871 by first upgrading the `tower` dependency everywhere, and then replacing our custom `BoxCloneService` struct with the new `tower::util::BoxCloneSyncService` implementation.


.clippy.toml
```diff
@@ -1,4 +1 @@
 allow-mixed-uninlined-format-args = false
-disallowed-types = [
-    { path = "tower::util::BoxCloneService", reason = "Use our internal BoxCloneService which is Sync" },
```
jplatte, line 3: I think we should still disallow it, but in favor of the other tower type.

jplatte (approved):
Thanks!

.clippy.toml
Turbo87: done :)

PR #3119, opened 2024-12-27
axum-extra/form: Add `FailedToDeserializeFormBody` rejection variant

Turbo87:
This brings the `From` extractor from `axum-extra` closer to the one from `axum`.

Related:

- https://github.com/tokio-rs/axum/pull/1683


jplatte (approved):
Interesting. Didn't know about this difference between querystring and body forms, but I guess the status code is a good reason to split them like that.

PR #3101, opened 2024-12-25
Release axum v0.8.0

jplatte:
I think we're all set. All issues in [the milestone](https://github.com/tokio-rs/axum/milestone/7) are closed, and I also created draft GH releases this time. Maintainers should be able to see them [here](https://github.com/tokio-rs/axum/releases).

PR #3120, opened 2024-12-27
Update docs.rs links to 0.8

Turbo87:
see https://github.com/tokio-rs/axum/pull/3101#issuecomment-2563604164


jplatte (approved):
Thanks!

Turbo87:
@syphar @guillaumegomez you don't happen to have an idea on how we could improve the situation?

tl;dr axum depends on axum-core, but in the axum-core docs we would ideally like to use intra-doc links to stuff in axum.

GuillaumeGomez:
Currently we cannot only link to downstream crates, not upstream. However relative paths work fine, but you will still need to have the version number in the path.

PR #3121, opened 2024-12-27
Refactor `TestClient` usage

Turbo87:
This PR changes how the `TestClient` struct from `axum` is shared with `axum-extra` and also makes it available to `axum-core`.

This change avoids the somewhat arcane `include!(concat!(...))` construct in the `axum-extra` crate, which some IDEs (JetBrains) don't like particularly much. The alternative proposed by this PR is a `__private` feature, similar to the one in `axum-macros`, that enables the public export of the `TestClient` struct. The `axum` dev-dependencies in the other two crates then enable that feature, so that it can be used in their test suites.

This enables us to move the `request_parts` tests into `axum-core`, next to the implementation that they are testing.


jplatte:
I like the idea! Will review later.

PR #3126, opened 2024-12-27
CI: Remove unnecessary toolchain install step

Turbo87:
If a `rust-toolchain` file exists, then that toolchain should be used by default anyway

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -101,12 +101,9 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v4
-    - name: Get rust-toolchain version
-      id: rust-toolchain
-      run: echo "version=$(cat axum-macros/rust-toolchain)" >> $GITHUB_OUTPUT
-    - uses: dtolnay/rust-toolchain@master
-      with:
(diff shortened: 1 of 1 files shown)


jplatte (commented):
I'm pretty sure the rust-cache action wants the toolchain you're going to be using for the CI job to be installed by the time it runs, for proper caching.

catuhana:
Instead caching with GitHub's official cache action could be used, though it requires a bit of manual *configuration*.

On another note, wouldn't it be better to move the toolchain file to the root?

jplatte:
> Instead caching with GitHub's official cache action could be used, though it requires a bit of manual *configuration*.

What problem would this be solving? 

> On another note, wouldn't it be better to move the toolchain file to the root?

No, I don't want to force contributors to have that specific nightly always. It's really only needed for stable output of the trybuild tests in the macro crate.

Turbo87:
> I'm pretty sure the rust-cache action wants the toolchain you're going to be using for the CI job to be installed by the time it runs, for proper caching.

hmm, you appear to be right. I was assuming that rust-cache looks at the `rust-toolchain` file, but maybe not, or maybe it doesn't care about a file being in a subfolder 🤔

PR #3127, opened 2024-12-28
feature: add `get_route_paths` to `path_router`

chris-windsor:
Added a `get_route_paths` method to the `path_router` to return a `Vec` of the current paths associated with a router.

## Motivation

There was a desire to create this feature to support logging frameworks that base their logging off the raw path rather than the path at time of request. I specifically am working on building a client for Apitally in Rust and wanted to support the Axum framework first. I will now be able to supply a list of the applications paths to Apitally. I can see a future where other frameworks desire this functionality as well!

## Solution

The paths are already stored in the `path_to_route_id` of the `path_router`. The `RouteId` is not strictly useful in this scenario and is also private. This implementation grabs the keys of `path_to_route_id` and collects them into a `Vec`.

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -328,6 +328,11 @@ where
         })
     }
 
+    /// List routes that the router contains.
+    pub fn get_route_paths(&self) -> Vec<Arc<str>> {
+        self.inner.path_router.get_route_paths()
+    }
+
     /// True if the router currently has at least one route added.
     pub fn has_routes(&self) -> bool {
         self.inner.path_router.has_routes()
--- a/axum/src/routing/path_router.rs
+++ b/axum/src/routing/path_router.rs
@@ -341,7 +341,7 @@ where
     }
 
     pub(super) fn get_route_paths(&self) -> Vec<Arc<str>> {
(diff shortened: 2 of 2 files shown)


jplatte:
I'm sorry but there's been multiple issues asking for this and it has been rejected as a feature.

chris-windsor:
@jplatte aide and utoipa are bloated for developers who don't need that. Developers need access to the inner router somehow. Writing a wrapper is also advocating for bloat. Seems like something that should be considered in the broader community and discussed to find a fair compromise. Could and probably should be behind a feature flag.

jplatte:
You can open a [discussion](https://github.com/tokio-rs/axum/discussions/categories/general) about this elaborating on your use case. Right now I'm fairly strongly against this, but maybe you or others can convince me or the other maintainers.

Turbo87:
> There was a desire to create this feature to support logging frameworks that base their logging off the raw path rather than the path at time of request.

FWIW you can use https://docs.rs/axum/0.7.9/axum/extract/struct.MatchedPath.html for that without having to touch the router.

chris-windsor:
@Turbo87 that is true, and I have implemented that for my use case. However, it would still be nice to retrieve a list of all the routes.

chris-windsor:
opened discussion #3139

Issue #3112, opened 2024-12-27
Tasks have lost their wakers


IlliaJoker:
While monitoring tasks using tokio-console, I see the following issue:
task has lost their wakers: <cargo>/axum-0.8.0-rc.1/src/serve.rs:268:13
![image](https://github.com/user-attachments/assets/32dbc26d-1034-4371-a322-735bc03d6eed)

Turbo87:
can you share a project that reproduces the issue? without more details this is a bit hard to debug

paolobarbolini:
That's probably just because the `Future` returned by `tokio::spawn` is dropped https://docs.rs/crate/axum/0.8.0-rc.1/source/src/serve.rs#268

IlliaJoker:
@paolobarbolini Is this correct behavior in this case?

paolobarbolini:
> @paolobarbolini Is this correct behavior in this case?

It might be, but I'm not sure. One thing I'd try doing, since I see you're using an outdated tokio version (1.39.2) is update everything via `cargo update` and see if anything changes. I don't have much experience with `tokio-console` but it might be just a false positive. Going through their issues they had some of those in the past.

Turbo87:
it looks like tokio-console points to https://github.com/tokio-rs/axum/blob/axum-v0.8.0-rc.1/axum/src/serve.rs#L268-L272, and, if I understand correctly, the only waker in this task is the `signal`, which is passed into the `with_graceful_shutdown()` fn. if that passed in `signal` loses its waker then that seems most likely not something that is controlled by `axum`, but by the surrounding code. you might want to check what you pass into `with_graceful_shutdown()` and whether that is dropped early or anything like that. as I mentioned above, without a project that reproduces the issue this is very hard to debug.

jplatte:
To me this seems pretty likely to be when *not* using graceful shutdown, where internally we use a graceful shutdown future of `std::future::Pending`. This means the wait-for-graceful-shutdown task leaks. We should probably fix that.

IlliaJoker:
I added graceful shutdown to my project, and it helped eliminate task leaks.
Thank you!

jplatte:
I think this should also fix it: https://github.com/tokio-rs/axum/pull/3129
Can somebody confirm?

PR #3129, opened 2024-12-29
Don't leak a tokio task when using serve without graceful shutdown

jplatte:
Closes https://github.com/tokio-rs/axum/issues/3112.


yanns (approved):
Looking good, but I'm no expert here.

Turbo87 (approved):
same for me. I assume it would be hard to write tests for this?

PR #3132, opened 2025-01-01
Remove warning about breaking changes from README.md

jplatte:
Looks like our "breaking changes on main" warning made it to crates.io 😅

### DIFF
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "axum"
-version = "0.8.0" # remember to bump the version that axum-extra depends on
+version = "0.8.1" # remember to bump the version that axum-extra depends on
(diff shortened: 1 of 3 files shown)


Turbo87 (approved):

axum/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "axum"
-version = "0.8.0" # remember to bump the version that axum-extra depends on
+version = "0.8.1" # remember to bump the version that axum-extra depends on
```
yanns, line 3: Can you explain why we need this change? Is it to prepare the next release?

yanns (approved):

axum/Cargo.toml
jplatte: Yeah, I want to publish right after this PR to update the crates.io page.

PR #3141, opened 2025-01-03
Properly respond with sec-websocket-protocol under http/2

coolreader18:
## Motivation

Was looking at the `ws` code and noticed that the subprotocol never gets properly communicated to the client if the connection is HTTP/2.

## Solution

Add the `sec-websocket-protocol` header irrespective of HTTP version. A response with that header is still a valid HTTP/2 websocket handshake, as per [RFC8441](https://www.rfc-editor.org/rfc/rfc8441#section-5.1).

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -1090,10 +1090,11 @@ mod tests {
     #[crate::test]
     async fn integration_test() {
         let addr = spawn_service(echo_app());
-        let (socket, _response) = tokio_tungstenite::connect_async(format!("ws://{addr}/echo"))
-            .await
-            .unwrap();
-        test_echo_app(socket).await;
+        let uri = format!("ws://{addr}/echo").try_into().unwrap();
+        let req = tungstenite::client::ClientRequestBuilder::new(uri)
+            .with_sub_protocol(TEST_ECHO_APP_REQ_SUBPROTO);
+        let (socket, response) = tokio_tungstenite::connect_async(req).await.unwrap();
+        test_echo_app(socket, response.headers()).await;
     }
 
     #[crate::test]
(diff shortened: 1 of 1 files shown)


jplatte:
cc @SabrinaJewson

SabrinaJewson:
Ah, looks good! Can you modify the `http2` test to check for this as well? (Or write a new test)

axum/src/extract/ws.rs
```diff
@@ -1152,11 +1158,20 @@ mod tests {
...
         )
     }
 
-    async fn test_echo_app<S: AsyncRead + AsyncWrite + Unpin>(mut socket: WebSocketStream<S>) {
+    const TEST_ECHO_APP_REQ_SUBPROTO: &str = "echo3, echo";
+    async fn test_echo_app<S: AsyncRead + AsyncWrite + Unpin>(
+        mut socket: WebSocketStream<S>,
+        headers: &http::HeaderMap,
+    ) {
+        dbg!(headers);
```
yanns, line 1172: The CI is complaining about this line. `dbg!` should only be used locally when debugging some issues.
Can you remove this line?
```suggestion
```

coolreader18: Oh whoops lol, yep.

yanns:
Same here, no knowledge in WS in HTTP2.
@SabrinaJewson would be more knowledgeable?

@Turbo87 You haven't approved AFAIS:
Screenshot 2025-01-29 at 11 28 06

yanns (approved):

Turbo87:
> You haven't approved AFAIS

yeah, that was intentional. I didn't want to approve anything I don't have a clue about :D

SabrinaJewson (approved):
Yeah, sorry I didn’t say before but this looks good.

yanns:
Should we add a changelog entry for this?

jplatte:
We probably should, but I merged because I think asking the original author to update their PR after it previously being stuck for a month is a bit annoying.

Feel free to open a PR; otherwise I'll do it myself at some point.

PR #3142, opened 2025-01-03
Implement `OptionalFromRequest` for the Json extractor

infiniteregrets:
## Motivation

I want to receive requests with an empty body when using the Json extractor without the request getting rejected. I saw that `OptionalFromRequest` has been implemented for `Query` and `Path` and I think this is quite useful to have as it allows me to do something like:

```rs
pub async fn create(
    State(account_service): State<Arc<AccountService>>,
    Path(name): Path<String>,
    WithRejection(payload, _): WithRejection<Option<Json<CreateRequest>>, ServiceError>,
) -> Result<impl IntoResponse, ServiceError> {
...

    let Json(payload) = payload.unwrap_or_default();
...
}
```

## Solution

Implement `OptionalFromRequest` for the Json extractor.

I have tested that this works locally for myself. I don't know if this is absolutely the right way/per standards of how something like this should be approached for axum. However very quick change, so I thought I'd open a PR. Please correct me if I am wrong anywhere, I'll be happy to make any changes/close the PR if not in plans!

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- **added:** `OptionalFromRequest` impl for `Json`.
+- **added:** Implement `OptionalFromRequest` for `Json` ([#3142])
+
+[#3142]: https://github.com/tokio-rs/axum/pull/3142
 
 # 0.8.0
 
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+- **added:** `OptionalFromRequest` impl for `Json`.
+
 # 0.8.0
 
 ## since rc.1
--- a/axum/src/json.rs
+++ b/axum/src/json.rs
@@ -121,15 +121,14 @@ where
     type Rejection = JsonRejection;
 
     async fn from_request(req: Request, state: &S) -> Result<Option<Self>, Self::Rejection> {
-        if json_content_type(req.headers()) {
-            let bytes = Bytes::from_request(req, state).await?;
-            if !bytes.is_empty() {
-                Ok(Some(Self::from_bytes(&bytes)?))
-            } else {
-                Ok(None)
-            }
-        } else {
-            Ok(None)
+        let is_json = json_content_type(req.headers());
+        let bytes = Bytes::from_request(req, state).await?;
+
+        match (is_json, bytes.is_empty()) {
+            (true, true) => Err(EmptyJsonBody.into()),
+            (true, false) => Ok(Some(Json::from_bytes(&bytes)?)),
+            (false, true) => Ok(None),
+            (false, false) => Err(MissingJsonContentType.into()),
         }
     }
 }
--- a/axum/src/json.rs
+++ b/axum/src/json.rs
@@ -1,5 +1,6 @@
 use crate::extract::Request;
 use crate::extract::{rejection::*, FromRequest};
+use axum_core::extract::OptionalFromRequest;
 use axum_core::response::{IntoResponse, Response};
 use bytes::{BufMut, Bytes, BytesMut};
 use http::{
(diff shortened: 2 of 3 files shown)


axum/src/json.rs
```diff
@@ -112,6 +113,27 @@ where
...
+{
+    type Rejection = JsonRejection;
+
+    async fn from_request(req: Request, state: &S) -> Result<Option<Self>, Self::Rejection> {
+        if json_content_type(req.headers()) {
+            let bytes = Bytes::from_request(req, state).await?;
+            if !bytes.is_empty() {
+                Ok(Some(Self::from_bytes(&bytes)?))
+            } else {
+                Ok(None)
```
Turbo87, line 129: if the header says that the content-type is JSON, but the body is empty, shouldn't that be an error?

axum/src/json.rs
```diff
@@ -112,6 +113,27 @@ where
...
+    T: DeserializeOwned,
+    S: Send + Sync,
+{
+    type Rejection = JsonRejection;
+
+    async fn from_request(req: Request, state: &S) -> Result<Option<Self>, Self::Rejection> {
+        if json_content_type(req.headers()) {
+            let bytes = Bytes::from_request(req, state).await?;
+            if !bytes.is_empty() {
+                Ok(Some(Self::from_bytes(&bytes)?))
```
Turbo87, line 127: this causes a mismatching JSON body to return an error instead of `None`, is that intentional?

Turbo87 (commented):
given the various ways in which `None` could be interpreted, I'm wondering whether this causes more confusion than it solves. it might be better to use `Result<Json, _>` as the extractor and explicitly handle the corresponding error cases 🤔

jplatte:
I like it but I think only a request with _no_ content-type header and an empty body should result in `None`. Would that fit your use case, @infiniteregrets?

axum/src/json.rs
infiniteregrets: yeah that makes sense, I added a new error, is that the right way to approach this?

axum/src/json.rs
infiniteregrets: is this in reference to the previous CR comment? by mismatch do you mean an empty body here?

infiniteregrets:
thanks for the review!

@jplatte that does fit my use case here!

axum/src/json.rs
Turbo87: what rejection is the non-optional extractor using in this case? I assume `JsonSyntaxError`? would it be possible to reuse that rejection?

infiniteregrets: I looked into that one, I am not quite sure what `Error` is supposed to be here in `JsonSyntaxError(Error)` ? Is it really just specific to serde?

Turbo87: I guess we should always call `Self::from_bytes()` even if `bytes.is_empty()`, because if the header was set then we expect valid JSON and an empty body is not valid JSON 🤔

infiniteregrets: do i need to create a deserializer and then use `serde_path_to_error::deserialize` to use that error variant?

infiniteregrets: > I guess we should always call Self::from_bytes() even if bytes.is_empty(), because if the header was set then we expect valid JSON and an empty body is not valid JSON 🤔

something like:

```rs
match Self::from_bytes(&bytes) {
  Ok(_) => unreachable!("JSON body can only be empty here"),
  Err(err) => Err(err),
}               
```

axum/src/json.rs
```diff
@@ -112,6 +113,26 @@ where
...
+    async fn from_request(req: Request, state: &S) -> Result<Option<Self>, Self::Rejection> {
+        let is_json = json_content_type(req.headers());
+        let bytes = Bytes::from_request(req, state).await?;
+
+        match (is_json, bytes.is_empty()) {
+            (true, true) => Err(EmptyJsonBody.into()),
+            (true, false) => Ok(Some(Json::from_bytes(&bytes)?)),
+            (false, true) => Ok(None),
+            (false, false) => Err(MissingJsonContentType.into()),
+        }
```
jplatte, line 132: No need to check whether the body is empty if there's a JSON content type / introduce a separate rejection variant for that. Also `!is_json` is not the condition I had in mind for returning `None`. Specifically, I only wanted `None` to be returned if there is no `content-type` header at all - not if there is a `content-type` header but it has a value that's not recognized by `json_content_type`.

infiniteregrets: right, let me make some changes and get back to you on this

infiniteregrets: @jplatte made changes, is this what you meant?

jplatte (approved):
Yeah, this looks good to me. Thanks!

jplatte:
Could you add a changelog entry for this?

infiniteregrets:
yup!

axum/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+- **added:** `OptionalFromRequest` impl for `Json`.
```
jplatte, line 10:
```suggestion
- **added:** Implement `OptionalFromRequest` for `Json` ([#3142])

[#3142]: https://github.com/tokio-rs/axum/pull/3142
```

Turbo87:
I guess the new behavior should also be documented in the doc comment?

jplatte:
Ah, yeah.

Dianne087 (commented):
> 1 | *<environment>up<Advancce.Analysising>
    | ^ > 1 | *<environment>up<Advancce.MISCELLANEOUS UTILITY AND INTERFACES similar in spirit to the standard templete Libarary(STL)define by > 1 | *<environment>up<Advancce.Analysising>
    | ^ > 1 | *<environment>up<Advancce.Analysis

PR #3143, opened 2025-01-03
axum: Update matchit to 0.8.6 and support capture prefixes and suffixes

### LINKED ISSUE #3140: Update matchit to 0.8.6
The latest release [adds support for parameter suffixes](https://github.com/ibraheemdev/matchit/pull/61). matchit is currently pinned to 0.8.5 because there was some concern about this being a breaking change but I am not sure about the details. Let me know if there is anything I can do to help with that.

mladedav:
## Motivation

Closes #3140 

`matchit` released a new version which now supports prefixes and suffixes in routes like `\{image}.png`.

## Solution

This updates the dependency and makes sure the stripping works properly with the new routes.

## Open Questions

- We have locked `matchit` version to `0.8.0` (and then later to `0.8.4`) because changes like these could introduce problems such as that it would be legal to register paths using new matchit features but `axum` would be broken because it wouldn't know how to work with it. We can continue in the same spirit, locking the version to `0.8.6` or we can relax the requirement to not require the same version. I suspect e.g. [this issue](https://github.com/ibraheemdev/matchit/issues/39) could have similar effects.
- The fact that `\foo` does not match `\foo{capture}` might be surprising to users. I've added it to docs about captures but I feel it can be easily overlooked. It is consistent with normal captures which must not be empty either. We could automatically register `\foo` along with `\foo{capture}` but I'm not sure about that.
- `nest` and `nest_service` will be subtly different. In short `nest` should be preferred as it will take into consideration the whole path while `nest_service` will decide which service to call just based on the prefix. More details can be found [here](https://github.com/tokio-rs/axum/issues/3140#issuecomment-2569683544). We might consider not allowing partial captures in `nest_service` prefixes. I wouldn't expect people to run into this but if someone does, the difference might be very surprising.
- Naming of `Path` can be confusing when using e.g. `\{file}.png` as the `Path<String>` extractor will contain only the capture. I don't think we want to change this but I wanted to flag it if anyone thinks we should.


mladedav:
I've turned this into a draft because we might wait for `matchit@0.9.0` because of ambiguities between `prefix{capture}` and `{capture}suffix`. See the linked issue for details.

rosymati:
bumping this since matchit 0.9.0 has been released and I am currently depending on a private fork to use it. Would be nice to see it implemented :)

ADD-SP:
Hi, @mladedav, any update on this PR? May I take over this task if you are not working on it. I also need the newer version of `matchit`.

mladedav:
Sure go for it, I forgot to come back to this after the 0.9 matchit release.

I think it should work the same way so it should be pretty straightforward to finish this.

ADD-SP:
I have opened a new PR to move forward this work: <https://github.com/tokio-rs/axum/pull/3679>.

mladedav:
Superseded by #3702

PR #3157, opened 2025-01-07
Extension: Implement `OptionalFromRequestParts` trait

Turbo87:
## Motivation

It was previously possible to use `Option<Extension<T>>` as an extractor, but with axum 0.8 we seem to have lost that functionality.

## Solution

This PR implements the `OptionalFromRequestParts` trait for the `Extension` wrapper struct, restoring the previous behavior.


jplatte (approved):
Thanks!

jplatte:
(missing changelog entry though)

Turbo87:
> missing changelog entry though

added

PR #3161, opened 2025-01-08
axum: Move `serve.rs` to `serve/mod.rs`

mladedav:
## Motivation

We use pretty consistently `x/mod.rs` insteadd of `x.rs` when there is the directory. When we added the `serve` directory, the `serve.rs` file wasn't moved.

## Solution

Move the file.


jplatte (approved):
Let's try enabling the `self_named_module_files` clippy lint. Can be a separate PR.

PR #3163, opened 2025-01-08
axum-macros: move from_request.rs to from_request/mod.rs

mladedav:
## Motivation

We want to use `mod.rs` if we already have a directory with the same name.

## Solution

Move the file.


jplatte (commented):
I don't like how most of the diff is about files that clippy had false-positives on. Let's just rename the axum-macros file and not enable the lint for now?

Found some existing bug reports, this one has most detail I think: https://github.com/rust-lang/rust-clippy/issues/11916

mladedav:
Yeah, I agree.

jplatte (approved):
Would make sense to also change the PR title.

PR #3165, opened 2025-01-10
Add ability to set read_buffer_size for websockets

Zarathustra2:
## Motivation

https://github.com/snapview/tungstenite-rs/pull/468 changed the read buffer from 64KB to 128KB. It also added a method `read_buffer_size` so the read buffer size can be set. Currently there is no way to my knowledge so that Axum users can change the read buffer size. 

After upgrading to axum 0.8 I noticed increased memory usage for my Websocket server and I assume the high read buffer is contributing to this. Since my websocket only receives small messages I would like to set it to something smaller than 128KB.

## Solution
Added `WebsocketUpgrade::read_buffer_size` so that the read buffer size can be set.

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -152,6 +152,12 @@ impl<F> std::fmt::Debug for WebSocketUpgrade<F> {
 }
 
 impl<F> WebSocketUpgrade<F> {
+    /// Read buffer capacity. The default value is 128 KiB.
+    pub fn read_buffer_size(mut self, size: usize) -> Self {
+        self.config.read_buffer_size = size;
+        self
+    }
+
     /// The target minimum size of the write buffer to reach before writing the data
     /// to the underlying stream.
     ///


Zarathustra2:
I was wrong the default was 4kb and then got upgrade to 64kb and then to 128kb, see here: https://github.com/snapview/tungstenite-rs/issues/480

I think there might be arguments for axum to change the default from 128kb to something lower. I could make that change but would need to know with what value the maintainers are comfortable with. 

What do you think?

xd009642:
It would be nice if this could land, working on a server with a lot of websocket usage means we have to be very cautious about upgrading to axum 0.8

yanns:
Is it the same change as https://github.com/tokio-rs/axum/pull/3178 ?

xd009642:
oh wait yeah it looks to be, this should possibly just be closed then :eyes:

yanns:
https://github.com/tokio-rs/axum/pull/3178 is missing a changelog. Maybe you can take care of this?

Zarathustra2:
u mean me? @yanns I can do this if u want

Issue #3164, opened 2025-01-09
Make <Serve as Future>::Output = !


jplatte:
See #2895. The PR conflicted with other ongoing work so it was not merged, but the idea was good and I think we should revisit it for the next breaking-change release. (unfortunately forgot to do it for 0.8)

jplatte:
@SabrinaJewson I guess we could have merged the docs aspect of the PR / still merge it now. Do you want to open a new PR for that?

SabrinaJewson:
Right, PR sent.

In writing that PR, a thought came to mind: currently, in `with_graceful_shutdown` we have `F: Future<Output = ()>`. It seems reasonable that we might want to have `F: Future`, and the return value of `WithGracefulShutdown`’s future be `F::Output`. There is a subtle change in behaviour in that we would be forced to propagate instead of swallow panics, but I think that’s closer to what the user desires anyway.

Another thought that comes to mind is that it’s a performance hazard to run `axum::serve` on the main thread – i.e. inside `#[tokio::main]` – which I believe is because (a) it’s slower to spawn a task from the main thread than a worker thread (as the main thread doesn’t have access to a local queue) and (b) it’s possible having more active threads than cores is bad for performance.

Which is to say, I could envisage an API that looks more like this:

```rs
let server = axum::serve(listener, router);
tokio::signal::ctrl_c().await?;
server.graceful_shutdown().await;
```

Of course, `Serve` could still implement `IntoFuture<Output = !>` for the case when no graceful shutdown is needed, though that implementation would just have `Future = Pending<!>` (well, I guess it would be `Pending<Infallible>` in practice). Like before, if `Serve` is dropped then new connections would no longer be accepted while old connections would continue to run.

There is a drawback to this API, however, which is that `axum::serve` is now eager, meaning that we can no longer extend it with methods that control aspects of the serving. I don’t know how much we want this door to be left open. Of course, we could always use a builder:

```rs
axum::serve_builder(listener, router)
    .set_some_config(some_value)
    .build()
    .await
````

Or alternatively, have a `.start()` method returning a `Handle` type:

```rs
// option A:
axum::serve(listener, router).await

// option B, for graceful shutdown:
let server = axum::serve(listener, router).start();
tokio::signal::ctrl_c().await?;
server.graceful_shutdown().await;
```

[...]

PR #3166, opened 2025-01-10
Guarantee return value of `serve`, Pt. 2

SabrinaJewson:
See #3164. This PR is the revival of #2895.

Like before, in order to ensure in the type system that errors cannot be returned, the logic of `WithGracefulShutdown` has been extracted to a separate `run` function.


jplatte (approved):
Thanks!

PR #3168, opened 2025-01-10
refactor: make `status` a const function in rejection handling

### LINKED ISSUE #3167: Make `status` function that generated by `__define_rejection` a const function

## Feature Request
Make `status` function that generated by `__define_rejection` a const function

https://github.com/tokio-rs/axum/blob/d07d129bafb182a54cbfbaa5708feed916d7ad45/axum-core/src/macros.rs#L56-L59

### Motivation
Allows the status code to be evaluated at compile time

TheAwiteb:
Fix #3167, Updated the `status` method to be a const function for improved performance and compile-time evaluation.

## Motivation

Allows the status code to be evaluated at compile time

## Solution

```diff
-pub fn status(&self) -> http::StatusCode {
+pub const fn status(&self) -> http::StatusCode {
```


Turbo87 (dismissed):
could you add the `const` keyword for the second macro variant too?

and do you know if we could do the same for `__composite_rejection`?

TheAwiteb:
> could you add the `const` keyword for the second macro variant too?

No, because it uses [`ToString::to_string`](https://doc.rust-lang.org/std/string/trait.ToString.html), which is not a const function.

> and do you know if we could do the same for __composite_rejection?

Yes I can, for the `status` function only

Turbo87:
> No, because it uses [`ToString::to_string`](https://doc.rust-lang.org/std/string/trait.ToString.html), which is not a const function.

I'm talking about the function in line 115, and I don't see a `to_string()` call in there :)

TheAwiteb:
`FailedToDeserializePathParams` and `InvalidUtf8InPathParam` are used in the `__composite_rejection` macro, but their `status` functions are not implemented by the `__define_rejection` macro. Therefore, I made their `status` functions `const`.

jplatte (approved):
Makes sense to me, but please note that this does not impact performance in any way. `const fn` means it can be evaluated in `const` contexts such as `const ITEM`s and `static`s, it has no influence on whether optimizations like constant folding are applied in code outside of a `const` context.

jplatte:
Could you add changelog entries to `axum` and `axum-extra` (where the macro-generated methods live)?

jplatte (approved):
Thanks!

Issue #3170, opened 2025-01-12
With the move of the `Host` extractor to the axum-extra crate, we've lost the `OptionalFromRequestParts` implementation for it


dnlsndr:
- [x] I have looked for existing issues (including closed) about this

### Version

v0.8.1

### Platform

UNIX

### Crates

axum
axum-extra

### Description

Now that the `Host` extractor was moved to the axum-extra crate, it's not possible to do something like this anymore:

```rust
use axum_extra::extract::Host;

async fn handler(host: Option<Host>){}
```

as the compiler complains with this error:
```
the trait bound `axum_extra::extract::Host: OptionalFromRequestParts<Arc<AppState>>` is not satisfied
you can use `cargo tree` to explore your dependency tree
required for `std::option::Option<axum_extra::extract::Host>` to implement `FromRequestParts<Arc<AppState>>`
```

taladar:
The `Query` extractor is still missing one of those too.

taladar:
Specifically I meant the one from axum but axum_extra seems to have one too that is also affected.

dnlsndr:
@taladar Check this out: https://github.com/tokio-rs/axum/issues/3079 although their reasoning does not apply to extractors such as `Host`

taladar:
I see, so in case someone comes across this `OptionalQuery` from axum_extra looks like the correct replacement for the situation where the whole query value is optional, not just individual fields.

Turbo87:
@taladar `OptionalQuery` is close to being deprecated too. I would recommend to wrap all your query fields with `Option` instead, since that best reflects reality of how query parameters work.

taladar:
Semantically 99% of the time when I even want several query parameters stored in the same value I want to know if all of them have been specified though, not a mess of several `Option` values, some of them potentially `None` that I have to check manually again anyway.

jplatte:
Could you give an example? I have the exact opposite experience, I have never seen a set of query parameters that are optional _as a group_, rather than individually.

jplatte:
@dnlsndr Want to send a PR?

taladar:
Usually that occurs when you have related information that is useless if you only have some of it, e.g. you have a map API with x, y and z parameters or a sort by and sort direction parameter.

dnlsndr:
I'd be happy to! PR incoming.

Turbo87:
> you have a map API with x, y and z parameters



```rust
let (x, y, z) = match (x, y, z) {
[...]

PR #3177, opened 2025-01-14
Implement the OptionalFromRequestParts trait for the Host extractor

dnlsndr:
Fixes: #3170 

## Motivation

To allow for `Option<Host>` extractors we need to implement the OptionalFromRequestParts trait for it.

## Solution

Implement OptionalFromRequestParts for the Host Extractor

### DIFF
--- a/axum-extra/src/extract/host.rs
+++ b/axum-extra/src/extract/host.rs
@@ -171,10 +171,7 @@ mod tests {
     async fn ip4_uri_host() {
         let mut parts = Request::new(()).into_parts().0;
         parts.uri = "https://127.0.0.1:1234/image.jpg".parse().unwrap();
-        let host =
-            <Host as axum::extract::FromRequestParts<_>>::from_request_parts(&mut parts, &())
-                .await
-                .unwrap();
+        let host = parts.extract::<Host>().await.unwrap();
         assert_eq!(host.0, "127.0.0.1:1234");
     }
 
@@ -182,21 +179,14 @@ mod tests {
     async fn ip6_uri_host() {
         let mut parts = Request::new(()).into_parts().0;
         parts.uri = "http://cool:user@[::1]:456/file.txt".parse().unwrap();
-        let host =
-            <Host as axum::extract::FromRequestParts<_>>::from_request_parts(&mut parts, &())
-                .await
-                .unwrap();
-
+        let host = parts.extract::<Host>().await.unwrap();
         assert_eq!(host.0, "[::1]:456");
     }
 
     #[crate::test]
     async fn missing_host() {
         let mut parts = Request::new(()).into_parts().0;
-        let host =
-            <Host as axum::extract::FromRequestParts<_>>::from_request_parts(&mut parts, &())
-                .await
-                .unwrap_err();
+        let host = parts.extract::<Host>().await.unwrap_err();
         assert!(matches!(host, HostRejection::FailedToResolveHost(_)));
     }
 
@@ -204,20 +194,14 @@ mod tests {
     async fn optional_extractor() {
         let mut parts = Request::new(()).into_parts().0;
         parts.uri = "https://127.0.0.1:1234/image.jpg".parse().unwrap();
-        let host = Option::<Host>::from_request_parts(&mut parts, &())
-            .await
-            .unwrap();
-
+        let host = parts.extract::<Option<Host>>().await.unwrap();
         assert!(host.is_some());
     }
 
     #[crate::test]
     async fn optional_extractor_none() {
         let mut parts = Request::new(()).into_parts().0;
-        let host = Option::<Host>::from_request_parts(&mut parts, &())
-            .await
-            .unwrap();
-
+        let host = parts.extract::<Option<Host>>().await.unwrap();
         assert!(host.is_none());
     }
 
--- a/axum-extra/src/extract/host.rs
+++ b/axum-extra/src/extract/host.rs
@@ -1,5 +1,8 @@
 use super::rejection::{FailedToResolveHost, HostRejection};
-use axum::extract::{FromRequestParts, OptionalFromRequestParts};
+use axum::{
+    extract::{FromRequestParts, OptionalFromRequestParts},
+    RequestPartsExt,
+};
 use http::{
     header::{HeaderMap, FORWARDED},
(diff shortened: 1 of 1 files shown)


axum-extra/src/extract/host.rs
```diff
@@ -148,18 +167,55 @@ mod tests {
     async fn ip4_uri_host() {
         let mut parts = Request::new(()).into_parts().0;
         parts.uri = "https://127.0.0.1:1234/image.jpg".parse().unwrap();
-        let host = Host::from_request_parts(&mut parts, &()).await.unwrap();
+        let host =
+            <Host as axum::extract::FromRequestParts<_>>::from_request_parts(&mut parts, &())
+                .await
+                .unwrap();
```
Turbo87, line 173:
```suggestion
        let host = parts.extract::<Host>().await.unwrap();
```

I think this should work too if the `RequestPartsExt` trait is imported

axum-extra/src/extract/host.rs
```diff
@@ -59,6 +60,24 @@ where
...
+
+    async fn from_request_parts(
+        parts: &mut Parts,
+        _state: &S,
+    ) -> Result<Option<Self>, Self::Rejection> {
+        Ok(
+            <Self as FromRequestParts<S>>::from_request_parts(parts, _state)
+                .await
+                .ok(),
+        )
```
Turbo87, line 77: if I understand correctly, this implementation ignores any rejections and turns them into `None`, which would match the previous behavior. I'm not sure if that's the behavior we want though.

it might be better to turn `OptionalFromRequestParts` into the primary implementation and in `FromRequestParts` we use `parts.extract::<Option<Host>>().await.unwrap_or(FailedToResolveHost)` or something roughly like that.

dnlsndr: Had to `.ok().flatten()` it because it's a `Result<Option<Host>, Infallible>` now

axum-extra/src/extract/host.rs
```diff
@@ -31,31 +35,50 @@ where
     type Rejection = HostRejection;
 
     async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
+        parts
+            .extract::<Option<Host>>()
+            .await
+            .ok()
+            .flatten()
+            .ok_or(HostRejection::FailedToResolveHost(FailedToResolveHost))
```
yanns, line 43: A nitpick: the `OptionalFromRequestParts` implementation contains less information than the `FromRequestParts` one. In the future, we could, for example, return other errors than just `HostRejection::FailedToResolveHost`.
I'd suggest to keep the implementation of `FromRequestParts` and to base the `OptionalFromRequestParts` implementation on top of it, removing the error information.

Turbo87: @yanns if we actually introduce other rejections for `Host` then we could change the `OptionalFromRequestParts` implementation to reject with `HostRejection` too. I don't see why we would need separate implementations for this 🤔

Turbo87: `OptionalFromRequestParts` returns `Result<Option<T>, Rejection>` 😉

yanns: ah right, then forget about my comment. I had the feeling that `OptionalFromRequestParts` is swallowing errors and was less precise than `FromRequestParts`. But I'm probably wrong here.

dnlsndr:
@jplatte feel free to review :)

mladedav:
I don' t hink we want `Option` to be the same as the normal extractor and just discard errors.

I think it would be better to surface the parsing errors and return `None` only if there weren't any. I believe that was the idea behind `OptionalFromRequestParts`, when you just want to discard errors, you should use `Result<T, _>`. Arguably the full extractor should also differentiate between missing and malformed headers so this might have to wait until the next breaking release.

Others' opinions might differ and we can go ahead with this in that case, if not I'm not sure where that would leave this PR. We could merge it and change it along with the other extractor or just wait with it.

jplatte:
My availability is limited for the next two weeks or so, but I'll get to reviewing it some time next month.

Turbo87:
> I don' think we want `Option` to be the same as the normal extractor and just discard errors.

@mladedav the `Host` extractor in its current state already swallows any parsing errors and just discards the header. the only rejection that it currently returns is `FailedToResolveHost`, which is equivalent to the `None` case, so this refactored implementation does not discard more errors than before due to this particular `OptionalFromRequestParts` implementation using `type Rejection = Infallible`.

see also https://github.com/tokio-rs/axum/pull/3177#discussion_r1916232970 😉

mladedav:
Yeah, I get that it doesn't discard _more_ errors than the other extractor, but that doesn't change that it does discard them. I appreciate we want the two to work as similarly as possible, but I'm not sure that this addition is really needed then.

The only reason I can see why people want this is because people are used to `Option` being just `Result` where all errors are conflated into `None`. Which I think we should not promote because it already is not the case for some extractors and it will only get more surprising. In the linked issue someone seemed genuinely surprised they can just extract `Result` to get basically the same thing so I think this is just an issue with education rather than users really wanting `Option`s specifically.

I generally don't think we should have any `OptionalFromRequestParts` extractors where the implementation can be done with `Result::ok`. That doesn't mean we should have the optional extractor here, I just think it should mean something slightly different which is arguably also different from what the main extractor does right now.

jplatte (approved):

jplatte:
I think given the current `FromRequestParts` impl, this `OptionalFromRequestParts` impl makes sense. But I don't know whether the former is all that sensible.

Honestly, I'd love to just deprecate `Host` (alongside the recently-added `Scheme` extractor 🥲) because there's just so many different ways users could want / expect the error handling to work.

FancyQian:
@dnlsndr can we merge this code? I need those fix in axum upstream since it help me a lot

dnlsndr:
It's not up to me to merge. It has been approved and probably will be merged, once deemed appropriate.

FancyQian:
> It's not up to me to merge. It has been approved and probably will be merged, once deemed appropriate.

Thanks for the quick response.

Turbo87:
IMHO this is ready to merge.

I agree though that the current implementation of the extractor in general seems problematic and might not be worth keeping around.

PR #3179, opened 2025-01-16
Update low-level-rustls example - added http2 feature flag

richardalanjones:
## Motivation

- Upon starting example server and running curl request as specified, request fails and server logs warning "HTTP/2 is not supported".

## Solution

- added "http2" feature flag to hyper-util, which is required feature in order to utilize HTTP2.


mladedav (approved):
Thanks

PR #3187, opened 2025-01-21
ci: Remove condition from typos job

jplatte:
This condition is kind of broken: Un-drafting a PR does not make the job run. I think the only solution to that is making the entire workflow (re-)trigger on draft status change, which seems overkill. I suppose we could split this job out into its own workflow file to make that less costly, but I'm skeptical about skipping this for draft PRs in the first place.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -248,4 +248,4 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Check the spelling of the files in our repo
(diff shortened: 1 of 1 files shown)


yanns (approved):

Turbo87 (approved):
with a runtime of 3 sec this seems to be one of our fastest jobs, so I'm not sure why this was skipped for draft PRs in the first place 😅

mladedav (approved):

PR #3188, opened 2025-01-21
Release axum 0.8.2 and related crates

jplatte:
We've had a couple of useful changes, release them.


evanrichter:
so you are aware, v0.8.2 is not on crates.io, despite the github release and tag

jplatte:
It is on crates.io, but in yanked state. I'm currently very busy with real life so haven't put time into 0.8.3.

If you or anybody else wants to move things along, you can read up on why the release was yanked in the closed issue about breaking changes. Then as a next step it would be good to open a discussion proposing a way forward regarding the changes that lead to breakage (a simple solution might be to just revert the `const fn` change that caused issues, pubpish new releases and revisit that change afterwards).

evanrichter:
ah sorry, I didn't see anything about the yank. maybe it could be added as a line in the github release.

I don't mean to add pressure to release 0.8.3, it just seemed like something was missed

mladedav:
I've added a line on top of that release to tell future visitor that it has been yanked. Thanks for the suggestion.

PR #3197, opened 2025-01-24
feat: Add `302 Found` redirect method

DSharifi:
## Motivation

In the HTTP handler I am implementing with Axum, I have the use case for doing redirects with [`302 Found`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302) HTTP status code. The use case is for temporary redirects where the subsequent HTTP method should always be `GET`.

## Solution

This PR adds a new method to `axum::response::Redirect` to create redirects with the `302` HTTP status code.


jplatte:
If you want the request method to be `GET`, use [`Redirect::to`](https://docs.rs/axum/latest/axum/response/struct.Redirect.html#method.to) (status code 303).

Status code 302 is intentionally not exposed because it's not defined what the request method after redirect should be.

PR #3205, opened 2025-01-31
axum: allow body types other than `axum::body::Body` in `Service`s passed to `serve`

mladedav:
## Motivation

A user in #3202 tried to use middlewares around the whole `Router` so that their layers run before routing happens. They ran into problems caused by their layers changing the `Body` type.

So far I've seen only the `NormalizePathLayer` used in this way which does not change the `Body` type so there were never issues. But I don't see a reason not to support this for other layers that do change the body, such as `CompressionLayer`.

## Solution

This PR adds another generic parameter to `Serve` to keep track of the `Response::Body` type returned by the service provided.

There are no breaking changes as far as I can tell as inference knows the body type. This could break someone who for some reason names the type parameters explicitly. There is no rush with this and we may want to wait with this and release it in the next breaking release.

## Alternatives

We may also choose not to support this. Users can just map the response body with `.map_response_body(axum::body::Body::new)` before callign `serve`. The price is just one box (and maybe a bit of ergonomics) as far as I can tell.

### DIFF
--- a/axum/src/serve/mod.rs
+++ b/axum/src/serve/mod.rs
@@ -2,6 +2,7 @@
 
 use std::{
     convert::Infallible,
+    error::Error as StdError,
     fmt::Debug,
     future::{Future, IntoFuture},
     io,
@@ -11,6 +12,7 @@ use std::{
 
 use axum_core::{body::Body, extract::Request, response::Response};
 use futures_util::FutureExt;
+use http_body::Body as HttpBody;
 use hyper::body::Incoming;
 use hyper_util::rt::{TokioExecutor, TokioIo};
 #[cfg(any(feature = "http1", feature = "http2"))]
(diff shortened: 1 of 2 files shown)


axum/src/serve/mod.rs
```diff
@@ -686,4 +701,31 @@ mod tests {
...
+            ) -> std::task::Poll<Option<Result<http_body::Frame<Self::Data>, Self::Error>>>
+            {
+                #![allow(clippy::unreachable)] // The implementation is not used, we just need to provide one.
+                unreachable!();
+            }
+        }
+
+        let app = ServiceBuilder::new()
+            .layer_fn(|_| tower::service_fn(|_| std::future::ready(Ok(Response::new(CustomBody)))))
+            .service(Router::<()>::new());
```
yanns, line 723: If we add some routes, which type will be the body? The axum body type? Or `CustomBody`?

yanns: Could we add an handle to check if this works? Or are our handlers tight to our body implementation?

mladedav: I can do that (in a few days).

The handler (and `Router`) would still return `axum::body::Body`, it's just the outer layer which changes that to something else.

ttys3:
seems this PR is a reverse of https://github.com/tokio-rs/axum/pull/1751 ?

axum/src/serve/mod.rs
mladedav: I've added a route there.

Note though that the whole `Router` is completely ignored by the outer layer. I could even write this instead and it would work the same way even without the router.
```
let app = tower::service_fn(|_| std::future::ready(Ok(Response::new(CustomBody))));
```

yanns: I see. What would be a better example?
A middleware changing the body type depending of the request?
(I have issues understanding the value)

mladedav: #3002 is good example I think. I just wanted to add a test that stuff like this compiles going forward.

We can add a real example or extend the one showing middlewares wrapping the whole router (assuming we already have one)

yanns: There's this example I wrote: https://github.com/tokio-rs/axum/blob/main/examples/compression/src/main.rs
(but I guess there's no need to layer outside the router for this)
I'll approve this PR code-wise. Feel free to enhance it with examples / tests showing its value as I cannot find any myself.

yanns (approved):

jplatte:
Can you add a changelog entry?

jplatte:
Can you rebase? Or do you want me to?

mladedav:
Rebased

PR #3224, opened 2025-02-14
feat: Add `struct Accepted` as a shortcut

anagno:
Similar to https://github.com/tokio-rs/axum/pull/2978 I am adding a struct for the Accepted response

## Motivation

We have a server that we have to trigger a request and it needs to return `Accepted`. Since there is already a shortcut for `No Content`, I thought of adding a similar one for `Accepted`

## Solution

Adding the additional stuct for shortcut, instead of having to define it always in my code

### DIFF
--- a/axum/src/response/mod.rs
+++ b/axum/src/response/mod.rs
@@ -84,6 +84,31 @@ impl IntoResponse for NoContent {
     }
 }
 
+/// An empty response with 202 Accepted status.
+///
+/// Due to historical and implementation reasons, the `IntoResponse` implementation of `()`
+/// (unit type) returns an empty response with 200 [`StatusCode::OK`] status.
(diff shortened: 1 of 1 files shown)


BeowulfR:
I understand the addition of the `NoContent` struct, because using the `NoContent`struct prevents adding a body - what the `NoContent` status should never have.

But `Accepted` could have a body, so I find the addition unnecessary and just a pollution 🤔  If this struct would be added, why wouldn't a struct be added for every status code? And this would be to much...

Where is the problem with just using `StatusCode` as return type and returning `StatusCode::ACCEPTED`? 🤔

anagno:
@BeowulfR You are absolute right. I thought that I always had to return a struct and focused on the wrong thing. Returning directly the StatusCode makes sense ! Thank you for the clarification !

PR #3234, opened 2025-02-24
docs: correct TypedHeader link in axum-core/extract/request_parts.rs

K-tecchan:
## Motivation

https://docs.rs/axum/0.8/axum/extract/struct.TypedHeader.html didn't seem to be working.

## Solution

I corrected it to https://docs.rs/axum-extra/latest/axum_extra/struct.TypedHeader.html.

### DIFF
--- a/axum-core/src/extract/request_parts.rs
+++ b/axum-core/src/extract/request_parts.rs
@@ -53,7 +53,7 @@ where
 ///
 /// Prefer using [`TypedHeader`] to extract only the headers you need.
 ///
-/// [`TypedHeader`]: https://docs.rs/axum/0.8/axum/extract/struct.TypedHeader.html
+/// [`TypedHeader`]: https://docs.rs/axum-extra/latest/axum_extra/struct.TypedHeader.html
 impl<S> FromRequestParts<S> for HeaderMap
 where
     S: Send + Sync,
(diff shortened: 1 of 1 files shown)


mladedav:
This changes the link from pointing to a (wrong) specific version to latest. Which means after we release new semi-major version of `axum-extra`, this link will be broken on this version.

So I'm not honestly sure what we should prefer, never having wrong links on the latest version or not breaking older docs but sometimes having a bad link like this.

I guess we could try adding some kind of CI check that all links are to expected versions.

axum-core/src/extract/request_parts.rs
```diff
@@ -53,7 +53,7 @@ where
 ///
 /// Prefer using [`TypedHeader`] to extract only the headers you need.
 ///
-/// [`TypedHeader`]: https://docs.rs/axum/0.8/axum/extract/struct.TypedHeader.html
+/// [`TypedHeader`]: https://docs.rs/axum-extra/latest/axum_extra/struct.TypedHeader.html
```
jplatte, line 56:
```suggestion
/// [`TypedHeader`]: https://docs.rs/axum-extra/0.10/axum_extra/struct.TypedHeader.html
```

jplatte (approved):

PR #3248, opened 2025-03-05
Add `WebSocketUpgrade::selected_protocol`

coolreader18:
## Motivation

This allows you to check the subprotocol before spawning the websocket task, letting you e.g. bail out with a `4XX` if no subprotocol was provided.

## Solution

Add a `WebSocketUpgrade::selected_protocol(&self) -> Option<&HeaderValue>` function, so you can access the same value as `WebSocket::protocol()` without yet calling `on_upgrade`.

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -265,6 +265,15 @@ impl<F> WebSocketUpgrade<F> {
         self
     }
 
+    /// Return the selected WebSocket subprotocol, if one has been chosen.
+    ///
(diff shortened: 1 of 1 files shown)


coolreader18:
Is there anyone who could take a look at this?

jplatte:
Ping @SabrinaJewson, as the resident WebSockets and SSE expert ^^

SabrinaJewson:
Yeah, this looks good to me. We already expose the functionality, so this is just exposing it a little earlier.

jplatte (approved):

PR #3251, opened 2025-03-06
feat(extract): prefixed signed&private cookies

### LINKED ISSUE #3246: [axum-extra] Cookie Name Prefixes

### Motivation

Cookie Name Prefixes is defined in the draft [RFC6265bis](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis#section-4.1.3) ([HTTP WG draft](https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#section-4.1.3))

### Proposal

`cookie` has https://docs.rs/cookie/latest/cookie/prefix/struct.PrefixedJar.html - can it be used somehow?

### Alternatives

Manually doing all this is error prone...

andrzejSulkowski:
closes #3246

# Cookie Prefixes Support

## Motivation
Cookie prefixes (__Host- and __Secure-) are a web security feature that help protect against cookie manipulation and session fixation attacks. This PR enhances Axum's cookie handling by adding proper support for cookie prefixes across all cookie jar types (`CookieJar`, `SignedCookieJar`, and `PrivateCookieJar`).

## Solution

This PR implements:
- Support for cookie prefixes with correct attribute enforcement (e.g., Secure flag, Path=/) using the underlying cookie crate's PrefixedJar
- Consistent prefix handling methods (`add_prefixed`, `get_prefixed`, `remove_prefixed`) for all cookie jar types
Integration of prefix support with signed and private cookies, ensuring the encryption/signing is maintained
- Extended cookie tests to verify cookie encryption/signing rather than plain-text value matching

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -102,9 +102,7 @@ tower-service = "0.3"
 
 # optional dependencies
 axum-macros = { path = "../axum-macros", version = "0.5.0", optional = true }
-cookie = { package = "cookie", version = "0.18.0", features = [
-    "percent-encode",
-], optional = true }
+cookie = { package = "cookie", version = "0.18.0", features = ["percent-encode"], optional = true }
 fastrand = { version = "2.1.0", optional = true }
 form_urlencoded = { version = "1.1.0", optional = true }
 headers = { version = "0.4.0", optional = true }
@@ -124,11 +122,7 @@ typed-json = { version = "0.1.1", optional = true }
 axum = { path = "../axum", features = ["macros", "__private"] }
 axum-macros = { path = "../axum-macros", features = ["__private"] }
 hyper = "1.0.0"
-reqwest = { version = "0.12", default-features = false, features = [
-    "json",
-    "stream",
-    "multipart",
-] }
+reqwest = { version = "0.12", default-features = false, features = ["json", "stream", "multipart"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.71"
 tokio = { version = "1.14", features = ["full"] }


Turbo87:
@jplatte do you know if there is a reason why the inner jars are not exposed as public API? seems like that would simplify the implementation quite a bit 🤔

axum-extra/Cargo.toml
```diff
@@ -16,7 +16,13 @@ default = ["tracing"]
...
 async-read-body = ["dep:tokio-util", "tokio-util?/io", "dep:tokio"]
 async-stream = [] # unused, remove before the next breaking-change release
-file-stream = ["dep:tokio-util", "tokio-util?/io", "dep:tokio", "tokio?/fs", "tokio?/io-util"]
+file-stream = [
+    "dep:tokio-util",
+    "tokio-util?/io",
+    "dep:tokio",
+    "tokio?/fs",
+    "tokio?/io-util",
+]
```
Turbo87, line 49: @andrzejSulkowski any reason for these toml file changes? are these only changing the formatting or are there also semantic changes hidden in them?

yanns: I bet this is automated formatting from zed...

axum-extra/src/extract/cookie/mod.rs
```diff
@@ -268,6 +345,19 @@ mod tests {
                     .await
                     .unwrap();
                 let cookie_value = res.headers()["set-cookie"].to_str().unwrap();
+                println!("Set Cookie value: {}", cookie_value);
```
yanns: One `println` to remove?

axum-extra/src/extract/cookie/mod.rs
```diff
@@ -302,17 +392,115 @@ mod tests {
...
+                {
+                    assert!(!cookie_value.contains("key=value"));
+                } else {
+                    assert!(cookie_value.contains("key=value"));
+                }
+
+                // Extract just the cookie part (before the first semicolon)
+                // Set-Cookie: __Host-key=value; Secure; Path=/ -> __Host-key=value
+                let cookie_header_value = cookie_value.split(';').next().unwrap().trim();
+                println!("Using Cookie header value: {}", cookie_header_value);
```
yanns: one `println` to remove?

axum-extra/src/extract/cookie/mod.rs
```diff
@@ -302,17 +392,115 @@ mod tests {
...
+                    .route("/remove", get(remove_cookie_prefixed))
+                    .with_state(state);
+
+                let res = app
+                    .clone()
+                    .oneshot(Request::builder().uri("/set").body(Body::empty()).unwrap())
+                    .await
+                    .unwrap();
+                let cookie_value = res.headers()["set-cookie"].to_str().unwrap();
+                println!("Set Cookie value: {}", cookie_value);
```
yanns: one `println` to remove?

yanns (approved):
The code is looking good to me.
Nitpick: I don't know if we want some `println` in the code through.

I don't know enough the API + our consensus about exposing the cookie API to make a decision here.

axum-extra/Cargo.toml
Turbo87: seems like it's just noise in the diff that shouldn't have been committed then. it's unfortunately also causing a merge conflict.

andrzejSulkowski:
> @jplatte do you know if there is a reason why the inner jars are not exposed as public API? seems like that would simplify the implementation quite a bit 🤔

Are you suggesting we expose the inner jars alongside the current convenience methods, or replace the current API entirely with direct access to the underlying cookie crate jars?

axum-extra/Cargo.toml
andrzejSulkowski: restored to the old formatting

axum-extra/src/extract/cookie/mod.rs
andrzejSulkowski: done

axum-extra/src/extract/cookie/mod.rs
andrzejSulkowski: done

axum-extra/src/extract/cookie/mod.rs
andrzejSulkowski: done

Turbo87:
> Are you suggesting we expose the inner jars alongside the current convenience methods

maybe :D

I don't know the history of this extractor, so before I can form an opinion it would be better to know why this wasn't done in the first place 😅

andrzejSulkowski:
> > Are you suggesting we expose the inner jars alongside the current convenience methods
> 
> maybe :D
> 
> I don't know the history of this extractor, so before I can form an opinion it would be better to know why this wasn't done in the first place 😅

I don't know the historical reasoning either, and I'm not really proficient enough to predict what users might need.
But the hybrid approach seems like a good middle ground - keep the convenience methods for quick development while exposing the inner jars for more tailored use-cases that need the full flexibility of the cookie crate.

What do you think?

jplatte:
I think the historical reason was that the cookie stuff used to be in axum and David didn't want the cookie crate as part of the public API, to be able to upgrade it across major versions within a minor version of axum.

However now the cookie crate is already part of the public API anyways, so I think we might as well.. Maybe?

It's been a while that I've actually looked at the details here. If you want, I can take a closer look before time is spent on refactoring stuff here that may be reversed again at some point.

andrzejSulkowski:
> I think the historical reason was that the cookie stuff used to be in axum and David didn't want the cookie crate as part of the public API, to be able to upgrade it across major versions within a minor version of axum.
> 
> However now the cookie crate is already part of the public API anyways, so I think we might as well.. Maybe?
> 
> It's been a while that I've actually looked at the details here. If you want, I can take a closer look before time is spent on refactoring stuff here that may be reversed again at some point.

I'd appreciate clear instructions 🙏

PR #3261, opened 2025-03-11
Use via extractor's rejection instead of `axum::response::Response`

### LINKED ISSUE #3160: `FromRequestParts` derive macro uses wrong rejection

### Version

0.8.1

### Crates

`axum-macros`

### Description

https://docs.rs/axum/latest/axum/extract/derive.FromRequest.html#the-whole-type-at-once claims that:

> The rejection will be the “via extractors”’s rejection. For the previous example that would be [axum::extract::rejection::ExtensionRejection](https://docs.rs/axum/0.8.1/axum/extract/rejection/enum.ExtensionRejection.html).

but from what I've found, this is incorrect. I've tried to use `#[from_request(via(Query))]` on a struct, assuming that the corresponding rejection would be `QueryRejection`, but instead `axum::response::Response` was put in the derived impl (documented as the default at https://docs.rs/axum/0.8.1/axum/extract/derive.FromRequest.html#the-rejection).

I assume that this is a bug in the derive macro for at least `FromRequestParts`, but probably `FromRequest` too.

kyle-nweeia:
Fixes #3160.

Makes the `from_request` attribute use the rejection type assigned to the extractor by the `FromRequest` or `FromRequestParts` implementation. This brings it in line with the behavior documented at https://docs.rs/axum/latest/axum/extract/derive.FromRequest.html#the-whole-type-at-once.

Updates two tests that expected the extractor rejection to be `axum::response::Response`.


mladedav (approved):
Looks good to me, thanks.

jplatte:
A changelog entry would be nice.

mladedav:
Regarding this being a breaking change - I guess we want to be on the safe side after the last release but it's true this was always just a bug since it's explicitly documented as using the other rejection.

But to be clear, do we want to wait for 0.9 before releasing this?

jplatte:
Yeah, looks good. Thanks!

jplatte (approved):

jplatte:
Looks like this needs a rebase.

Issue #3263, opened 2025-03-12
Improving the examples - adding tests


szabgab:
Looking at the examples in the `examples` folder I see that many don't come with tests.
Having tests with the examples will both help make sure that they stay correct as axum changes and it will help the would be developers to learn how to test.

As I see many of the examples have 

```rust
async fn main() {
   ...
   let app = Router::new()
   ...
}
```

This, as far as I understand, makes it impossible to use the router in the tests. So I think it would be a good idea to refactor these examples to defined a function

```rust
fn app() -> Router {
}
```

and then write

```rust
async fn main() {
   ...
   let app: Router = app();
   ...
}
```

I'd be happy to do this work for some of the examples, but I'd like to know if you even consider this a good idea.

yanns:
For people who would like starting contributing, it can be a nice way to add tests for one example. Step by step, we could reach completion.

tidely:
I think having additional indirection in examples complicates the learning process. I think most people want to learn incrementally and having an additional 50-100 lines of tests on examples equivalent to hello world doesn't really achieve that goal. I'd restrict testing to more advanced examples only. If the goal is to help users write tests for their routers, there are ready examples for that.

Axum tries not to introduce implicit behaviour changes, meaning examples are unlikely to break once written. If there are breaking changes, they can be caught using `cargo check --workspace`. I think it'd be more beneficial to make the examples seem like examples to cargo as well, so we can run `cargo check --examples` separately.

darth-raijin:
> I think having additional indirection in examples complicates the learning process. I think most people want to learn incrementally and having an additional 50-100 lines of tests on examples equivalent to hello world doesn't really achieve that goal. I'd restrict testing to more advanced examples only. If the goal is to help users write tests for their routers, there are ready examples for that.

Examples aren’t only for beginners learning the absolute basics.
For many users, examples are the only place to learn:
* how to structure tests
* how to create a shared `app()` in a good way to propegate through tests
* best practices for doing assertions in a safe and conclusive manner

These concepts aren’t trivial, and they’re not covered anywhere else in Axum.

By having community maintained tests, everyone can write safer and better software, because we can learn from each others patterns. Personally, the first thing I looked for was test and logging examples.

[...]

PR #3264, opened 2025-03-12
[test] the hello-world example

szabgab:
See #3263

### DIFF
--- a/examples/hello-world/src/main.rs
+++ b/examples/hello-world/src/main.rs
@@ -22,3 +22,16 @@ async fn main() {
 async fn handler() -> Html<&'static str> {
     Html("<h1>Hello, World!</h1>")
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use axum::response::Html;
+
+    #[tokio::test]
+    async fn test_handler() {
+        let response = handler().await;
+        let Html(content) = response;
+        assert_eq!(content, "<h1>Hello, World!</h1>");
+    }
+}


yanns:
Does this test really provides value?
In my opinion, a test sending a HTTP request to the `app` and checking the http response would provide more value.
You can use https://github.com/tokio-rs/axum/blob/da3539cb0e5eed381361b2e688a776da77c52cd6/examples/compression/src/tests.rs#L15-L34 as an example of such test.

szabgab:
If I am not mistaken for that I'll have to create the `app` function and move the Router creation there. I think that would be the better way to do this but I was not sure if that is desirable in this most basic "Hello World" example.

yanns:
Yes you're right.
As the name implies, the goal of this example is to be as minimal as possible. So we should avoid adding structure just for testing.
I keep thinking that the new unit test does not provide very much value. And if we want to keep this file as minimal as possible, we should consider not having this test at all.
Maybe you can check if other examples if you can find more valuable testing to add?

szabgab:
I am a newbie to axum, but I have used and taught a number of other frameworks in other languages.
If I understand correctly the idiomatic way is to write axum is to create a function that returns the `app`.

It would not be much more complex to have that function and IMHO it would be better start with an idiomatic example. Assuming it is really the more idiomatic way to write axum code.

szabgab:
OK, I'll close this PR for now. I'll move forward testing the other examples and the book #3268 and then we'll see if you like my suggestions.

PR #3269, opened 2025-03-14
[test] the templates example

szabgab:
* Move the router definition to its own function.
* Add test
* See #3263


yanns:
thank you!

PR #3273, opened 2025-03-17
Implement `From<Bytes>` for `Message`

manifest:
## Motivation

Implementation `From<Bytes>` for `Message` is missing.

The internal representation of the binary message is `Bytes`. It would be convenient to have related `From` implementation, especially for application code that is designed to work with `Bytes` as opposed to `Vec<u8>`.

## Solution

Added an implement `From<Bytes>` for `Message`.


jplatte (approved):
Thanks!

PR #3275, opened 2025-03-17
[test] the versioning example

szabgab:
See #3263

### DIFF
--- a/examples/versioning/src/main.rs
+++ b/examples/versioning/src/main.rs
@@ -7,7 +7,7 @@
 use axum::{
     extract::{FromRequestParts, Path},
     http::{request::Parts, StatusCode},
-    response::{IntoResponse, Response},
+    response::{Html, IntoResponse, Response},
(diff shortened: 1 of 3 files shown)


examples/versioning/src/main.rs
```diff
@@ -68,3 +72,51 @@ where
...
+            .oneshot(
+                Request::builder()
+                    .uri("/v4/foo")
+                    .body(Body::empty())
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(response.status(), StatusCode::NOT_FOUND);
```
yanns, line 115: why is it not found?

szabgab: Because the enum has v1,v2, v3
I could use `"/blabla/foo"` in the example if you think that makes it clearer.

yanns (approved):
thanks for the explanation

PR #3276, opened 2025-03-17
[test] the validator example

szabgab:
See #3263

### DIFF
--- a/examples/validator/src/main.rs
+++ b/examples/validator/src/main.rs
@@ -99,9 +99,18 @@ impl IntoResponse for ServerError {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use axum::{body::Body, http::Request, http::StatusCode};
+    use axum::{
+        body::Body,
+        http::{Request, StatusCode},
+    };
     use http_body_util::BodyExt;
     use tower::ServiceExt;
+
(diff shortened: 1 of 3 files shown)


examples/validator/src/main.rs
```diff
@@ -91,3 +95,86 @@ impl IntoResponse for ServerError {
...
+    async fn test_no_param() {
+        let response = app()
+            .oneshot(Request::builder().uri("/").body(Body::empty()).unwrap())
+            .await
+            .unwrap();
+
+        assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+        let body = response.into_body();
+        let bytes = body.collect().await.unwrap().to_bytes();
+        let html = String::from_utf8(bytes.to_vec()).unwrap();
```
yanns, line 116: Nitpick: you could extract this into a local function

szabgab: Do you mean like this?

```rust
    async fn get_html(response: Response<Body>) -> String {
        let body = response.into_body();
        let bytes = body.collect().await.unwrap().to_bytes();
        String::from_utf8(bytes.to_vec()).unwrap()
    }
```

yanns (approved):

PR #3278, opened 2025-03-19
Add test to chat example

kyle-nweeia:
Refs #3263.

### DIFF
--- a/examples/chat/src/main.rs
+++ b/examples/chat/src/main.rs
@@ -159,3 +159,27 @@ fn check_username(state: &AppState, string: &mut String, name: &str) {
 async fn index() -> Html<&'static str> {
     Html(std::include_str!("../chat.html"))
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_check_username() {
+        let user_set = Mutex::new(HashSet::new());
+        let (tx, _rx) = broadcast::channel(100);
+        let app_state = Arc::new(AppState { user_set, tx });
+        let mut username = String::new();
+        let name = "Ferris";
+
+        check_username(&app_state, &mut username, name);
+
+        assert!(app_state.user_set.lock().unwrap().contains(name));
+        assert_eq!(username, name);
+
+        username.clear();
+        check_username(&app_state, &mut username, name);
+
+        assert_eq!(username, "");
+    }
+}


examples/chat/src/main.rs
```diff
@@ -159,3 +159,27 @@ fn check_username(state: &AppState, string: &mut String, name: &str) {
...
+
+        check_username(&app_state, &mut username, name);
+
+        assert!(app_state.user_set.lock().unwrap().contains(name));
+        assert_eq!(username, name);
+
+        username.clear();
+        check_username(&app_state, &mut username, name);
+
+        assert_eq!(username, "");
```
darth-raijin, line 183: The assertion could be simplified by doing the following
```suggestion
        assert!(username.is_empty());
```

examples/chat/src/main.rs
```diff
@@ -159,3 +159,27 @@ fn check_username(state: &AppState, string: &mut String, name: &str) {
 async fn index() -> Html<&'static str> {
     Html(std::include_str!("../chat.html"))
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_check_username() {
```
darth-raijin, line 168: I think it would be cool if the test reflected the behavior you're asserting in the test, maybe something like:
`check_username_registers_unique_usernames`

examples/chat/src/main.rs
```diff
@@ -159,3 +159,27 @@ fn check_username(state: &AppState, string: &mut String, name: &str) {
...
+        let mut username = String::new();
+        let name = "Ferris";
+
+        check_username(&app_state, &mut username, name);
+
+        assert!(app_state.user_set.lock().unwrap().contains(name));
+        assert_eq!(username, name);
+
+        username.clear();
+        check_username(&app_state, &mut username, name);
```
darth-raijin, line 181: You could also assert on `app_state.user_set` again, to ensure the `HashSet` was not changed

darth-raijin (commented):
Hey! Minor ideas for lifting the quality of the test

davidpdrsn:
I don't think this example benefits much from having tests. Like I wrote [here](https://github.com/tokio-rs/axum/issues/3263#issuecomment-4458454377) I don't think our general goal should be to have tests for all examples as that clutters them up.

So I'll close this for now.

Issue #3279, opened 2025-03-20
Allow calling into a Router with non-() state


goffrie:
- [x] I have looked for existing issues (including closed) about this

### Motivation
I want to be able to dynamically select the state for a router before calling into it.
I know the documentation suggests using `Extension` for this, but I don't want to lose type safety for state.

Right now it is possible to use something like `router.clone().with_state(my_state).oneshot(...)`, but this ends up cloning the entire interior of the `Router` including e.g. the route table (because `with_state` uses `into_inner`), which is pretty expensive.

### Proposal
Expose `Router::call_with_state`. Right now this method exists generically but it's never called except with `()` as the state. This is explicit and it can be documented that typical use cases shouldn't require it.

### Alternatives
This could also be exposed as a `tower::Service` implementation with a different input type (e.g. `(Request, S)`).

mladedav:
What do you mean by selecting the state dynamically? Can you provide an example of what you're trying to achieve?

What parameters determine how the state is constructed and how do you plan to call the method?

goffrie:
What I'm trying to do is efficiently serve multiple instances of the same Axum app. The state is determined by HTTP headers.

Currently there are two ways to achieve this:
- Enumerate all of the instances in advance and bind their state separately. This is memory-inefficient because it creates a copy of the PathRouter for each instance.
- Bind state at HTTP request handling time (basically `router.clone().with_state(instance_state).oneshot(http_request)`). This is CPU-inefficient because it creates a copy of PathRouter during the `with_state` call.

`call_with_state` solves this problem by never cloning the PathRouter.

jplatte:
Sounds like your "state" should be an extension instead, tbh.

goffrie:
Extensions could be used to solve my problem, but I find them unsatisfying because 1. they lose type safety (failing at runtime if the extension isn't provided), 2. they'd require changing the entire app and all its route handlers, even when built in a single-instance (static state) configuration.

mladedav:
What you're describing sounds to me more like request processing than application state and I think you should actually create an extractor that looks at the headers (or anything else you need) and then use that in your handlers.

If the caveat is that based on the headers you need something with interior mutability to affect future requests, I'd say you should still put all of this state inside a single global application state and then in the extractor select the right one.

If your requirement is not having to make code changes to the routes (i.e. not adding a custom extractor) then I'd say having multiple copies of the router doesn't sound that bad to me if you don't have thousands of different states.

[...]

PR #3280, opened 2025-03-20
Expose Router::call_with_state

goffrie:
## Motivation
See #3279.

## Solution
Rename the existing non-generic `call_with_state` function and add a publicly exposed, generic wrapper around it.


jplatte:
Closing as per discussion on the issue.

PR #3287, opened 2025-03-25
Revert "Make `status` a const function in rejection handling"

jplatte:
This reverts commit e09cc593655de82d01971b55130a83842ac46684 / PR #3168.

Also updates changelogs to note the crate yanks.

## Motivation

`main` has been in an un-releasable state for a while due to the accidental breakage from the linked PR.

## Solution

Revert that change, as it's not clear how we would best make it non-breaking, and nobody has stepped up to fix the issue in another way.


axum-extra/CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog],
 and this project adheres to [Semantic Versioning].
 
-# 0.11.0
+# Unreleased
```
mladedav, line 8: There's still a line about making `status` into `const`.

mladedav (changes requested):

mladedav (changes requested):
Sorry I didn't notice those during the first pass.

mladedav (approved):

PR #3292, opened 2025-03-27
Release axum 0.8.3 and related crates

jplatte:
Release log (to be done post merge):

- [ ] Run cargo publish in the right order
- [ ] Create git tags
- [ ] Create GitHub releases

### DIFF
--- a/axum-core/CHANGELOG.md
+++ b/axum-core/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+# 0.5.2
+
+- **added:** Implement `Stream::size_hint` for `BodyDataStream` ([#3195])
+
+[#3195]: https://github.com/tokio-rs/axum/pull/3195
+
 # 0.5.1
 
 Yanked from crates.io due to unforeseen breaking change, see [#3190] for details.
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -9,7 +9,7 @@ license = "MIT"
 name = "axum-extra"
 readme = "README.md"
 repository = "https://github.com/tokio-rs/axum"
-version = "0.11.0"
+version = "0.10.1"
 
 [features]
 default = ["tracing"]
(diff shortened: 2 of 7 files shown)


jplatte:
Switching to draft because I want https://github.com/tokio-rs/axum/pull/3294 to land first, and I want to include axum-extra as well now that it's no longer going to be a breaking change.

axum/CHANGELOG.md
```diff
@@ -10,11 +10,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **added:** Implement `From<Bytes>` for `Message` ([#3273])
```
mladedav, line 10: Should this PR already change the "Unreleased" header to "0.8.3"?

`axum-core` already has version in this PR, `axum` and `axum/extra` have "Unreleased".

axum-core/CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+# 0.5.2
+
+- **added:** Implement `Stream::size_hint` for `BodyDataStream` ([#3195])
+- **change:** Some documentation improvements
```
mladedav, line 11: I don't see the benefit of mentioning this so I wouldn't include it. Feel free to ignore this though, it doesn't really harm anyone either.

axum-extra/Cargo.toml
```diff
@@ -48,8 +48,8 @@ __private_docs = [
 ]
 
 [dependencies]
-axum = { path = "../axum", version = "0.8.2", default-features = false, features = ["original-uri"] }
-axum-core = { path = "../axum-core", version = "0.5.1" }
+axum = { path = "../axum", version = "0.8.3", default-features = false, features = ["original-uri"] }
```
mladedav, line 51: `axum-extra` itself is missing a version bump.

jplatte: Yeah, this was intentional as I made this PR before https://github.com/tokio-rs/axum/pull/3293. Now that that's merged, there's no more reason to not release axum-extra.

mladedav (approved):
Should we also release `axum-extra` as `0.11.1`? If someone already upgraded (and wasn't hit by the breaking change) they'll be on the yanked version until we release another version which I assume we don't plan to do ever and with the next breaking version we'll jump directly to 0.12. Re-adding the feature is not a breaking change so that should not be a problem.

Otherwise, we can just ignore that and potentially point people to using 0.10 instead of 0.11 again.

jplatte:
I don't see much of a point. If it's working, no need to do anything about it.

PR #3293, opened 2025-03-27
Bring back no-op async-stream feature for axum-extra

jplatte:
The next release does not have to be a breaking-change release.


mladedav (approved):
Should we create a new PR for that and mark it as breaking so we don't forget?

jplatte:
Sounds like a good idea, want to take care of that?

PR #3295, opened 2025-03-28
Improve vpath macro validation logic

willianricardo:
## Motivation

The current `vpath!` macro only performs basic validation on static paths. It does not catch several edge cases like:
- paths ending with a slash (`"/api/"`),
- consecutive slashes (`"/api//users"`),
- relative segments like `"."` or `".."`.

These invalid paths can cause routing issues and unexpected behavior at runtime. Catching them earlier improves developer experience and avoids subtle bugs.

## Solution

Extended the validation logic inside `validate_static_path()` to cover:
- Non-root paths must not end with a slash
- No empty path segments (e.g. consecutive slashes)
- No relative segments like `"."` or `".."`

Added new tests to assert each validation condition and updated the `vpath!` macro documentation with examples using `compile_fail`.

### DIFF
--- a/axum-extra/src/routing/mod.rs
+++ b/axum-extra/src/routing/mod.rs
@@ -30,25 +30,126 @@ pub use self::typed::{SecondElementIs, TypedPath};
 #[doc(hidden)]
 pub const fn __private_validate_static_path(path: &'static str) -> &'static str {
     if path.is_empty() {
-        panic!("Paths must start with a `/`. Use \"/\" for root routes")
+        panic!("Paths cannot be empty. Use \"/\" for root routes");
     }
     if path.as_bytes()[0] != b'/' {
-        panic!("Paths must start with /");
+        panic!("Paths must start with \"/\"");
+    }
+    if path.len() != 1 {
+        if path.as_bytes()[path.len() - 1] == b'/' {
+            panic!("Non-root paths must not end with a \"/\"");
+        }
+        let bytes = path.as_bytes();
+        let mut start = 1;
+        let mut i = 1;
+        while i <= bytes.len() {
+            if i == bytes.len() || bytes[i] == b'/' {
+                let seg_len = i - start;
+                if seg_len == 0 {
+                    panic!("Paths must not contain empty segments (i.e. no consecutive \"/\")");
+                }
+                if seg_len == 1 && bytes[start] == b'.' {
+                    panic!("Paths must not contain relative segments like \".\" or \"..\"");
+                }
+                if seg_len == 2 && bytes[start] == b'.' && bytes[start + 1] == b'.' {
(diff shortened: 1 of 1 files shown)


axum-extra/src/routing/mod.rs
```diff
@@ -30,25 +30,126 @@ pub use self::typed::{SecondElementIs, TypedPath};
...
 pub const fn __private_validate_static_path(path: &'static str) -> &'static str {
     if path.is_empty() {
-        panic!("Paths must start with a `/`. Use \"/\" for root routes")
+        panic!("Paths cannot be empty. Use \"/\" for root routes");
     }
     if path.as_bytes()[0] != b'/' {
-        panic!("Paths must start with /");
+        panic!("Paths must start with \"/\"");
+    }
+    if path.len() != 1 {
```
mladedav, line 38: You can replace this whole logic with
```rust
            path[1..].split('/').for_each(|segment| {
                assert!(!segment.is_empty(), "error msg");
                assert_ne!(segment, ".", "error msg");
                assert_ne!(segment, "..", "error msg");
            });
```

It covers even the trailing slash though I'm not sure we want to reject that but that's separate discussion.

mladedav (commented):
I'm not entirely convinced we want to reject all these.

What's rejected now are things that would cause a runtime panic during setup while the additions here are more of a best practice but they're all valid URL paths.

I'd be inclined to agree that having `/./` or `/../` inside statically set routes is improbable, but trailing slashes are not that rare for various reasons.

So I guess the question here is what's the purpose of the macro, if it's just to change runtime panics into compilation errors (which was my impression) or if it should be opinionated about the routes and if so, how far it should or should not go.

mladedav:
By the way, you're using different email in your commit than what you have linked to your GitHub account. If you're fine with that that's ok, just so you know that you might have your git settings misconfigured.

axum-extra/src/routing/mod.rs
willianricardo: Firstly, thank you for reviewing my code. This is my first contribution to an open-source project. I attempted to compile using the suggested code, but it doesn’t work within a const function. Considering the points raised, should I apply the changes involving “.” or “..”?

willianricardo: I mistakenly committed using the wrong account, and I only realized it after pushing the branch.

mladedav:
I think we should not include this with the reasons stated before. I can see this being part of another crate that is more opinionated that would provide more best practices to `axum`'s users, but I don't think we should reject more than we have to in a macro in `axum-extra`.

As to where else to put, this, I really don't know. Maybe it could match any project in `ECOSYSTEM.md` or if you decide to release your own crate, feel free to add the project there.

If anyone has other opinions on this, feel free to share them.

In any case, thanks for trying to improve the project.

PR #3299, opened 2025-03-31
Better sessions in oauth example

JonathanWoollett-Light:
Updates the oauth example to use a better approach to session management.

## Motivation

The current approach to sessions within the oauth example is not fit for any circumstance, and if anyone where to take it at face value and use `async_sessions` their application would encounter severe performance issues. The current approach not only fails to lead users down the right path but actively leads them down the wrong path.

## Solution

By updating the example to use a viable approach it helps people understand the right way to solve the problem, rather than potentially misleading them.


mladedav:
Would you mind expanding a bit on the motivation? Is the only issue performance?

JonathanWoollett-Light:
> Would you mind expanding a bit on the motivation? Is the only issue performance?

The main motivation is that the current example while simpler doesn't really give an example of what someone needs to know.

Personally I used this example in testing, but recently discovered that this is not usable in production and thus needed to figure out what was usable. It would've saved me a lot of time if the example was usable to begin with.

I think we can save a lot of people time by simply making the example include what they will need rather than forcing them to look at this example, then realise it can't be used, then find a different example that can be used.

Cargo.lock
```diff
@@ -54,12 +54,12 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.11"
+version = "0.8.12"
```
mladedav: Remove all the unrelated updates from the PR please.

examples/oauth/src/main.rs
```diff
@@ -40,12 +58,23 @@ async fn main() {
...
+    // Create session in-memory cache.
+    let moka_store = MokaStore::new(Some(100));
+    let store = CachingSessionStore::new(moka_store, sql_store);
+
+    // Create app state.
     let oauth_client = oauth_client().unwrap();
     let app_state = AppState {
-        store,
         oauth_client,
+        store,
```
mladedav, line 77: Try not making unrelated changes like this for no apparent reason. It makes the whole process much harder because you want to improve the example by adding some stuff (like an actual SQL backing store) but then there is too much noise we have to sift through and there's no easy way to tell what's important and what just happened.

examples/oauth/src/main.rs
```diff
@@ -132,40 +149,59 @@ struct User {
 }
 
 // Session is optional
-async fn index(user: Option<User>) -> impl IntoResponse {
-    match user {
+async fn index(
+    State(AppState { store, .. }): State<AppState>,
+    TypedHeader(cookies): TypedHeader<headers::Cookie>,
```
mladedav, line 154: With this change the cookie is not optional anymore.

Please leave the `User` extractor as is unless you have a reason to delete it.

JonathanWoollett-Light: I couldn't find a way to make it work without changing it.

Cargo.lock
JonathanWoollett-Light: Running `cargo test --no-run` changes the lock file. I didn't change it manually. Cargo thinks these changes are related.

examples/oauth/src/main.rs
JonathanWoollett-Light: Could you be more specific to the change you disagree with here?

mladedav: Here specifically it's moving `store` by one line. I wrote other larger examples elsewhere.

PR #3300, opened 2025-04-01
Further downstream code size improvements

svix-jplatte:
## Motivation

Help code size and compile times of downstream crates.

## Solution

Extract some non-generic functionality out of generic `IntoResponse` and `FromRequest` impls (and related fns).


yanns (approved):
I did not know this was working with inner functions. I learned something new!

svix-jplatte:
Yeah I wasn't _entirely_ certain either but I knew that inner functions can not reference generic parameters of the outer scope so it made sense when I saw no difference at all between having the helpers inside or outside the impl.

PR #3308, opened 2025-04-07
add missing track_caller to PathRouter method

Laifsyn:
## Motivation
Make internal panics to return the actual offending line instead of redirecting to somewhere inside the library.

```rs
pub fn product() -> axum::Router<()> {
    use axum::routing::post;
    axum::Router::new()
        .route("/", post(|| async {}))
        .route("/", post(|| async {}))
         // The resulting message for panic location will display somewhere 
         // internal in the library instead of the user's offendin line.
}
```
## Solution

Add the missing `#[track_caller]` attribute in the call stack (?)

### DIFF
--- a/axum/src/routing/path_router.rs
+++ b/axum/src/routing/path_router.rs
@@ -80,6 +80,7 @@ where
         self.v7_checks = false;
     }
 
+    #[track_caller]
     pub(super) fn route(
         &mut self,
         path: &str,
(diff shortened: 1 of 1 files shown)


Laifsyn:
I embedded a panic in `merge_for_path` to find what other functions called it yet didn't return the actual callsite at the tests.
So, here's another place that adding the track_caller would allow it to report the callsite of the "offending" test
```shell
thread 'routing::tests::merging_routers_with_same_paths_but_different_methods' panicked at axum\src\routing\path_router.rs:198:67: # Inside a protected Method (protected as in `pub(super)`)
```

jplatte:
Thanks for bringing this up! I took a look, and I think the better fix is to avoid panicking in internal functions. I implemented that and added a test in #3319.

PR #3311, opened 2025-04-13
Update `route.md` to avoid the confusion of applying a `method_router` for a `path`

DeepSpaceTravel:
# Problem
Confusing a "period"`.` for an end of line to a continuation of a function. (e.g. path.someOtherFunc)

## Motivation

As I was trying to understand the syntax of how to apply RESTful actions to a route, I couldn't grasp this part for 5 minutes. Then I realized the dot is a period instead of the continuation of a function.

## Solution

~~Added line break for `path`. `method_router`~~
From the examination of building the doc locally, a line break that uses two spaces at the end, along with a sigle line break works. However, as suggested by @mladedav, Rephrasing the setence is much easier to understand.

Therefore, changing the setence to "Usually, `method_router` will be..."

to avoid confusing it as 

`path.method_router`

### DIFF
--- a/axum/src/docs/routing/route.md
+++ b/axum/src/docs/routing/route.md
@@ -4,7 +4,8 @@ Add another route to the router.
 can be either static, a capture, or a wildcard.
 
 `method_router` is the [`MethodRouter`] that should receive the request if the
-path matches `path`. `method_router` will commonly be a handler wrapped in a method
+path matches `path`.  
+`method_router` will commonly be a handler wrapped in a method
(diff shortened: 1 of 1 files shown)


mladedav:
I can't check right now, but I think a single line break is ignored in markdown so the change will not take effect. You can try it by building the docs and looking at the rendered page.

You can also just reword the second sentence like "Usually, `method_router` will be..." to break it up.

mladedav (approved):

PR #3312, opened 2025-04-18
Fix recursive call leading to stack overflow when tracing level filter is `TRACE`

myFavShrimp:
## Motivation

My server application crashed, when I set a low `DefaultBodyLimit` and tried to turn the `MultipartError` into a response. After debugging I found the recursive call of `body_text` in the tracing event capture causes a stack overflow when a tracing subscriber with level filter `TRACE` is set, which lead to the crash. 

## Solution

Create the response body before capturing the tracing event.

### DIFF
--- a/axum-extra/src/extract/multipart.rs
+++ b/axum-extra/src/extract/multipart.rs
@@ -401,6 +401,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[cfg(feature = "tracing")]
     async fn body_too_large_with_tracing() {
         tracing::subscriber::set_global_default(
             tracing_subscriber::FmtSubscriber::builder()
--- a/axum-extra/src/extract/multipart.rs
+++ b/axum-extra/src/extract/multipart.rs
@@ -403,21 +403,27 @@ mod tests {
     #[tokio::test]
     #[cfg(feature = "tracing")]
     async fn body_too_large_with_tracing() {
-        tracing::subscriber::set_global_default(
-            tracing_subscriber::FmtSubscriber::builder()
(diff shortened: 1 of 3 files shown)


axum-extra/src/extract/multipart.rs
```diff
@@ -398,4 +399,36 @@ mod tests {
         let res = client.post("/").multipart(form).await;
         assert_eq!(res.status(), StatusCode::PAYLOAD_TOO_LARGE);
     }
+
+    #[tokio::test]
+    async fn body_too_large_with_tracing() {
+        tracing::subscriber::set_global_default(
+            tracing_subscriber::FmtSubscriber::builder()
```
mladedav, line 406: You might want to use something like [this](https://docs.rs/tracing/latest/tracing/subscriber/fn.with_default.html) instead. If someone added a second test like this, the initialization would panic.

Also, I don't think we need `tracing` dev dependency just `tracing-subscriber`.

mladedav (changes requested):

axum-extra/src/extract/multipart.rs
jplatte: If the `tracing` dev-dep is removed, this test needs to have `#[cfg(feature = "tracing")]`. But that probably makes sense anyways.

myFavShrimp:
Very sensible changes. Thank you for the feedback! Let me know if there is anything else that can be improved.

jplatte (approved):
I'll wait to hear whether @mladedav has any extra feedback, but looks good to me.

jplatte:
@mladedav are you good with the changes now?

mladedav (approved):
Yep, it's good. Sorry I missed the last message.

PR #3314, opened 2025-04-19
Generalize `Connected` implementation to all `Listener`s

Voultapher:
The previous implementation implemented the trait `Connected` for `TcpListener` and `TapIo`. This was both needlessly specific *and* blocked other implementers of the `Listener` trait from themselves implementing `Connected`.

For example when using axum with
`into_make_service_with_connect_info::<SocketAddr>()` a `Listener` implementation that isn't `TcpListener` will yield a compiler error that `Connected` is not implemented for this other `Listener`. Implementing it for `SocketAddr` directly isn't possible because of the orphan rule. Even with a wrapper type, a library abstraction can't implement it generically for all `Listener`s because it would run into conflicting trait implementations.

### DIFF
--- a/axum/src/extract/connect_info.rs
+++ b/axum/src/extract/connect_info.rs
@@ -85,21 +85,13 @@ pub trait Connected<T>: Clone + Send + Sync + 'static {
 #[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
 const _: () = {
     use crate::serve;
-    use tokio::net::TcpListener;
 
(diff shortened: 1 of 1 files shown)


jplatte:
Well, as you can see from the CI failures, with the change you suggested (I think I tried exactly the same thing previously), _all_ user-supplied `Connected` implementations become ambiguous.

I think wrapper types might be the solution for what you were trying to do originally, but I don't know for sure. Could you please open a [Discussion](https://github.com/tokio-rs/axum/discussions/new?category=q-a) about that, if you need help?

PR #3316, opened 2025-04-22
Refactor signal transmission by removing unnecessary Arc wrapping in …

harry0349:
…`mod.rs`

## Motivation

Remove the unused wrapper Arc.

### DIFF
--- a/axum/src/serve/mod.rs
+++ b/axum/src/serve/mod.rs
@@ -6,7 +6,6 @@ use std::{
     future::{poll_fn, Future, IntoFuture},
     io,
     marker::PhantomData,
-    sync::Arc,
 };
 
 use axum_core::{body::Body, extract::Request, response::Response};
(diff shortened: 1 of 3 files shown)


mladedav:
As you can see in the CI, the issue is that the `Sender` wasn't always `Clone` so we can't do this unless we bump MSRV.

So I think this PR can wait until we have to bump MSRV for other reasons and then we can get rid of the `Arc`.

jplatte:
Is that really the case? I want to look into it more closely, soon-ish.

jplatte:
I'm pretty sure this is just a case of the MSRV CI job running with minimal dependency versions, and our minimum version of tokio not having the required `Clone` implementation for the `watch` channel sender.

@harry0349 could you bump the tokio version in `axum/Cargo.toml` to `1.44.1` (latest)? That one still has a lower MSRV than axum itself.

axum/src/serve/mod.rs
```diff
@@ -293,7 +292,7 @@ where
         } = self;
 
         let (signal_tx, signal_rx) = watch::channel(());
-        let signal_tx = Arc::new(signal_tx);
+        let signal_tx = signal_tx;
```
jplatte, line 295: If we're no longer wrapping it, this line can be removed entirely.

harry0349: Yes， you are right. I'll bump the tokio version.

jplatte (approved):
Thanks!

mladedav:
Right sorry, that's `tokio`, not `std` :facepalm:

PR #3319, opened 2025-04-24
Fix panic location for overlapping method routes

jplatte:
## Motivation

The panic message currently points into library internals, when it should be pointing at where the user made a mistake.

## Solution

Change one internal function to use `Result` error propagation like the rest, so the actual panicking happens inside a function that's `#[track_caller]`.


mladedav (approved):
The test is really cool.

PR #3323, opened 2025-04-25
axum-extra: implement `FromRequest` for `Either*`

mladedav:
## Motivation

`Either` can currently be used only for `FromRequestParts` extractor. We might want to support extraction of different kinds of bodies. Example of a user wanting something like this is #3315 

## Solution

Have `Either` buffer the whole body into `Bytes` and then try the extractors in turn.


jplatte:
This seems like a little bit of a footgun potentially. Using `Either<Json<T>, Form<T>>` is going to buffer the request body regardless of the content type. I'm also wondering whether / how to support `Either<impl FromRequest, impl FromRequestParts>` or `Either<impl FromRequestParts, impl FromRequest>`. I'm pretty sure those don't work with this PR as-is.

Any comments / thoughts?

mladedav:
Yeah, it buffers the whole requests. Another option is to instead drive all the extractors at the same time. But I'd expect something like `Either3<Json<T>, String, Bytes>` to then actually have three copies of all the data in memory so I'd say one buffered reused copy would be preferable.

We might also consider not using `Either` for this at and make another more descriptive type specifically for this, something like `BufferedEither`/`FromRequestBufferredEither`?

As for supporting `Either<impl FromRequest, impl FromRequestParts>`, I've put together something working, but I originally thought we might not want that. But we can use it.

Then again we probably do want to have a speicifc type for this, because if we implement both `FromRequestParts` and `FromReqeust` for the same type, it has two different `FromRequest` implementations, one with `ViaParts` and the explicit one. So the way it's set up now, this would break people who have handlers where the last argument is `Either*` containing only `impl FromRequestParts` types.

So I guess another type for this would be the way to go? Or just not do this at all and let people clone/try extract requests explicitly?

davidpdrsn:
I think this is pretty neat but I agree that the foot guns are real. I kinda think if users really need they'll likely be better off writing their own custom extractor.

Because of this I'll close this for now.

PR #3333, opened 2025-04-29
Improve documentation about sharing state with handlers

jplatte:
I think task-local variables could have better docs. This is a start. We should minimize the example as well, but I feel like that we'll lose some information now. Maybe we need more complete examples for each of the state-sharing mechanisms, outside of the crate root docs. Anyways, this change should make sense regardless of any bigger doc improvements.

### DIFF
--- a/axum/src/lib.rs
+++ b/axum/src/lib.rs
@@ -148,10 +148,12 @@
 //! pool of database connections or clients to other services may need to
 //! be shared.
 //!
-//! The three most common ways of doing that are:
+//! The four most common ways of doing that are:
+//!
 //! - Using the [`State`] extractor
 //! - Using request extensions
 //! - Using closure captures
+//! - Using task-local variables
 //!
 //! ## Using the [`State`] extractor
 //!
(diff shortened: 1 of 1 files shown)


yanns (approved):

axum/src/lib.rs
```diff
@@ -337,6 +338,12 @@
 //!     .route_layer(middleware::from_fn(auth));
 //! ```
 //!
+//! The main downside to this approach is that it only works when the async executor being used
+//! has the concept of task-local variables. The example above uses
+//! [tokio's `task_local` macro](https://docs.rs/tokio/1/tokio/macro.task_local.html).
+//! smol does not yet offer equivalent functionality at the time of writing (see
+//! [this GitHub issue](https://github.com/smol-rs/async-executor/issues/139)).
```
mladedav, line 345: I'm late, but just to check, axum only supports tokio and can't really run under smol anyway, right? I haven't seen any plans to make it runtime-agnostiv either. Just checking if I'm missing something?

jplatte: No, actually. `axum::serve` uses tokio, but tokio is an optional dependency. There's [`smol-axum`](https://docs.rs/smol-axum/latest/smol_axum/) for an easy way of running an axum service on smol.

PR #3335, opened 2025-04-30
Release axum 0.8.4

jplatte:
Release log (to be done post merge):

- [x] Run cargo publish in the right order
- [x] Create git tags
- [ ] Create GitHub releases

After this release, I'd start to merge breaking changes. We can still make 0.8.x releases then, but from a separate release branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -302,7 +302,7 @@ dependencies = [
 
 [[package]]
 name = "axum"
-version = "0.8.3"
+version = "0.8.4"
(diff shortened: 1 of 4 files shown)


Turbo87 (approved):
looks like the lockfile update triggered a couple of cargo-deny failures 😅

jplatte:
cargo-deny is another good reason to split the friggin examples into their own repo... 😒

I guess for now I'll just revert that commit.

PR #3339, opened 2025-05-02
axum: add router nesting fallback changelog

mladedav:
#3158 was missing a changelog entry so I'm adding it here.

I don't know how much info should I add there though, if this is enough or we want more since we consider it breaking.


mladedav:
Maybe a link to [this comment](https://github.com/tokio-rs/axum/pull/3158#issuecomment-2578712553) would be sufficient since those are the only changes we expect?

yanns (approved):

PR #3340, opened 2025-05-02
Further refactor `json_content_type`

daalfox:
I think this is more understandable. Reject it if you disagree, up to you.


daalfox:
One question, is it fine if I make more of these refactoring PRs if I catch more or is it annoying? :)

yanns (approved):

Issue #3330, opened 2025-04-27
Remove State in favor of task locals?


jplatte:
I would like to drop `State`, probably not in the next release line (0.9) but in the one after it (0.10). Why:

- It makes `FromRequestParts` and `FromRequest` more complicated
  - Users have to put stupid `S: Send + Sync` bounds, often without understanding why
  - In particular, an implementation that does not use any state must be generic over `S`, which means it is only monomorphized when a handler function using it is turned into a boxed service (typically via `routing::{get, post, ...}`), likely in the final binary crate for a service
- People regularily get confused about the meaning of `Router<AppState>` ("requires `AppState` before it can be used")
- Because it doesn't exist at the tower layer, we have to add ugly APIs like `middleware::from_fn_with_state`
- It accounts for a significant part of the implementation complexity, while not being all that necessary (extensions were working okay-ish before)

However, because `Extension` is a pretty sloppy way of sharing state, I don't think it is a good replacement. What __is__ a good replacement in my opinion, is task locals:

```rust
// This particular way of setting task locals is not yet merged:
// https://github.com/tower-rs/tower/pull/821
use tower::task_local::SetTaskLocalLayer;

tokio::task_local! {
    static MY_STATE: MyState;
}

let app = Router::new()
    .route("/", get(|| async {
        let state = MY_STATE.get(); // get a clone of the state
        // TODO: create a response using state
    }))
    .route("/whatever", post(|| async {
        // can also access the state without cloning
        let foo = MY_STATE.with(|state| state.foobar());
        // TODO: create a response using foo
    }))
    // supplying state by cloning, once per request
    .layer(SetTaskLocalLayer::new(&MY_STATE, MyState::default()))
    // can also initialize new state per request
    .layer(axum::middleware::from_fn(move |req: Request, next: Next| {
        async move {
            let state = todo!();
            MY_STATE.scope(state, next.run(req)).await
        }
    }));
```

cc @mladedav @yanns @Turbo87

mladedav:
I personally see the main advantage of `State` in that `axum` type-checks that the state will be present where it is needed. Neither `Extension`s nor task locals can guarantee that I did not forget to set a middleware somewhere and if I did, I'll get runtime panics when a given handler is called.

I would rather keep `State` and try to help the issues in some other way.

jplatte:
Yeah, that is the one advantage of `State` compared to task-locals. Personally I don't think that advantage is bigger than the downsides.

[...]

PR #3343, opened 2025-05-06
Stateless FromRequestParts

mladedav:
## Motivation

This could address part of the issues listed in #3330. Namely, majority of extractors do not need to use state but are still forced to have it in the trait definition, be generic over the state with bounds they might not understand and then the extractor is monomorphized only when it is used.

People can use instead a variant of `FromRequestParts` that will not take `State` into account at all.

This however makes worse the last point in the linked issue, i.e. even higher implementation complexity. I think it should be easier on users since most of them would just move to this new variant, but then again we have a new trait which is not free in terms of education and some users might get confused about which to use when.

## Solution

The same solution is used as for `FromRequest` which either can have an explicit implementation or have an implementation based on `FromRequestParts`. It has a generic parameter (with default value) so that we can have one blanket implementation for things implementing `FromRequestParts` and specific implementation for other types. Then anything using this can be generic over this hidden parameter and rustc will figure out which type to use there.

I've added a new trait similar to `FromRequestParts` but without `State` and changed `FromRequestParts` to also have another generic argument similar to what `FromRequest` has now.

This addition is propagated to a lot of different spaces, e.g. `RequestPartsExt::extract`. This also has the prior art of `RequestExt::extract` which also already has another argument.

As an example, I've taken `Query` and changed its `FromRequestParts` implementation to use the new trait to verify that this can work.

## Next Steps

At this point I'd rather first talk about this, if this is something we believe would make usage of axum easier or if it just complicates things both for us and for users.

If this is something we won't reject at this point, we should then

- [ ] Change all other extractors to use the new trait (unless they use state)
- [ ] Benchmark that compilation times do not suffer from this
- [ ] Clean this up a bit


mladedav:
The idea is to have somthing like
```rust
impl<T> FromStatelessRequestParts for Query<T>
where
    T: DeserializeOwned,
{
    type Rejection = QueryRejection;

    async fn from_request_parts(parts: &mut Parts) -> Result<Self, Self::Rejection> {
        Self::try_from_uri(&parts.uri)
    }
}
```
The usage (well exactly this query impl) can be seen in the last commit, the rest of it are really technical details.

instead of stateful `FromRequestParts` implementation.

Any thoughts @jplatte @yanns @Turbo87 ?

Turbo87:
at first glance I'm not sure it's worth the extra complexity just to avoid the extra `S` generic. I've personally not experienced many issues with the state parameter when implementing these traits.

if we move forward with @jplatte's plan of getting rid of `State` again, this would create another bit of churn for our users too.

mladedav:
> could we reach the same API by using a default type for State?

No, because then the extractor could then only be used with the concrete default type (i.e. handlers that used any state other than `()` would not be able to use these extractors).

yanns:
I understand the value.
I'm worrying that we're adding complexity for us and also for the users who have to choose between `FromRequest`, `FromRequestParts`, `OptionalFromRequest` and `OptionalFromRequestParts`.
I wish there would be a user-friendly way to combine them. But I don't see any. If others are ok for this change, I'll follow.

jplatte:
Yeah, thanks for the API sketch but while this kind of helps the first point I made on that issue, it also increases the API surface and could lead to extra confusion by itself. It could be a bit better than doing nothing, or a bit worse. I'm leaning towards favoring nothing.

mladedav:
It's more complexity without clear advantage, so I'm closing it.

PR #3347, opened 2025-05-09
Add "Run with" comment to example-low-level-openssl

StevenBlack:
If merged, the low-level-openssl example will include the helpful comment, found in most other examples, instructing how to run this example.

## Motivation

Axum noob here 🙂 systematically iterating through examples. I see this particular example does not have the helpful comment in `main.rs` to instruct how to run this example. This PR fixes that.

## Solution

This is a helpful code comment, added at the top of `main.rs`, that mimics the comments found in other examples.

### DIFF
--- a/examples/low-level-openssl/src/main.rs
+++ b/examples/low-level-openssl/src/main.rs
@@ -1,3 +1,9 @@
+//! Run with
+//!
+//! ```not_rust
+//! cargo run -p example-low-level-openssl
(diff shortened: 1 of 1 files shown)


StevenBlack:
cc Romain @erebe — merci!

erebe (approved):

jplatte (approved):
Thanks!

PR #3348, opened 2025-05-11
Added translations links to the README.md

Sanjar-swe:
## Motivation

I have translated the documentation of the Axum repository into several languages, including Russian, Chinese, Japanese, Hindi, and Arabic, to make the project more accessible to a broader audience. This is necessary so that developers speaking these languages can better understand the project and get started faster with using it in their applications.

## Solution

I have added new documentation files in the specified languages and updated the main README.md by adding links to the translated versions of the documentation. These changes help improve the accessibility of the project for the international community and provide a better understanding of how to work with Axum in different language groups.


jplatte:
Hey, thanks for the PR! The CI failure is unrelated, I'll try to fix it later today in a separate PR.

I like the idea of this, but I'm a bit hesitant because I can't validate any of these translations myself. @Turbo87 @mladedav @yanns any thoughts?

mladedav:
I don't think that translating just readme provides much value since all actual documentation and examples are in English but if we had someone on hand who speaks all those languages, then sure. Otherwise simple changes like putting a warning that `main` contains breaking changes become more complicated.

The translated versions also link back to English documents so I wonder if we shouldn't at least prune some sections or add information that we e.g. only accept contributions in English (since contributing is one of the sections in the translated readmes).

@Sanjar-swe just out of curiosity, do you speak all those languages or are the translations machine generated?

Sanjar-swe:
@mladedav  it's GPT generated, it can be added until next serious changes, to main README, to call more auditory.
More maintainers - users - more work power.

jplatte:
Sorry, but I'm definitely not accepting LLM-generated translations.

PR #3362, opened 2025-05-27
Add hot-patching example

pickfire:
## Motivation

Add an example to do hot-patching using dioxus subsecond.

## Solution

A simple hello-world experimental example to get people started, no `#[hot]` macro supported yet.

Still waiting from dioxus maintainer to fix the serde_json conflict.

image

See https://github.com/DioxusLabs/dioxus/issues/4195

It works when running locally, outside of the examples folder.

### DIFF
--- a/examples/hot-patch/Cargo.toml
+++ b/examples/hot-patch/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "hot-patch"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+axum = { path = "../../axum" }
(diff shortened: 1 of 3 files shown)


stefnotch:
Wouldn't updating the lockfile to use the latest `serde_json` also fix the error? 
`cargo update serde_json` should do the trick, I believe.

mladedav:
I don't think we want examples that depend on this much experimental packages (i.e. alpha version and git dependency). But we could just wait till it stabilizes.

But in any case I'm also not entirely convinced that the example belongs here instead of next to the tool. I get that it can be argued that it teaches people how to easier develop axum apps, but it feels like showcasing third-party tooling, not anything axum has to offer.

PR #3370, opened 2025-06-03
Dm/split examples

mladedav:
## Motivation

Partly #2942, partly trying to make the minimum dependencies checks better (which is similar to the first point in that discussion.

Currently, both `cargo deny` and our MSRV checks use Cargo.lock file which has unified features and versions across both the axum crates and all the examples. This can hide some issues as usually when someone adds an example, they might use `cargo add` which will silently update the dependency for the whole repository. Some compilation errors (like `axum` requiring `bytes@1.0` while it uses features from `bytes@1.7`) will then be hidden.

I don't think most users would ever need to use the minimal versions anyway so this is not as severe, but someone might run into compilation errors.

## Solution

Split the examples out of the top-level workspace. I think this improves the situation and if we want to really move the examples to their own repository at any time, this would be the first step anyway.

In the discussion there was a consideration that this makes it harder to test changes locally because CI might error on the examples which will not be built locally with `cargo test`, but due to `default-members` in the workspace this has always been the case (unless you use `--workspace`).

The changes that are here regarding versions:
- `crc32fast` is a transitive dependency which doesn't build with the lowest possible version. This was patched in CI manually, I just moved the patch into `Cargo.toml`.
- `lazy_static` has the same issue but the minimum version seemed to be higher because of examples.
- `serde` - we use features that were added after `1.0.0`. I didn't look for the lowest possible version and just used whatever was the lowest version we used until now (the lowest version including tests). This could be lowered but I don't think there's a point.
- `bytes` - we use `impl From<Bytes> for BytesMut` in `sse.rs` which was added in 1.7.
- `tracing-subscriber` - last release accidentally did not update `tracing-core` to the correct and this optional feature forces the correct versions to be selected.


jplatte:
It used to be this way, we moved it together because local development was just super annoying with two workspaces in one repo. I forgot to update examples all the time.

mladedav:
Then I'll probably just fix the versioning issues here and then drop that commit.

yanns:
> It used to be this way, we moved it together because local development was just super annoying with two workspaces in one repo. I forgot to update examples all the time.

I guess that the CI is screaming then for each PR?

jplatte:
Yes, this would lead to CI failures when everything seemed fine locally.

yanns:
> Yes, this would lead to CI failures when everything seemed fine locally.

AFAIC, I'd be ready to pay this "annoyance price".

jplatte:
Alright, if you both think this is an improvement, I'm happy to try it again. Actually these days rust-analyzer also has the `linkedProjects` config option, so maybe that gets around my previous annoyances.

yanns:
@mladedav do you have time to make this PR merge-able?

mladedav:
I'll try to update this soon. Hopefully sometime this week.

mladedav:
I've rebased this but dropped the dev-dependency version fixes because I don't think they work after all. We can look into that in another PR.

For users this should be identical but we might encounter more failing MSRV checks which try to resolve minimum versions which may have been previously set higher due to versions set in examples.

mladedav:
I had to fix the minimal versions after all to fix the MSRV CI job.

One was already being patched in CI but it feels better to make the `Cargo.toml` use the correct version without manual `cargo update`. I similarly bumped `lazy-static` as a transitive dep dependency. I also had to update `tracing-subscriber` (also just a dev dependency) because there seems to be a bug that it tries to implement a new method on a trait that doesn't exist in the minimal versions.

Then I had to bump the minimal version of `hyper` and `hyper-util` to fix the `test_with_graceful_shutdown_request_header_timeout` test introduced in #3478.

axum-core/Cargo.toml
```diff
@@ -58,5 +58,9 @@ hyper = "1.0.0"
 tokio = { version = "1.25.0", features = ["macros"] }
 tower-http = { version = "0.6.0", features = ["limit"] }
 
+# fix minimum versions of transitive dependencies
+__crc32fast = { package = "crc32fast", version = "1.1.1" }
+__lazy_static = { package = "lazy_static", version = "1.1.0" }
```
jplatte: How about `cargo update -p package --exact version` steps in the CI job instead?

mladedav: I felt this is better since it's easier to run the check manually since you don't need to explicitly `cargo update` the two versions.

What you're proposing was there originally for `crc32fast`, I thought it would be better to not have there two of those and do it this way instead, but I don't mind changing it back and extending it that way.

jplatte: Yeah I'd prefer the CI job solution. I understand the local testing perspective, but this solution has the downside that it's easy to forget to clean these up if they ever stop being transitive dependencies.

mladedav: Ok, I've done it as a step in CI. I had to manually also bump `version_check` which is a build dependency of `lazy_static` because `1.1.0` requires higher than the minimal version based on our `Cargo.toml` and `cargo update` seems to not want to update dependency versions on its own and `--precise` and `--recursive` cannot be combined.

Issue #3365, opened 2025-05-28
Make `Redirect` testable


MidasLamb:
- [x] I have looked for existing issues (including closed) about this

### Motivation

One of the advantages of axum is that the handlers (can) return strongly typed values, that get turned into HTTP later down the line, this means you can write unit tests that don't have to reinterpret the HTTP to see if the inner logic was correct.
However, when writing tests for something that returns a `Redirect`, there was no way of inspecting the `Redirect` once constructed (without relying on `IntoResponse`) 

### Proposal

Rework `Redirect` to not have status code internally, but rather be an enum (or contain an enum) and move the "status code" logic
https://github.com/tokio-rs/axum/blob/7eabf7e645caf4e8e974cbdd886ab99eb2766d1d/axum/src/response/redirect.rs#L41-L43

into the `IntoResponse` implementation

https://github.com/tokio-rs/axum/blob/7eabf7e645caf4e8e974cbdd886ab99eb2766d1d/axum/src/response/redirect.rs#L87-L91

With this, we can add introspection methods easily (`is_to()`, `location()`).

I do think that this approach would also be more in line with the philosophy of axum about how the `IntoResponse` trait is meant to be used (but I might be wrong here) .

I'd be glad to put up a PR for this, but I'd thought a discussion around it first would be fitting.

### Alternatives

Keep the logic as is, but add introspection that "reverses" the initial logic (i.e. `StatusCode::TEMPORARY_REDIRECT` means it was constructed with the `temporary` function.

jplatte:
How about adding just `status_code` and `location` methods?

zeonzip:
I've drafted a PR at [#3377](https://github.com/tokio-rs/axum/pull/3377). We could also consider reworking the Redirect so the URI is stored in the Redirect as a &str, and only makes a header at IntoResponse, or make it a enum as mentioned.

 For now I only added a ```location()``` which returns a parsed ```to_str()``` result of the ```HeaderValue```, and a ```status_code()``` which returns the current status code without rework of the structure. I also added some tests which all run successfully.

zeonzip:
Also wouldn't making Redirect contain a &str as the URI and making it a enum make it so that we can guarantee the Redirect is valid and Result's aren't required in the internal Redirect? Could also .expect on the ```location()```as it's impossible to not be UTF-8, and these changes would be a breaking change, but a big plus to safety too.

zeonzip:
Believe this issue got resolved in #3377, and can be closed now. 👍

PR #3377, opened 2025-06-18
New redirect inspection

zeonzip:
## Motivation

This PR was made in response to the [issue #3365](https://github.com/tokio-rs/axum/issues/3365). 

Highlights that in axum most responses are strongly typed (and therefore inspectable) and is turned into HTTP at the IntoResponse stage.

Redirect does not have any methods to inspect StatusCode or the URI after being constructed, and constructs the http headers when created.

## Solution

This PR adds testability and inspection of ```Redirect```.
It adds ```status_code()```which returns the current status code of the Redirect, and ```location()``` which returns the result of parsing the Http Header to a &str.
It also adds tests using the new inspection methods.

This PR should also be iterated on, but currently doesn't have any breaking changes to the internals of Redirect, but only adds new methods. We could still make these changes (discussed in the issue above):

- Instead of Redirect constructing the HeaderValue when Redirect is constructed, it could be constructed when Redirect is made into a response in ```IntoResponse```.
- Make redirect a enum to ensure that StatusCode isn't invalid.

### DIFF
--- a/axum/src/response/redirect.rs
+++ b/axum/src/response/redirect.rs
@@ -74,7 +74,8 @@ impl Redirect {
 
     /// Returns the parsed location header as a &str.
     ///
-    /// This function shouldn't ever return a Err() since Redirect takes a &str, and will be valid UTF-8.
+    /// This function shouldn't ever return a Err() since Redirect takes a &str in it's constructors,
+    /// and will be valid UTF-8.
     pub fn location(&self) -> Result<&str, http::header::ToStrError> {
         self.location.to_str()
     }
--- a/axum/src/response/redirect.rs
+++ b/axum/src/response/redirect.rs
@@ -72,9 +72,9 @@ impl Redirect {
         self.status_code
     }
 
-    /// Attempts to parse the location header to a &str, and returns the result.
+    /// Returns the parsed location header as a &str.
     ///
-    /// This function shouldn't ever return a Err() since Redirect takes a &str.
+    /// This function shouldn't ever return a Err() since Redirect takes a &str, and will be valid UTF-8.
     pub fn location(&self) -> Result<&str, http::header::ToStrError> {
         self.location.to_str()
     }
--- a/axum/src/response/redirect.rs
+++ b/axum/src/response/redirect.rs
@@ -72,9 +72,9 @@ impl Redirect {
         self.status_code
     }
 
-    /// Returns the parsed location header as a &str.
+    /// Returns the parsed Location header as a &str.
     ///
-    /// This function shouldn't ever return a Err() since Redirect takes a &str in it's constructors,
+    /// This function shouldn't ever return a Err since Redirect takes a &str in it's constructors,
     /// and will be valid UTF-8.
     pub fn location(&self) -> Result<&str, http::header::ToStrError> {
         self.location.to_str()
--- a/axum/src/response/redirect.rs
+++ b/axum/src/response/redirect.rs
@@ -68,7 +68,7 @@ impl Redirect {
     }
 
     /// Returns the current Redirect StatusCode
-    pub fn status(&self) -> StatusCode {
+    pub fn status_code(&self) -> StatusCode {
         self.status_code
     }
 
--- a/axum/src/response/redirect.rs
+++ b/axum/src/response/redirect.rs
@@ -67,7 +67,7 @@ impl Redirect {
         Self::with_status_code(StatusCode::PERMANENT_REDIRECT, uri)
     }
 
-    /// Returns the current Redirect StatusCode
+    /// Returns the HTTP status code of the Redirect.
     pub fn status_code(&self) -> StatusCode {
         self.status_code
     }
--- a/axum/src/response/redirect.rs
+++ b/axum/src/response/redirect.rs
@@ -67,6 +67,11 @@ impl Redirect {
         Self::with_status_code(StatusCode::PERMANENT_REDIRECT, uri)
     }
 
+    /// Returns the current Redirect StatusCode
+    pub fn status(&self) -> StatusCode {
+        self.status_code
+    }
+
     // This is intentionally not public since other kinds of redirects might not
     // use the `Location` header, namely `304 Not Modified`.
     //
--- a/axum/src/response/redirect.rs
+++ b/axum/src/response/redirect.rs
@@ -72,7 +72,9 @@ impl Redirect {
         self.status_code
     }
 
-    /// Parses the location header to a String.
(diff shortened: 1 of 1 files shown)


zeonzip:
Could also .expect the ```location()``` and return the &str since it's guaranteed to be a valid UTF-8 string. Also a better way to store the URI could be to store the raw &str given, and turn the URI into a header in IntoResponse.

zeonzip:
(suggested in the issue above) We could introduce a RedirectKind which means we don't have to assert the StatusCode, and also turn that into the valid status code when using IntoResponse meaning we can eliminate any code that may in any way panic. And also we are going towards 0.9, and since breaking changes were allowed, it's a better time to do it before 0.9

jplatte:
I think we should just update the location field to be a `String`. I don't see what good the eager conversion to `HeaderValue` (and panic if that fails) does.

zeonzip:
> I think we should just update the location field to be a `String`. I don't see what good the eager conversion to `HeaderValue` (and panic if that fails) does.

Sure, I'll draft it to the branch

zeonzip:
I finished the change changing HttpHeader in Redirect to a String and updated the new public API with it too. Also added another test for Redirect. Also seems like public api checks are failing in the CI, it seems to be happening in other PR's too so I'm gonna take a guess that it's not my fault.

zeonzip:
Also possible is storing the HeaderValue in addition to the String to avoid lazily constructing, It's probably negligble but It's a suggestion.

jplatte:
I don't think the lazy construction is a problem. I think maybe we should make conversion failure an internal server error instead of a panic though. cc @Turbo87 @yanns @mladedav for opinions on that.

W.r.t. CI, I'll try to fix it this evening (CET).

axum/src/response/redirect.rs
```diff
@@ -21,7 +21,7 @@ use http::{header::LOCATION, HeaderValue, StatusCode};
 #[derive(Debug, Clone)]
 pub struct Redirect {
     status_code: StatusCode,
-    location: HeaderValue,
+    location: String,
```
Turbo87, line 24: wouldn't it be better to use `Url` here or something like that? I assume URLs would always be valid header values 🤔

zeonzip: Agree it's a better idea since it can also be validated earlier than until it's used as a response, and a Url could also be used in other contexts, and also means the URL can give a result rather than panic, and then structs like Redirect could take the 100% valid URL instead.

I'm happy to implement that change unless someone else wants to!

(ideally just change to use `http::Uri` or `url::Url`, and temporarily deprecate the older API)

jplatte: I'm not convinced on this, actually. I'm pretty sure the redirect `location` can be relative. I don't think a `url::Url` or `http::Uri` can be.

zeonzip: > I'm not convinced on this, actually. I'm pretty sure the redirect `location` can be relative. I don't think a `url::Url` or `http::Uri` can be.

Ah, makes sense!

Could also make a custom wrapper around &str to perform valid HTTP Header checks before construction, or perform sanitization in Redirect.

jplatte: I'd like to keep things simple for now, that way it can be trivially backported.

Turbo87: > I don't think a `url::Url` or `http::Uri` can be.

FWIW `http::Uri` does seem to support relative paths. see https://docs.rs/http/latest/http/uri/index.html#examples

zeonzip: > > I don't think a `url::Url` or `http::Uri` can be.
> 
> FWIW `http::Uri` does seem to support relative paths. see https://docs.rs/http/latest/http/uri/index.html#examples

Well that's good, but what's the plan as of right now?

jplatte: What I meant by relative is page-relative, not host-relative (i.e. not starting with `/`). I don't think that `http::Uri` supports this - it wouldn't make sense to me.

zeonzip: > What I meant by relative is page-relative, not host-relative (i.e. not starting with `/`). I don't think that `http::Uri` supports this - it wouldn't make sense to me.

yeah, it doesn't, it only supports /page so what's the plan right now if I may ask?

jplatte: Implementation mostly looks good, though the `Panics` doc sections are now incorrect (it's actually going to panic when calling `into_response`).

Since the panic is kind of awkward anyways, I think the best fix is to remove all those `Panics` sections, and change `IntoResponse` to return `(StatusCode::INTERNAL_SERVER_ERROR, error.to_string()).into_response()` if the `HeaderValue` conversion fails.

zeonzip: > Implementation mostly looks good, though the `Panics` doc sections are now incorrect (it's actually going to panic when calling `into_response`).
> 
> Since the panic is kind of awkward anyways, I think the best fix is to remove all those `Panics` sections, and change `IntoResponse` to return `(StatusCode::INTERNAL_SERVER_ERROR, error.to_string()).into_response()` if the `HeaderValue` conversion fails.

Added those changes with a match expression in the `into_response`

axum/src/response/redirect.rs
```diff
@@ -47,26 +43,28 @@ impl Redirect {
...
-    /// # Panics
-    ///
-    /// If `uri` isn't a valid [`HeaderValue`].
-    ///
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308
     pub fn permanent(uri: &str) -> Self {
         Self::with_status_code(StatusCode::PERMANENT_REDIRECT, uri)
     }
 
+    /// Returns the HTTP status code of the Redirect.
```
jplatte:
```suggestion
    /// Returns the HTTP status code of the `Redirect`.
```

axum/src/response/redirect.rs
```diff
@@ -47,26 +43,28 @@ impl Redirect {
...
+    pub fn status_code(&self) -> StatusCode {
+        self.status_code
+    }
+
+    /// Returns the Redirect's URI as a &str.
```
jplatte:
```suggestion
    /// Returns the `Redirect`'s URI.
```

No need to repeat parts of the type signature.

jplatte (commented):
Would you mind re-adding the test including a newline in the location, just updating the assertion to be about the 500 status code?

zeonzip:
> Would you mind re-adding the test including a newline in the location, just updating the assertion to be about the 500 status code?

Added it now!

PR #3394, opened 2025-07-02
Activate more lints

theodorebje:
## Motivation

Draft PR.
Earlier discussion: #3393

The purpose of this PR is to activate and fix more lints for the codebase. This will improve codebase readability and maintainability. The ultimate goal is to have as many lints from the `pedantic` and `nursery` lint groups as possible. For now, only a subset of lints are enabled.

Discussions welcome.

## Solution

New lints:

- [`must_use_candidate`](https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate)
- [`use_self`](https://rust-lang.github.io/rust-clippy/master/index.html#use_self)
- [`redundant_else`](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_else)
- [`unnecessary_semicolon`](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_semicolon)
- [`manual_let_else`](https://rust-lang.github.io/rust-clippy/master/index.html#manual_let_else)
- [`option_if_let_else`](https://rust-lang.github.io/rust-clippy/master/index.html#option_if_let_else)
- [`missing_const_for_fn`](https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn)

I will update this list once I activate more lints.

## Sub-PRs

- [x] #3395
- [x] #3396
- [x] #3399
- [x] #3406
- [ ] #3408

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ verbose_file_reads = "warn"
 must_use_candidate = "warn"
 use_self = "warn"
 redundant_else = "warn"
+unnecessary_semicolon = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ semicolon_if_nothing_returned = "warn"
 manual_assert = "warn"
 ignored_unit_patterns = "warn"
 wildcard_imports = "warn"
+return_self_not_must_use = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ must_use_candidate = "warn"
 use_self = "warn"
 redundant_else = "warn"
 unnecessary_semicolon = "warn"
+manual_let_else = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,7 @@ if_not_else = "warn"
 map_unwrap_or = "warn"
 semicolon_if_nothing_returned = "warn"
 manual_assert = "warn"
+ignored_unit_patterns = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ unnecessary_semicolon = "warn"
 manual_let_else = "warn"
 option_if_let_else = "warn"
 missing_const_for_fn = "warn"
+if_not_else = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,7 @@ use_self = "warn"
 redundant_else = "warn"
 unnecessary_semicolon = "warn"
 manual_let_else = "warn"
+option_if_let_else = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,7 @@ map_unwrap_or = "warn"
 semicolon_if_nothing_returned = "warn"
 manual_assert = "warn"
 ignored_unit_patterns = "warn"
+wildcard_imports = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,7 @@ option_if_let_else = "warn"
 missing_const_for_fn = "warn"
 if_not_else = "warn"
 map_unwrap_or = "warn"
+semicolon_if_nothing_returned = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,6 +52,7 @@ redundant_else = "warn"
 unnecessary_semicolon = "warn"
 manual_let_else = "warn"
 option_if_let_else = "warn"
+missing_const_for_fn = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ missing_const_for_fn = "warn"
 if_not_else = "warn"
 map_unwrap_or = "warn"
 semicolon_if_nothing_returned = "warn"
+manual_assert = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ unused_self = "warn"
 verbose_file_reads = "warn"
 must_use_candidate = "warn"
 use_self = "warn"
+redundant_else = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,7 @@ manual_let_else = "warn"
 option_if_let_else = "warn"
 missing_const_for_fn = "warn"
 if_not_else = "warn"
+map_unwrap_or = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ unnested_or_patterns = "warn"
 unused_self = "warn"
 verbose_file_reads = "warn"
 must_use_candidate = "warn"
+use_self = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,6 +60,7 @@ manual_assert = "warn"
 ignored_unit_patterns = "warn"
 wildcard_imports = "warn"
 return_self_not_must_use = "warn"
+doc_markdown = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ uninlined_format_args = "warn"
 unnested_or_patterns = "warn"
 unused_self = "warn"
 verbose_file_reads = "warn"
+must_use_candidate = "warn"
 
 # configuration for https://github.com/crate-ci/typos
 [workspace.metadata.typos.default.extend-identifiers]
--- a/axum-core/src/body.rs
+++ b/axum-core/src/body.rs
@@ -48,6 +48,7 @@ impl Body {
     }
 
     /// Create an empty body.
+    #[must_use]
     pub fn empty() -> Self {
         Self::new(http_body_util::Empty::new())
     }
@@ -72,6 +73,7 @@ impl Body {
     /// you need a [`Stream`] of all frame types.
     ///
     /// [`http_body_util::BodyStream`]: https://docs.rs/http-body-util/latest/http_body_util/struct.BodyStream.html
+    #[must_use]
     pub fn into_data_stream(self) -> BodyDataStream {
         BodyDataStream { inner: self }
     }
--- a/axum-core/src/body.rs
+++ b/axum-core/src/body.rs
@@ -26,11 +26,10 @@ where
     K: Send + 'static,
 {
     let mut k = Some(k);
-    if let Some(k) = <dyn std::any::Any>::downcast_mut::<Option<T>>(&mut k) {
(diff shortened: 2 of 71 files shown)


theodorebje:
For reference, here are all the applicable `pedantic` and `nursery` lints not included in this PR (yet), sorted by the number of occurrences.
```rust
43 redundant_pub_crate
28 needless_pass_by_value
26 return_self_not_must_use
24 missing_errors_doc
16 similar_names
13 manual_assert
12 missing_panics_doc
11 ignored_unit_patterns
09 redundant_closure_for_method_calls
08 wildcard_imports
08 semicolon_if_nothing_returned
08 default_trait_access
07 doc_markdown
07 doc_link_with_quotes
06 trivially_copy_pass_by_ref
06 map_unwrap_or
05 too_many_lines
05 ref_option
04 used_underscore_binding
04 match_same_arms
03 or_fun_call
03 no_effect_underscore_binding
03 missing_fields_in_debug
03 items_after_statements
03 cast_possible_truncation
02 single_match_else
02 redundant_clone
02 mismatching_type_param_order
02 implicit_clone
02 format_push_string
02 explicit_iter_loop
02 equatable_if_let
01 unused_peekable
01 unused_async
01 needless_pass_by_ref_mut
01 if_not_else
01 future_not_send
01 derive_partial_eq_without_eq
01 debug_assert_with_mut_call
```
_Edited because the previous table was wrong_

theodorebje:
Hi, I'm working on `doc_markdown`. According to Clippy, technical words such as _WebSockets_ should be wrapped in backticks. Do you agree with Clippy? Here's an example line (stolen from [here](https://github.com/tokio-rs/axum/blob/main/axum/src/extract/ws.rs#L135))
```rust
/// `None` if HTTP/2+ WebSockets are used.
```
What parts do you think should be in backticks? `WebSockets`? `HTTP/2+`? Or only `None`? All of them?

mladedav:
Only `None`. Backticks are for code. The lint description also mentions that.

theodorebje:
Updated list of non-activated lints:
```rust
43 redundant_pub_crate
28 needless_pass_by_value
24 missing_errors_doc
16 similar_names
12 missing_panics_doc
09 redundant_closure_for_method_calls
08 default_trait_access
07 doc_markdown
07 doc_link_with_quotes
06 trivially_copy_pass_by_ref
05 too_many_lines
05 ref_option
04 used_underscore_binding
04 match_same_arms
03 no_effect_underscore_binding
03 missing_fields_in_debug
03 items_after_statements
03 cast_possible_truncation
02 single_match_else
02 redundant_clone
02 mismatching_type_param_order
02 implicit_clone
02 format_push_string
02 explicit_iter_loop
02 equatable_if_let
01 unused_peekable
01 unused_async
01 needless_pass_by_ref_mut
01 future_not_send
01 derive_partial_eq_without_eq
01 debug_assert_with_mut_call
```

axum-core/src/error.rs
```diff
@@ -16,6 +16,7 @@ impl Error {
     }
 
     /// Convert an `Error` back into the underlying boxed trait object.
+    #[must_use]
     pub fn into_inner(self) -> BoxError {
```
mladedav, line 20: For errors (and possibly other cases such as builders) we might want to add `#[must_use]` to the type rather than methods.

axum-core/src/extract/option.rs
```diff
@@ -57,7 +57,7 @@ where
 {
     type Rejection = T::Rejection;
 
-    async fn from_request(req: Request, state: &S) -> Result<Option<T>, Self::Rejection> {
+    async fn from_request(req: Request, state: &S) -> Result<Self, Self::Rejection> {
```
mladedav, line 60: Personally, I'm not a fan of this on 3rd party structs like `Option<T>` here or `String` somewhere else. `Self::from_utf8` looks just weird to me.

So I wouldn't add this lint, but I get that this is stylistic, so if other people have different opinions, I don't mind.

axum-core/src/body.rs
```diff
@@ -48,6 +48,7 @@ impl Body {
     }
 
     /// Create an empty body.
+    #[must_use]
     pub fn empty() -> Self {
```
jplatte, line 51: Should `Body` itself be `#[must_use]` maybe?

axum-core/src/body.rs
```diff
@@ -72,6 +73,7 @@ impl Body {
     /// you need a [`Stream`] of all frame types.
     ///
     /// [`http_body_util::BodyStream`]: https://docs.rs/http-body-util/latest/http_body_util/struct.BodyStream.html
+    #[must_use]
     pub fn into_data_stream(self) -> BodyDataStream {
```
jplatte: Same for `BodyDataStream`.

axum-extra/src/extract/cookie/mod.rs
```diff
@@ -118,6 +118,7 @@ impl CookieJar {
     /// run extractors. Normally you should create `CookieJar`s through [`FromRequestParts`].
     ///
     /// [`FromRequestParts`]: axum::extract::FromRequestParts
+    #[must_use]
     pub fn from_headers(headers: &HeaderMap) -> Self {
```
jplatte, line 122: I think `CookieJar` can be `#[must_use]`, then we don't need to attribute on any of its `Self`-returning fns. Same for the other cookie jar types.

axum-extra/src/response/multiple.rs
jplatte, line 1: Same for `MultipartForm`.

axum/src/test_helpers/test_client.rs
```diff
@@ -50,39 +50,45 @@ impl TestClient {
         TestClient { client, addr }
     }
 
+    #[must_use]
     pub fn get(&self, url: &str) -> RequestBuilder {
```
jplatte, line 54: This seems like another instance of the same thing where the type itself should be `#[must_use]`.

Cargo.toml
```diff
@@ -47,6 +47,7 @@ unnested_or_patterns = "warn"
 unused_self = "warn"
 verbose_file_reads = "warn"
 must_use_candidate = "warn"
+use_self = "warn"
```
jplatte, line 50: Nice, I didn't know about this lint!

axum-macros/src/debug_handler.rs
```diff
@@ -536,9 +536,8 @@ fn check_input_order(item_fn: &ItemFn, kind: FunctionKind) -> Option<TokenStream
             return Some(quote_spanned! {*span=>
                 compile_error!(#error);
             });
-        } else {
-            return None;
         }
+        return None;
```
jplatte, line 541: Could also change the whole `if-else` to something like

```rust
return (types_that_consume_the_request[0].0 != number_of_inputs - 1).then(|| {
    // construct the compile_error
});
```

jplatte:
This PR is pretty large. Can you maybe split it up? Doesn't have to be one PR for every lint, but ones that lead to lots of changes like the `must_use` one seem worth splitting.

axum-macros/src/debug_handler.rs
```diff
@@ -52,20 +52,23 @@ pub(crate) fn expand(attr: Attrs, item_fn: ItemFn, kind: FunctionKind) -> TokenS
...
+                        #check_future_send
+                    }
+                },
+                |check_input_order| {
+                    quote! {
+                        #check_input_order
+                        #check_future_send
+                    }
+                },
+            )
```
mladedav, line 71: I'm honestly not a fan of this either. This is nice when it's just one-liners (or at least the default branch is a one-liner), but these bigger blocks are just more indented versions of the original, effectively with else branch first and after that the if branch and with worse readability as to where does it end and that it is an if...else (instead of keyword just a block ends and then another closure is provided).

axum-extra/src/routing/mod.rs
```diff
@@ -28,13 +28,13 @@ pub use self::typed::{SecondElementIs, TypedPath};
...
-        panic!("Paths must start with a `/`. Use \"/\" for root routes")
-    }
-    if path.as_bytes()[0] != b'/' {
-        panic!("Paths must start with /");
-    }
+    assert!(
+        !path.is_empty(),
+        "Paths must start with a `/`. Use \"/\" for root routes"
+    );
+    assert!(path.as_bytes()[0] == b'/', "Paths must start with /");
```
mladedav, line 37: You can use `assert_eq` for this. There is another case like this in this file.

I think there might also be a lint for this.

axum-extra/src/response/file_stream.rs
```diff
@@ -118,20 +118,24 @@ where
...
-    pub fn content_size(mut self, len: u64) -> Self {
+    #[must_use]
+    pub const fn content_size(mut self, len: u64) -> Self {
         self.content_size = Some(len);
         self
     }
 
     /// Return a range response.
     ///
+    /// ```rust
```
mladedav, line 136: I think you need to add `no_run` here so that it's not considered as a doctest.

But regardless, it's not valid rust either, so I'd probably mark it as such (maybe text instead of rust? Not sure).

axum-core/src/macros.rs
```diff
@@ -54,7 +54,7 @@ macro_rules! __define_rejection {
             }
 
             /// Get the status code used for this rejection.
-            pub fn status(&self) -> http::StatusCode {
+            pub const fn status(&self) -> http::StatusCode {
```
mladedav, line 57: I don't think most of these really make sense. I don't think anyone would build a rejection in a const context to then call this.

I did not go through all of this commit though, there might be some useful ones here, but this lint might also be one to split out of this PR.

mladedav (commented):
The stylistic opinions I wrote are just mine and other people may view things differently, but I kind of don't agree with some of the lints. That said if other people prefer e.g. `Self` over typenames in all permissible contexts, I can live with that (as long as clippy reminds me to use that).

In any case, thank you for trying to improve the codebase.

mladedav:
From the list, one thing I think we really want is the `future_not_send` lint. It's just horrible to suddenly find out I can't use something because I'm in a future to be `tokio::spawn`ed so I'd rather prevent that.

theodorebje:
Thanks for the constructive feedback, everyone! I'll look through it all tomorrow.

theodorebje:
> From the list, one thing I think we really want is the `future_not_send` lint.
@mladedav The problematic line for that one seems to be:
```rust
pub async fn from_path(path: impl AsRef<Path>) -> io::Result<FileStream<ReaderStream<File>>> { … }
```
[(here)](https://github.com/tokio-rs/axum/blob/main/axum-extra/src/response/file_stream.rs#L96). `AsRef` doesn't implement `Send`.

I don't have that deep knowledge about the codebase yet, so before I edit that line to fix the lint warning, what do you think would be the best type to take?

axum-core/src/extract/option.rs
theodorebje: @jplatte What do you think about this? Should we only use `Self` for Axum's types?

axum-core/src/error.rs
theodorebje: I'm not sure how to avoid `#[must_use]` for the methods here, since `BoxError` is a type alias and not a "real" type. Adding `#[must_use]` to the `Error` struct doesn't do anything, either. Any ideas?

axum-core/src/extract/option.rs
theodorebje: Moving this discussion to #3396

mladedav:
You can just change it to `path: impl AsRef<Path> + Sync`.

I wouldn't expect anyone to use anything that isn't `Sync` there so I wouldn't expect to break someone with this, but it is a breaking change regardless. The good news is that since people wouldn't use `!Sync` types as parameters here, it's not really a problem the method can theoretically be `!Send`.

So if you have more breaking changes in a separate PR, you can put that there.

theodorebje:
Thanks! Since this PR already includes breaking-ish changes, and I’m trying to group the (big) commits by function, I think I'll include this one-line fix in the main draft PR instead, if that's OK for you?

mladedav:
Sure, feel free to put it here.

Now that I look at it closer, I think it should also include `+ Send` but clippy doesn't seem to mind when that's missing.

axum-macros/src/debug_handler.rs
theodorebje: Wouldn't that cause the `if` statement at the bottom of the function to become dead code? Should I refactor how the construction of the `compile_error` is done?
```rust
if types_that_consume_the_request.len() == 2 { /* compile error */ } else { /* compile error */ }
```
Taken from [here](https://github.com/tokio-rs/axum/pull/3394/files#diff-bc3e16dd51fba2c8ee677d1ab1eef9d2150037f7f309185b1c654f3f1a168d24R544-R574).

theodorebje:
Regarding `redundant_pub_crate`, it seems to be a bunch of false positives by Clippy: rust-lang/rust-clippy#5369.

axum-extra/src/routing/mod.rs
jplatte: I don't think we should, in this case. We don't really want to be printing the first byte of the path in case the assertion fails, we just want the message.

axum-extra/src/response/file_stream.rs
jplatte: What the heck even is this line?

jplatte:
`option_if_let_else`, `map_unwrap_or`: these rewrite code to use `map_or` / `map_or_else`. Unlike David, I don't just dislike these methods when one or both of the arguments is a big closure. I dislike these methods in general. I'd much rather go the opposite direction and disallow both 😄

`missing_const_for_fn`: let's add `const` where people have an actual use case for it, not in every place possible up-front

`ignored_unit_patterns`: just looks weird and I don't think there's a big advantage

theodorebje:
@jplatte Do you also dislike `manual_let_else` (ddf5830) or only the ones you mentioned above?

jplatte:
Only the ones mentioned.

davidpdrsn:
I think this gonna be hard to land given all the merge conflicts. If you're still motivated to work on this I'd rather that you submit smaller more focused PRs.

So I'll close this for now but if there are still high value lints you think we should enable then you're welcome to open new PRs.

PR #3395, opened 2025-07-03
Add #[must_use] to types and methods

theodorebje:
## Motivation

Sometimes, users forget to use their newly created values after calling certain methods. To make Axum more user-friendly, we should help them remember to use these values.

## Solution

The `#[must_use]` attribute does exactly that. It causes a warning if the resulting value is never used.

This PR also adds 3 lints to `Cargo.toml`.

## Blockers

What types should receive the `#[must_use]` attribute? More importantly, which ones should not? I imagine a [GitHub search might help](https://github.com/search?q=%22let%20_%20%3D%22%20language%3Arust%20axum&type=code).


theodorebje:
I wonder if 8a561f3 would make the `return_self_not_must_use` lint more accurate? Maybe we should try and merge that first?

axum-extra/src/extract/cookie/signed.rs
```diff
@@ -170,6 +170,7 @@ impl SignedCookieJar {
     /// run extractors. Normally you should create `SignedCookieJar`s through [`FromRequestParts`].
     ///
     /// [`FromRequestParts`]: axum::extract::FromRequestParts
+    #[must_use]
     pub fn from_headers(headers: &HeaderMap, key: Key) -> Self {
```
jplatte: Here the type can also become `#[must_use]` rather than the methods.

axum-extra/src/response/file_stream.rs
jplatte, line 1: I don't see why one would ever get a `FileStream` as a side effect, so let's also make this type `#[must_use]`.

axum-extra/src/response/multiple.rs
jplatte, line 1: Unrelated to this PR, but I'm pretty sure this module was meant to be named `multipart`, not `multiple` 😄

axum-extra/src/routing/mod.rs
```diff
@@ -84,6 +85,7 @@ pub trait RouterExt<S>: sealed::Sealed {
     ///
     /// See [`TypedPath`] for more details and examples.
     #[cfg(feature = "typed-routing")]
+    #[must_use]
```
jplatte: These are useless. `RouterExt` is only implemented for `Router`, which is itself already `#[must_use]`. Can you add an `#[allow]` to the trait declaration?

axum/src/extract/ws.rs
```diff
@@ -152,6 +152,7 @@ impl<F> std::fmt::Debug for WebSocketUpgrade<F> {
 
 impl<F> WebSocketUpgrade<F> {
     /// Read buffer capacity. The default value is 128KiB
+    #[must_use]
     pub fn read_buffer_size(mut self, size: usize) -> Self {
```
jplatte: Why not make the type `#[must_use]`?

Cargo.toml
```diff
@@ -46,6 +46,8 @@ uninlined_format_args = "warn"
 unnested_or_patterns = "warn"
 unused_self = "warn"
 verbose_file_reads = "warn"
+return_self_not_must_use = "warn"
```
jplatte: Reminder to keep the list sorted :)

axum-extra/src/response/multiple.rs
theodorebje: Haha, noticed that too. I thought it was intended, though 😄

axum-extra/src/routing/mod.rs
theodorebje: I tried to do that, but it seems like `#[must_use]` on traits do nothing.

jplatte: I think you misread my comment :)

theodorebje: Sorry, I am tired. Will do.

axum-extra/src/extract/cookie/mod.rs
```diff
@@ -153,6 +154,7 @@ impl CookieJar {
     ///         .map(|cookie| cookie.value().to_owned());
     /// }
     /// ```
+    #[must_use]
     pub fn get(&self, name: &str) -> Option<&Cookie<'static>> {
```
mladedav, line 158: Isn't `Option` already `must_use`? Did clippy lint this? (This happens Ina few other places too)

axum-extra/src/extract/cookie/mod.rs
```diff
@@ -84,6 +84,7 @@ pub use cookie::Key;
 ///     .route("/me", get(me));
 /// # let app: Router = app;
 /// ```
+#[must_use]
```
mladedav:
```suggestion
#[must_use = "`CookieJar` should be returned as part of a `Response`, otherwise it does nothing."]
```
 This one can use a bit of clarity why it must be used if all `add` and `remove` methods and so on return `Self`.

And the same thing applies to the other jars.

axum-extra/src/extract/cookie/mod.rs
theodorebje: [Option is not `#[must_use]`](https://doc.rust-lang.org/src/core/option.rs.html#582), unfortunately. [The individual methods for Option are](https://doc.rust-lang.org/src/core/option.rs.html#760), though. Result is `#[must_use]`, however.

mladedav: Interesting, I didn't know that, thanks.

jplatte:
I agree. Any reason this is still marked as draft?

theodorebje:
Sorry, forgot to convert it to a PR.

theodorebje:
I'm on GitHub mobile right now, I can't find a way to convert it... Are any of you able to convert it for me? Otherwise I'll convert it in a few hours when I get home.

theodorebje:
Ready to merge now :ok_hand:

PR #3396, opened 2025-07-03
style: use `Self` to avoid unnecessary repetition

theodorebje:
## Motivation

Repeating type names over and over again quickly becomes a problem if the name of the type ever changes, or a refactored is required. It also makes it more difficult to copy code from one method to another.

## Solution

In `impl`s, `Self` can be used instead of the struct/enum being implemented, to reduce unnecessary repetition. This PR uses `Self` as often as possible. This fixes the `use_self` clippy lint.

### DIFF
--- a/axum/src/extract/connect_info.rs
+++ b/axum/src/extract/connect_info.rs
@@ -105,7 +105,7 @@ const _: () = {
     }
 };
 
-impl Connected<SocketAddr> for SocketAddr {
+impl Connected<Self> for SocketAddr {
     fn connect_info(remote_addr: Self) -> Self {
         remote_addr
     }


theodorebje:
Copying over the previous discussion from #3394
>>_Originally posted by mladedav in https://github.com/tokio-rs/axum/pull/3394#discussion_r2180558513_
>>
>>Personally, I'm not a fan of this on 3rd party structs like `Option<T>` here or `String` somewhere else. `Self::from_utf8` looks just weird to me.
>>
>>So I wouldn't add this lint, but I get that this is stylistic, so if other people have different opinions, I don't mind.
>_Originally posted by theodorebje in https://github.com/tokio-rs/axum/pull/3394#discussion_r2182467375_
>
>jplatte What do you think about this? Should we only use `Self` for Axum's types?

jplatte:
Yeah, let's be a bit more conservative and not change any of the places that David had some concerns about. You can still enable the lint, just `#[allow]` it in those specific places (there weren't many, right?).

Cargo.toml
```diff
@@ -46,6 +46,7 @@ uninlined_format_args = "warn"
 unnested_or_patterns = "warn"
 unused_self = "warn"
 verbose_file_reads = "warn"
+use_self = "warn"
```
jplatte: This table is alphabetically sorted ;)

```suggestion
use_self = "warn"
verbose_file_reads = "warn"
```

axum/src/extract/connect_info.rs
```diff
@@ -105,8 +105,8 @@ const _: () = {
     }
 };
 
-impl Connected<SocketAddr> for SocketAddr {
-    fn connect_info(remote_addr: SocketAddr) -> Self {
+impl Connected<Self> for SocketAddr {
+    fn connect_info(remote_addr: Self) -> Self {
```
theodorebje, line 109: @jplatte [`SocketAddr` is a foreign type](https://doc.rust-lang.org/stable/core/net/enum.SocketAddr.html). But I didn't add the `-> Self`. Should I go ahead and replace the return type with `SocketAddr` too? E.g.
```rust
impl Connected<SocketAddr> for SocketAddr {
    fn connect_info(remote_addr: SocketAddr) -> SocketAddr {
```

jplatte: @mladedav wdyt?

mladedav: I'd say uncovered return type `-> Self` is always ok. I'm not entirely sure how I feel about generic types such as `Vec<Self>`, but I think it's pretty standard to use things like `Result<Self, Self::Error>` at least in traits, so I guess anything goes for return types and parameters in traits so doing the same in normal methods might also be fine.

Which is inconsistent with how I felt about `Option<T>`. So if you feel like doing it more consistently and using this more, I'd be ok with that. I think that might even be better in that at least I'll find out if that's actually better or if I really don't like that style.

What I'd shy away from is stuff like `Self::new()`. I think that's unnecessarily less readable similarly to how there's a lint to not use `Default::default()` to instead use the concrete type. I feel like `Self::default()` wouldn't be much better (though objectively it is at least a little better).

jplatte: Huh, that's weird, I like `Self` in expression position a lot too. Maybe you can review the diff as-is and note any places that you're still not a fan of?

axum-extra/src/extract/cookie/mod.rs
```diff
@@ -321,12 +321,14 @@ mod tests {
...
     impl FromRef<AppState> for Key {
+        #[allow(clippy::use_self)]
         fn from_ref(state: &AppState) -> Key {
             state.key.clone()
         }
     }
 
     impl FromRef<AppState> for CustomKey {
+        #[allow(clippy::use_self)]
         fn from_ref(state: &AppState) -> CustomKey {
```
mladedav: These would be fine to have as `Self`

mladedav (approved):
I think everything here is fine, if I didn't miss something, it's really just constructors and enum variants (and a few signatures).

theodorebje:
With all checks passing, this is ready to merge :ok_hand:

mladedav:
@jplatte if you want to change some of the allows to `Self`, we can, otherwise let's merge this.

axum/src/extract/connect_info.rs
```diff
@@ -105,6 +105,7 @@ const _: () = {
     }
 };
 
+#[allow(clippy::use_self)]
 impl Connected<SocketAddr> for SocketAddr {
     fn connect_info(remote_addr: SocketAddr) -> Self {
```
jplatte:
```suggestion
impl Connected<SocketAddr> for SocketAddr {
    fn connect_info(remote_addr: Self) -> Self {
```

axum/src/middleware/map_request.rs
```diff
@@ -380,6 +380,7 @@ where
 }
 
 impl<B> IntoMapRequestResult<B> for Request<B> {
+    #[allow(clippy::use_self)]
     fn into_map_request_result(self) -> Result<Request<B>, Response> {
```
jplatte:
```suggestion
    fn into_map_request_result(self) -> Result<Self, Response> {
```

axum/src/extract/connect_info.rs
```diff
@@ -106,7 +106,7 @@ const _: () = {
 };
 
 impl Connected<SocketAddr> for SocketAddr {
```
jplatte:
```suggestion
impl Connected<Self> for SocketAddr {
```

theodorebje:
Thanks everyone!

PR #3399, opened 2025-07-07
Simplify `if` statements

theodorebje:
## Motivation

There are many different ways to express the same pattern matching and equality checking. This PR standardises those two things using clippy lints, and also makes the code more readable.

## Solution

This PR adds 4 new lints to `Cargo.toml`:

- [x] `manual_let_else`
- [x] `match_same_arms`
- [x] `single_match_else`
- [x] `equatable_if_let`


axum/src/routing/mod.rs
```diff
@@ -187,14 +187,11 @@ where
...
-                panic!(
-                    "Invalid route: `Router::route_service` cannot be used with `Router`s. \
-                     Use `Router::nest` instead"
-                );
-            }
-            Err(service) => service,
+        let Err(service) = try_downcast::<Self, _>(service) else {
+            panic!(
+                "Invalid route: `Router::route_service` cannot be used with `Router`s. \
+                      Use `Router::nest` instead"
```
jplatte:
```suggestion
                Use `Router::nest` instead"
```

jplatte (approved):
Thanks!

Issue #3253, opened 2025-03-07
Supporting two-phase `Listener`s


SabrinaJewson:
- [x] I have looked for existing issues (including closed) about this

### Motivation

Some kinds of listener, like TLS listeners, will want to perform work after the connection has been accepted, but before HTTP requests can actually be served. This logic can be moved into the listener itself via complex multiplexing logic – for example, the `TlsListener` crate uses `FuturesUnordered` to do this – however I think that ideally it would be performed on the same task that serves the connection.

### Proposal

Modify the `Listener` trait as follows:

```rs
pub trait Listener: Send + 'static {
    type Io: AsyncRead + AsyncWrite + Unpin + Send + 'static;
    type Addr: Send;
    fn accept(&mut self) -> impl Future<Output = impl Future<Output = Option<(Self::Io, Self::Addr)>> + Send + 'static> + Send;
    fn local_addr(&self) -> io::Result<Self::Addr>;
}
```

The outer future of `accept` will resolve when a connection has been accepted, while the inner future runs on the spawned task. `Option<(Self::Io, Self::Addr)>` is used to account for errors: if `None` is returned, then the task is aborted.

### Alternatives

- Use a `type Error: Into<Box<dyn Error + Send + Sync>>;` and `Result<(Self::Io, Self::Addr), Self::Error>` instead of `Option`. This might be more convenient for the user, as they would have to handle the errors less themselves.
- Say that users should define their own “lazy TLS stream” type to handle this case. It would be an `enum` over the handshake future and the underlying TLS stream, and its implementation of `AsyncRead`/`AsyncWrite` would first drive the handshake to completion before performing I/O (or report EOF should the handshake fail). I can’t particularly think of practical reasons to avoid this other than “it feels weird”.
- Say that users who want TLS in this way should not use `axum::serve`. That function is, however, a very convenient abstraction, taking care of graceful shutdown and constructing the `server::conn::auto::Builder` with the right configuration.

SabrinaJewson:
To prove the utility of this API, here’s pseudocode for what a TLS listener API might look like:

<details><sumary>TLS listener pseudocode</summary>

```rs
pub trait TlsAcceptor<C> {
    fn begin_handshake(&mut self) -> impl TlsHandshake<C>;
}

pub trait TlsHandshake<C>: Send + 'static {
    type Stream: AsyncRead + AsyncWrite + Unpin + Send + 'static;
[...]

Issue #2494, opened 2024-01-07
Built-in TLS Support 🎯


NiiightmareXD:
- [X] I have looked for existing issues (including closed) about this

## Feature Request
I think it would be better if Axum had built-in TLS support right now we can use [low-level-rustls](https://github.com/tokio-rs/axum/tree/main/examples/low-level-rustls) but it's unnecessarily hard I think it would be better if we had something like in (Actix)[https://actix.rs/docs/http2] which has bind_openssl function and doesn't need any other crates other than OpenSSL and Actix itself.

I can attempt to implement it myself if you believe it would be beneficial.

### Motivation
Pretty much every web server uses TLS.

### Proposal
The implementation should be easy it has been already done in (axum-server)[https://github.com/programatik29/axum-server] but it would be better if we didn't need to add another crate to project just to get TLS.

### Alternatives
using [axum-server](https://github.com/programatik29/axum-server), [low-level-openssl](https://github.com/tokio-rs/axum/tree/main/examples/low-level-openssl), [low-level-rustls](https://github.com/tokio-rs/axum/tree/main/examples/low-level-rustls) and [tls-rustls](https://github.com/tokio-rs/axum/tree/main/examples/tls-rustls).

davidpdrsn:
axum intentionally doesn’t have that built-in and I think I want it to stay that way.

In the future axum::serve will be generic over the listener and IO which will make it easier to plug-in a TLS acceptor. See https://github.com/tokio-rs/axum/pull/2479

NiiightmareXD:
> axum intentionally doesn’t have that built-in and I think I want it to stay that way.
> 
> In the future axum::serve will be generic over the listener and IO which will make it easier to plug-in a TLS acceptor. See [#2479](https://github.com/tokio-rs/axum/pull/2479)

Hey, now that the PR you mentioned has merged, is there any example of how to do it without the `axum-server` crate?

mladedav:
Without having seen the internals of the crate, you could try something like [this](https://github.com/tmccombs/tls-listener/blob/main/examples%2Faxum.rs).

If you're willing to create an example out of it (either by pulling in the crate or by implementing something like that in the example, if it would be easy enough), feel free to send a PR.

NiiightmareXD:
> Without having seen the internals of the crate, you could try something like [this](https://github.com/tmccombs/tls-listener/blob/main/examples%2Faxum.rs).
> 
[...]

PR #3400, opened 2025-07-08
Add `ListenerExt::handshake` for TLS and other handshakes

NiiightmareXD:
## Motivation
This change adds a convenient, built-in mechanism for applying asynchronous handshake protocols (most notably TLS) to the incoming connections handled by axum::serve. It aims to offer a potential solution for issues like #3253 and #2494, which discuss the desire for first-class TLS support directly within Axum's serving infrastructure.

Currently, to use TLS with axum::serve, one must implement a custom Listener or manage the connection accept loop manually. This PR proposes a more ergonomic and composable alternative that integrates smoothly with the existing Listener trait.

## Solution
This PR introduces a new method, handshake, to the ListenerExt trait. It acts as a generic adapter, similar in spirit to the existing tap_io method.

### I chose `ListenerExt` because:
- It allows chaining operations on a listener. A user can create a `TcpListener`, `tap_io` to set `TCP_NODELAY`, and then `handshake` to apply TLS, all before passing it to `axum::serve`.
- The core `axum::serve` function remains completely unchanged.
- The `handshake` method is generic and works with any async function that takes an I/O stream and returns a `Future<Output = Option<HandshakeIo>>`.

The `handshaker` closure now returns an `Option<Io>`. If the handshake succeeds, it returns `Some(io)`. If it fails (e.g., due to an invalid TLS client hello), it returns `None`. The `Handshake` listener will silently drop this connection and await the next one, preventing log spam from common client-side errors.

### Example Usage
```rs
    // Configure Rustls certificate and key
    let certs = CertificateDer::pem_file_iter(
        PathBuf::from(env!("CARGO_MANIFEST_DIR"))
            .join("self_signed_certs")
            .join("cert.pem"),
    )
    .unwrap()
    .collect::<Result<Vec<_>, _>>()
    .unwrap();

    let key = PrivateKeyDer::from_pem_file(
        PathBuf::from(env!("CARGO_MANIFEST_DIR"))
            .join("self_signed_certs")
            .join("key.pem"),
    )
    .unwrap();

    let config = rustls::ServerConfig::builder()
        .with_no_client_auth()
        .with_single_cert(certs, key)
        .unwrap();

    let acceptor = TlsAcceptor::from(Arc::new(config));

    let app = Router::new().route("/", get(handler));

    let addr = SocketAddr::from(([127, 0, 0, 1], 443));
    let listener = TcpListener::bind(addr).await.unwrap();
    debug!("listening on https://127.0.0.1:443");

    // Use the `handshake` adapter to apply TLS to incoming connections
    let listener = listener.handshake(move |stream| {
        let acceptor = acceptor.clone();
        async move {
            match acceptor.accept(stream).await {
                Ok(stream) => Some(stream),
                Err(err) => None,
            }
        }
    });
```

***I've been using this approach personally and have found it to be a clean and effective way to handle TLS in Axum. However, I understand the project may have a different vision for this feature, so please feel free to deny this pull request if it doesn't align with the maintainers' goals. I am happy to make any requested changes.***

### DIFF
--- a/axum/src/serve/listener.rs
+++ b/axum/src/serve/listener.rs
@@ -1,7 +1,14 @@
-use std::{fmt, future::Future, marker::PhantomData, time::Duration};
+use std::{
+    fmt,
+    future::Future,
+    marker::PhantomData,
+    pin::Pin,
+    task::{Context, Poll},
+    time::Duration,
+};
 
 use tokio::{
-    io::{self, AsyncRead, AsyncWrite},
+    io::{self, AsyncRead, AsyncWrite, ReadBuf},
     net::{TcpListener, TcpStream},
 };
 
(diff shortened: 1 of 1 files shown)


mladedav:
This has the same issue I pointed out in #2494, namely that you're `await`ing the handshake in the same loop where you're accepting new connections.

If one bad (or just badly written) actor connects to the server and then hangs in the middle of a TLS handshake, all new connections would be blocked.

---

What was proposed in #3253 was changing the `Listener` to return `impl Future<Output = impl Future<..>>` exactly so that once the outer future resolves, the inner one can be polled while new `accept` is already being polled, essentially having TLS handshakes race concurrently, not processing them sequentially as is done in this PR.

One of the alternatives is what `tls-acceptor` currently does, which is (from what I remember from #3253) to have the handshakes spawned in another thread, signalling back to the `Io`s that they can continue.

The other, and what I think would be preferable, is to map the returned `Io` so that when `AsyncRead`/`AsyncWrite` methods are called, it first drives the handshake to completion (returning `Poll::Pending`) and then doing the requested operations.

NiiightmareXD:
Hi @mladedav,

Thank you for reminding me... again.

I've updated the PR to implement your preferred solution. The listener now immediately returns a wrapper future that implements AsyncRead/AsyncWrite. This drives the handshake to completion when the stream is first used, allowing multiple handshakes to run concurrently without blocking the accept loop.

A Box::pin was necessary to solve the related Unpin issue. The new approach should now align with what you had in mind.

I am unsure whether using Box::pin could cause any performance issues.

Ready for another look when you get a chance!

NiiightmareXD:
We can add `Unpin` trait bound to remove the `Box::pin` requirement: `H: Future<Output = Option<HandshakeIo>> + Send + Unpin`

But I'm not sure if this is the correct way to do this

NiiightmareXD:
We might be able to avoid `Box::pin` here. We could create something like `with_graceful_shutdown` (for example, with_handshaker) function. This function would accept a struct that implements a Handshaker trait and is clonable. This approach could make the implementation a bit cleaner and more efficient. What do you think about this?

axum/src/serve/listener.rs
```diff
@@ -95,6 +102,24 @@ pub trait ListenerExt: Listener + Sized {
             tap_fn,
         }
     }
+
+    /// Add an async handshaking step to the listener.
+    ///
+    /// This is useful for implementing TLS. The handshaker closure is handed
+    /// ownership of the I/O object and is expected to return a new I/O
+    /// object, possibly of a different type.
```
mladedav, line 110: Also mention that by returning `None` you can abort the connection.

axum/src/serve/listener.rs
```diff
@@ -137,6 +162,199 @@ where
...
+/// See that method for details.
+pub struct Handshake<L, F, H, HandshakeIo> {
+    listener: L,
+    handshaker: F,
+    _phantom: PhantomData<fn() -> (H, HandshakeIo)>,
+}
+
+impl<L, F, H, HandshakeIo> fmt::Debug for Handshake<L, F, H, HandshakeIo>
+where
+    L: Listener + fmt::Debug,
```
mladedav, line 176:
```suggestion
    L: fmt::Debug,
```
I don't think we should care about it being a `Listener` here.

axum/src/serve/listener.rs
```diff
@@ -95,6 +102,24 @@ pub trait ListenerExt: Listener + Sized {
...
+
+    /// Add an async handshaking step to the listener.
+    ///
+    /// This is useful for implementing TLS. The handshaker closure is handed
+    /// ownership of the I/O object and is expected to return a new I/O
+    /// object, possibly of a different type.
+    fn handshake<F, H, HandshakeIo>(self, handshaker: F) -> Handshake<Self, F, H, HandshakeIo>
+    where
+        F: FnMut(Self::Io) -> H + Send + 'static,
+        H: Future<Output = Option<HandshakeIo>> + Send,
```
mladedav, line 114:
```suggestion
        H: Future<Output = Option<HandshakeIo>> + Send + 'static,
```
To mirror the bounds that are present later for `impl Listener`.

axum/src/serve/listener.rs
```diff
@@ -137,6 +162,199 @@ where
...
+
+    fn local_addr(&self) -> io::Result<Self::Addr> {
+        self.listener.local_addr()
+    }
+}
+
+pin_project_lite::pin_project! {
+    pub struct HandshakeFuture<F, T>
+    where
+        F: Future,
```
mladedav, line 210: We don't need the bound here and I don't think it would help anyone as we're the only ones building this type.

axum/src/serve/listener.rs
```diff
@@ -137,6 +162,199 @@ where
...
+    enum State<F, T> {
+        Handshaking { #[pin] fut: F },
+        Ready { stream: T },
+    }
+}
+
+impl<F, T> HandshakeFuture<F, T>
+where
+    F: Future<Output = Option<T>>,
+    T: AsyncRead + AsyncWrite,
```
mladedav, line 228: I'd get rid of these too.

Generally, it's better not to put bounds anywhere that are not strictly needed. It can make sense sometimes so that users know right away that something wouldn't work but you're risking that the bounds will drift (like they did even here for the handshaker method. Or even the `T` here which should also be `Unpin + Send`).

axum/src/serve/listener.rs
```diff
@@ -137,6 +162,199 @@ where
...
+        loop {
+            match this.state.as_mut().project() {
+                StateProj::Handshaking { fut } => {
+                    // Poll the handshake future
+                    let stream = match fut.poll(cx) {
+                        Poll::Ready(Some(stream)) => stream,
+                        Poll::Ready(None) => {
+                            return Poll::Ready(Err(io::Error::new(
+                                io::ErrorKind::ConnectionAborted,
+                                "handshake failed",
```
mladedav, line 258: How about having the handshaker return `io::Result` directly? It will have a better idea of what went wrong, maybe it wasn't actually a connection aborted error but something else?

axum/src/serve/listener.rs
```diff
@@ -137,6 +162,199 @@ where
...
+                            )));
+                        }
+                        Poll::Pending => return Poll::Pending,
```
mladedav, line 261: Nit, if you're doing this, you might want to use the `ready!` macro instead.

axum/src/serve/listener.rs
```diff
@@ -137,6 +162,199 @@ where
...
+{
+    fn poll_write(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<io::Result<usize>> {
+        let mut this = self.project();
+        loop {
+            match this.state.as_mut().project() {
+                StateProj::Handshaking { fut } => {
```
mladedav, line 289: This branch is repeated 4 times, extract it into a helper function please.

axum/src/serve/listener.rs
```diff
@@ -137,6 +162,199 @@ where
...
+                }
+                StateProj::Ready { stream } => {
+                    return Pin::new(stream).poll_read(cx, buf);
+                }
+            }
+        }
+    }
+}
+
+impl<F, T> AsyncWrite for HandshakeFuture<F, T>
```
mladedav, line 276: We should ideally also implement any default methods to pass through to the lower specialized methods if provided.

examples/tls-rustls/Cargo.toml
```diff
@@ -7,7 +7,7 @@ publish = false
 [dependencies]
 axum = { path = "../../axum" }
 axum-extra = { path = "../../axum-extra" }
-axum-server = { version = "0.7", features = ["tls-rustls"] }
+tokio-rustls = "0.26"
```
mladedav, line 10: I'd rather see a new example. Even though people can use the `Listener` for this, having also the option and example of using `axum-server` is good I think.

jplatte:
If one of you could extract just the public API being added and put it in the PR description, I'd love to review it (pretty busy so reviewing the whole thing, even if not that big, would be harder to find the time for).

jplatte:
Also, thanks for the PR and sorry for me not responding even longer than David, while I'm still the primary maintainer ([looking to improve the maintenance situation](https://github.com/tokio-rs/axum/discussions/3431)).

mladedav:
I'll leave the description to the author for now, but this adds the following into `ListenerExt`:

```rust
    /// Add an async handshaking step to the listener.
    ///
    /// This is useful for implementing TLS. The handshaker closure is handed
    /// ownership of the I/O object and is expected to return a new I/O
    /// object, possibly of a different type.
    fn handshake<F, H, HandshakeIo>(self, handshaker: F) -> Handshake<Self, F, H, HandshakeIo>
    where
        F: FnMut(Self::Io) -> H + Send + 'static,
        H: Future<Output = Option<HandshakeIo>> + Send,
        HandshakeIo: AsyncRead + AsyncWrite + Unpin + Send + 'static,
    {
        // ...
    }
```

Where the returned struct also implements `Listener` but any call to `AsyncRead::poll_read`, `AsyncWrite::poll_write`, etc. drive the provided future first.

NiiightmareXD:
I've been thinking about how to resolve the potential issues with the Boxed trait, and I have two alternative approaches in mind:

1. **Trait-based generic parameter:** We could introduce a `Handshaker` trait and require a Clone-able implementation to be passed to `ListenerExt`. This mirrors the existing pattern where `TlsAcceptor` must be cloned for each connection, so the user could implement this pattern themselves.

2. **Builder-style configuration:** A cleaner alternative might be to add a `with_handshaker(H: Handshaker)` method, similar in spirit to `with_graceful_shutdown`. The primary advantage here is that the handshaking logic could be executed once per connection before the resulting stream is used. This would entirely remove the need for a stateful `AsyncRead` wrapper that has to check if the handshake is complete on every call to `poll_read`.

3. **Keep the current approach:** We can proceed with the existing implementation using `Box` and address the other feedback. This would be the most direct path to finishing the PR, but it means we must accept the performance hit.

I'm leaning towards Option 2 as it seems to offer a cleaner architecture and better performance by avoiding the per-read state check and the heap allocation.

What are your thoughts on this?

mladedav:
I've taken a closer look and I don't think there is much to be done. We require the `Io` to be `Unpin` because that's what `hyper` requires. So if we want to have the future as a part of the `Io`, it also has to be `Unpin`.

However, I think the proper solution would be to add `H: Unpin` on line 189 and let the user explicitly pin the future or provide one that is `Unpin` and add the `Box::pin` into the example (and remove it from `listener.rs`).

---

I don't really understand what you mean by the second option. So maybe you see something I don't so feel free to pursue it if you think it would lead to a better solution if you think it's worth it.

NiiightmareXD:
Based on the discussion about the performance concerns with the Box wrapper, I've put together a draft of an alternative approach using a Handshaker trait.

You can see the new proposal here: #3440

This new design attempts to resolve the overhead. The builder pattern implementation in the new PR is only a proof-of-concept; I will refine it and ensure it's correct if the general approach is approved.

I'd be really grateful if you could take a quick look and advise if this new direction seems more viable. If so, I'm happy to close this PR and focus our efforts on the new one. Thank you!

PR #3401, opened 2025-07-08
Update and centralize the `LICENSE` file to the repo root 

theodorebje:
## Motivation

Currently, the project `LICENSE` file exists in each crate's directory only. This causes problems since [GitHub is unable to figure out which license the project uses](https://github.com/tokio-rs/axum/community).

This repo also includes multiple different variations of the same file (different years, different wrapping, different symbols, etc). Simply fixing each file to include the same content is not enough, as one `LICENSE` file may be forgotten about in the future.

## Solution
Centralizing the `LICENSE` file in the repo root and then using symlinks in each crate to point to the `LICENSE` file allows us to only have to change one file when updating the `LICENSE`. This helps follow the principle of DRY.

I used `2019–2025` as the copyright year for this centralized `LICENSE` file, since the year needed updating as well. Note that the dash is an en dash (`U+2013`), which requires working Unicode support to display correctly.

For the word wrapping, I used the original from when the repo was originally set up (`002e3f9`). This is different from how [Tokio's `LICENSE` file is wrapped](https://github.com/tokio-rs/tokio/blob/master/LICENSE).

This PR retains the original Unicode symbol (`(c)`, used in [`axum/LICENSE`](https://github.com/tokio-rs/axum/blob/fb64e72de98d229fc3911a6c441514d9c452b108/axum/LICENSE)). Note that the other license files [did not use any symbol at all](https://github.com/tokio-rs/axum/blob/fb64e72de98d229fc3911a6c441514d9c452b108/axum-core/LICENSE).

Related, should I use the proper Unicode Copyright symbol (`©`, `U+00A9`) instead? What do you think?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,5 @@
+MIT License
+
 Copyright (c) 2019–2025 axum Contributors
 
 Permission is hereby granted, free of charge, to any


theodorebje:
License header added due to earlier discussion in #3238.

mladedav (approved):
Sure, though I don't really have an opinion about this. If it helps someone, why not.

One thing to note is not every system supports symlinks but we already use them for the README, so I don't think that matters much.

PR #3403, opened 2025-07-08
Update sse.rs: add space between duration and `:`

GlenDC:
## Motivation

retry was the only field for which we did not add a space

### DIFF
--- a/axum/src/response/sse.rs
+++ b/axum/src/response/sse.rs
@@ -307,7 +307,7 @@ impl Event {
         self
     }
 
-    /// Set the event's retry timeout field (`retry:<timeout>`).
+    /// Set the event's retry timeout field (`retry: <timeout>`).
     ///
     /// This sets how long clients will wait before reconnecting if they are disconnected from the
     /// SSE endpoint. Note that this is just a hint: clients are free to wait for longer if they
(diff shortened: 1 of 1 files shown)


jplatte:
Thanks for the PR! Pinging @SabrinaJewson for review since I still have close to no knowledge of SSE.

SabrinaJewson:
Seems reasonable either way to me. FWIW, Spaces are mandatory for all the other fields because the field value itself may start with a space (including `json` in case the user uses `RawValue` for example), and `retry` is the only field where it’s not necessary. But having consistency also makes sense.

GlenDC:
> Seems reasonable either way to me. FWIW, Spaces are mandatory for all the other fields because the field value itself may start with a space (including `json` in case the user uses `RawValue` for example), and `retry` is the only field where it’s not necessary. But having consistency also makes sense.

strictly speaking, https://html.spec.whatwg.org/multipage/server-sent-events.html, defines spaces as always optional. If the first character of the value is a space it is to be removed, regardless of the field.

That said any other field in Axum's SSE implementation uses a space so there's little reason not to do it for retry. On top of that all other implementations I am aware of use a space, and some test suites rely on this space to be there.

Therefore I would argue to just add a space here as well. It doesn't hurt, it makes it more consistent, aligns with other implementations and doesn't do harm anywhere.

jplatte:
I wonder whether this should have a changelog entry 🤔

GlenDC:
> I wonder whether this should have a changelog entry 🤔

are you planning to do a release soon? how often are these releases? Otherwise I might, if you are ok with it, want to still squeeze in the original issue I posted (about exposing the writer)

jplatte:
No immediate plans for a release, no. Feel free to include the changelog entry in your next PR. Thanks!

PR #3406, opened 2025-07-10
Improve value and reference passing efficiency

theodorebje:
## Motivation

This PR addresses several (small) performance flaws in the current code regarding references and values:

- `&Option<T>` is less common, idiomatic, and allows for fewer memory optimisation strategies by the Rust compiler.
- Passing arguments by value is less flexible and causes unnecessary allocations.
- However, passing arguments by reference for tiny types (2 bytes or fewer) is also inefficient.
- Marking a parameter as `mut` when it is, in fact, not `mut` is less flexible and can make the Rust compiler optimize badly in asynchronous code.
- Including redundant `clone` calls allocates memory for it to immediately be dropped.

It also improves semantics by replacing `to_owned` with `copy` to make the code more readable.

## Solution

This PR adds the following new Clippy lints, and also fixes them in the code:

- [x] `ref_option`
- [x] `needless_pass_by_value`
- [x] `needless_pass_by_ref_mut`
- [x] `trivially_copy_pass_by_ref`
- [x] `redundant_clone`
- [x] `implicit_clone`

See each individual commit's message for more information.

### DIFF
--- a/axum-extra/src/protobuf.rs
+++ b/axum-extra/src/protobuf.rs
@@ -229,9 +229,7 @@ mod tests {
 
         #[axum::debug_handler]
         async fn handler(Protobuf(input): Protobuf<Input>) -> Protobuf<Output> {
-            let output = Output {
-                result: input.foo,
-            };
+            let output = Output { result: input.foo };
 
             Protobuf(output)
         }


theodorebje:
This is a breaking change, by the way.

axum-extra/src/protobuf.rs
```diff
@@ -172,7 +172,7 @@ mod tests {
 
         let app = Router::new().route(
             "/",
-            post(|input: Protobuf<Input>| async move { input.foo.to_owned() }),
+            post(|input: Protobuf<Input>| async move { input.foo.clone() }),
```
mladedav: This is slightly misleading, because the commit message says that we own the value so `clone` is better, but I think `clippy` is just confused by deref here. When I try `(&input.foo).to_owned()`, the lint also goes away.

Anyway, here in both cases the correct solution is `input.0.foo` as we really do have access to the owned value so there's no need to clone implicitly or otherwise.

axum-macros/src/debug_handler.rs
```diff
@@ -8,6 +8,8 @@ use proc_macro2::{Ident, Span, TokenStream};
 use quote::{format_ident, quote, quote_spanned};
 use syn::{parse::Parse, spanned::Spanned, FnArg, ItemFn, ReturnType, Token, Type};
 
+// I'm not sure how to make [`ItemFn`] be passed by reference without causing errors...
+#[allow(clippy::needless_pass_by_value)]
```
mladedav: [This](https://github.com/mladedav/axum/commit/5426b806e60e9d737b76b860c66758f1eed53e54) seems to work for me or am I missing something?

For the record this seems fine, changing the other one feels like a downgrade due to the change to the callsite.

I'd be fine with keeping `allow` for both of these, but I'll let you decide, what you like more.

mladedav (commented):
:+1:

axum-extra/src/protobuf.rs
theodorebje: My bad, didn't even know that `Protobuf<Input>` contained a tuple :sweat_smile:. Haven't had to dig around much with `Protobuf`'s yet.

axum-macros/src/debug_handler.rs
theodorebje: Can't believe I missed that! Was a bit tired when I worked on that commit :sweat_smile:

axum-extra/src/protobuf.rs
```diff
@@ -172,7 +172,7 @@ mod tests {
 
         let app = Router::new().route(
             "/",
-            post(|input: Protobuf<Input>| async move { input.foo.to_owned() }),
+            post(|input: Protobuf<Input>| async move { input.0.foo }),
```
jplatte:
```suggestion
            post(|Protobuf(input): Protobuf<Input>| async move { input.foo }),
```

axum-extra/src/protobuf.rs
```diff
@@ -230,7 +230,7 @@ mod tests {
         #[axum::debug_handler]
         async fn handler(input: Protobuf<Input>) -> Protobuf<Output> {
             let output = Output {
-                result: input.foo.to_owned(),
+                result: input.0.foo,
```
jplatte: Same here.

axum/src/routing/mod.rs
```diff
@@ -347,14 +347,14 @@ where
     }
 
     #[doc = include_str!("../docs/routing/method_not_allowed_fallback.md")]
-    pub fn method_not_allowed_fallback<H, T>(self, handler: H) -> Self
+    pub fn method_not_allowed_fallback<H, T>(self, handler: &H) -> Self
```
jplatte: Let's not have people pass handlers by reference please.

jplatte:
Are there still any breaking changes? If so, should add them to the changelog.

theodorebje:
I don't think so, passing handlers by reference is the only breaking change that I am aware of, and that has been reverted.

PR #3410, opened 2025-07-16
docs: Refine the example comments for `Router::route_service()`

### LINKED ISSUE #2536: `Router::route_service()` example code is out of date
[The current documentation for `axum::Router::route_service()`](https://docs.rs/axum/0.7.4/axum/struct.Router.html#method.route_service) has an example that contains the following lines:

```
        // Services whose response body is not `axum::body::BoxBody`
        // can be wrapped in `axum::routing::any_service` (or one of the other routing filters)
        // to have the response body mapped
        any_service(service_fn(|_: Request| async {
            let res = Response::new(Body::from("Hi from `GET /`"));
            Ok::<_, Infallible>(res)
        }))

        ...

        // This service's response body is `axum::body::BoxBody` so
        // it can be routed to directly.
        service_fn(|req: Request| async move {
            let body = Body::from(format!("Hi from `{} /foo`", req.method()));
            let res = Response::new(body);
            Ok::<_, Infallible>(res)
        })
```

However, despite what the comments in the code say, both responses appear to have `axum::body::Body` as their bodies, and there is currently no `axum::body::BoxBody` in the crate.  I assume that `BoxBody` was removed in an earlier version and the example was only partially updated, but that just raises the question: What is `any_service()` good for now?

unixzii:
## Motivation

Comments in the example code of `Router::route_service()` might be confusing. `axum::body::BoxBody` is not a public item and the description of what `axum::routing::any_service` does also seems inaccurate.

## Solution

The PR fixes some factual errors and clarifies the purpose of `axum::routing::any_service`.

This should fix #2536.

### DIFF
--- a/axum/src/docs/routing/route_service.md
+++ b/axum/src/docs/routing/route_service.md
@@ -20,9 +20,9 @@ let app = Router::new()
     .route(
         // Any request to `/` goes to a service
         "/",
-        // Services whose response body is not `axum::body::BoxBody`
-        // can be wrapped in `axum::routing::any_service` (or one of the other routing filters)
-        // to have the response body mapped
+        // Services whose responses implement `axum::response::IntoResponse` can
+        // be wrapped in `axum::routing::any_service` (or one of the other routing filters)
+        // to turn them into routes.
         any_service(service_fn(|_: Request| async {
             let res = Response::new(Body::from("Hi from `GET /`"));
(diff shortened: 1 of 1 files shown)


wyf027:
Opened a rebased copy as #3766 — happy to close mine if you rebase #3410 instead.

PR #3417, opened 2025-07-27
axum-extra: implement http_body::Body for Field

joeydewaal:
## Motivation
When trying to do streaming uploads to aws s3 with `aws-sdk-s3` you currently have to do something like this.
```rust
async fn upload_to_s3(client: Client, field: Field) {
    // use `TryStreamExt` to map the stream of `Bytes` to a `Stream` of `Frame<Bytes>`.
    let field = field.map_ok(Frame::data);

    // Create something that implements `http_body::Body` from a `Stream`.
    let body = http_body_util::StreamBody::new(field);

    // Create a `ByteStream` from a `Body` (also needs to be 'static).
    let body = ByteStream::from_body_1_x(body);

    let _ = client.put_object().body(body).send().await;
}
```
This is only a few loc extra but also requires to user to take `http_body_util` and `futures` as a dependency.

## Solution
Implement `http_body::Body` for `axum_extra::extract::multipart::Field`. Now we can create a `ByteStream` from the `Field`:

```rust
async fn upload_to_s3(client: Client, field: Field) {
    let body = ByteStream::from_body_1_x(field);

    let _ = client.put_object().body(body).send().await;
}
```

### DIFF
--- a/axum-extra/src/extract/multipart.rs
+++ b/axum-extra/src/extract/multipart.rs
@@ -15,6 +15,7 @@ use http::{
     header::{HeaderMap, CONTENT_TYPE},
     Request, StatusCode,
 };
+use http_body::{Body as HttpBody, Frame};
 use std::{
     error::Error,
     fmt,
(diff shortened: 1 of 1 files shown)


mladedav:
I'm not entirely convinced this makes sense semantically. `Field` itself is not an HTTP body, it's part of a body. It just seems you want to use it as a body in an unrelated request but I don't think that that's enough to warrant the trait implementation.

joeydewaal:
Thats fair

PR #3422, opened 2025-07-30
Update documentation for route layer usage

### LINKED ISSUE #3421: Misleading docs for `route_layer`
The [docs](https://docs.rs/axum/latest/axum/struct.Router.html#method.route_layer) for the `route_layer` method are mentioning that 

> So you have to first add your routes (and / or fallback) and then call route_layer afterwards

Which implies that the middleware would also run for fallback, and it does not. I'm not sure if it's a docs issue or an actual bug that it does not match the fallback

omer9564:
Fixes #3421

### DIFF
--- a/axum/src/docs/method_routing/route_layer.md
+++ b/axum/src/docs/method_routing/route_layer.md
@@ -2,9 +2,12 @@ Apply a [`tower::Layer`] to the router that will only run if the request matches
 a route.
 
 Note that the middleware is only applied to existing routes. So you have to
-first add your routes (and / or fallback) and then call `route_layer`
-afterwards. Additional routes added after `route_layer` is called will not have
-the middleware added.
+first add your routes and then call `route_layer` afterwards. Additional routes
+added after `route_layer` is called will not have the middleware added.
+
+**Important**: The middleware is NOT applied to fallback handlers. Fallback handlers
+will run without the middleware applied. This is because `route_layer` only applies
+to matched requests, while fallbacks handle unmatched requests.
 
 This works similarly to [`MethodRouter::layer`] except the middleware will only run if
 the request matches a route. This is useful for middleware that return early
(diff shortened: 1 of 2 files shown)


axum/src/docs/method_routing/route_layer.md
```diff
@@ -2,9 +2,12 @@ Apply a [`tower::Layer`] to the router that will only run if the request matches
 a route.
 
 Note that the middleware is only applied to existing routes. So you have to
-first add your routes (and / or fallback) and then call `route_layer`
-afterwards. Additional routes added after `route_layer` is called will not have
-the middleware added.
+first add your routes and then call `route_layer` afterwards. Additional routes
+added after `route_layer` is called will not have the middleware added.
+
+**Important**: The middleware is NOT applied to fallback handlers. Fallback handlers
```
jplatte, line 8: This paragraph is redundant with the one after it. Maybe that one isn't quite clear, but I don't think leaving it as-is and adding an extra one before it is the right fix. Can you think about how to merge these two paragraphs?

axum/src/docs/routing/route_layer.md
```diff
@@ -26,10 +29,11 @@ use tower_http::validate_request::ValidateRequestHeaderLayer;
 
 let app = Router::new()
     .route("/foo", get(|| async {}))
+    .fallback(|| async { "fallback" })
     .route_layer(ValidateRequestHeaderLayer::bearer("password"));
 
 // `GET /foo` with a valid token will receive `200 OK`
 // `GET /foo` with a invalid token will receive `401 Unauthorized`
-// `GET /not-found` with a invalid token will receive `404 Not Found`
+// `GET /not-found` with a invalid token will receive `200 OK` (fallback)
```
jplatte, line 37: I don't think this example benefits from having an explicit fallback handler. Why did you add that? Maybe just add `(default fallback handler)` after the `404 Not Found`?

jplatte (commented):
Thanks for the PR! Left some comments.

jplatte:
Hey @omer9564, do you plan to get back to this and address the review feedback?

yanns:
@omer9564 any news on this?

PR #3425, opened 2025-08-05
support custom (binary) data to be written into SSE Event

### LINKED ISSUE #3385: provide writer access to SSE event

### Motivation

For json SSE event data there's a nice function, but in case you have structured data
in a format different then json you are forced to first serialize into something that `AsRef<str>` (https://github.com/tokio-rs/axum/blob/6d78e843982ca299b5d15b7b9095c249e6cdf978/axum/src/response/sse.rs#L202-L217) so that you can pass it as "string-like" data.

This means that for the `datastar` rust sdk crate we are forcing users of axum to double allocate or else roll our own SSE which that crate first did.

### Proposal

Somehow expose the buffer to write into it. This could be done in an opaque way. E.g. in pseudo code:

```
// this writer needs to write a new `data: ` prefix when a newline is getting written into it
struct EventWriter<'a>(&'a mut Buffer);

impl Event {
    pub fn data_writer(&mut self) -> EventWriter<'_> {
         // 1. do your same checks if data wasn't written yet
         // 2.if all good write `data:` prefix and set the state that we have data
         // 3.return writer 
}
```

This could then be used even by the `json` function.

But even better is that now outside users such as `datastar` can write data directly into the buffer,
avoiding the extra data allocations for `datastar`.

### Alternatives

Roll our own SSE or accept useless allocations for each event sent...

GlenDC:
Fixes #3385


GlenDC:
@SabrinaJewson and @jplatte this PR is now ready for review

axum/src/response/sse.rs
```diff
@@ -174,6 +176,23 @@ pub struct Event {
...
+/// such that any form of data can be written as data safely.
+///
+/// This also ensures that newline characters `\r` and `\n`
+/// correctly trigger a split with a new `data: ` prefix.
+///
+/// # Panics
+///
+/// Panics if any `data` has already been written prior to the first write
+/// of this [`EventDataWrite`] instance.
+pub struct EventDataWrite<'a> {
```
SabrinaJewson: By analogy of the naming guideline [C-ITER-TY](https://rust-lang.github.io/api-guidelines/naming.html#iterator-type-names-match-the-methods-that-produce-them-c-iter-ty), it might make more sense to call this struct `EventDataWriter`, since it is returned by the method `Event::data_writer`. Also, this would be inline with e.g. `BufWriter`.

axum/src/response/sse.rs
```diff
@@ -195,24 +225,20 @@ impl Event {
...
-        if self.flags.contains(EventFlags::HAS_DATA) {
-            panic!("Called `Event::data` multiple times");
-        }
+        {
+            use std::io::Write;
 
-        for line in memchr_split(b'\n', data.as_ref().as_bytes()) {
-            self.field("data", line);
+            let mut write = self.data_writer();
+            let _ = write!(write, "{}", data.as_ref());
```
SabrinaJewson: It would be more performant to use:

```rs
writer.write_all(data.as_ref().as_bytes());
```

axum/src/response/sse.rs
```diff
@@ -407,6 +406,48 @@ impl Event {
...
+            self.buffer.write_all(b"data: ")?;
+            last_split = delimiter + 1;
+        }
+        self.buffer.write_all(&buf[last_split..])?;
+
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        self.buffer.flush()
```
SabrinaJewson: [This is a no-op](https://docs.rs/bytes/latest/src/bytes/buf/writer.rs.html#85-87), so maybe just do `Ok(())` here?

axum/src/response/sse.rs
```diff
@@ -407,6 +406,48 @@ impl Event {
     }
 }
 
+impl std::io::Write for EventDataWrite<'_> {
```
SabrinaJewson: Why are we implementing `std::io::Write` instead of `std::fmt::Write`? Given that event streams are required to be valid UTF-8, I think it makes more sense to implement the latter.

For the JSON case, we could maybe avoid UTF-8 checks with a wrapper, since we know the concatenated output will be valid UTF-8?

axum/src/response/sse.rs
```diff
@@ -407,6 +406,48 @@ impl Event {
...
+    fn flush(&mut self) -> std::io::Result<()> {
+        self.buffer.flush()
+    }
+}
+
+impl Drop for EventDataWrite<'_> {
+    fn drop(&mut self) {
+        // ASSUMPTION: data has been written if flags is None
+        // (see impl of `std::io::Write` for `EventDataWrite`)
+        if self.flags.is_none() {
```
SabrinaJewson: The trouble with this is that we can’t technically rely on `Drop` being called. But on the other hand, getting a corrupted SSE stream from `forget`ing this type, which no one should be doing anyway, does not seem like the worst outcome. So maybe just add a comment here explaining the tradeoff?

GlenDC:
@SabrinaJewson ready for review once again. I believe I resolved all feedback and suggestions. Perhaps not in the way you had in mind though, so do let me know if anything else is to be done.

axum/src/response/sse.rs
```diff
@@ -407,6 +421,85 @@ impl Event {
...
+        }
+        writer.write_all(&buf[last_split..])?;
+
+        Ok(buf.len())
+    }
+}
+
+impl std::fmt::Write for EventDataWriter {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        if s.is_empty() {
```
SabrinaJewson: Can we not just call `self.write_buf` in here?

axum/src/response/sse.rs
```diff
@@ -174,6 +175,27 @@ pub struct Event {
     flags: EventFlags,
 }
 
+#[derive(Debug)]
+#[must_use]
+/// Expose [`Event`] as a [`std::io::Write`]
```
SabrinaJewson: as a `std::fmt::Write`

axum/src/response/sse.rs
GlenDC: Sure we could, to me it doesn't make much difference personally. Reason I didn't do is because it are different error types and we would have to choose in that case how to bridge the error types. Any opinion or suggestion regarding that @SabrinaJewson ?

SabrinaJewson: Well, the error types are fake anyway – `write` [cannot fail](https://docs.rs/bytes/latest/src/bytes/buf/writer.rs.html#79-82). I’d just make `write_buf` entirely infallible and either `.unwrap()` or `let _` or `.ok()` in its body.

GlenDC: Fair enough! Sadly it's a static error type, so I suppose your suggestion is best that can be done here. Will rectify it.

GlenDC:
Ready for review once again @SabrinaJewson thanks for the guidance so far, it improved the PR a lot!

GlenDC:
how do you see this moving forward @SabrinaJewson ?

Btw the CI is failing but afaik it's because of the new rust version that introduced some new warnings/lints. Want me to open a PR for from another branch or is that something I can just ignore and you or some colleague will deal with that yourself later?

SabrinaJewson:
Well, I’m not a maintainer so we’ll just have to wait for someone else to review :sweat_smile:

GlenDC:
Ah ok 😓 I suppose they still want you to approve and agree none the less.
For the final approval I suppose it will be @jplatte as usual suspect?

jplatte:
Yeah, a separate PR to fix the lint would be appreciated :)

GlenDC:
> Yeah, a separate PR to fix the lint would be appreciated :)

Hehe, somehow I figured you would take me up on that offer 😅 👯 

@jplatte https://github.com/tokio-rs/axum/pull/3432 it's open ;)
Once that one is approved and merged I'll merge main into this branch and I guess we can discuss about how to move this PR over the finish line as well.

GlenDC:
@jplatte all ci checks pass now

axum/src/response/sse.rs
```diff
@@ -174,6 +176,27 @@ pub struct Event {
     flags: EventFlags,
 }
 
+#[derive(Debug)]
+#[must_use]
+/// Expose [`Event`] as a [`std::fmt::Write`]
```
jplatte: This is pretty unusual, I don't think we put regular attributes above doc comments anywhere else. Can you swap it?

axum/src/response/sse.rs
```diff
@@ -407,6 +422,60 @@ impl Event {
...
+            let _ = writer.write_all(b"data: ");
+            last_split = delimiter + 1;
+        }
+        let _ = writer.write_all(&buf[last_split..]);
+
+        buf.len()
+    }
+}
+
+impl std::fmt::Write for EventDataWriter {
```
jplatte:
```suggestion
impl fmt::Write for EventDataWriter {
```

axum/src/response/sse.rs
GlenDC: For you can I can move the moon Jonas. Consider it done.

jplatte (approved):

PR #3430, opened 2025-08-12
examples/prometheus-metrics: run upkeep task

### LINKED ISSUE #3426: examples/prometheus-metrics: missing call to run_upkeep()?

## Bug Report
I am looking at the example code for setting up prometheus metrics in https://github.com/tokio-rs/axum/blob/main/examples/prometheus-metrics/src/main.rs. It creates a handle like so:
https://github.com/tokio-rs/axum/blob/2977bd4adfe096905059e79eb9c596059da2427e/examples/prometheus-metrics/src/main.rs#L84-L92

Reading the docs for https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/#upkeep-and-maintenance it states
> However, when using lower-level builder methods [PrometheusBuilder::build_recorder](https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/struct.PrometheusBuilder.html#method.build_recorder) or [PrometheusBuilder::install_recorder](https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/struct.PrometheusBuilder.html#method.install_recorder), this upkeep task is not spawned automatically. Users are responsible for keeping a handle to the recorder ([PrometheusHandle](https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/struct.PrometheusHandle.html)) and calling the [run_upkeep](https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/struct.PrometheusHandle.html#method.run_upkeep) method at a regular interval.

I see no calls to `run_upkeep()` in the example code, is this something that should be added? From looking at upstream code it seems this is done like so:
https://docs.rs/metrics-exporter-prometheus/latest/src/metrics_exporter_prometheus/exporter/builder.rs.html#472-478:
```
let recorder_handle = handle.clone();
        tokio::spawn(async move {
            loop {
                tokio::time::sleep(upkeep_timeout).await;
                recorder_handle.run_upkeep();
            }
        });
```

### Version

`main`

eest:
Run upkeep task in the background as this is not done automatically when using the lower-level builder method install_recorder().

The 5 second duration mirrors the PrometheusBuilder default upkeep_timeout.

Fixes #3426.


mladedav (approved):
Thanks, looks good.

The CI failure is unrelated, we'll fix that and then merge this.

PR #3440, opened 2025-08-19
Implement `Handshaker` trait with `with_handshaker`.

NiiightmareXD:
## Motivation

This PR introduces first-class support for TLS termination and other connection-level handshakes directly within `axum::serve`. It is an alternative, more performant approach to the one proposed in #3400.

The previous approach wrapped the I/O stream in a boxed trait object that performed the handshake lazily on the first read. This had two main drawbacks:
1.  **Performance Overhead**: A state check was required on every `poll_read` call to see if the handshake had been completed.
2.  **Heap Allocation**: It required a `Box<dyn ...>` for the stream wrapper, adding allocation and dynamic dispatch overhead for every connection.

This new proposal aims to solve the same problem with a more robust and efficient architecture by decoupling the one-time handshake process from the continuous I/O operations of the resulting stream. This pattern is also flexible enough to support other protocols that require a handshake, such as the PROXY protocol.

## Solution

This PR introduces a new `Handshaker` trait in `axum::serve` and a corresponding `.with_handshaker()` method on the `Serve` builder.

1.  **`Handshaker` Trait**:
    This trait defines an interface for performing a handshake on an incoming raw I/O stream.
    ```rust
    pub trait Handshaker<L: Listener> {
        type Io: AsyncRead + AsyncWrite;
        type Error;
        type Future: Future<Output = Result<Self::Io, Self::Error>>;

        fn handshake(&self, io: L::Io) -> Self::Future;
    }
    ```
    An implementation takes the listener's raw stream (`L::Io`) and returns a future that resolves to a new, potentially wrapped, I/O stream.

2.  **`.with_handshaker()` Method**:
    The `Serve` builder gets a new method, `with_handshaker(H: Handshaker)`. For each connection accepted by the listener, `axum::serve` will first invoke the provided handshaker. The resulting stream (e.g., a `TlsStream`) is then passed to Hyper to handle the HTTP protocol.

This design provides several key advantages:

* **Performance**: The handshake is performed exactly once per connection before any HTTP processing begins, eliminating the per-read overhead of the previous wrapper approach.
* **Zero-Allocation**: The stateful stream wrapper is no longer needed, avoiding heap allocation and dynamic dispatch for I/O operations.
* **Clean Architecture**: It creates a clear separation of concerns between establishing a secure transport layer and handling application-level protocols.
* **Ergonomics**: The builder-style API is consistent with other `axum::serve` configurations like `.with_graceful_shutdown()`, making it intuitive to use.

An example implementation for `tokio-rustls` is included in the examples directory to demonstrate the pattern and provide a clear path for users to implement TLS.


mladedav:
The main problem with this is that we can't combine `with_handshaker` and `with_graceful_shutdown` as it's written now.

We could add for example `with_graceful_shutdown` to `WithHandshaker`, but then people will get confused that they can add graceful shutdown after handshaker but not the other way around. So we can also support the other way, but then they can infinitely chain `with_graceful_shutdown` and `with_handshaker` probably with surprising results.

I think the only usable solution would be to have one struct that contains both graceful shutdown and handshaker and any other extension we would want to support.

Anyway, I wanted to go the other way around and make the whole connectin loop more configurable, you can look at #3441 though I don't see it as being ready (I really pushed it just because of this, otherwise I'd just work on it by myself more later). But I'll definitely look into whether it could support this use-case as well (I don't see a reason it shouldn't). It would probably be mostly solved just by people having easier access to `hyper-rustls` and friends, but I do think easier and more direct control over TLS has some merit. Though as I see it, the TLS implementation should be preferrably in the `Listener`/`Io` part than in the later connection handling logic, but we'll see what the future holds.

NiiightmareXD:
Thanks for the great feedback and for #3441, that makes a lot of sense.

You're right about the builder chaining problem. My idea was to use a typestate pattern, where calling `.with_handshaker()` would return a new type that then offers `.with_graceful_shutdown()`, and the combined type shouldn't have any more with_x to prevent weird combinations.

However, the more configurable connection loop you're proposing in #3441 definitely feels like a better long-term solution.

My main goal with these two PRs was simply to see if I could help advance the TLS support discussion with a proposal. I'm definitely not attached to this specific implementation if there's a better path forward.

Happy to close this PR (and #3400) to support your approach instead. Just let me know what you'd prefer 😊.

Thanks again!

PR #3443, opened 2025-08-20
websocket: add a wrapper around is_terminated

loikki:
Adding a wrapper function around `is_terminated` for `WebSocket`.

## Motivation

In some of my code, I need to check if the websocket is still open or not. From what I understood there is currently no way to do it without pulling some data from it.

## Solution

Just calling the function provide by tungstenite.

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -531,12 +531,10 @@ impl WebSocket {
 }
 
 impl FusedStream for WebSocket {
-
     /// Returns true if the websocket has been terminated.
     fn is_terminated(&self) -> bool {
         self.inner.is_terminated()
     }
-
 }
 
 impl Stream for WebSocket {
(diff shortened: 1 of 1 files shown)


0x676e67:
like:

```rust
impl FusedStream for WebSocket {
    fn is_terminated(&self) -> bool {
        self.inner.is_terminated()
    }
}
```

loikki:
Oh right, I did not realize that it was coming from a single function trait. Thanks for the hint :)

axum/src/extract/ws.rs
```diff
@@ -530,6 +530,13 @@ impl WebSocket {
     }
 }
 
+impl FusedStream for WebSocket {
+    /// Returns true if the websocket has been terminated.
```
mladedav, line 534: This is just a nit for your future reference, but doc comments on trait implementations are pretty hard to find to the point that I'd argue they're mostly useless. They don't hurt though.

mladedav (approved):

PR #3445, opened 2025-08-21
ci: Refactor dependencies-are-sorted job

tottoto:
## Motivation

Refactors the dependencies-are-sorted job.

## Solution

Installs and uses cargo-sort binary directly.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -229,19 +229,12 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v5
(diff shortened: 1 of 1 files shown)


jplatte:
How about using https://github.com/taiki-e/install-action?

tottoto:
Since `taiki-e/install-action` does not support to track the latest version of `cargo-sort` v2, uses a way to get the latest binary file so as to keep the existing behavior. I'll update to use it if it is your preference.

jplatte:
IME it's updated frequently enough. I think I'd slightly prefer it since it seems less likely to break over time.

jplatte (approved):
Thanks!

PR #3447, opened 2025-08-22
Improve tests, tracing defaults, examples, and deps; minor fix in path test

kands-code:
## Motivation

Update the example projects and fix minor issues discovered during the cleanup.

## Solution

- tests(extract/path): fix `captures_match_empty_inner_segments` by using `param.clone()` instead of `param.to_string()` to satisfy clippy
- examples(readme): configure tracing to emit DEBUG-level logs
- examples: lower default tracing level from TRACE to DEBUG; set `axum::rejection=trace` for selected examples to retain useful diagnostics
- examples: standardize tracing macros (`tracing::debug!`, etc.) and format HTML/JS
- examples: rename `auto-reload` to `example-auto-reload` to match example naming conventions
- examples(cors): add `"http://127.0.0.1:3000".parse::<HeaderValue>().unwrap()` to avoid localhost vs 127.0.0.1 CORS mismatch
- deps(examples): bump some example dependencies; set `tokio = "1"` (allow any 1.x) to avoid implying a 1.0.x pin
- examples(oauth): migrate `oauth2` from 4.4.2 to 5.0.0 (**not runtime-verified due to lack of Discord account**)

### Notes

- No behavior change for library code beyond the test fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -112,9 +112,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.95"
+version = "1.0.99"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "34ac096ce696dc2fcabef30516bb13c0a68a11d30131d3df6f04711467681b04"
+checksum = "b0674a1ddeecb70197781e945de4b3b8ffb61fa939a5597bcf48503737663100"
 
 [[package]]
 name = "arc-swap"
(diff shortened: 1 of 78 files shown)


mladedav:
> deps(examples): bump some example dependencies; set tokio = "1" (allow any 1.x) to avoid implying a 1.0.x pin

This is a misconception. Writing `1`, `1.0`, and `1.0.0` is identical and all of those will resolve to the highest available `1.x.y`.

> examples: lower default tracing level from TRACE to DEBUG; set axum::rejection=trace for selected examples to retain useful diagnostics

I don't think this improves anything. The tracing setup in examples is complicated enough as is and I don't think that this really helps in any scenario so I don't think we should do that.

But huge thanks for the oauth bump, I've recently randomly noticed that's the last thing that makes us depend on `http@0.2` so it'll be nice to get rid of that. I'll just try to find some time to verify it works.

Just next time when making so many unrelated changes, try to split them into different commits. It's easier to review that way (and drop selected parts like the tracing example changes).

kands-code:
I apologize for submitting a somewhat disorganized PR. While reviewing axum's example projects, I noticed that many dependencies were "outdated", and some projects did not function as expected (e.g., the `tracing::debug!` in the `readme` example was not working, and the `cors` example had issues with localhost vs. 127.0.0.1). As such, I manually updated the `Cargo.toml` files for each example project, with the following main changes:

1. Formatted `Cargo.toml` files to make them more consistent and readable.
2. Updated outdated dependencies to their latest versions, such as upgrading the oauth dependency to 5.0.

> This is a misconception. Writing `1`, `1.0`, and `1.0.0` is identical and all of those will resolve to the highest available `1.x.y`.

I understand that for Cargo, specifying a version as `1`, `1.0`, or `1.0.0` is equivalent. However, influenced by version constraints in other languages, I had a misconception that "`1.0` implies the dependency is limited to `1.0.x`, while `1.0.0` refers only to the `1.0.0` version." To avoid similar confusion for others, I unified the `tokio` version to `1`.

> > examples: lower default tracing level from TRACE to DEBUG; set axum::rejection=trace for selected examples to retain useful diagnostics
> 
> I don't think this improves anything. The tracing setup in examples is complicated enough as is and I don't think that this really helps in any scenario so I don't think we should do that.

These changes aim to fix the `readme` example project, where the lack of a tracing level caused `tracing::debug!` statements to fail to log properly, and to standardize tracing configurations across all example projects. Additionally, I changed the TRACE level to DEBUG in one or two projects because TRACE-level logs can overwhelm beginners with excessive information, obscuring key points. If you think this change isn’t suitable, I can revert it.

> Just next time when making so many unrelated changes, try to split them into different commits. It's easier to review that way (and drop selected parts like the tracing example changes).

I apologize again! As a newcomer who rarely submits PRs, I overlooked the review process after submission. Sorry for the inconvenience caused.

jplatte:
Would you be up for splitting this into smaller PRs? I'm also okay with @mladedav re-reviewing and merging as-is, but if you have the time, smaller would certainly be nicer.

kands-code:
I apologize for any inconvenience caused, as the content is not well-organized and could benefit from further improvements. I plan to refine it when I have more time and resubmit a better version. Grateful for your understanding and support!

Closing this PR for now.

PR #3454, opened 2025-09-02
Bump tracing-subscriber

avandecreme:
## Motivation

My other PR #3453 had the pipeline fail because of RUSTSEC-2025-0055
https://rustsec.org/advisories/RUSTSEC-2025-0055

## Solution

Ran `cargo update tracing-subscriber`


jplatte (approved):
Thanks!

PR #3459, opened 2025-09-07
fix typo in file_stream error message

lauk20:
## Motivation

Fixed type in error message:
`"build FileStream responsec error: {e}"` should be `"build FileStream response error: {e}"`

## Solution

Fixed the typo


jplatte (approved):
Thanks

PR #3463, opened 2025-09-11
clarify that AddExtension is not the actual `Layer`

2ndDerivative:
This was a bit of a doozie to find, as the fact that the underlying service can't be publicly created isn't trivial to read from the docs, so I added this tip to the documentation.

### DIFF
--- a/axum/src/extension.rs
+++ b/axum/src/extension.rs
@@ -159,7 +159,7 @@ where
 /// [request extensions]: https://docs.rs/http/latest/http/struct.Extensions.html
 ///
 /// If you need a layer to add an extension to every request,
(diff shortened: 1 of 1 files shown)


axum/src/extension.rs
```diff
@@ -157,6 +157,9 @@ where
 /// for more details.
 ///
 /// [request extensions]: https://docs.rs/http/latest/http/struct.Extensions.html
+///
+/// If you need a layer to add an extension to every request,
+/// use the `Layer` implementation of `Extension`.
```
mladedav: Make the extension a link please. Just enclosing it in `[]` should work as it's in scope.

Ideally make `Layer` a link too, but that's not as important.

PR #3464, opened 2025-09-12
Support skipping the `Allow` header in method routing

othelot:
## Motivation
There is no way to remove the added 'Allow' Header from Method Router. It is not possible to operate in middleware, and even if possible, it is a code that causes overhead.

## Solution

Add public access to skip_allow_header.

### DIFF
--- a/axum/src/docs/method_routing/skip_allow_header.md
+++ b/axum/src/docs/method_routing/skip_allow_header.md
@@ -1,0 +1,12 @@
+Skip `Allow` Header for not implemented method
+
+```rust
+use axum::routing::get;
+use axum::Router;
+
+let app = Router::new().route("/", get(|| async {}).skip_allow_header());
+
+// Our app now accepts
+// - GET /
+# let _: Router = app;
+```
--- a/axum/src/routing/method_routing.rs
+++ b/axum/src/routing/method_routing.rs
@@ -1111,6 +1111,7 @@ where
         self.layer(HandleErrorLayer::new(f))
     }
 
+    #[doc = include_str!("../docs/method_routing/skip_allow_header.md")]
     pub fn skip_allow_header(mut self) -> Self {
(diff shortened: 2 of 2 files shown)


jplatte:
What's the actual problem with the header being sent?

othelot:
> What's the actual problem with the header being sent? 

I am developing an undocumented internal API.
It replies 404 Not Found for unimplemented methods. Also, I don't want to let anyone know of an accessible method.

mladedav:
I think this would be brittle for what you're trying to achieve. You would have to call the method on every single method router in your app.

I don't see why a middleware approach wouldn't work. You can just remove the header if you choose to, or is there a problem with that?

Finally, I think what you're really looking for is [`method_not_allowed_fallback`](https://docs.rs/axum/latest/axum/routing/struct.Router.html#method.method_not_allowed_fallback). I didn't test it right now, but I think that it should disable the allow headers. And it also allows you to return 404 as you seem to want rather than the default 405.

othelot:
> I think this would be brittle for what you're trying to achieve. You would have to call the method on every single method router in your app.
> 
> I don't see why a middleware approach wouldn't work. You can just remove the header if you choose to, or is there a problem with that?
> 
> Finally, I think what you're really looking for is [`method_not_allowed_fallback`](https://docs.rs/axum/latest/axum/routing/struct.Router.html#method.method_not_allowed_fallback). I didn't test it right now, but I think that it should disable the allow headers. And it also allows you to return 404 as you seem to want rather than the default 405.

yes, you're right. i'm already using [method_not_allowed_fallback](https://docs.rs/axum/latest/axum/routing/struct.Router.html#method.method_not_allowed_fallback).
but it is still insert `Allow` header. i'll open pull request for skip allow header when use method_not_allowed_fallback.
thank you for comment

PR #3465, opened 2025-09-12
routing: omit Allow header when method_not_allowed_fallback is config…

othelot:
Related: [#3464](https://github.com/tokio-rs/axum/pull/3464)

## Motivation

Some applications configure method_not_allowed_fallback specifically to return 404 for unimplemented HTTP methods and to avoid revealing which methods are supported. Today, MethodRouter still injects an Allow header even when a method_not_allowed_fallback is present, which leaks the set of supported methods and undermines this use case. Removing the header in middleware is brittle and adds overhead, and introducing a per-route “skip allow” knob would require touching every route. Per RFC 9110, Allow is required on 405 responses but not on 404, so omitting it when a 404 fallback is used is compliant.

## Solution

When a MethodRouter has a method_not_allowed_fallback configured, do not automatically insert the Allow header. 
In this mode the fallback is considered authoritative and can set status and headers as it sees fit (including explicitly sending 405 + Allow if desired). 
The default behavior is unchanged: without a fallback, MethodRouter continues to respond with 405 and includes the Allow header. This avoids information leakage for internal/undocumented APIs while preserving RFC-compliant behavior by default.

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning].
 - **breaking:** Remove unused `async-stream` feature, which was accidentally
   introduced as an implicit feature through an optional dependency which was no
   longer being used ([#3298])
+- **breaking:** `option_layer` now maps the `Response` body type to `axum::body::Body` ([#3469])
 
 [#3298]: https://github.com/tokio-rs/axum/pull/3298
 
--- a/axum-extra/src/middleware.rs
+++ b/axum-extra/src/middleware.rs
@@ -1,6 +1,7 @@
 //! Additional middleware utilities.
 
 use crate::either::Either;
+use axum::middleware::ResponseAxumBodyLayer;
 use tower_layer::Identity;
 
 /// Convert an `Option<Layer>` into a [`Layer`].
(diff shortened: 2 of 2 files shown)


jplatte:
I think our current default makes more sense, and would not want to ship this in a patch version, if at all. What's the purpose of not revealing supported methods?

othelot:
> Thanks! Could you cover your change with some tests?

Of course! I just committed tests in the latest commit.

othelot:
> I think our current default makes more sense, and would not want to ship this in a patch version, if at all. What's the purpose of not revealing supported methods?

I think `method_not_allowed_fallback` as an option to customize the default `405` handling. If that customization returns a status other than `405`, the Allow header is not required by RFC 9110 (it’s only required for `405`) and can be semantically confusing. When a fallback is used, it’s more consistent for the router to not inject the `Allow` header and to respect the fallback’s response as-is. 

I really appreciate the flexibility of the `tokio-rs` ecosystem and I respect your perspective; however, this specific behavior feels less flexible in practice because it forces the router to inject Allow even when the fallback intentionally changes the semantics (e.g., to `404`), undermining the purpose of the customization, and it leaks information about supported methods on non-405 responses that some users explicitly try to avoid for internal/undocumented APIs. 

Relying on middleware to strip the header is also brittle, and adds avoidable overhead where the router could simply refrain from inserting it. 

More broadly, it conflicts with the intuition that the fallback is authoritative. 

Once a fallback is configured, users expect its status and headers to be taken as-is.

yanns:
Maybe we could change the signature of `method_not_allowed_fallback` to provide the list of supported methods so that the callback can decide to use them in an `Allow` header or not.
That would be a breaking change though.

othelot:
> Maybe we could change the signature of `method_not_allowed_fallback` to provide the list of supported methods so that the callback can decide to use them in an `Allow` header or not.
> 
> That would be a breaking change though.

That’s good news, I‘m looking forward to it.

Thank you for considering my proposal 😌

jplatte:
> > I think our current default makes more sense, and would not want to ship this in a patch version, if at all. What's the purpose of not revealing supported methods?
> 
> I think `method_not_allowed_fallback` as an option to customize the default `405` handling. If that customization returns a status other than `405`, the Allow header is not required by RFC 9110 (it’s only required for `405`) and can be semantically confusing. When a fallback is used, it’s more consistent for the router to not inject the `Allow` header and to respect the fallback’s response as-is. 

How about setting the allow header if the status code is 405, and omitting it otherwise? (regardless of whether `MethodRouter::fallback` or `Router::method_not_allowed_fallback` was used)

mladedav:
This is the current documentation so we might need to update that too.

```
## Setting the `Allow` header

By default `MethodRouter` will set the `Allow` header when returning `405 Method
Not Allowed`. This is also done when the fallback is used unless the response
generated by the fallback already sets the `Allow` header.

This means if you use `fallback` to accept additional methods, you should make
sure you set the `Allow` header correctly.

```

othelot:
> > > I think our current default makes more sense, and would not want to ship this in a patch version, if at all. What's the purpose of not revealing supported methods?
> > 
> > 
> > I think `method_not_allowed_fallback` as an option to customize the default `405` handling. If that customization returns a status other than `405`, the Allow header is not required by RFC 9110 (it’s only required for `405`) and can be semantically confusing. When a fallback is used, it’s more consistent for the router to not inject the `Allow` header and to respect the fallback’s response as-is.
> 
> How about setting the allow header if the status code is 405, and omitting it otherwise? (regardless of whether `MethodRouter::fallback` or `Router::method_not_allowed_fallback` was used)

that's good idea. i just commited with tests.

othelot:
> This is the current documentation so we might need to update that too.
> 
> ```
> ## Setting the `Allow` header
> 
> By default `MethodRouter` will set the `Allow` header when returning `405 Method
> Not Allowed`. This is also done when the fallback is used unless the response
> generated by the fallback already sets the `Allow` header.
> 
> This means if you use `fallback` to accept additional methods, you should make
> sure you set the `Allow` header correctly.
> ```

I think it makes sense to just use it.
It says "~ **when** returning `405 Method Not Allowed`."

mladedav:
But right after that it says

> This is also done when the fallback is used unless the response
generated by the fallback already sets the `Allow` header.

Which is wrong by my reading.

Just the fact we can argue about whether it is or isn't right means it's not clear, in which case, let's make it clearer if not more correct please.

othelot:
> But right after that it says
> 
> > This is also done when the fallback is used unless the response
> generated by the fallback already sets the `Allow` header.
> 
> Which is wrong by my reading.
> 
> Just the fact we can argue about whether it is or isn't right means it's not clear, in which case, let's make it clearer if not more correct please.

you're right, thanks.

How about correcting this to:
This is also done when the fallback returns `405 Method Not Allowed` 
unless the response generated by the fallback already sets the `Allow` header.

but I'm not good at english, please correct it if there is clearer expression ;(

mladedav:
Sure, that sounds fine.

othelot:
> Sure, that sounds fine.

I just committed changes.
Please review and approve, thanks!

PR #3469, opened 2025-09-12
axum-extra: make `option_layer` map body type to `axum::body::Body`

mladedav:
## Motivation

[`option-layer` does not work with services that map the body type](https://github.com/tokio-rs/axum/discussions/3439) such as `Trace`.

The comment there also seems to be obsolete as `tower::util::option_layer` no longer maps the error types.

## Solution

Now the body will be converted back right away. This will be what most users will most likely want to have happened though there may be advanced scenarios, where someone might want to map the body with another layer and then pay the price of converting to `axum::body::Body` just once. In that case, users can just use `tower::util::option_layer`.


jplatte (approved):
Nice :)

PR #3475, opened 2025-09-13
Update to cargo-deny api version 2

tottoto:
Updates to cargo-deny API version 2. This allow to use features such as `exclude-unpublished` which could make the config simple.


jplatte (approved):
Nice!

PR #3476, opened 2025-09-13
axum: correct typo in append_allow_header

K-tecchan:
## Motivation
I assumed uft-8 was a typo for utf-8, so I corrected it.


tottoto (approved):
Thanks.

Issue #2741, opened 2024-05-16
How to avoid Slowloris DoS Attack?


josecelano:
This is a very critical issue. I opened an issue one month ago but have not yet found a complete answer.

When you use Axum there is no way to set up a timeout for the time the server waits until the client sends the first request. YOu can reproduce it with:

1. Setup a basic server with axum.
2. Open a connection to the server with: `telnet localhost 3000`.

The server will never close the connection even if the client does not send any request.

You can find more info in the [discussion](https://github.com/tokio-rs/axum/discussions/1383) and an [example project I have published](https://github.com/josecelano/axum-server-timeout).

I have converted the discussion into an issue because I think this is a critical issue for some people. I know people who have migrated from Axum to ActixWeb because of this security problem.

I'm even considering it since the main purpose of web framework is to abstract away the details of HTTP operations. And I'm having a lot of trouble trying to patch this problem. I will keep trying and I will post my solution here (if I find it) if I find a complete solution. I know that maintaining this type of library takes considerable effort so I'm not complaining. I just wanted to give more visibility to this problem because I think it's not only my problem but a problem that all users have without even knowing it.

If you want to know what I have tried. I'm trying to use a custom Accetor written by @programatik29 but it does not work when you enable TSL. Details [here](https://github.com/torrust/torrust-index/pull/584).

### Discussed in https://github.com/tokio-rs/axum/discussions/2716

<div type='discussions-op-text'>

<sup>Originally posted by **josecelano** April 18, 2024</sup>
### Summary

Relates to: https://github.com/tokio-rs/axum/discussions/1383

I'm trying to set a timeout for the time the server keeps a connection open while waiting for the client to **start sending** a request.

**IMPORTANT**: it's **NOT** a timeout for:

- The time receiving the request headers (after the client starts sending the headers)
- The time processing the request (because it takes too long)
- The time building the response body.

I've created a repo to reproduce the problem with a detailed description:

https://github.com/josecelano/axum-server-timeout

It's very easy to perform a [slowloris attack ](https://hexadix.com/slowloris-dos-attack-mitigation-nginx-web-server/).

### axum version

0.7.5</div>

mladedav:
`axum` could theoretically add a timeout for the connection that would be cancelled once the `tower` service for the connection (i.e. `axum::Router`) was first polled. This should cover both connections not sending anything (which is what you seem to be describing) as well as attackers sending single bytes with large intervals (the slowloris attack).

jplatte:
[...]

PR #3478, opened 2025-09-14
axum::serve: Enable Hyper request header timeout

kwi-dk:
## Motivation

It's possible for an HTTP request to get stuck (due to network issues or malicious clients) halfway through the request line or header, in which case no amount of timeouts configured by an application using axum/tower will help, since Hyper hasn't yet handed off the request to the service.

This has two consequences:

- Wasted memory: Up to ~1 MB per connection in the worst case, compared to the ~2.5 kB used by a regular idle connection.

- A `with_graceful_shutdown` signal will cause the server to stop accepting new requests, but then hang instead of actually stopping.

## Solution

This changes axum::serve to configure a Timer for HTTP/1 connections, which activates Hyper's default timeout (currently 30 seconds).

The timeout applies only to the request line and request header, not the request body or subsequent response (where axum applications can instead simply configure a timeout using `tower_http::timeout::TimeoutLayer`).

The timeout does not currently apply to newly opened connections, even though Hyper's `header_read_timeout` _should_ apply here too since 1.4.0; discussion on #2741 suggests a possible `TokioIo` issue. However, the graceful shutdown still works as expected for such connections.

The timer is only enabled for HTTP/1 here since similar functionality does not currently appear to exist for HTTP/2 connections in Hyper.

**This is a breaking change** for any axum::serve users who currently rely on being able to begin sending a HTTP/1 request, and then take more than 30 seconds to finish the request line and headers, or rely on keeping connections idle for 30+ seconds between requests.

### DIFF
--- a/axum/src/serve/mod.rs
+++ b/axum/src/serve/mod.rs
@@ -511,7 +511,7 @@ mod tests {
     #[cfg(unix)]
     use tokio::net::UnixListener;
     use tokio::{
-        io::{self, AsyncRead, AsyncWrite, AsyncWriteExt},
+        io::{self, AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt},
         net::TcpListener,
     };
     use tower::ServiceBuilder;
@@ -740,6 +740,39 @@ mod tests {
             tokio::time::timeout(Duration::from_secs(timeout + 2), server_task)
                 .await
                 .expect("server_task didn't exit in time");
+        }
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn test_hyper_header_read_timeout_is_enabled() {
+        let header_read_timeout_default = 30;
+        for req in [
+            "GET / HT",                   // stall during request line
+            "GET / HTTP/1.0\r\nAccept: ", // stall during request headers
+        ] {
+            let (mut client, server) = io::duplex(1024);
+            client.write_all(req.as_bytes()).await.unwrap();
+
+            let server_task = async {
+                serve(ReadyListener(Some(server)), Router::new())
+                    .await
+                    .unwrap();
+            };
+
+            let wait_for_server_to_close_conn = async {
+                tokio::time::timeout(
+                    Duration::from_secs(header_read_timeout_default + 1),
+                    client.read_to_end(&mut Vec::new()),
+                )
+                .await
+                .expect("timeout: server didn't close connection in time")
+                .expect("read_to_end");
+            };
+
+            tokio::select! {
+                _ = server_task => unreachable!(),
+                _ = wait_for_server_to_close_conn => (),
+            };
         }
     }
 
--- a/axum/src/serve/mod.rs
+++ b/axum/src/serve/mod.rs
@@ -28,7 +28,8 @@ pub use self::listener::{Listener, ListenerExt, TapIo};
 /// Serve the service with the supplied listener.
 ///
 /// This method of running a service is intentionally simple and doesn't support any configuration.
-/// Use hyper or hyper-util if you need configuration.
+/// hyper's default configuration applies (including [timeouts]); use hyper or hyper-util if you
+/// need configuration.
 ///
 /// It supports both HTTP/1 as well as HTTP/2.
 ///
@@ -88,6 +89,7 @@ pub use self::listener::{Listener, ListenerExt, TapIo};
 /// error. Errors on the TCP socket will be handled by sleeping for a short while (currently, one
 /// second).
 ///
+/// [timeouts]: hyper::server::conn::http1::Builder::header_read_timeout
 /// [`Router`]: crate::Router
 /// [`Router::into_make_service_with_connect_info`]: crate::Router::into_make_service_with_connect_info
 /// [`MethodRouter`]: crate::routing::MethodRouter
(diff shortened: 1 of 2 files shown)


axum/src/serve/mod.rs
```diff
@@ -686,6 +713,31 @@ mod tests {
...
+            let server_task = async {
+                serve(ReadyListener(Some(server)), Router::new())
+                    .with_graceful_shutdown(tokio::time::sleep(Duration::from_secs(1)))
+                    .await
+                    .unwrap();
+            };
+
+            // Hyper's default request header timeout is 30 secs as of this writing, but
+            // may change in the future, so accommodate a potentially higher timeout.
+            tokio::time::timeout(Duration::from_secs(305), server_task)
```
mladedav: Put here just 30 or 35 if we should be worried that the clock goes forward unexpectedly under some circumstances. I don't think this test should accommodate changes to the timeout from 30 seconds to 5 minutes.

Then again it might be better to just explicitly set the timeout to 30 seconds instead of using the hyper default, then this whole conundrum goes away.

axum/src/serve/mod.rs
```diff
@@ -686,6 +713,31 @@ mod tests {
         assert_ne!(address.port(), 0);
     }
 
+    #[tokio::test(start_paused = true)]
+    async fn test_with_graceful_shutdown_request_header_timeout() {
+        for req in [
+            "",                           // don't send a request
```
mladedav: > The timeout does not currently apply to newly opened connections

I thought that talked about this case but here it seems it's being handled? Or am I misunderstanding?

mladedav:
Can you also please describe the timeout in `serve` documentation and add a changelog entry?

axum/src/serve/mod.rs
kwi-dk: I thought about just configuring an explicit number, but then I figured it was better to leave the concern of picking a reasonable default to Hyper, instead of having a possibly diverging setting in Axum.

But I guess there's no harm in making the test explicit about the timeout currently being 30 seconds, the worst that'll happen is that the test may need a trivial change at some point in the future: dec738dd

axum/src/serve/mod.rs
kwi-dk: This test doesn't actually test the timeout directly, it tests the fact that a graceful shutdown is possible (within the set time). And while newly opened connections don't time out, they can still be gracefully shut down (no timer or timeout required). I've tried to make the test more explicit about this: dec738dd

mladedav: Right, I get it now. So basically these connections will hang indefinitely, but will finish instantly once graceful shutdown is triggered, while when the request was partially written, it will time out after 30 seconds regardless of shutdown.

On that note, there's no way to observe the behavior directly in axum itself without graceful shutdown, right? We could observe it just in a client by opening a connection (or rather a custom `Listener::Io`) that would write partial data like this and then await on some form of `close` event on the client's reader half. If you feel like it, I guess a direct test like that would be slightly better than just the graceful shutdown (though I would of course like to keep this one in either case), but I'm happy to merge as is.

mladedav (approved):

axum/src/serve/mod.rs
kwi-dk: Took a look (d2a0dc66) at adding such a test. Its two "stall during request" cases matches the results of manual testing over real TCP:

    $ printf 'GET / HT' | time nc localhost 3000
    nc localhost 3000  0.00s user 0.01s system 0% cpu 30.010 total

In the "idle before request sent" case, there's no timeout, and the test fails if that case is added, also matching the results of manual testing.

Strangely, in the "idle after complete exchange" case, the test I have here _also_ fails, even though manual testing confirms that hyper's 30 second timeout _is_ active:

    $ ( printf 'GET / HTTP/1.1\r\n'; sleep 10; printf '\r\n' ) | time nc localhost 3000
    HTTP/1.1 404 Not Found
    content-length: 0
    date: Mon, 15 Sep 2025 21:22:54 GMT

    nc localhost 3000  0.00s user 0.00s system 0% cpu 40.005 total

It's unclear to me why the test result doesn't match the real-world result.

mladedav: Thanks for adding it! I'll let you know if I get some time to investigate but having this is already good enough I think.

mladedav: I just tried it and what you did did not time out for me. I had to interrupt the first one, the second one finished on its own after 40 seconds as expected.

```
~ > (printf 'GET / HTTP/1.1\r\n'; sleep 10; printf '\r\n' ) | time nc localhost 3000
HTTP/1.1 200 OK
content-type: text/html; charset=utf-8
content-length: 22
date: Thu, 18 Sep 2025 07:24:09 GMT

<h1>Hello, World!</h1>^Cnc localhost 3000  0,00s user 0,01s system 0% cpu 1:36,32 total
~ >
~ > (printf 'GET / HTTP/1.1\r\n'; sleep 10; printf '\r\nGET /' ) | time nc localhost 3000
HTTP/1.1 200 OK
content-type: text/html; charset=utf-8
content-length: 22
date: Thu, 18 Sep 2025 07:26:04 GMT

<h1>Hello, World!</h1>nc localhost 3000  0,00s user 0,00s system 0% cpu 40,008 total
~ >
```

PR #3479, opened 2025-09-14
chore: move docs from markdown files to rust source

### LINKED ISSUE #929: Documentation not showing up through rust-analyzer when include_str! is used
The way [documentation is defined for certain functions](https://docs.rs/axum/latest/src/axum/routing/mod.rs.html#407) prevents it from showing up through rust-analyzer (see screenshot).

This is what you see when you use `into_make_service()`:
image

This is what you see when you use `into_make_service_with_connect_info()`:
image

I don't know if this should count as a rust-analyzer issue, but it could be worth considering to improve usability of APIs with the current state of tooling.

mladedav:
Closes #929 

## Motivation

Documentation created with `include_str` does not show in `rust-analyzer` tooltips. The issue is tracked by `rust-analyzer` but there's no indication anyone would be able to work on it anytime soon.

## Solution

Move the documentation inline. `handlers_intro.md` and `debugging_handler_type_errors.md` were used in two places so there is some duplication but not too much, one is 7 and the other 40 lines including empty lines.


jplatte:
Thanks for taking care of this. I'll try to review in the next couple days.

yanns:
What is the best way to review this? Generating doc locally and comparing with the current version?

jplatte:
That's a good question 😅 

… there are some git diff settings for highlighting moved code, and for ignoring whitespace. That's probably what I would have tried.

jplatte:
We may have waited long enough w/ finishing this that it won't be necessary anymore.. 😅

https://github.com/rust-lang/rust-analyzer/pull/21303

mladedav:
If it's fixed in rust analyzer and we close this, that's still the best case scenario I'd say.

joshka:
> We may have waited long enough w/ finishing this that it won't be necessary anymore.. 😅

It's still IMHO nicer to read it as code than in a hover pane. I would advocate for accepting this over closing it even with the r-a change.

To review that the docs are the same, perhaps use `cargo doc --output-format json` and compare before and after? Hopefully the identifiers would end up the same (if not then the comparison would have to use a tool - perhaps there's something that does this already against the json?)

mladedav:
I'll close this for now because the docs have changed since then and I suspect it'll be easier to do this again rather than hunting down all changes.

This has been helpful anyway, I didn't think of diffing generated docs so next time it might be even easier.

joshka:
What are your thoughts on:

>It's still IMHO nicer to read it as code than in a hover pane. I would advocate for accepting this over closing it even with the r-a change.

Happy to redo this change if there's some agreement there on this point.

mladedav:
I'm honestly fine with either. My editor has a shortcut to open the file under the cursor but also to fold the whole doc comment when it's inline. I'm not entirely thrilled when there's too long wall of inline docs, but it's what almost all crates do so at this point I don't mind that much.

The main issue for me was ra not showing the docs, so now, I'm really fine with either but it's far from being a priority again. Reading again some of your earlier arguments (doc links and browsing elsewhere than in editor), I am again a bit more inclined to the inline docs, so unless someone else strongly disagrees for some reason, I'd say go ahead.

joshka:
Thinking this through more - maybe it would be better to find a way to let vscode link through to the include files - that's my missing piece on this. I do prefer inline docs examples as a bunch of things work better (intradoc links etc.), but that's in my nice to have bucket - mostly being able to read it, in some inline view (i.e. hover for small things, click through for longer things) without too much extra effort.

mladedav:
I use vscode with the vim extension, so if you happen to use that, `gf` works to open the file under under your cursor. If you're not using that extension, then I have no idea.

PR #3481, opened 2025-09-16
Spelling misstake in Using closure capture

sedrik:
`The downside to this approach is that it’s a the most verbose approach.` -> `The downside to this approach is that it’s the most verbose approach.`


jplatte (approved):
Thanks!

Issue #3305, opened 2025-04-04
Configure Max Header Count


Adam-Alj:
- [x] I have looked for existing issues (including closed) about this

### Motivation

Working on a large enterprise system and using axum for one of our services.
The system, for better or worse, passes around a *large* number of headers, over the [default number of 100 headers as declared in hyper.](https://github.com/hyperium/hyper/blob/master/src/proto/h1/role.rs#L31)

The hyper library [exposes a function to override this default value](https://github.com/hyperium/hyper/blob/c449528a33d266a8ca1210baca11e5d649ca6c27/src/client/conn/http1.rs#L460), however this is not exposed through axum.

### Proposal

Expose this configuration through axum configuration settings.

Would you be receptive to a pull request to expose this as a config option?

### Alternatives

We maintain a patched fork to set this in `serve.rs`, but would rather have this configurability present in the unpatched library.

We considered using plain hyper, which is suggested in the axum docs, however such a relatively innocuous http-server configuration seems valid for inclusion in axum configuration.

mladedav:
There are more hyper configurations that should be available so I don't think we should go with exposing this one option but rather find a way to let users have more control over the underlying hyper connections.

You might also want to look at `axum-server`, I believe you can set things like these through that but it's another (3rd party) dependency.

jplatte:
I don't think there is a way to give people access to all the hyper settings at once without making `hyper-util` a public dependency.

Not sure what the best solution is for this. At my workplace we have also vendored `axum::serve` for some setting (I forget which one). It's probably not actually that bad to add a couple of extra settings as people ask for them?

mladedav:
I worry it'll become a whack-a-mole with different configs to be exposed. But sure, we can expose some settings.

I don't see a problem with making `hyper-util` public though. I don't feel like it's that unstable or anything.

Adam-Alj:
Great to hear that there's interest and openness to this. We'll work on getting a PR up.

Issue #2980, opened 2024-10-11
Handler stuck when receiving request from ffmpeg


marcoNjoHarmonic:
- [x] I have looked for existing issues (including closed) about this

### Version

```
├── axum v0.6.20
│   ├── axum-core v0.3.4
│   ├── axum-macros v0.3.8 (proc-macro)
``` 

### Platform

```
Linux 0bd88cd80067 6.8.0-39-generic #39-Ubuntu SMP PREEMPT_DYNAMIC Sat Jul  6 02:50:39 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux
``` 

### Description

I tried upload files (.m3u8 and .m4v) to the following code with ffmpeg 4.2.10

```rust
mod app_settings;
mod app_state;
mod database;
mod errors;
mod event;
mod http_handlers;
mod jwt;
mod metrics;
mod models;
mod processes;
mod stream;
mod utilities;

use axum::extract::{BodyStream, Path};
use axum::routing::put;
use futures::StreamExt;
use hyper::StatusCode;

use crate::app_settings::AppSettings;

use axum::{Router, Server};

use foundations::{
    cli::{Arg, ArgAction, Cli},
    telemetry::{init_with_server, log},
    BootstrapResult,
};

#[tokio::main]
async fn main() -> BootstrapResult<()> {
    let service_info = foundations::service_info!();

    let cli = Cli::<AppSettings>::new(
        &service_info,
        vec![Arg::new("dry-run")
            .long("dry-run")
            .action(ArgAction::SetTrue)
            .help("Validate or generate config without running the server")],
    )?;

    if cli.arg_matches.get_flag("dry-run") {
        return Ok(());
    }

    let tele_serv_fut = init_with_server(&service_info, &cli.settings.telemetry, vec![])?;

    if let Some(tele_serv_addr) = tele_serv_fut.server_addr() {
        log::info!("Telemetry server is listening on http://{}", tele_serv_addr);
    }

    let router = Router::new().route("/*filepath", put(upload_file));

    let server = Server::bind(&cli.settings.bind).serve(router.into_make_service());
    log::info!("Listening on http://{}", server.local_addr());

    tokio::select! {
        _ = server => {
            log::error!("Server has stopped unexpectedly");
        }
        _ = tele_serv_fut => {
            log::error!("Telemetry server has stopped unexpectedly");
        }
    }
    return Ok(());
}

async fn upload_file(
[...]

PR #3484, opened 2025-09-18
Make serve generic over the HTTP implementations

mladedav:
## Motivation

Currently, only serving with hyper with predefined configuration is supported out of the box. This runs into issues where people want to configure hyper settings (#3305, #2980, #2939). It's fairly complicated to use anything but the default hyper setup, though it is of course possible even now as shown in several low-level examples.

I see this as the natural extension of #2941. We can allow `serve` to not only be generic over the way connections are accepted, but also how they're handled before a `Request` is created for our router to handle (and how to handle the returned `Response`).

## Solution

This adds a `ConnectionBuilder` trait that receives an `Io` and `Service` and connects them to an underlying logic that handles HTTP, e.g. a hyper connection.

The current implementation adds `hyper-util` as a public dependency which might [not be desirable](https://github.com/tokio-rs/axum/issues/3305#issuecomment-2780643872), in which case we may want to not expose the constructor with `Builder`, and have a 3rd party crate like `axum-server` create something like this with exposed internals. That would go a bit against the issues listed in motivation, but even with that, I think it would be nicer to expose hyper-specific settings this way as well as letting others to build custom `ConnectionBuilder`s without having to copy the whole `serve` implementation from here.

I've also added two examples. One is TLS usage with otherwise the default hyper implementation and the other is for showcasing using something else than hyper, in this case Rama.


jplatte:
That's a big diff (even excluding `Cargo.lock`). I'll see about finding time to review it. Also the `Cargo.lock` makes me once again want to split examples into their own repo...

yanns:
This is looking very interesting. Should we move forward with this?

mladedav:
I'd be happy to iterate on this, if you have any feedback, I can update the code or just start cleaning up the code/docs (I remember writing some, but it would probably deserve a bit more).

jplatte:
As before, I'd really like to see the new public API extracted into a comment so we can build consensus on that separately from implementation details :)

mladedav:
There are two new traits that users can implement:

```rust
/// Types that can handle connections accepted by a [`Listener`].
pub trait ConnectionBuilder<Io, S>: Clone {
    /// Take an accepted connection from the [`Listener`] (for example a `TcpStream`) and handle
    /// requests on it using the provided service (usually a [`Router`](crate::Router)).
    fn build_connection(&mut self, io: Io, service: S) -> impl Connection;

    /// Signal to all ongoing connections that the server is shutting down.
    fn graceful_shutdown(&mut self);
}

/// A connection returned by [`ConnectionBuilder`].
///
/// This type must be driven by calling [`Connection::poll_connection`].
///
/// Note that each [`Connection`] may handle multiple requests.
pub trait Connection: Send {
    /// Poll the connection.
    fn poll_connection(
        self: Pin<&mut Self>,
        cx: &mut Context<'_>,
    ) -> Poll<Result<(), Box<dyn StdError + Send + Sync>>>;
}
```

all of these methods are expected to be called by `axum` directly, not by users (that's something that I should add to the docs explicitly).

There is a new public struct `axum::serve::Hyper` which implements `ConnectionBuilder` similarly to what we did until now in the connection loop. It returns `HyperConnection` which implements `Connection` but users should not interact with that type directly in any way (it is public, I _think_ it doesn't have to be).

`Hyper` can be built either with `Hyper::default` which uses the same configuration as we're using now. Otherwise, users can create it with `Hyper::new(builder: hyper_util::server::conn::auto::Builder<TokioExecutor>)` where they've already set all settings they want on the `Builder`.

I know we historically didn't want to have `hyper_util` in public API so we might not want to provide this in `axum` itself and move it elsewhere, be it `axum-extra` or a 3rd party crate.

There's a new `axum::serve::serve_with_connection_builder` method that works just like `axum::serve::serve` except that it takes an additional argument of `impl ConnectionBuilder` which handles the connections. `axum::serve::serve` has the same API as until now and should behave the same way, it just internally uses `serve_with_connection_builder` with the default hyper-specific implementation of `ConnectionBuilder`.

The idea being people who do not care about this can keep using `axum::serve::serve` as they did until now and nothing changes. If they want to tweak hyper configuration, they can use custom `axum::serve::Hyper` along with `serve_with_connection_builder`. If they want to do something more advanced (TLS/not using hyper at all), they have to implement `ConnectionBuilder`.

---

Technically, `Serve` (as in the return value of `axum::serve::serve`) is changed by having an additional type parameter. The same applies to `WithGracefulShutdown`. I don't think most people will notice, but it is a change to the public API.

jplatte:
I'm guessing it's `serve_with_connection_builder` rather than `serve(...).with_connection_builder()` because `serve` continues to be feature-gated on `tokio`? I think it should be possible to make `serve(...)` compile but not `.await`able w/o the connection builder if that feature is disabled, to support the more composable form.

mladedav:
I didn't think of composability but yeah, I think we can do that. Just to make sure, you mean `serve(...)` would produce a `Serve` struct that would use hyper connections and then calling `with_connection_builder` would just swap that out, right? So this
```rust
serve(listener, router).await;
```
and this
```rust
serve(listener, router).with_connection_builder(conn).await;
```
would work?

In any case, the `serve` implementation still uses internally `tokio::spawn`, after an `Io` is obtained from a `Listener`, a task is spawned which hands the `Io` to the `ConnectionBuilder` and then the resulting `Connection` is handled. If we wanted to support other runtimes, I'd say we should add another method like `with_runtime`. From the top of my head, I think we depend just on being able to spawn tasks. So everything to do with `serve` still needs the `tokio` feature regardless of the `ConnectionBuilder` used.

mladedav:
I've redone it as `serve(listener, router).with_connection_builder(conn).await;`.

Just note that `with_graceful_shutdown` does not return `Serve` so currently, one can do
```rust
serve(listener, router).with_connection_builder(conn).with_graceful_shtudown().await;
```
but not
```rust
serve(listener, router).with_graceful_shutdown().with_connection_builder(conn).await;
```

We could add `with_connection_builder` to `WithGracefulShutdown` too but we're duplicating a bit more stuff, which is not great but so far possibly still managable. This was [raised before](https://github.com/tokio-rs/axum/pull/2803#pullrequestreview-2189186400) so I'm just flagging it as an alternative.

Another alternative would be to add a different `WithConnectionBuilder` type so users cannot call `with_connection_builder` and we force the order of `Serve (-> WithConnectionBuilder) -> WithGracefulShutdown` where each transition can be called just once. Currently, users can chain `with_connection_builder` as many times as they want.

Or we can just keep it as is but it just feels a little weird to not be consistent with how the two functions are handled (one changing just a genering on `Serve` and the other returning a different type).

Another open question still is what to do with the public reexports of hyper/hyper-util types. We can have the base implementation without the option to tweak it in axum and have another with better control in `axum-extra`.

davidpdrsn:
I think this is a good direction and I much prefer it over playing whack-a-mole with all of hyper's settings.

I think axum having a public dependency on hyper-util is a hard blocker though and doesn't seem this would be very usable without that.

I like what you suggested [in Discord](https://discord.com/channels/500028886025895936/1406236163819180185/1504463566449934538)

> How about we really extracted axum::serve out to another crate and exposed part of it through axum as it is now (so simple HTTP, no fancy TLS, maybe listeners) and for everyone else, there's a "blessed" dependency that can do anything and everything? Most people wouldn't explicitly pull another crate and those that would would do so because they want something more than just a simple server which to me seems fair. And at least for me preferrable to writing the explicit hyper loop most advanced examples use.

If we were to agree on that what do you think about waiting with this until 0.9 is out the door and then make this PR part of the API of our axum-server-something crate?

mladedav:
Yeah, sounds good.

I'll keep this open if you don't mind for better visibility but it'll probably be easier to create a new PR when the time comes. But I don't mind that much if you'd rather close this now since it looks like you're trying to clean up open issues (thanks for that)

davidpdrsn:
Lets just keep it open. I've moved it to the 0.10 milestone though.

PR #3485, opened 2025-09-20
axum-extra: Make axum optional dependency

tottoto:
## Motivation

It seems that `axum` is only used in a part of functionalities in `axum-extra`. If it can be turned off when it isn't used, it will improve compilation parallelism.

## Solution

Makes `axum` an optional dependency and adds some features which enable `axum`.

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -9,8 +9,16 @@ and this project adheres to [Semantic Versioning].
   introduced as an implicit feature through an optional dependency which was no
   longer being used ([#3298])
 - **breaking:** `option_layer` now maps the `Response` body type to `axum::body::Body` ([#3469])
+- **breaking:** Some new features are added which need to be opted in ([#3485]).
+  - `Cached` extractor requires `cached` feature.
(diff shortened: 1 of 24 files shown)


jplatte:
I like it! Haven't reviewed closely yet though.

axum-extra/CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning].
   introduced as an implicit feature through an optional dependency which was no
   longer being used ([#3298])
 - **breaking:** `option_layer` now maps the `Response` body type to `axum::body::Body` ([#3469])
+- **breaking:** `axum` is an optional dependency.
```
mladedav: For user-facing changelog, I think it would be better to just say we have added new features which may need to be explicitly opted in. The list following this already just lists them.

The fact that axum became an optional dependency, while being the driving reason behind this change, is more of a result of the new features.

mladedav (approved):

axum-extra/CHANGELOG.md
tottoto: Updated, and added a link to this pull request.

PR #3486, opened 2025-09-20
axum-extra: Remove unused tower dependency

tottoto:
## Motivation

`tower` crate doesn't seem to be used in `axum-extra`.

## Solution

Removes `tower` dependency from `axum-extra`.

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -79,7 +79,7 @@ typed-routing = [
 
 # Enabled by docs.rs because it uses all-features
 # Enables upstream things linked to in docs
-__private_docs = ["axum/json", "dep:serde"]
+__private_docs = ["axum/json", "dep:serde", "dep:tower"]
 
 [dependencies]
 axum = { path = "../axum", version = "0.8.4", default-features = false, features = ["original-uri"] }
@@ -94,7 +94,6 @@ mime = "0.3"
 pin-project-lite = "0.2"
 rustversion = "1.0.9"
 serde_core = "1.0.221"
-tower = { version = "0.5.2", default-features = false, features = ["util"] }
 tower-layer = "0.3"
 tower-service = "0.3"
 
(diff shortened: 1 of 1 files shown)


axum-extra/src/middleware.rs
```diff
@@ -30,6 +30,7 @@ use tower_layer::Identity;
 /// `axum_extra::middleware::option_layer` makes sure that the output `Body` is [`axum::body::Body`].
 ///
 /// [`Layer`]: tower_layer::Layer
+/// [`tower::util::option_layer`]: https://docs.rs/tower/latest/tower/util/fn.option_layer.html
```
jplatte: Instead of the http link, can you make tower optional and put it behind the `__private_docs` feature?

tottoto: Yes. Since it isn't used in the code in the document, used the http link, but it can be `__private_doc` feature as well. Updated to use it.

axum-extra/Cargo.toml
```diff
@@ -113,6 +112,7 @@ serde_path_to_error = { version = "0.1.8", optional = true }
 tokio = { version = "1.19", optional = true }
 tokio-stream = { version = "0.1.9", optional = true }
 tokio-util = { version = "0.7", optional = true }
+tower = { version = "0.5.2", default-features = false, features = ["util"], optional = true }
```
jplatte: Can you move this down to the "doc dependencies", next to serde? ^^'

axum-extra/src/middleware.rs
tottoto: Yes. Since it isn't used in the document code block, used the http link, but it can be the `__private_doc` feature as well. Updated to use it.

axum-extra/Cargo.toml
tottoto: Updated.

jplatte (approved):
Thanks!

PR #3487, opened 2025-09-20
axum-extra: Make rustversion and serde_core optional dependency

tottoto:
## Motivation

`rustversion` and `serde_core` crates seem to be used in a part of `axum-extra`.

## Solution

Makes `rustversion` and `serde_core` optional dependency.


jplatte (approved):
Sure, if CI says it's good. Feel free to merge yourself :)

PR #3489, opened 2025-09-21
axum: add ListenerExt::limit_connections

kwi-dk:
## Motivation

It's sometimes desirable to cap concurrent connections, in case keeping server memory usage reasonable is more important than strict availability. (Local testing suggests that a HTTP/1 connection can consume up to 1 MB of memory just for reading request URL and headers with the default hyper configuration used by `axum::serve`.)

On Unix, concurrent connections can be limited with `setrlimit(RLIMIT_NOFILE, …)`, but that interferes with opening files and such.

This feature was previously asked about in discussion #2561.

## Solution

Add `ListenerExt::limit_connections`, which allows limiting the overall number of concurrent connections that `axum::serve` will accept. This is an optional wrapper around `Listener`, and thus carries no overhead if not used.

There's nothing here that requires access to `axum` internals, so theoretically it could live somewhere else, e.g. `axum-extra`. YMMV.


axum/src/serve/listener.rs
```diff
@@ -64,6 +73,24 @@ impl Listener for tokio::net::UnixListener {
 
 /// Extensions to [`Listener`].
 pub trait ListenerExt: Listener + Sized {
+    /// Limit the number of concurrent connections. Once the limit has
+    /// been reached, no additional connections will be accepted until
+    /// an existing connection is closed. (Listener implementations will
+    /// typically continue to queue incoming connections, up to an OS
+    /// and implementation-specific listener backlog limit.)
```
mladedav: Remove these parentheses please (keep the content, just the parentheses).

axum/src/serve/listener.rs
```diff
@@ -99,6 +126,83 @@ pub trait ListenerExt: Listener + Sized {
...
+    /// See [`ListenerExt::limit_connections`] for details.
+    pub struct ConnLimiterIo<T> {
+        #[pin]
+        io: T,
+        permit: OwnedSemaphorePermit,
+    }
+}
+
+// Simply forward implementation to `io` field.
+impl<T: AsyncRead> AsyncRead for ConnLimiterIo<T> {
```
mladedav, line 166: You correctly implemented all provided methods for `AsyncWrite`, do the same thing here too please.

axum/src/serve/listener.rs
```diff
@@ -165,3 +269,55 @@ fn is_connection_error(e: &io::Error) -> bool {
...
+        assert_eq!(COUNT.load(Ordering::SeqCst), 0);
+
+        // First 'accept' succeeds immediately.
+        let conn1 = listener.accept().await;
+        assert_eq!(COUNT.load(Ordering::SeqCst), 1);
+
+        time::timeout(time::Duration::from_secs(1), listener.accept())
+            .await
+            .err()
+            .expect("Second 'accept' should time out.");
```
mladedav: Nit, you can call `expect_err` instead.

axum/src/serve/listener.rs
kwi-dk: 97c00b0a

axum/src/serve/listener.rs
kwi-dk: [That's it](https://docs.rs/tokio/latest/tokio/io/trait.AsyncRead.html), though. It also surprised me, but everything else is provided by AsyncReadExt.

axum/src/serve/listener.rs
kwi-dk: Added `Debug` impl for `ConnLimiterIo`, and switched to `expect_err` in 97c00b0a,

axum/src/serve/listener.rs
mladedav: Right, sorry about this. I googled to check but it seems I got to a random page with docs of the 0.2 version which had some provided methods.

axum/src/serve/listener.rs
mladedav: I didn't realize this wouod need a new trait implementation. That said having the `Debug` impl is probably what we want anyway, so thanks.

PR #3491, opened 2025-09-21
axum: Make futures-sink optional dependency

tottoto:
## Motivation

`futures-sink` seems to be used only in the `ws` feature.

## Solution

Makes `futures-sink` an optional dependency.


yanns (approved):
💯

PR #3492, opened 2025-09-21
axum-extra: Add link definition for pull request to changelog

tottoto:
## Motivation

It doesn't seem there is a definition which corresponds to the link in the changelog.

## Solution

Adds a link definition for the pull request to the changelog.


yanns (approved):
thanks!

PR #3496, opened 2025-09-25
add axum-conditional-requests to ECOSYSTEM.md

2ndDerivative:
Hi!
I'm making this library and as it has its first features I'd like to add it here.

I will update this entry appropriately when I build up the conditional request library


ECOSYSTEM.md
```diff
@@ -60,6 +60,7 @@ If your project isn't listed here and you would like it to be, please feel free
 - [static-serve](https://crates.io/crates/static-serve): A helper macro for compressing and embedding static assets in an axum webserver.
 - [datastar](https://crates.io/crates/datastar): Rust implementation of the Datastar SDK specification with Axum support
 - [axum-governor](https://crates.io/crates/axum-governor): An independent Axum middleware for rate limiting, powered by [lazy-limit](https://github.com/canmi21/lazy-limit) (not related to tower-governor).
+- [axum-conditional-requests](https://crates.io/crates/axum-conditional-requests): A library for handling client-side caching HTTP headers (Only supports Last-Modified so far)
```
mladedav: I think it's fine if you don't include the "Only supports Last-Modified so far" part. If someone looks to the project, they can find out which parts it does or doesn't support and keeping that information here could become needlessly verbose.

It also means this won't drift away from reality as you continue to add features.

yanns: @2ndDerivative can you update the PR following the previous comment?

PR #3502, opened 2025-09-28
Remove unused rustversion dependency of axum-core

jplatte:


mladedav (approved):
Huh, that's nice, I thought we still had it for the `unimplemented` compilation hints on `FromRequestParts`and other traits but I guess the msrv bumps also got rid of that.

PR #3508, opened 2025-09-30
Release axum 0.8.6 and related crates

mladedav:
This is practically empty release just to fix docs.rs build.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -297,7 +297,7 @@ dependencies = [
 
 [[package]]
 name = "axum"
-version = "0.8.5"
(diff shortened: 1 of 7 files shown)


jplatte (approved):
axum-core wasn't actually changed, so there is no need to release it. But I guess I'll just do it anyways, so there's no need to update the branch again.

**DO NOT SQUASH-MERGE**
I'll do it on the commandline.

jplatte:
I've ran `cargo publish` for all three crates and created the git tags. Do you want to do the GH releases?

PR #3509, opened 2025-09-30
Use Vec for PathRouter

pickfire:
Routes in path seemed to work with linear memory instead of needing the extra hash and tracking of prev_route_id, a Vec should work just fine while having the benefits of cache locality.

## Motivation

`Vec` seemed to be able to fulfill routes without needing `HashMap`, as well as better cache locality.

## Solution

Change `PathRouter` to use `Vec`.

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -58,7 +58,7 @@ macro_rules! panic_on_err {
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub(crate) struct RouteId(u32);
+pub(crate) struct RouteId(usize);
 
 /// The router type for composing handlers and services.
 ///
--- a/axum/src/routing/path_router.rs
+++ b/axum/src/routing/path_router.rs
@@ -14,9 +14,8 @@ use super::{
 };
 
 pub(super) struct PathRouter<S> {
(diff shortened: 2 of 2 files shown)


yanns:
Under which scenarios would we iterate over all elements instead of using the optimized cache access?

mladedav:
I think we iterate just when we merge/nest routers. I'm not sure what do you mean by optimized cache access.

Disrgarding any cache locality arguments, I would still merge this because it cleans up the code by not having to track `prev_route_id` explicitly.

This is on me that I did not do so earlier but I'm happy to do that now unless you see there some issues (or prefer the original for some reason).

yanns:
> I'm not sure what do you mean by optimized cache access.

Usually, one uses a hashmap to avoid a complete iteration, and to get "optimized" access by key.

I don't see any issues with it. I also think it's a nice change. You can go forward with it. ❤️

yanns (approved):

mladedav:
For merging & nesting there is really no way around the full iteration, we have to empty one vector/map and insert those into the other, there's no iteration to find a given value if I'm not mistaken if that's the iteration you mean. When looking up a given route, we always know the route ID so now instead of hashing it and using that for the lookup, we index by that directly.

pickfire:
Yes, we don't need to do the extra hashing, using index is enough.

PR #3514, opened 2025-10-02
ci: Check axum-extra typed-routing feature

tottoto:
## Motivation

Since `axum-extra`'s `typed-routing` feature requires `routing` feature, both are needed to be enabled to be checked.

## Solution

Adds an check of `axum-extra` with enabling both of `routing` and `typed-routing` features.


jplatte:
I see, this is due to #3485. But I don't think this is the right solution. Shouldn't we just make the `typed-routing` feature pull in the `routing` feature, instead of being a no-op on its own?

tottoto:
Yes, that is also possible. The original change was intended to fix the current check. Updated.

PR #3517, opened 2025-10-03
Upgrade axum-extra to prost v0.14

xumaple:
## Motivation

Upgrade to newest version of prost so it's compatible with tonic 0.14


jplatte:
Mind adding a changelog entry?

PR #3518, opened 2025-10-05
examples: Update to diesel-async 0.7

tottoto:
Updates to diesel-async 0.7.


examples/diesel-async-postgres/Cargo.toml
```diff
@@ -6,8 +6,9 @@ publish = false
 
 [dependencies]
 axum = { path = "../../axum", features = ["macros"] }
-diesel = "2"
-diesel-async = { version = "0.6", features = ["postgres", "bb8"] }
+diesel = "~2.3"
```
mladedav, line 9: Why tilde, is `2.4` somehow broken?

tottoto: `diesel-async` only supports versions which satisfy `~2.3`.

tottoto: `diesel-async` `0.7` only supports versions which satisfy `~2.3`.

mladedav: In that case cargo should be able to find the version on its own without us having the tilde here too I think.

Then again it doesn't matter, it's fine if we have this here.

tottoto: While the tilde is technically unnecessary, there is other example that uses `diesel`, so I suppose that making it clearer by explicitly stating that there is the restriction on the `diesel` version would make the examples easier to understand.

PR #3527, opened 2025-10-10
refactor error logging in error-handling example

pemattern:
## Motivation

It's clumsy to log as a side-effect in a type conversion function.

## Solution

Ensure a separation of concerns by optionally adding an `AppError` `Extension` to the `Response` when logging is desired, which can be used by a middleware to log the `AppError`

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -88,6 +88,7 @@ query = [
 tracing = ["axum-core/tracing", "axum/tracing", "dep:tracing"]
 typed-header = ["dep:headers"]
 typed-routing = [
+    "routing",
     "dep:axum-macros",
     "dep:percent-encoding",
     "dep:serde_core",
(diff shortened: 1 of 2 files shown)


examples/error-handling/src/main.rs
```diff
@@ -163,31 +166,37 @@ impl IntoResponse for AppError {
...
+        }
+        response
     }
 }
 
 impl From<JsonRejection> for AppError {
     fn from(rejection: JsonRejection) -> Self {
-        Self::JsonRejection(rejection)
+        // Arc enables cloning of a JsonRejection
+        Self::JsonRejection(Arc::new(rejection))
```
mladedav: I think it would be better to not use the `Arc` here but instead add `Arc<AppError>` to the extensions.

Though that might be just my preference to not change the errors themselves to enable this scenario.

pemattern: Agreed, much better.

examples/error-handling/src/main.rs
```diff
@@ -163,30 +164,36 @@ impl IntoResponse for AppError {
...
+            // Insert our error into the response, our logging middleware will use this.
+            response.extensions_mut().insert(Arc::new(err));
+        }
+        response
     }
 }
 
 impl From<JsonRejection> for AppError {
     fn from(rejection: JsonRejection) -> Self {
+        // Arc enables cloning of a JsonRejection
```
jplatte, line 196: This comment is now in the wrong place.

pemattern: my bad..

PR #3539, opened 2025-10-25
examples: Update to sqlx 0.9

tottoto:
Updates to `sqlx` 0.9.

### DIFF
--- a/examples/sqlx-postgres/Cargo.toml
+++ b/examples/sqlx-postgres/Cargo.toml
@@ -10,4 +10,4 @@ tokio = { version = "1.0", features = ["full"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 
-sqlx = { version = "0.8", features = ["runtime-tokio-rustls", "any", "postgres"] }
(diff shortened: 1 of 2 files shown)


examples/sqlx-postgres/Cargo.toml
```diff
@@ -10,4 +10,4 @@ tokio = { version = "1.0", features = ["full"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 
-sqlx = { version = "0.8", features = ["runtime-tokio-rustls", "any", "postgres"] }
+sqlx = { version = "=0.9.0-alpha.1", features = ["runtime-tokio", "any", "postgres"] }
```
davidpdrsn: Would it make sense to wait until 0.9 is fully released and out of testing phase?

If people do `cargo add sqlx` they'll still get 0.8

tottoto: I think we should wait until 0.9 is released.

davidpdrsn (approved):
Thank you!

PR #3541, opened 2025-10-27
Release axum-extra 0.12

mladedav:
I've tried to cherry pick anything that was in the changelog and then whatever started with `axum-extra:` in the commit message.

I then just merged main back into this keeping the version update in `Cargo.toml` and the changelog for 0.12.

I hope I didn't miss anything (though I wouldn't be surprised if I did) and if I should do something else than the merge back, I can change this.


axum-extra/CHANGELOG.md
```diff
@@ -16,12 +21,10 @@ and this project adheres to [Semantic Versioning].
   - `OptionalPath` extractor requires `optional-path` feature.
   - The routing utilities require `routing` feature.
   - `WithRejection` extractor requires `with-rejection` feature.
-- **breaking:** Upgraded `prost` dependency to v0.14. ([#3517])
```
tottoto, line 19: I think that it would be good to include upgrading `prost` in `axum-extra` 0.12.

mladedav: I'll check again just in case, but it should be there, the changelog should be wrong.

jplatte: Maybe forgot to cherry-pick the changelog entry? Either way, this PR will remove the changelog entry as-is so please make sure that it's updated (and the release branch should also have the right changelog).

mladedav: You were right, I lost that commit somewhere on the way. Thanks for catching it.

tottoto (approved):

jplatte:
Okay, so I've done the release. For the merge back to main, I usually do it the other way around (not sure whether there is any difference other than the order of the parents of the merge commit + how the default commit message mentions them). Probably doesn't matter.

What does matter though is that I don't want to merge a merge commit, so would you mind redoing the merge against the latest `main`, and once CI passes fast-forward `main` to it? (even though the web UI doesn't allow you to fast-forward merge a PR, PR-based branch restrictions can be met when pushing from the command line - GitHub looks for a matching PR that has the commit you're trying to push as its head and checks the conditions based on that)

jplatte:
Closing, did this as part of https://github.com/tokio-rs/axum/pull/3547.

PR #3542, opened 2025-10-30
Use extensions directly in from_request_parts

braddunbar:
## Motivation

We don't need to go through `Extension<T>::from_request_parts` and build an error to throw away when we have direct access to `parts.extensions`.

## Solution

Use `parts.extensions` directly to get the cached value.


jplatte (approved):
This also gets rid of a pointless await point. Thanks!

yanns:
nice!

PR #3550, opened 2025-11-10
Make it easier to visually scan for default features

silverlyra:
## Motivation

Tables with yes–no columns (or other kinds of binary distinctions) can be difficult to 👀 visually scan if the column in each row contains “yes” or “no”. It’s usually easier to spot the “yes” rows if they have a symbol there that the “no” rows do not.

The tables of available features on `axum` and `axum-extra` have this issue on their _Default?_ column where the `default` features are indicated.

## Solution

Before | After
-------|------
Screenshot 2025-11-09 at 11 43 28 PM | Screenshot 2025-11-09 at 11 43 55 PM

This PR removes the “No” values, and replaces “Yes” with a checkmark. Scaled down here to fit within the PR description space, the easier readability is (imo!) even more obvious.

For screen readers, I wrapped the checkmark in a `<span>` which sets the `aria-label` to “Default feature”, so this should be clearer even when the table is not being visually read.

I appreciate everyone who’s worked to ship and maintain such a delightful library as Axum! 🙏🏻

### DIFF
--- a/axum-extra/src/lib.rs
+++ b/axum-extra/src/lib.rs
@@ -9,30 +9,30 @@
 //!
 //! Name | Description | Default?
 //! ---|---|---
-//! `async-read-body` | Enables the [`AsyncReadBody`](crate::body::AsyncReadBody) body | No
-//! `attachment` | Enables the [`Attachment`](crate::response::Attachment) response | No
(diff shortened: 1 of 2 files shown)


yanns (approved):
I like it!

jplatte (approved):
Yeah, this is great 💚

PR #3555, opened 2025-11-13
Relax implicit Send / Sync bounds

jplatte:
Implicit bounds were unnecessarily restrictive.
Closes #3552.

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -605,7 +605,7 @@ where
 /// See [`Router::as_service`] for more details.
 pub struct RouterAsService<'a, B, S = ()> {
     router: &'a mut Router<S>,
-    _marker: PhantomData<B>,
+    _marker: PhantomData<fn(B)>,
 }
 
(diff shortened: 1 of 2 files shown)


axum/src/routing/mod.rs
```diff
@@ -605,7 +605,7 @@ where
 /// See [`Router::as_service`] for more details.
 pub struct RouterAsService<'a, B, S = ()> {
     router: &'a mut Router<S>,
-    _marker: PhantomData<B>,
+    _marker: PhantomData<fn(B)>,
```
tottoto, line 608: Why is the variance of the type about `B` contravariant?

jplatte: Because this is the request body, and `Router` receives requests, it doesn't produce them. So, it only makes sense to me to use the same variance as `fn(B)`. Practically, I think we require `'static` anyways so this doesn't matter?

tottoto: Understood. Makes sense.

yanns (approved):

PR #3566, opened 2025-11-19
Fix integer underflow in `try_range_response` for empty files

Darksonn:
Thanks to @mufeedvh for reporting this bug.

### DIFF
--- a/axum-extra/src/response/file_stream.rs
+++ b/axum-extra/src/response/file_stream.rs
@@ -603,43 +603,34 @@ mod tests {
 
     #[tokio::test]
     async fn response_range_empty_file() -> Result<(), Box<dyn std::error::Error>> {
-        struct TempFile(&'static str);
-
-        impl Drop for TempFile {
-            fn drop(&mut self) {
-                let _ = std::fs::remove_file(self.0);
-            }
-        }
-
-        let filename = "test_empty_file.txt";
-        std::fs::write(filename, []).unwrap();
-        let _cleanup = TempFile(filename);
+        let file = tempfile::NamedTempFile::new()?;
+        file.as_file().set_len(0)?;
+        let path = file.path().to_owned();
 
         let app = Router::new().route(
             "/range_empty",
-            get(move |headers: HeaderMap| async move {
-                let range_header = headers
-                    .get(header::RANGE)
-                    .and_then(|value| value.to_str().ok());
+            get(move |headers: HeaderMap| {
+                let path = path.clone();
+                async move {
+                    let range_header = headers
+                        .get(header::RANGE)
+                        .and_then(|value| value.to_str().ok());
 
-                let (start, end) = if let Some(range) = range_header {
-                    if let Some(range) = parse_range_header(range) {
-                        range
+                    let (start, end) = if let Some(range) = range_header {
+                        if let Some(range) = parse_range_header(range) {
+                            range
+                        } else {
+                            return (StatusCode::RANGE_NOT_SATISFIABLE, "Invalid Range")
+                                .into_response();
+                        }
                     } else {
-                        return (StatusCode::RANGE_NOT_SATISFIABLE, "Invalid Range")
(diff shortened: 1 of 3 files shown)


Range request DoS via integer underflow in try_range_response - axum.pdf
martin-g, line 1: This file seems unused.
But even it should be used in a test then most probably it should not be in the project root.

Darksonn: Sorry I didn't mean to commit that file.

seanmonstar:
Just to add a note from spec:

> For a [GET](https://httpwg.org/specs/rfc9110.html#GET) request, a valid bytes [range-spec](https://httpwg.org/specs/rfc9110.html#rule.ranges-specifier) is [satisfiable](https://httpwg.org/specs/rfc9110.html#satisfiable) if it is either:
>
> - an [int-range](https://httpwg.org/specs/rfc9110.html#rule.int-range) with a [first-pos](https://httpwg.org/specs/rfc9110.html#rule.int-range) that is less than the current length of the selected representation

So, with a file of 0 bytes, the first-pos of 0 is not less than the length, and thus is now rejected.

yanns:
From CI:
```
error: the borrowed expression implements the required traits
   --> axum-extra/src/response/file_stream.rs:615:34
    |
615 |         std::fs::write(filename, &[]).unwrap();
    |                                  ^^^ help: change this to: `[]`
```

Darksonn:
Fixed thanks.

axum-extra/src/response/file_stream.rs
```diff
@@ -596,4 +600,61 @@ mod tests {
...
+    async fn response_range_empty_file() -> Result<(), Box<dyn std::error::Error>> {
+        struct TempFile(&'static str);
+
+        impl Drop for TempFile {
+            fn drop(&mut self) {
+                let _ = std::fs::remove_file(self.0);
+            }
+        }
+
+        let filename = "test_empty_file.txt";
```
jplatte: Instead of writing to the project directory when running tests, can you add a dev-dep on `tempfile` and use a `NamedTempFile`, which will be written to the system's temp directory?

Darksonn: Done.

yanns (approved):

axum-extra/src/response/file_stream.rs
```diff
@@ -191,6 +191,10 @@ where
         let metadata = file.metadata().await?;
         let total_size = metadata.len();
 
+        if total_size == 0 {
+            return Ok((StatusCode::RANGE_NOT_SATISFIABLE, "Range Not Satisfiable").into_response());
```
darth-raijin, line 195: Question:
I noticed that the error body `"Range Not Satisfiable"` is hard-coded directly into the response. Would it make sense to extract this into a shared constant so tests can assert on the *exact* same value?

Having the literal inline makes tests fragile, since they need to duplicate the exact string. If the message ever changes, that could also unintentionally break consumers relying on the current wording or cause test drift.

In particular, there are two branches in this PR that both return `StatusCode::RANGE_NOT_SATISFIABLE`. It might be useful if the [test](https://github.com/tokio-rs/axum/pull/3566/files#diff-572de7d1dc9a91415d797b7a1fdb59e7ce3113da6f2f123739dff2f25becf601R648) also asserted on the error body, to ensure the `StatusCode::RANGE_NOT_SATISFIABLE` comes from the branch in [`file_stream.rs`](https://github.com/tokio-rs/axum/pull/3566/files#diff-572de7d1dc9a91415d797b7a1fdb59e7ce3113da6f2f123739dff2f25becf601R195) with the expected response body, and not the [test branch](https://github.com/tokio-rs/axum/pull/3566/files#diff-572de7d1dc9a91415d797b7a1fdb59e7ce3113da6f2f123739dff2f25becf601R623).

PR #3567, opened 2025-11-21
docs: clarify `route_layer` does not apply middleware to the fallback handler

darth-raijin:
## Motivation
The current `route_layer` documentation incorrectly implies that middleware applied via `route_layer` will also run on the fallback route. This is not the case, and this PR clarifies the intent.

Issue #3421 was created, with a PR to correct the bug in the documentation. However, that PR has been stale for multiple weeks.

## Solution
* Remove misleading section `(and / or fallback)` in `route_layer.md`
* No functional changes, the PR updates documentation only.


jplatte (approved):
Thank you!

PR #3568, opened 2025-11-21
Allow latest version of matchit (v0.8.6)

sferik:
This PR relaxes the version constraint on `matchit` to include the latest v0.8.6 release, bringing in the latest improvements and fixes from the routing library. I've verified that all tests still pass on the latest version.


mladedav:
I believe there's already 0.9 out though. Also note that the prefix/sufix rules have been reworked between that and the last 0.8.x release so it's better to skip that.

mladedav:
You might also take a look at #3143. Just the fact that current tests pass isn't sufficient because axum has to handle the new matchit features introduced in the new version.

sferik:
Ah, I didn't see that https://github.com/tokio-rs/axum/pull/3143 was already in the works. I'll close this one. Thanks for the feedback.

PR #3571, opened 2025-11-25
fix: sse event do not skip data field when empty

co42:
## Motivation

Since https://github.com/tokio-rs/axum/releases/tag/axum-v0.8.5 SSE events with empty data are sent without a `data` field.

```rust
Event::default().event("deleted").data("")
```

Before :
```
event: deleted
data: 

```

After
```
event: deleted

```

According to https://html.spec.whatwg.org/multipage/server-sent-events.html (I did not found an RFC)

> If the data buffer is an empty string, set the data buffer and the event type buffer to the empty string and return.

Which means these events are ignored (if I understand it correctly).

Before this change the SSE client I use was receiving the event, it does not anymore.

More context : https://github.com/jpopesculian/eventsource-stream/issues/6

## Solution

Write empty data field if set to an empty string as before.


jplatte:
Seems likely to be related to #3154, any thoughts @nanoqsh / @mladedav?

co42:
Can it be merged ?

kafji:
Stumble on this while porting from other code.

Any update on this @jplatte, @nanoqsh, @mladedav?

The proposed change is sensible and the change is small enough, tiny even.

jplatte:
I'm no longer a maintainer. @Turbo87 @yanns @GlenDC @tottoto might be able to help.

axum/src/response/sse.rs
```diff
@@ -653,22 +649,24 @@ mod tests {
...
 
+    #[test]
+    fn empty_data_not_omitted() {
+        let no_leading_space = Event::default().data("");
+        assert_eq!(&*no_leading_space.finalize(), b"data: \n\n");
+    }
+
     #[test]
     fn write_data_writer_str() {
         // also confirm that nop writers do nothing :)
```
mladedav, line 660: This comment is obsolete.

Though I think the test should keep working if we don't delete the empty writes?

mladedav (commented):
Sorry, I don't really know the SSE implementation but I think we can merge it. If you want to reopen, I'd be happy to merge this PR, otherwise anyone is welcome to open a new one.

However, I think that this is a subtle change in behavior that I'd rather do in a breaking release.

PR #3574, opened 2025-11-27
test(axum-extra): improve `response_not_set_filename` test invariants

darth-raijin:
Strengthen test invariants for `file_stream.rs` by deriving values where possible and replacing magic values with constants.

## Motivation

To improve test quality and reduce brittleness when changing test data, this PR updates the `response_not_set_filename` test to derive values (like `content-length`) from the test payload and to use existing constants (like mime types and headers).

## Solution
- Define `file_content` and `size` once at the top of the test to enable reuse for assertions and usage in the `/file` handler
- Replace hard coded header names, with constants from `http::header`
- Replace hard coded MIME string with constant from `mime`

### DIFF
--- a/axum-extra/src/response/file_stream.rs
+++ b/axum-extra/src/response/file_stream.rs
@@ -305,6 +305,11 @@ mod tests {
     use std::io::Cursor;
     use tokio_util::io::ReaderStream;
     use tower::ServiceExt;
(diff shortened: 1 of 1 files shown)


davidpdrsn:
I don't know... Don't really think this meaningfully improves the test so think I'll close this for now.

PR #3576, opened 2025-12-03
ci: Switch cargo-public-api-crates to cargo-check-external-types

tottoto:
## Motivation

https://github.com/tower-rs/tower-http/pull/613#issuecomment-3605940637

## Solution

Switches `cargo-public-api-crates` to `cargo-check-external-types`.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -61,22 +61,22 @@ jobs:
     - name: cargo hack check
       run: cargo hack check --each-feature --no-dev-deps --all
 
-  cargo-public-api-crates:
+  external-types:
     runs-on: ubuntu-24.04
-    strategy:
-      matrix:
-        crate: [axum, axum-core, axum-extra, axum-macros]
     steps:
     - uses: actions/checkout@v6
-    - uses: dtolnay/rust-toolchain@nightly
+    - uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: nightly-2025-08-06
+    - name: Install cargo-check-external-types
+      uses: taiki-e/cache-cargo-install-action@v2
+      with:
+        tool: cargo-check-external-types@0.3.0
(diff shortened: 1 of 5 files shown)


.github/workflows/CI.yml
```diff
@@ -61,22 +61,22 @@ jobs:
...
     runs-on: ubuntu-24.04
-    strategy:
-      matrix:
-        crate: [axum, axum-core, axum-extra, axum-macros]
     steps:
     - uses: actions/checkout@v6
-    - uses: dtolnay/rust-toolchain@nightly
+    - uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: nightly-2025-08-06
```
jplatte, line 70: I'm fine with pinning nightly, but does newer not work?

axum/src/lib.rs
```diff
@@ -455,6 +455,7 @@ pub mod serve;
 
 #[cfg(any(test, feature = "__private"))]
 #[allow(missing_docs, missing_debug_implementations, clippy::print_stdout)]
```
jplatte: I wonder if the `missing_docs` lint also applies inside `doc(hidden)`, or we can remove it.

jplatte (approved):
Thanks!

axum/src/lib.rs
tottoto: This seems to be required.

https://github.com/tokio-rs/axum/actions/runs/19897287499/job/57030931349

.github/workflows/CI.yml
tottoto: This is the documented known working version, though newer nightly versions might work too.

https://github.com/awslabs/cargo-check-external-types/releases/tag/v0.3.0

jplatte:
Oh, I didn't notice that we had required CI checks that no longer exist with this. I removed those now, wanna merge?

PR #3585, opened 2025-12-19
chore: Remove shuttle from docs

joshua-mo-143:
Hi, unfortunately Shuttle is stopping their deployment platform which means any guides written with a Shuttle deployment in mind are mostly pointless to have now (hence me opening this PR).

Context: https://docs.shuttle.dev/docs/shuttle-shutdown

## Motivation
Motivation is as above. 

## Solution
Just deleting the required links.

I am hoping to salvage most if not all of the articles I wrote for them and re-purpose them somehow, but not really sure where since there are other, better guides now. Probably. People are also using LLMs to get their answers which means that nobody's really reading them anymore.


joshua-mo-143:
> Also, what a shitshow. Announcing shutdown right before Christmas and leaving users just weeks to migrate. All that without even mentioning _why_ it's shutting down.

The reason why they are shutting down is that they are pivoting to making an AI devops agent called Neptune. They made the announcement on Discord a couple of weeks ago or so stating that they'll still be around and then yesterday shipped the email announcement that the Shuttle platform is going to cease operations.

EDIT: Posted is the Discord announcement. EDIT2: Just realised I accidentally cropped the date out, but this was posted on the 4th of December.

image

Actual link to the Discord announcement (you'll probably need to join the Shuttle server to see this): https://discord.com/channels/803236282088161321/1029328917292666931/1446191190754988102

PR #3594, opened 2025-12-25
axum: Use serde_html_form for Query and Form

### LINKED ISSUE #3075: Move axum-extra's Query type to axum

jplatte:
Closes #3075.


Turbo87 (approved):
I'm wondering if it would be better to release this in two stages. first update axum-extra to use the new serde_html_form release to have people try it out. and then later move it into axum itself once it's proven that everything works as intended.

if you prefer to release it as proposed here I won't block it though :)

jplatte:
I would probably do that if I wasn't so sure that virtually nobody would go and try it 😅

I think I'd rather make a pre-release of axum 0.9 that people can try ahead of time, that seems more likely to get people's attention.

PR #3597, opened 2025-12-27
Add customizable WebSocket subprotocol selection

Lethe10137:
## Motivation

Token-encoded subprotocol is a common practice for authentication in websocket, and though not a standard usage, it is even used by [kubernetes,](https://github.com/kubernetes/kubernetes/commit/714f97d7baf4975ad3aa47735a868a81a984d1f0)

However, current implementation of axum does not support read such ws subprotocol.

## Solution
This adds a getter for the Sec-WebSocket-Protocol requested by the
client and a setter for the chosen subprotocol.

This enables more flexible selection logic (e.g. Token-encoded subprotocols or runtime selection) and avoids unnecessary allocations present in the existing protocols() helper.

And,  in the use case of [k8s](https://github.com/kubernetes/kubernetes/commit/714f97d7baf4975ad3aa47735a868a81a984d1f0), where the client init the ws connection as:

```js
var ws = new WebSocket(
  "wss://<server>/api/v1/namespaces/myns/pods/mypod/logs?follow=true",
  [
    "base64url.bearer.authorization.k8s.io.bXl0b2tlbg",
    "base64.binary.k8s.io"
  ]
);
```

 example usage could be:
```rust
async fn ws_handler(
    mut ws: WebSocketUpgrade,
    ConnectInfo(addr): ConnectInfo<SocketAddr>,
) -> impl IntoResponse {
    let token = ws
        .requested_protocols()
        .find_map(|protocol| protocol.strip_prefix("base64url.bearer.authorization.k8s.io."));

    if token.is_some_and(|token| token == "bXl0b2tl3bg") {
        ws.set_selected_protocol(HeaderValue::from_static("base64.binary.k8s.io"));
        ws.on_upgrade(move |socket| handle_socket(socket, addr))
    } else {
        // Reject the connection
        (axum::http::StatusCode::UNAUTHORIZED, "Unauthorized").into_response()
    }
}
```

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -265,14 +265,16 @@ impl<F> WebSocketUpgrade<F> {
         self
     }
 
-    /// Return the WebSocket subprotocols requested by the client as an iterator of &str.
+    /// Return the WebSocket subprotocols requested by the client.
     ///
     /// # Examples
     ///
     /// If the client sends the following HTTP header in the WebSocket upgrade request:
+    ///
     /// ```txt
     /// Sec-WebSocket-Protocol: soap, wamp
     /// ```
+    ///
     /// this method returns an iterator yielding `"soap"` and `"wamp"`.
     pub fn requested_protocols(&self) -> impl Iterator<Item = &str> {
         self.sec_websocket_protocol


axum/src/extract/ws.rs
```diff
@@ -265,9 +265,117 @@ impl<F> WebSocketUpgrade<F> {
...
+    ///
+    /// let app = Router::new().route("/ws", any(handler));
+    ///
+    /// async fn handler(ws: WebSocketUpgrade) -> Response {
+    ///     ws.select_protocol(|proto| {
+    ///             (proto == "my-protocol")
+    ///                 .then(|| HeaderValue::from_static("my-protocol"))
+    ///         }).on_upgrade(|socket| async {
+    ///             // ...
+    ///         })
```
jplatte:
```suggestion
    ///     ws.select_protocol(|proto| {
    ///         (proto == "my-protocol")
    ///             .then(|| HeaderValue::from_static("my-protocol"))
    ///     }).on_upgrade(|socket| async {
    ///         // ...
    ///     })
```

axum/src/extract/ws.rs
```diff
@@ -265,9 +265,117 @@ impl<F> WebSocketUpgrade<F> {
...
+    /// ```
+    ///
+    /// # Notes
+    ///
+    /// - The predicate is evaluated lazily and stops at the first accepted protocol.
+    /// - The selected protocol is echoed back in the WebSocket upgrade
+    ///   response as required by RFC 6455. Some browsers may reject a
+    ///   value that was not present in the client's request.
+    /// - No allocation is performed unless the predicate constructs a dynamic
+    ///   `HeaderValue`.
```
jplatte: Why is this here? I think we generally shouldn't make any promises about allocations.

jplatte:
Extra reviews appreciated, because I don't know much about ws, especially this subprotocol stuff. cc @SabrinaJewson who has reviewed a lot of WS + SSE stuff before. Also @Turbo87 maybe?

axum/src/extract/ws.rs
```diff
@@ -265,9 +265,117 @@ impl<F> WebSocketUpgrade<F> {
...
+    ///             (proto == "my-protocol")
+    ///                 .then(|| HeaderValue::from_static("my-protocol"))
+    ///         }).on_upgrade(|socket| async {
+    ///             // ...
+    ///         })
+    /// }
+    /// # let _: Router = app;
+    /// ```
+    ///
+    /// ## Accept token based auth
```
SabrinaJewson: Very minor grammar nit: should be “token-based” :smile:

axum/src/extract/ws.rs
```diff
@@ -265,9 +265,117 @@ impl<F> WebSocketUpgrade<F> {
...
+    /// let app = Router::new().route("/ws", any(handler));
+    /// async fn handler(ws: WebSocketUpgrade) -> impl IntoResponse {
+    ///     let ws = ws.select_protocol(|proto| {
+    ///         proto
+    ///             .strip_prefix("token.subprotocol.bearer.")
+    ///             .map(|token| {
+    ///                 (token == "bXl0b2tlbgac7G") // replace with real authenticate logic
+    ///                     .then_some(HeaderValue::from_static("token.subprotocol.auth"))
+    ///             })
+    ///             .flatten()
```
SabrinaJewson: Again very minor, but could potentially be replaced with:

```rs
let token = proto.strip_prefix("token.subprotocol.bearer.")?;
(token == "bXl0b2tlbgac7G")
    .then_some(HeaderValue::from_static("token.subprotocol.auth"))
```

SabrinaJewson:
Hmmm. I can’t help but wonder if we should just offer an explicit iterator over the request’s protocols together with a setter for the chosen protocol header, in case for example the token authentication logic is async or fallible? e.g.

```rs
pub fn requested_protocols(&self) -> impl Iterator<Item = &str> {
    self.sec_websocket_protocol
        .as_ref()
        .and_then(|p| p.to_str().ok())
        .into_iter()
        .flat_map(|s| s.split(','))
        .map(|s| s.trim())
}
pub fn set_protocol(&mut self, protocol: HeaderValue) {
    self.protocol = Some(protocol);
}
```

axum/src/extract/ws.rs
Lethe10137: https://github.com/tokio-rs/axum/blob/07f831203864c6e178777511d6cd6f0e1499bfbe/axum/src/extract/ws.rs#L251-L258
Actually,  I tried to respond to the `FIXME` in `protocols()`.  Yet indeed we can just remove these two lines.

Lethe10137:
> Hmmm. I can’t help but wonder if we should just offer an explicit iterator over the request’s protocols together with a setter for the chosen protocol header, in case for example the token authentication logic is async or fallible? e.g.
> 
> ```rust
> pub fn requested_protocols(&self) -> impl Iterator<Item = &str> {
>     self.sec_websocket_protocol
>         .as_ref()
>         .and_then(|p| p.to_str().ok())
>         .into_iter()
>         .flat_map(|s| s.split(','))
>         .map(|s| s.trim())
> }
> pub fn set_protocol(&mut self, protocol: HeaderValue) {
>     self.protocol = Some(protocol);
> }
> ```
Thanks for your time.

Indeed this would be much easier to understand.
I would do the following in the next commit:
- add these two interfaces
- remove the `select_protocol()` I added

Should I close this PR and open a new one, or just force-push on this PR?

jplatte:
Force-push is fine.

Lethe10137:
> Force-push is fine.


I have force-pushed the code, with updated doc comments. And the use example in this PR has been correspondingly updated. Could you please have a review of that?

Lethe10137:
@SabrinaJewson  Could you please have a quick review of this?  Following your insightful opinion, I added a getter and setter (as you wrote in the comment) in the current PR.   These two methods are really useful for a project of mine, 
where neither cookie nor url param is not feasible places to place the auth token for websocket.

SabrinaJewson:
It looks good to me. Minor comment: I know I suggested the name originally, but since the getter is called `selected_protocol`, maybe `set_selected_protocol` is a more appropriate name for the setter?

Lethe10137:
> It looks good to me. Minor comment: I know I suggested the name originally, but since the getter is called `selected_protocol`, maybe `set_selected_protocol` is a more appropriate name for the setter?

Happy new year!   I have changed the name for the setter in the latest commit.

Lethe10137:
@SabrinaJewson  Could you please tell me what further modifications are needed for this PR to be merged?

SabrinaJewson:
Sorry – to be clear I’m not a maintainer, so you’ll have to page @jplatte.

axum/src/extract/ws.rs
```diff
@@ -265,11 +265,43 @@ impl<F> WebSocketUpgrade<F> {
         self
     }
 
+    /// Return the WebSocket subprotocols requested by the client as an iterator of &str.
```
jplatte:
```suggestion
    /// Return the WebSocket subprotocols requested by the client.
```

(no need to restate the return type in words)

axum/src/extract/ws.rs
```diff
@@ -265,11 +265,43 @@ impl<F> WebSocketUpgrade<F> {
...
+    ///
+    /// # Examples
+    ///
+    /// If the client sends the following HTTP header in the WebSocket upgrade request:
+    /// ```txt
+    /// Sec-WebSocket-Protocol: soap, wamp
+    /// ```
+    /// this method returns an iterator yielding `"soap"` and `"wamp"`.
```
jplatte, line 278:
```suggestion
    /// If the client sends the following HTTP header in the WebSocket upgrade request:
    ///
    /// ```txt
    /// Sec-WebSocket-Protocol: soap, wamp
    /// ```
    ///
    /// this method returns an iterator yielding `"soap"` and `"wamp"`.
```

axum/src/extract/ws.rs
```diff
@@ -265,11 +265,43 @@ impl<F> WebSocketUpgrade<F> {
...
+    pub fn requested_protocols(&self) -> impl Iterator<Item = &str> {
+        self.sec_websocket_protocol
+            .as_ref()
+            .and_then(|p| p.to_str().ok())
```
jplatte, line 282: I'm not a big fan of just returning an empty iterator when the header value is non-utf8, but I guess it can be fixed later (seems extremely unlikely to happen in practice unless the client is actively testing edge case behavior).

PR #3603, opened 2025-12-27
Fix `IntoResponse` for tuples overriding error response codes

### LINKED ISSUE #2287: `IntoResponse` implementation for `(StatusCode, R) where R: IntoResponse` overrides error status codes and might be a footgun

jplatte:
Closes #2287.
Continuation of #2468.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -171,6 +171,7 @@ dependencies = [
  "hyper",
  "mime",
  "pin-project-lite",
+ "serde",
  "sync_wrapper",
  "tokio",
  "tower-http",
--- a/axum-core/Cargo.toml
+++ b/axum-core/Cargo.toml
@@ -55,6 +55,7 @@ axum = { path = "../axum", features = ["__private"] }
 axum-extra = { path = "../axum-extra", features = ["typed-header"] }
 axum-macros = { path = "../axum-macros", features = ["__private"] }
 hyper = "1.0.0"
+serde = { version = "1.0.200", features = ["derive"] }
 tokio = { version = "1.25.0", features = ["macros"] }
 tower-http = { version = "0.6.0", features = ["limit"] }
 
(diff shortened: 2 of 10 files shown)


jplatte:
I can generate a coverage report locally, but no clue how to _view_ it.. Don't have time to set up continous coverage tracking right now.

PR #3609, opened 2025-12-30
ci: Make examples cached and checked with rustfmt

tottoto:
Makes `examples` are cached and checked with rustfmt in the check job.


jplatte (approved):
Makes a lot of sense ^^

yanns (approved):

PR #3610, opened 2026-01-01
add a TypedMethod trait/derive similar to TypedPath

Zizico2:
## Motivation

Defining an entire path in the same place, on the same struct (missing query params?) seems useful to me. It also matches what `actix-web`, `rocket` or `poem` macros provide.

## Solution

A `TypedMethod` trait/derive easily allows the method to be co-located with the path. 

Usage:

```rust
#[tokio::main]
async fn main() {
    let app = Router::new().typed(get_user);

    let listener = TcpListener::bind("127.0.0.1:3000")
        .await
        .expect("failed to bind 127.0.0.1:3000");

    println!("Listening on http://127.0.0.1:3000");

    axum::serve(listener, app).await.expect("server error");
}

async fn get_user(UsersMember { id }: UsersMember) -> String {
    format!("User {id} is ready")
}

#[derive(TypedMethod, TypedPath, Deserialize)]
#[typed_method(MethodFilter::GET)]
#[typed_path("/users/{id}")]
struct UsersMember {
    id: u32,
}
```

Would this be considered? It's missing comments and tests if needed.


svenstaro:
For what it's worth, I'd love to have this.

mladedav:
Yes, this seems intriguing and it is something we could consider.

That said, the way I think you're proposing this would mean that people who want e.g. GET and DELETE on the same endpoint would have to either not use this or duplicate the structs.

Of course we can't have one type that allows both method and then call `typed()` on that because there is no way to pick which one to choose. But maybe it could at least create a compilation failure when someone tries to pass a typed method with GET and DELETE into a `typed_put`?

This is me just exploring the options a bit, but we can also just do it your way. I would just like to see our options.

Zizico2:
Actually, that made me realize a flaw in my mental model, so I changed the implementation a bit.

The current mental model is that `TypedMethod` and `TypedPath` are blissfully unaware of each other:

1. If you implement `TypedPath`, you get access to `typed_*(...)`.
2. If you implement `TypedMethod`, you could theoretically get access to `typed_with_path<const PATH: &str>(...)`.
3. If you implement both, you get access to `typed(...)`, but you do not lose access to 1 or 2.

For your use case, you can choose a `TypedPath` type - say, `SharedType` - with `typed_put` and `typed_get`. Or you can also wrap it in `GetSharedType(SharedType)` and `PutSharedType(SharedType)` and implement `TypedMethod` and `TypedPath` on those wrappers.

And yes, you could pass a type with `#[typed_method(MethodFilter::GET)]` to `typed_put`, just as you can pass a type that implements Serde's `Serialize` or `Deserialize` to `typed_put`. `typed_put` only cares about `TypedPath`.

To make the newtype pattern more convenient, I added `Get<T>`, `Put<T>`, `Post<T>`, and related wrappers. They implement `TypedMethod` unconditionally, and `TypedPath` when `T` implements it.

I also added tests (generated with Codex).

I opened a follow-up PR, [#3859](https://github.com/tokio-rs/axum/pull/3859), which I think should probably supersede this one. It takes a stricter approach: it disallows the exact case of feeding a `Get<_>` into `typed_put`, but introduces a small breaking change.

PR #3613, opened 2026-01-03
ci: Refactor ci style

tottoto:
## Motivation

Aligns the CI yaml style across workflows.

## Solution

Alines the followings.

- The Indent style.
- The blank lines and description.


jplatte:
I would prefer standardizing on the indent style with dashes being more indented than their parent key.

tottoto:
Updated to the indent style.

PR #3615, opened 2026-01-05
docs: document that `vpath` and `json` macros are feature-gated

matheus-consoli:
`vpath` is missing the documentation that it is only available with `-F routing`, making it not obvious when users are reading its documentation from docs.rs

## Motivation

No warning about feature gating (screenshot from docs.rs)

image

## Solution

Add `#[cfg_attr(docsrs, doc(cfg(feature = "routing")))]` to the macro

image


matheus-consoli:
I just realised that `axum-extra::json` is missing the docs too

matheus-consoli:
I found https://github.com/rust-lang/rust/issues/100916

PR #3616, opened 2026-01-05
Stop calling axum a 'framework'

jplatte:
It's not very opinionated or batteries-included like Ruby on Rails, Django or Spring, nor does it intend to be. Because of that, I think the term 'framework' is not fitting.

Inspired by <https://status.pointless.one/@pointlessone/115804828862541092>.


Turbo87 (commented):
I guess it depends a bit on the individual definition of "framework". Rocket probably comes closer to being a proper framework, but I'm not sure axum is that far off. I think I'd keep it as is, even if just for SEO purposes.

yanns (approved):

jplatte:
I think axum is definitely far off what even Rocket provides / aims to provide. There's so many major thing we delegate to other parts of the ecosystem, like

- DB access
- CSRF protection
- templating
- logging

I'm happy to keep discussing this but with three approvals I think this can be merged for now.

PR #3618, opened 2026-01-06
Improve typed path by disallowing deprecated variable definitions

tcanabrava:
I was staring at my old toy project in axum, and aparently my paths are now deprecated, and hit a runtime assert, instead of a compile assert. 
So, I improved the vpath! macro to make sure we catch those earlier.


tcanabrava:
ping?

yanns:
Could you add a changelog entry: https://github.com/tokio-rs/axum/blob/main/axum-extra/CHANGELOG.md ?

tcanabrava:
sure.

yanns:
Can you resolve the conflicts?

PR #3620, opened 2026-01-07
Rewrite sec-websocket-protocol handling

jplatte:
Follow-up to #3597. cc @Lethe10137 @SabrinaJewson


SabrinaJewson:
It’s a shame about https://github.com/hyperium/http/issues/459, otherwise the parsing could be zero-copy. Otherwise I think this looks good.

PR #3627, opened 2026-01-10
ci: Improve test-msrv job cache

tottoto:
Improves the `test-msrv` job cache.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -125,14 +125,8 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v6
-      - uses: dtolnay/rust-toolchain@master
-        with:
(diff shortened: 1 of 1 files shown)


tottoto:
The `test-msrv` job currently checks both the MSRV build and the minimal dependencies versions build. If it is changed to check only the MSRV build using that approach, a separate job will be needed to verify the build with the minimal dependencies versions, and that job will still require nightly, which is the same approach to the hyper project.

jplatte:
Right, I guess testing min-versions as well is kinda nice. Though the amount of workarounds needed there is a bit ridiculous. How do you feel about testing msrv in the way I proposed, plus min-direct-versions? I'm pretty sure there's a separate nightly-only flag for that.

tottoto:
I think it's reasonable. I'll open a pull request to separate these checks.

PR #3630, opened 2026-01-12
chore: Remove unused package metadata

tottoto:
Removes the unused package metadata.


axum-core/Cargo.toml
```diff
@@ -22,9 +22,6 @@ allowed_external_types = [
     "http_body::*",
 ]
 
-[package.metadata.cargo-machete]
```
jplatte, line 25: Just because this isn't currently used in CI, doesn't mean it's not useful. I think we should add cargo-machete as a job instead.

tottoto: Removed this diff.

PR #3631, opened 2026-01-12
feat: Add full feature

tottoto:
## Motivation

Makes it easy to turn on all features, like `tokio` and `hyper`.

## Solution

Adds `full` feature, which enables all features.

### DIFF
--- a/axum-core/Cargo.toml
+++ b/axum-core/Cargo.toml
@@ -29,6 +29,7 @@ ignored = ["tower-http"] # See __private_docs feature
 all-features = true
 
 [features]
+full = ["tracing"]
 tracing = ["dep:tracing"]
 
 # Required for intra-doc links to resolve correctly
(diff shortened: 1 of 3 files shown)


jplatte:
I'm not a big fan to be honest. What motivated this PR?

tottoto:
This allows users to easily enable all features and start development.
As an additional benefit, I think it makes it easier in CI to test jobs that should not enable private features, instead of relying on `--all-features`.

jplatte:
I understand the general idea behind this Cargo feature, but what made you open this PR? Was there a project where you actually wanted to use this yourself? Did you start writing some tutorial or similar material where you wanted to simplify things? Something else?

tottoto:
When testing, I sometimes create new environments and try out several features. In such cases, I think it would be convenient to have a way to enable multiple features at once, instead of looking up the required features one by one and enabling them individually, or maintaining a full list of all features and copying it each time.

davidpdrsn:
I don't know :/ Don't see this as being super valuable. axum has so many disjoint features that add new dependencies and I think its rare that you _actually_ want to enable all the features. I don't think that is the case for tokio so `full` makes more sense there.

So I'm inclined to not add this for now.

PR #3633, opened 2026-01-12
ci: Use msrv aware dependency resolver

tottoto:
Uses the MSRV aware dependency resolver.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -142,6 +142,29 @@ jobs:
       - name: Run tests
         run: cargo +${{ env.MSRV }} test --locked --all-features
 
+  check-minimal-versions:
+    needs: check
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


mladedav:
I always though that we also tried to check that the minimum versions work (e.g. we're not using API from newer versions). Do we want to add another test for that or do we care just about MSRV? I'm honestly not sure if it's worth it so we can maybe merge as is.

tottoto:
I think the minimal dependency version should be checked. I would think it would be worth maintaining, if it can be implemented more simply.

mladedav:
To have both, we can either keep it as is or have two different jobs, one for msrv (what you did here) and one for minimum dependencies (either what this was before or the same with newer toolchain or even just nightly since we need nightly cargo anyway). So two jobs I guess?

tottoto:
> So two jobs I guess?

Yes. I think so.

And, as an additional benefit of separating them, the effects can be removed, which are caused by dev-dependencies and private features. On the other hand, it looks more cumbersome since we need to specify all features instead of `--all-features` option.

PR #3636, opened 2026-01-21
axum-extra: remove unused extract::rejection module

martinetd:
previous commit removed extract/host which was its sole user, rejection can go as well.

Fixes: e3b32f48b5bc ("Remove deprecated extractors from axum-extra (#3599)")

## Motivation

Follow up of https://github.com/tokio-rs/axum/pull/3599

Rejection is pub but HostRejection::FailedToResolveHost is not, so the enum could not actually be used outside of axum-extra either, and it's safe to remove
(I noticed copying the old extract/host.rs out and making it work)


jplatte:
Can you add a changelog entry? Maybe as a sub-item of the other one.

martinetd:
> Can you add a changelog entry? Maybe as a sub-item of the other one.

Given the other one wasn't released yet I considered that part of the above, but happy to comply -- done!

PR #3647, opened 2026-02-06
docs: update deprecated TimeoutLayer::new in graceful-shutdown example

ghost:
Hello, this is my first Pull Request to a major library/framework, so I am very open to any feedback or changes you might suggest!

## Motivation

I am updating the graceful-shutdown example to replace the deprecated `TimeoutLayer::new()` method with `TimeoutLayer::with_status_code()`. Currently, when users run or check this example, the compiler generates a deprecation warning from tower-http.

The main problem is "developer experience" (DX). New users like myself often look to the examples as a source of truth for best practices. Seeing deprecation warnings in an official example can be confusing and suggests the example might be out of date.

## Solution

Modified examples/graceful-shutdown/src/main.rs to use `TimeoutLayer::with_status_code(StatusCode, Duration)`.
Added the necessary axum::http::StatusCode import.
I have verified the change with cargo check and ensured formatting is correct with cargo fmt.

### DIFF
--- a/examples/key-value-store/src/main.rs
+++ b/examples/key-value-store/src/main.rs
@@ -123,6 +123,7 @@ fn admin_routes() -> Router<SharedState> {
         state.write().unwrap().db.remove(&key);
     }
 
+    #[allow(deprecated)] // FIXME
     Router::new()
         .route("/keys", delete(delete_all_keys))
         .route("/key/{key}", delete(remove_key))


ghost:
Updated tower-http to 0.6.8 to make the changes work, yet because `auth::require_authorization` is deprecated in 0.6.7 the check failed. I tried reverting to 0.6.6, but then my primary changes failed because the method is missing in that version. I'm eager to help fix the other examples to satisfy the CI, but as this is my first PR, I'm a bit unsure of the best way to handle these deprecations. I also noticed it automatically updated some of the windows-sys packages.

Since this is my first PR, I'm still learning how to manage dependencies across multiple examples. I’d love to fix these warnings to get the CI passing—would you recommend I update the other examples as well, or is there a better way to handle this?

Edit again:

Apologies, I just realized I did an amend and force-push earlier to fix a typo in the commit message. I’ve just re-read the contributing guide about rebase and realized I should have avoided that once the PR is open. It won't happen again.

jplatte:
I have a thing locally for replacing some other deprecated calls. Though if you want you can add `#[allow(deprecated)] // FIXME` to those places in this PR and we'll handle that separately.

ghost:
Thanks for the guidance, @jplatte. I’ve added the #[allow(deprecated)] // FIXME to the key-value-store example as you suggested to clear the CI warnings.

PR #3650, opened 2026-02-09
Restrict Mutlipart::next_field's Field lifetime

Niedzwiedzw:
Restricting it (making it shorter) allows for more operations on the field itself, for example returning the Field from a function which holds a &mut Multipart reference. I don't think this is a breaking change, as it actually enables more legal code to compile.


Niedzwiedzw:
Actually this solves nothing. sorry about the noise

Issue #3641, opened 2026-01-30
Remove unwrap() in Router::fallback_endpoint by using a take-once service wrapper


next-n:
-[yes] I have looked for existing issues (including closed) about this
Summary

Router::fallback_endpoint currently relies on an Option + unwrap() pattern inside service_fn to ensure the fallback route is only consumed once. This assumes the service is never called more than once; if that invariant is violated, it panics in the request path.

Location

router/mod.rs — Router::fallback_endpoint (see existing TODO: “get rid of the unwraps”).

Problem (conceptual)

The fallback service is wrapped in Option and consumed via service.take().unwrap() inside a FnMut.
If the service is ever called more than once (due to misuse, future refactors, or edge cases), this results in a panic instead of safe handling.

Expected

The fallback route is consumed once, and any unexpected reuse is handled safely (no panic in prod).

Actual

A second call triggers unwrap() and panics.

Proposed approach

Replace the Option + unwrap() pattern with a small internal “take-once” service wrapper that:

consumes the route on first call (preserving the no-extra-clone optimization),

safely handles subsequent calls instead of panicking.

Note

There is already a TODO in the code about removing the unwraps. I’d like to implement this refactor along with a small test if that’s acceptable.

yanns:
Thanks for the issue. You can of course implement it and open a PR.

next-n:
PR incoming.

PR #3652, opened 2026-02-09
routing: avoid unwrap in fallback path

next-n:
Fixes https://github.com/tokio-rs/axum/issues/3641

Removed a panic path from fallback request handling.

Before, axum assumed the internal fallback service would only be called once and enforced that with unwrap(), which could panic if the invariant was violated.

Now, the invariant is still assumed, but a violation no longer panics on the request path: it returns a normal 500 Internal Server Error response. In debug builds it triggers a debug_assert!

### DIFF
--- a/axum/src/routing/mod.rs
+++ b/axum/src/routing/mod.rs
@@ -57,6 +57,21 @@ macro_rules! panic_on_err {
     };
 }
 
+const TAKE_ONCE_ROUTE_PANIC_MSG: &str =
+    "TakeOnceRoute called more than once; if this was not triggered by an intentional test, this should never happen. Please file an issue.";
+
+fn take_route_or_internal_error(service: &mut Option<Route>) -> Route {
+    service.take().unwrap_or_else(|| {
+        if cfg!(debug_assertions) {
+            panic!("{TAKE_ONCE_ROUTE_PANIC_MSG}");
+        }
+
+        Route::new(service_fn(|_req: Request| async move {
+            Ok::<_, Infallible>(http::StatusCode::INTERNAL_SERVER_ERROR.into_response())
+        }))
+    })
+}
+
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub(crate) struct RouteId(usize);
 
@@ -392,18 +407,7 @@ where
                                 move |mut request: Request| {
                                     #[cfg(feature = "matched-path")]
                                     request.extensions_mut().remove::<MatchedPath>();
-                                    let route = service.take().unwrap_or_else(|| {
-                                        debug_assert!(
-                                            false,
-                                            "TakeOnceRoute called more than once; if this was not triggered by an intentional test, this should never happen. Please file an issue.",
-                                        );
-                                        Route::new(service_fn(|_req: Request| async move {
-                                            Ok::<_, Infallible>(
-                                                http::StatusCode::INTERNAL_SERVER_ERROR
-                                                    .into_response(),
-                                            )
-                                        }))
-                                    });
+                                    let route = take_route_or_internal_error(&mut service);
                                     route.oneshot_inner_owned(request)
                                 }
                             )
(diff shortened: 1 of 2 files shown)


axum/src/routing/mod.rs
```diff
@@ -57,6 +58,59 @@ macro_rules! panic_on_err {
...
+        &mut self,
+        #[cfg_attr(not(feature = "matched-path"), allow(unused_mut))] mut req: Request,
+    ) -> Self::Future {
+        #[cfg(feature = "matched-path")]
+        req.extensions_mut().remove::<MatchedPath>();
+
+        let route = if let Some(route) = self.route.take() {
+            route
+        } else {
+            #[cfg(not(test))]
```
mladedav: Let the debug assert also work in tests if we happen to break this, it would be better to already have the information instead of a generic 500. If you want to test the behavior, add a test that expects a panic with the assert text instead.

axum/src/routing/mod.rs
```diff
@@ -57,6 +58,59 @@ macro_rules! panic_on_err {
...
+            debug_assert!(
+                false,
+                "TakeOnceRoute called more than once; returning an internal error route",
```
mladedav: If we hit the assert, we're not returning anything so this is slightly false.

Just drop the second part or maybe add something like "this should never happen" or "file a bug". There are a few other panics like these around but I don't think we're really consistent with the wording.

axum/src/routing/mod.rs
next-n: I changed the test to two parts.

Debug test (take_once_route_panics_on_second_call) is marked #[should_panic].
It passes only if panic happens.
- cargo test -p axum take_once_route_panics_on_second_call -- --nocapture


Release test (take_once_route_returns_internal_error_on_second_call) expects 500 and no panic.
It passes only if that response behavior happens.

 - cargo test -p axum --release take_once_route_returns_internal_error_on_second_call -- --nocapture

axum/src/routing/mod.rs
next-n: I corrected by changing the assert text to invariant/bug wording:

"TakeOnceRoute called more than once; this is a bug in axum. Please file an issue."

That avoids the inaccurate “returning internal error route” phrasing on the assert path.

axum/src/routing/mod.rs
mladedav: Frankly, I wouldn't include the test in release mode. I don't think we compile those and compile the whole project again in a release build just to run a single test seems wasteful, especially to check code that we want (and assume) to be unreachable.

jplatte:
I haven't yet looked at this in detail, but I'm wondering.. If we still end up with a runtime "assertion", couldn't we just update the current unwrap to be a 500 in release mode directly? Does all the refactoring buy us anything?

axum/src/routing/mod.rs
next-n: I did keep only debug panic test. please check my update code.

next-n:
@jplatte, I kept the existing service.take() pattern in fallback_endpoint. On second call, it falls back to 500 INTERNAL_SERVER_ERROR. In debug, it hits debug_assert! (panic). I also add test for this too.

axum/src/routing/mod.rs
```diff
@@ -392,7 +392,19 @@ where
                                 move |mut request: Request| {
                                     #[cfg(feature = "matched-path")]
                                     request.extensions_mut().remove::<MatchedPath>();
-                                    service.take().unwrap().oneshot_inner_owned(request)
+                                    let route = service.take().unwrap_or_else(|| {
+                                        debug_assert!(
+                                            false,
+                                            "TakeOnceRoute called more than once; if this was not triggered by an intentional test, this should never happen. Please file an issue.",
+                                        );
+                                        Route::new(service_fn(|_req: Request| async move {
```
jplatte: I think `if cfg!(debug_assertions) { panic!("..."); }` would be cleaner.

axum/src/routing/mod.rs
```diff
@@ -411,7 +423,19 @@ where
                                 move |mut request: Request| {
                                     #[cfg(feature = "matched-path")]
                                     request.extensions_mut().remove::<MatchedPath>();
-                                    service.take().unwrap().oneshot_inner_owned(request)
+                                    let route = service.take().unwrap_or_else(|| {
+                                        debug_assert!(
+                                            false,
+                                            "TakeOnceRoute called more than once; if this was not triggered by an intentional test, this should never happen. Please file an issue.",
+                                        );
```
jplatte: Same here.

jplatte: Actually, can you extract out the `service.take().unwrap_or_else(...)` bit into a separate `fn`? Then you can also test that function in the `tests` module, rather than have a test that doesn't really test anything from the module.

axum/src/routing/mod.rs
```diff
@@ -821,3 +845,51 @@ fn traits() {
     assert_send::<RouterIntoService<Body, ()>>();
     assert_sync::<RouterIntoService<Body, ()>>();
 }
+
+#[cfg(all(test, feature = "tokio", debug_assertions))]
+mod take_once_route_tests {
+    use super::*;
```
jplatte: Could you please expand this (only import what's actually used in the tests module)?

jplatte:
Looked at the old diff now, it's actually less stuff than I thought. Still, I think a simple function will be better rather than `TakeOnceRoute` where the new code was spread out over two inherent functions + one trait impl.

next-n:
@jplatte I updated the new commit, can you review it?

PR #3653, opened 2026-02-09
examples: Refactor Request and Response construct

tottoto:
Refactors `Request` and `Response` construct.

### DIFF
--- a/examples/anyhow-error-response/src/main.rs
+++ b/examples/anyhow-error-response/src/main.rs
@@ -70,7 +70,7 @@ mod tests {
     #[tokio::test]
     async fn test_main_page() {
         let response = app()
-            .oneshot(Request::builder().uri("/").body(Body::empty()).unwrap())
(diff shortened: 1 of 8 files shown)


examples/anyhow-error-response/src/main.rs
```diff
@@ -69,10 +69,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_main_page() {
-        let response = app()
-            .oneshot(Request::builder().uri("/").body(Body::empty()).unwrap())
-            .await
-            .unwrap();
+        let response = app().oneshot(Request::new(Body::empty())).await.unwrap();
```
yanns: I'd personally prefer staying explicit about the uri `"/"` rather than relying on default.

jplatte:
Agree with @yanns' comment. Otherwise LGTM.

examples/anyhow-error-response/src/main.rs
tottoto: Updated.

Issue #3282, opened 2025-03-22
Cancellation safety for WebSocket's async methods is not documented


ProgramCrafter:
Docs don't say if the following code is correct or may produce unexpected results:

```
  let update_receiver: tokio::sync::broadcast::Receiver;
  let sock: axum::extract::ws::WebSocket;
  
  let (mut ws_write, mut ws_read) = sock.split();

  tokio::select! {
    // Cancel-safe. https://docs.rs/tokio/1.43.0/tokio/sync/broadcast/struct.Receiver.html#cancel-safety
    clientbound = update_receiver.recv() => {...}
    // ??? https://docs.rs/axum/0.8.1/axum/extract/ws/struct.WebSocket.html#method.recv
    serverbound = ws_read.recv() => {...}
  }
```

wyf027:
Opened # documenting that `WebSocket::recv` is cancel-safe (via `StreamExt::next`), including `split` + `tokio::select!` guidance. Builds on the #3657 discussion.

wyf027:
Combined documentation PR: #3767 (supersedes #3765).

vilgotf:
Fixed by #3786

mladedav:
Yes, thanks again!

PR #3657, opened 2026-02-14
docs: document cancellation safety for WebSocket::recv

ghost:
## Motivation

Wanting to help with #3282

## Solution

I have updated the documentation for WebSocket::recv in axum/src/extract/ws.rs to:
- Explicitly state that the method is not cancel-safe.
- Explain the consequence (potential message loss).
- Provide the recommended alternative (using the Stream implementation directly with pinning).

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -551,7 +551,7 @@ impl WebSocket {
     /// Returns `None` if the stream has closed.
     /// 
     /// # Cancel safety
-    ///
+    ///cargo fmt --all --check --manifest-path ./Cargo.toml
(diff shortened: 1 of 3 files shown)


vilgotf:
But `recv` _is_ cancel-safe (transitively via `StreamExt::next`)

ghost:
Oh sh*t you're right. i feel stupid now, sorry but also thanks for pointing it out! Closing this. I'll be more careful and read the docs and code more closely in the future. Really thanks a lot 🙏🏼

davidpdrsn:
@vilgotf would you be up for submitting a PR that documents this?

Issue #3658, opened 2026-02-15
Warning of `empty_enums = "warn" in Cargo.toml` when cargo clippy


next-n:
- [yes] I have looked for existing issues (including closed) about this

###Problem

When running cargo clippy -p axum --lib --tests -- -D warnings

I got the following 

Blocking waiting for file lock on build directory
warning[E0602]: unknown lint: `clippy::empty_enums`
  |
  = help: did you mean: `clippy::empty_enum`
  = note: requested on the command line with `-W clippy::empty_enums`
  = note: `#[warn(unknown_lints)]` on by default

In Cargo.toml, need to change from empty_enums = "warn" to empty_enum = "warn"

###Expected

After changing to empty_enum = "warn".

It removes this warning and does actual enforcement of finding possible cases like enum Never {}.

next-n:
PR incoming

jplatte:
Turns out `empty_enums` is actually the current name and `empty_enum` is the previous one. The warning only happens on older toolchains.

next-n:
oh yes, you are right. Sorry for confusion about that.

PR #3642, opened 2026-02-02
fix: empty_enum renamed to empty_enums

canmi21:
## Motivation

```text
  ⎿  warning: lint `clippy::empty_enum` has been renamed to `clippy::empty_enums`                                               
       |                                                                                                                        
       = help: use the new name `clippy::empty_enums`  
```

## Solution

empty_enum -> empty_enums

PR #3659, opened 2026-02-15
chore: rename clippy empty_enums lint to empty_enum

next-n:
Fixes https://github.com/tokio-rs/axum/issues/3658

I did change into 

- empty_enum = "warn" in Cargo.toml

So now 

cargo clippy -p axum --lib --tests -- -D warnings complete with no warning of unknown lint

like this

warning[E0602]: unknown lint: clippy::empty_enums

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ type_complexity = "allow"
 
 await_holding_lock = "warn"
 dbg_macro = "warn"
-empty_enums = "warn"
+empty_enum = "warn"
 enum_glob_use = "warn"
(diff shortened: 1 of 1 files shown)


sirreidlos:
Am I missing something, or isn't this just reverting https://github.com/tokio-rs/axum/pull/3642? Plus the check job for this commit shows the warning again, so the unknown lint warning probably comes from an older version.

jplatte:
Right, sorry about the noise.. Force-pushed away the latest commit (to keep the history somewhat clean).

Thanks for catching this.

PR #3662, opened 2026-02-16
directly use inner service

yanns:
In `FromFn` middleware, directly use `inner` service.

I'm wondering why we need this `std::mem::replace`.

### DIFF
--- a/axum/src/middleware/from_fn.rs
+++ b/axum/src/middleware/from_fn.rs
@@ -281,8 +281,7 @@ macro_rules! impl_service {
             }
 
             fn call(&mut self, req: Request) -> Self::Future {
-                let not_ready_inner = self.inner.clone();
-                let ready_inner = std::mem::replace(&mut self.inner, not_ready_inner);
+                let inner = self.inner.clone();
 
                 let mut f = self.f.clone();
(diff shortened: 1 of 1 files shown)


yanns:
It's a pattern used in all tower `Service`, and I'd like to understand why this is necessary.

yanns:
I've also asked on [discord](https://discord.com/channels/500028886025895936/627651185406509067/1472902981597921334)

mladedav:
There's an answer on discord now. Risking that I'm really just repeating the answer there, the `clone` may not clone the readiness state, e.g. something like this:
```rust
struct PermitService {
    permit: Option<OwnedPermit>,
}

impl Service for PermitService {
    fn poll_service(...) -> Poll<()> {
        let permit = ready!(get_permit(cx));
        self.permit = Some(permit);
        Poll::Ready(())
    }

    fn call(...) {
        let permit = self.permit.take().expect("ups");
        ...
    }
}

impl Clone for PermitService {
    fn clone(&self) -> Self {
        Self { permit: None }
    }
}

```

yanns:
Thanks a lot for the help & clarification!

PR #3664, opened 2026-02-17
fix(axum-extra): escape filename in Content-Disposition header

### LINKED ISSUE #3646: axum-extra: Content-Disposition parameter injection via unescaped filename in Attachment/FileStream
First, thank you for your continued work on axum and axum-extra.

### Version

- `axum v0.8.8`
- `axum-core v0.5.5`
- `axum-extra v0.12.3`

### Platform

`Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:30 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6020 arm64`

### Crates

- `axum-extra` (`response::Attachment`)
- possibly similar behavior in `axum-extra::response::file_stream::FileStream`

### Description

Thanks for maintaining axum/axum-extra.

I may be misunderstanding the intended contract of `Attachment::filename(...)`, but I found behavior that looks surprising from a header-safety perspective.

I tried this code:

```rust
use axum::{extract::Path, routing::get, Router};
use axum_extra::response::Attachment;

async fn download(Path(name): Path<String>) -> Attachment<&'static str> {
    Attachment::new("header-check path: /download\n").filename(name)
}

let app = Router::new().route("/download/{name}", get(download));
```

Request:
`curl -i "http://127.0.0.1:4000/download/evil%22%3B%20filename*%3DUTF-8''pwned.txt%3B%20x%3D%22"`

Observed:
`content-disposition: attachment; filename="evil"; filename*=UTF-8''pwned.txt; x=""`

Image

Image

Image

Expected:
The input should remain inside filename="..." and should not allow adding extra disposition parameters.
This appears related to direct quoted-string construction in:

- `axum-extra/src/response/attachment.rs:90-92`
  - `let mut bytes = b"attachment; filename=\"".to_vec();`
  - `bytes.extend_from_slice(filename.as_bytes());`
  - `bytes.push(b'\"');`
- `axum-extra/src/response/file_stream.rs:279`
  - `format!("attachment; filename=\"{file_name}\"")`

For context, this looks very similar in class to CVE-2023-29401 (Gin file attachment filename handling):

I might be misunderstanding intended behavior, so please feel free to correct me.

If this behavior is expected, I’d appreciate guidance on the recommended safe usage pattern (and I can open a docs-focused follow-up if helpful).

If you think this should be adjusted in code, I’d be happy to prepare a small PR with a proposed escaping approach and regression tests.

4Mathleu:
## Motivation

`Attachment::filename()` and `FileStream::file_name()` insert the filename directly into a `Content-Disposition` quoted-string without escaping. A crafted filename containing `"` can break out of the quoted-string and inject arbitrary disposition parameters:

This is the same class of vulnerability as CVE-2023-29401 (Gin framework).

## Solution

Escape `\` as `\\` and `"` as `\"` in filenames before inserting them into the `Content-Disposition` header, per RFC 2616 quoted-string rules. This ensures the filename stays contained within the quoted-string boundary.

Changes:
- `axum-extra/src/response/attachment.rs`: byte-level escaping in `into_response()`
- `axum-extra/src/response/file_stream.rs`: string-level escaping in `into_response()`
- Added regression tests for both files

Fixes #3646

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning].
 
 # Unreleased
 
+- **fixed:** Escape backslashes and double quotes in `Content-Disposition` filenames
+  to prevent header parameter injection in `Attachment` and `FileStream` ([#3664])
 - **breaking:** Remove the deprecated `Host`, `Scheme` and `OptionalPath`
   extractors ([#3599])
   - Also remove `HostRejection` which only had `FailedToResolveHost`
@@ -23,6 +25,7 @@ and this project adheres to [Semantic Versioning].
       .route_with_tsr("/path", post(/* handler */))
   ```
 
+[#3664]: https://github.com/tokio-rs/axum/pull/3664
 [#3599]: https://github.com/tokio-rs/axum/pull/3599
 [#3586]: https://github.com/tokio-rs/axum/pull/3586
 


axum-extra/src/response/file_stream.rs
```diff
@@ -274,9 +274,12 @@ where
         let mut resp = Response::builder().header(header::CONTENT_TYPE, "application/octet-stream");
 
         if let Some(file_name) = self.file_name {
+            // Escape backslashes and double quotes in the filename to prevent
+            // Content-Disposition header parameter injection (similar to CVE-2023-29401)
+            let escaped = file_name.replace('\\', "\\\\").replace('"', "\\\"");
```
yanns: Could we extract this into its own struct and do the escaping in the format method? That way, we could re-use this with https://github.com/tokio-rs/axum/pull/3664/changes#diff-440bd42710daa4aa996f51e5e4e6034870aa1a3d12660c99f36199ca7a78376aR93-R97 ?

4Mathleu: Done. Extracted EscapedFilename struct in a new content_disposition module with a Display impl that handles the escaping. Both attachment.rs and file_stream.rs now reuse it.

axum-extra/src/response/file_stream.rs
```diff
@@ -273,13 +273,13 @@ where
     fn into_response(self) -> Response {
         let mut resp = Response::builder().header(header::CONTENT_TYPE, "application/octet-stream");
 
-        if let Some(file_name) = self.file_name {
-            // Escape backslashes and double quotes in the filename to prevent
-            // Content-Disposition header parameter injection (similar to CVE-2023-29401)
-            let escaped = file_name.replace('\\', "\\\\").replace('"', "\\\"");
+        if let Some(ref file_name) = self.file_name {
```
mladedav: Nit, don't use `ref` in the matching pattern unless it's necessary (which it almost never is).

You can just take the value as it was taken until now and take a reference later or explicitly use `.as_str()`. I believe edition 2024 will reject this when we update.

4Mathleu: thanks mate, removed the ref and moved to taking the value by move with an explicit & reference.

mladedav:
One last thing, could you also add one line to the changelog please?

PR #3665, opened 2026-02-18
fix(axum-macros): use OptionalFromRequest for Option<T> fields in derive

### LINKED ISSUE #3623: `FromRequest` derive with `Option<T>` field unexpected semantics

### Version

0.8.8

### Platform

`25.2.0 Darwin Kernel Version 25.2.0: Tue Nov 18 21:07:05 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T6020 arm64`

### Crates

`axum-macros`

### Description

```rust
#[derive(FromRequest)]
#[from_request(rejection(MyError))]
pub struct Args {
    #[from_request(via(Path))]
    something: String,
    #[from_request(via(Json))]
    request: Option<T>
}
```

**What I expect:** for the optional `request` field, I get a `Some(T)` only if a valid JSON body is provided. If it is invalid, it should cause a `JsonRejection` with the `From<JsonRejection> for MyError` impl.

**What is happening:** the expanded version is using `FromRequest` and `.ok()` to convert a `JsonRejection` into `None`.

**What I think should happen:** optional field should require the extractor to implement `OptionalFromRequest`, and use that.

4Mathleu:
**Motivation**

`Option<T>` fields with `#[from_request(via(...))]` were using `.ok()` to convert any rejection into `None`, silently swallowing errors like invalid JSON.

Fixes #3623

**Solution**

Use `OptionalFromRequest` / `OptionalFromRequestParts` traits instead of `.ok()`, letting each extractor decide when to return `None` vs propagate an error.

### DIFF
--- a/axum-macros/CHANGELOG.md
+++ b/axum-macros/CHANGELOG.md
@@ -9,10 +9,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **breaking:** `#[from_request(via(Extractor))]` now uses the extractor's
   rejection type instead of `axum::response::Response` ([#3261])
+- **breaking:** `Option<T>` fields in `#[derive(FromRequest)]` and
+  `#[derive(FromRequestParts)]` now use `OptionalFromRequest` /
+  `OptionalFromRequestParts` instead of calling `.ok()` on the result. This
+  means the extractor decides when to return `None` vs an error, instead of
+  silently converting all rejections to `None`. Extractors used with
+  `Option<T>` must implement `OptionalFromRequest` or
+  `OptionalFromRequestParts`. ([#3623])
 - **fixed:** Fix compilation error when deriving `TypedPath` with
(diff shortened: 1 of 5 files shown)


jplatte:
What's with all the backslashes in the PR description?

4Mathleu:
> What's with all the backslashes in the PR description?

Formatting got mangled when writing the description (also maybe translation shit)

themavik (commented):
Routing `Option<T>` through `OptionalFromRequest` / `OptionalFromRequestParts` fixes the old `.ok()` swallow; `option_via_json.rs` locks in the #3623 behavior. nit: when no custom `rejection` type is set, `map_err` still uses `IntoResponse::into_response` — changelog could mention that for hand-rolled optional extractors.

Samielakkad (commented):
Retracted.

mladedav:
#3760 fixed the same thing.

PR #3678, opened 2026-03-05
relax the version of `matchit`

ADD-SP:
Pinning the `matchit` version in the library crate is inconvenient for downstream. In case of bugfix or new feature for axum's deps, downstream has to bump the version in axum crate instead of their own binary crate.

This PR relaxes the version, so that the axum only requires `matchit >= 0.8.4`. This is not a breaking change as downstream can keep using the `0.8.4`. 

<https://github.com/tokio-rs/axum/pull/3143> is an example shows how inconvenient to pinning the `matchit`.


mladedav:
This was a conscious decision because we are parsing the path in runtime. We need matchit to reject everything we expect it to reject, so we can't allow new matchit features without explicit support in our extractors.

ADD-SP:
> This was a conscious decision because we are parsing the path in runtime. We need matchit to reject everything we expect it to reject, so we can't allow new matchit features without explicit support in our extractors.

Yeah, I realized this after opening this PR, so I closed it, and opened a new one(<https://github.com/tokio-rs/axum/pull/3679>).

PR #3679, opened 2026-03-05
axum: Update matchit to 0.8.6 and support capture prefixes and suffixes

ADD-SP:
Copy-pasting from <https://github.com/tokio-rs/axum/pull/3143> to move forward this work.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -1,0 +1,4 @@
+[default]
+extend-ignore-re = [ 
+    "`([^`]+)`",
+]


.typos.toml
```diff
@@ -0,0 +1,4 @@
+[default]
+extend-ignore-re = [ 
+    "`([^`]+)`",
```
ADD-SP, line 3: This is to suppress the false positive of type checker

jplatte: I think this is much too broad of an exclusion. Just update the docs sample to something else?

ADD-SP: Content wrapped in backticks usually consists of proper nouns and inline code, so I think it’s reasonable to exclude this pattern.

axum/src/docs/routing/route.md
```diff
@@ -38,6 +42,31 @@ regular expression. You must handle that manually in your handlers.
...
+match. If multiple of these were defined in a single router, the topmost one would be used.
+
+- `/abcdef`
+- `/abc{x}ef`
+- `/abc{x}f`
+- `/abc{x}`
+- `/a{x}def`
+- `/a{x}`
+- `/{x}def`
+- `/{x}`
```
mladedav, line 61: This is the reason we didn't merge the 0.8.6 update. The logic is simply too surprising.

If you want to push this forward, can you please update matchit to 0.9 and change the documentation (and possibly the new tests) to align with that version? That version disallows these semi-conflicting routes in a single router.

mladedav:
Closing in favor of #3702

Issue #2634, opened 2024-03-06
Rate Limit Layer not Respected


cloud303-cholden:
- [x] I have looked for existing issues (including closed) about this

### Version

v0.7.4

### Platform

Linux pop-os 6.6.10-76060610-generic

### Description

I am running into an issue with using the `RateLimitLayer` from `tower` with `axum`. I noticed that I can keep sending requests and the rate isn't actually limited. Below is a complete example.

```toml
[package]
name = "axum-rate-limit-test"
version = "0.1.0"
edition = "2021"

[dependencies]
axum = "0.7.4"
serde = { version = "1.0.193", features = ["derive"] }
serde_json = { version = "1.0.108", features = ["std"] }
tokio = { version = "1.35.0", features = ["full"] }
tower = { version = "0.4.13", features = ["limit", "buffer"] }
```

```rust
use axum::{
    error_handling::HandleErrorLayer,
    http::StatusCode,
    routing::get,
    BoxError,
    Router,
    Json,
};
use serde::Serialize;
use tokio::net::TcpListener;
use tower::ServiceBuilder;

#[derive(Debug, Serialize)]
pub struct Health {
    pub ok: bool,
}

pub async fn health() -> Result<Json<Health>, (StatusCode, String)> {
    Ok(Json(Health { ok: true }))
}

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let app = Router::new()
        .route("/health", get(health))
        .route_layer(
            ServiceBuilder::new()
                .layer(HandleErrorLayer::new(|err: BoxError| async move {
                    (
                        StatusCode::INTERNAL_SERVER_ERROR,
                        format!("Unhandled error: {}", err),
                    )
                }))
                .buffer(1024)
                .rate_limit(1, std::time::Duration::from_secs(5))
        );
    let listener = TcpListener::bind("127.0.0.1:3000")
        .await
        .expect("failed to bind TCP listener");
    axum::serve(listener, app)
        .await
        .expect("failed to serve service");

    Ok(())
}
```

I expected the request rate to be limited to 1 per 5 seconds. But the below script executes without any rate limiting.

```bash
for n in {1..10}; do curl localhost:3000/health; done
```

dayvejones:
https://github.com/tokio-rs/axum/discussions/2520

PacoDu:
I've tested this snippet with #2586 and it doesn't seem to fix the issue, the request rate is not limited

[...]

PR #3682, opened 2026-03-08
examples: add rate-limiting example

Sim-hu:
Adds a `rate-limiting` example. Rate limiting comes up a lot for axum and there wasn't an example for it, so this fills that gap.

The example is a small per-IP token-bucket limiter implemented as a `from_fn` middleware over a shared `Arc<Mutex<..>>`. Each client gets a burst allowance that refills over time; once it runs out, requests get `429 Too Many Requests` with a `Retry-After` header.

A note on the approach: I first wrote this with tower's `RateLimitLayer` behind a `BufferLayer`, but as #2634 points out that doesn't actually hold a limit under axum — the state lives inside the service, so each per-connection clone gets its own. I confirmed it with the old version: a configured cap of 5 req/s let 1500 concurrent requests all return 200. Keeping the state in a shared `Arc` is what makes the limit real, so the example does that now and stays dependency-free. The doc comment points at `tower-governor` for production use.

Verified locally: 10 concurrent requests return 5x 200 and 5x 429, and a sustained loop settles at the configured rate.

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -1503,6 +1503,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "example-rate-limiting"
+version = "0.1.0"
+dependencies = [
+ "axum",
+ "tokio",
+ "tower",
+ "tracing",
+ "tracing-subscriber",
+]
+
+[[package]]
 name = "example-readme"
 version = "0.1.0"
 dependencies = [
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -1508,7 +1508,6 @@ version = "0.1.0"
 dependencies = [
  "axum",
  "tokio",
- "tower",
  "tracing",
  "tracing-subscriber",
 ]
--- a/examples/rate-limiting/Cargo.toml
+++ b/examples/rate-limiting/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "example-rate-limiting"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+axum = { path = "../../axum" }
+tokio = { version = "1.0", features = ["full"] }
+tower = { version = "0.5.2", features = ["limit", "buffer", "load-shed"] }
+tracing = "0.1"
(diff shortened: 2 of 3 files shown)


themavik (commented):
`BufferLayer` around `RateLimitLayer` is the standard way to satisfy `Clone`. nit: `handle_error` always maps to 500 — `BufferLayer` full looks like a generic fault; matching `tower::buffer::error::ServiceError` and returning 503 would separate overload from real bugs.

Sim-hu:
Thanks, that's a good call — mapping everything to 500 loses useful signal. I'll update handle_error to match on the error type and return 503 for buffer/rate-limit overload, keeping 500 for unexpected failures.

Sim-hu:
Friendly ping — I've addressed the error handling feedback and the CI is green. Happy to make any further changes if needed\!

davidpdrsn:
Have you seen https://github.com/tokio-rs/axum/issues/2634? I've never used `RateLimitLayer` myself but it seems its not compatible with axum in its current form.

Sim-hu:
Good catch, and you were right to be skeptical. I dug into #2634 and reproduced it with the old version of this example: with a 5 req/s cap configured, 1500 concurrent requests all came back 200 and a sustained loop ran at ~36 req/s. `RateLimitLayer` keeps its state inside the service, so once axum clones the service per connection each clone gets its own limit and the cap never really applies.

So I've rewritten the example to not use `RateLimitLayer` at all. It's now a small `from_fn` middleware with a per-IP token bucket behind a shared `Arc<Mutex<..>>`, so every clone shares one limiter. 10 concurrent requests now give 5x 200 + 5x 429, and sustained traffic settles at the configured rate. No extra dependencies, and the doc comment points to `tower-governor` for production use.

Updated the PR description to match. Let me know if you'd rather not have a rate-limiting example in-tree at all — but if one is welcome, this version actually enforces the limit.

davidpdrsn:
Hmm I'm not sure I like this sorry 😕 I think there are lots of details that one has to think about to get rate limiting right and that makes me think its out of scope for an example.

The right answer for the target audience of this example is probably not to roll their own but use something like https://crates.io/crates/tower_governor.

PR #3683, opened 2026-03-08
docs: consolidate state management docs in crate root

kaze:
## Motivation

The `Router<S> -> Router<()>` transformation and the `FromRef` substate pattern are correct and well-designed, but non-obvious. Three failure modes recur in issues and discussions:

1. Forgetting `.with_state()` - produces a cryptic type mismatch at `.into_make_service()` or `serve()` rather than a clear message.
2. Not knowing `FromRef` exists - users default to one big `Arc<AppState>` and manually field-access in every handler.
3. No single entry point - the `Arc` pattern, `FromRef`, and the `Router<S>` type invariant are each explained somewhere, but there is no place that tells the full story in one read.

## Solution

Add a consolidated subsection under "Sharing state with handlers" in the crate root docs (`axum/src/lib.rs`) covering:

- Why `Arc` is used (state is cloned per request, so `Arc` makes clones cheap)
- Substates via `FromRef` with a worked example
- The `Router<S>` ->` Router<()>` type invariant

Add cross-links back to that section from `State`, `FromRef`, and `Router::with_state` so users can find the full picture from any entry point.

No code changes. All 161 doc tests pass.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -203,7 +203,7 @@ jobs:
       - name: Install cargo-sort
         uses: taiki-e/install-action@v2
         with:
-          tool: cargo-sort@2.0.2
+          tool: cargo-sort@2.1.3
       - name: Check dependency tables
         run: cargo-sort --workspace --grouped --check
       - name: Check examples dependency tables
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1744,9 +1744,9 @@ dependencies = [
 
 [[package]]
 name = "tokio-tungstenite"
-version = "0.28.0"
+version = "0.29.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d25a406cddcc431a75d3d9afc6a7c0f7428d4891dd973e4d54c56b46127bf857"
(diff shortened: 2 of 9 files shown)


axum/src/lib.rs
```diff
@@ -183,6 +183,51 @@
...
+//!     .with_state(AppState { api_state: ApiState {} });
+//!
+//! // This handler receives only the `ApiState` slice; it never sees `AppState`.
+//! async fn handler(State(api_state): State<ApiState>) {}
+//! # let _: Router = app;
+//! ```
+//!
+//! ### The `Router<S>` type parameter
+//!
+//! `Router<S>` means a router that is _missing_ a state of type `S`. Calling
```
yanns:
```suggestion
//! `Router<S>` when `S` is not `()` means a router that is _missing_ a state of type `S`. Calling
```

yanns (approved):
Looking good, thanks!
The `State` part is longer and longer.
Maybe we should extract it, like we do for error handling or middleware for example

axum/src/lib.rs
kaze: Thanks @yanns, I've incorporated the change you suggested.

PR #3689, opened 2026-03-18
Update to tokio-tungstenite 0.29

tottoto:
Updates to `tokio-tungstenite` 0.29.


yanns:
Is there something relevant that we should document in the release notes?

tottoto:
According to the CHANGELOG, there doesn’t seem to be anything in particular to mention.

https://github.com/snapview/tungstenite-rs/blob/v0.29.0/CHANGELOG.md#0290

Issue #3691, opened 2026-03-20
Docs mention query-params-with-empty-strings which doesn't exist


moubctez:
- [x] I have looked for existing issues (including closed) about this

### Version

0.8.8

### Platform

any

### Description

Docs https://docs.rs/axum/latest/axum/extract/struct.Query.html mention [query-params-with-empty-strings](https://github.com/tokio-rs/axum/blob/main/examples/query-params-with-empty-strings/src/main.rs) example, but it doesn't exist.

wyf027:
This was addressed in #3695 (merged): the example was removed in 0.9 and doc links now point at the correct release tag instead of `main`. Safe to close if that matches what you still see on docs.rs for your version.

PR #3695, opened 2026-03-23
[FIX] link to query-params-with-empty-string

mart-e:
The example was removed for axum 0.9 at cae6bc37097ea61fcdf9416 but the link in the doc hardcoded “main”

Fixes tokio-rs/axum#3691

## Motivation

https://docs.rs/axum/0.8.8/axum/extract/struct.Query.html has a link to https://github.com/tokio-rs/axum/blob/main/examples/query-params-with-empty-strings/src/main.rs

## Solution

Hardcode the version in the URL

Let me know if you want me to port this to 0.6 and 0.7 version as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -725,9 +725,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.10.1"
+version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d71b6127be86fdcfddb610f7182ac57211d4b18a3e9c82eb2d17662f2227ad6a"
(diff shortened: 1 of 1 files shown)


yanns:
Could you have a look at the errors in the CI pipeline?
Maybe you'd need to rebase on the latest `main`.

mart-e:
>  Maybe you'd need to rebase on the latest main.

But my goal is to correct https://docs.rs/axum/0.8.8/axum/extract/struct.Query.html. That needs to be done on `v0.8.x` branch right? Or you are no longer updating past versions?
I see you have just created the `v0.8.9` branch but the comments are no longer present there (as on the `main` branch).

yanns:
yes possible, sorry, I don't fully know how this work. I'll ask.

yanns:
You're right. This fix should only aim the `v0.8.x` branch. I've merged with the last version of this branch to fix the CI.

PR #3702, opened 2026-03-31
Bump matchit to 0.9.2

AMDmi3:
## Motivation

Explained in #3140 - primarily to be able to use very desired prefixed or suffixed captures in routes, such as `router.route("/{id}.png")`.

## Solution

The work had already been done by @mladedav, but as I understand there were issues regarding conflicting routes handling in `matchit`, which seem to have been resolved in `0.9.2`. I've rebased David's commit, bumped `matchit` to the latest version, and updated tests and documentation, mostly in regard to that prefixed and suffixed captures may no longer be mixed.

### DIFF
--- a/axum/src/routing/tests/mod.rs
+++ b/axum/src/routing/tests/mod.rs
@@ -552,6 +552,42 @@ fn colliding_fallback_with_wildcard() {
         .route("/{*wild}", get(|| async { "wildcard" }));
 }
 
+#[should_panic(
+    expected = "Invalid route \"/{wild}-bar\": Insertion failed due to conflict with previously registered route: /foo-{wild}"
+)]
+#[test]
+fn colliding_prefix_suffix() {
+    _ = Router::<()>::new()
+        .route("/foo-{wild}", get(|| async { "wildcard" }))
+        .route("/foo", get(|| async { "wildcard" }))
+        .route("/{wild}", get(|| async { "wildcard" }))
+        .route("/{wild}-bar", get(|| async { "wildcard" }));
+}
+
+#[should_panic(
+    expected = "Invalid route \"/foo-{wild}\": Insertion failed due to conflict with previously registered route: /foo-{wild}-bar"
+)]
+#[test]
+fn colliding_prefixsuffix_prefix() {
+    _ = Router::<()>::new()
+        .route("/foo-{wild}-bar", get(|| async { "wildcard" }))
+        .route("/foo", get(|| async { "wildcard" }))
+        .route("/{wild}", get(|| async { "wildcard" }))
+        .route("/foo-{wild}", get(|| async { "wildcard" }));
+}
+
+#[should_panic(
+    expected = "Invalid route \"/{wild}-bar\": Insertion failed due to conflict with previously registered route: /foo-{wild}-bar"
+)]
+#[test]
+fn colliding_prefixsuffix_suffix() {
+    _ = Router::<()>::new()
+        .route("/foo-{wild}-bar", get(|| async { "wildcard" }))
+        .route("/foo", get(|| async { "wildcard" }))
+        .route("/{wild}", get(|| async { "wildcard" }))
+        .route("/{wild}-bar", get(|| async { "wildcard" }));
+}
+
 // We might want to reject this too
 #[crate::test]
 async fn colliding_wildcard_with_fallback() {


AMDmi3:
Updated to matchit 0.9.2, which has fixed route conflict issue discovered on the previous iteration. Previously ignored tests now pass.

AMDmi3:
Remaining issues (typo checker false positives, formatting issue, messed examples Cargo.lock update) fixed, CI now green.

axum/src/docs/routing/route.md
```diff
@@ -45,20 +45,14 @@ regular expression. You must handle that manually in your handlers.
...
-- `/a{x}`
-- `/{x}def`
-- `/{x}`
+You may have either capture(s) with static prefixes, capture(s) with suffixes, or a single
+capture with both prefix and suffix, but these kinds of captures may not be mixed. You may mix
+these with static routes and a standalone capture though. If multiple patterns match, static
+segment takes precedence, then the capture with longest static prefix or suffix.
+
+Example valid mixed route sets:
+- `/logo.png`, `/author.jpg`, `/{id}.png`, `{other_file}` (but you may not add `/{id}.jpg` or `/post-{id}`).
```
mladedav: Nit, `{other_file}` is invalid, it must start with a forward slash so it should be `/{other_file}`.

mladedav: `/{id}.jpg` is valid to be added here, there's no conflict since it also uses just a suffix, same as `/{id}.png`. The other one with prefix is invalid though.

Maybe you wanted to also have a prefix in the first one?

mladedav: Since this needs other changes anyway, could you please add something with a longer but matching suffix like `/{id}.old.png` so that it's clear that this does not collide with `/{id}.png`?

axum/src/docs/routing/route.md
```diff
@@ -45,20 +45,14 @@ regular expression. You must handle that manually in your handlers.
...
+- `/logo.png`, `/avatar-{id}.png`, `{other_file}` (but you may not add `/{id}.jpg`, `/avatar-{id}.png`).
```
mladedav: The second example that cannot be added is the same as the already registered partial match?

Also, at this point, it's illegal to register _any_ other partial matches, right?

axum/src/routing/tests/mod.rs
```diff
@@ -461,15 +460,64 @@ async fn prefix_suffix_match() {
...
+    assert_eq!(get("/").await, "fallback");
+
+    // Empty capture is not allowed
+    assert_eq!(get("/start--end").await, "fallback");
+
+    assert_eq!(get("/start-regex-end").await, "regex");
+    assert_eq!(get("/foo/start-regex-end").await, "fallback");
 }
 
 #[crate::test]
```
mladedav, line 523: Can you also add a few collision tests please? You can pretty much copy `colliding_fallback_with_wildcard` just for partial capture routes.

.typos.toml
```diff
@@ -0,0 +1,5 @@
+[default]
+extend-ignore-re = [ 
+    "OnFailedUpdgrade`",
+	"/[a-z][a-z]\\{",
```
mladedav: Do I understand it correctly you've allowed one specific string (ending with a backtick?) and any two letter words directly followed by `\{`?

mladedav: There was already [this comment](https://github.com/tokio-rs/axum/pull/3679/changes#r2887792664) so if you wouldn't mind, could you just change the docs/wherever there's stuff like `/fo{o}` to something more sensible?

axum/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **changed:** `serve` has an additional generic argument and can now work with any response body
   type, not just `axum::body::Body` ([#3205])
 - **changed:** `Redirect` constructors now accept any `impl Into<String>` ([#3635])
+- **changed:** Updated `matchit` allowing for routes with captures and static prefixes and suffixes ()
```
mladedav: This one's probably on me, but add a link to this PR please.

axum/src/docs/routing/route.md
AMDmi3: > Also, at this point, it's illegal to register any other partial matches, right?

Correct

.typos.toml
AMDmi3: > Do I understand it correctly you've allowed one specific string (ending with a backtick?)

Yes, since it fires on these changelog items and the regexp covers both cases.

```
Rename `DefaultOnFailedUpdgrade` to `DefaultOnFailedUpgrade` 
Rename `OnFailedUpdgrade` to `OnFailedUpgrade`
```

>  and any two letter words directly followed by `\{`?

Yes, for `/fo{o}` cases. It's a great idea to make these into sensible strings instead.

AMDmi3:
Comments addressed
- Changed routes in tests and doc in a way they don't alarm typo checker.
- Fixed mixed route examples in doc.
- Added route collision tests.
- Mentioned the PR on the changelog.
- For reasons unknown, type checker no longer fires on `Updgrade` either, so I've removed its config from the PR.

BTW, this shouldn't be a breaking change, should it be rebased onto 0.8.x instead? Or will the next 0.8.x be released from main?

mladedav:
Thanks for the updates.

We handle backports for 0.8.x releases, main already contains breaking changes.

@yanns I guess this is actually a good time to ask since you released new versions today, do you think it would be better to keep doing the 0.8.x releases as until now by cherry picking commits when the release happens, or we can start adding the commits after PRs? Or maybe add a checkbox to the PR template?

yanns:
> We handle backports for 0.8.x releases, main already contains breaking changes.
> 
> @yanns I guess this is actually a good time to ask since you released new versions today, do you think it would be better to keep doing the 0.8.x releases as until now by cherry picking commits when the release happens, or we can start adding the commits after PRs? Or maybe add a checkbox to the PR template?

Let's wait for the new versions to be released.
And then, yes, I think every maintainer could cherry-pick the PRs in the `v0.8.x` branch.

PR #3704, opened 2026-04-01
Add Serve::with_executor and Executor trait for custom task spawning

### LINKED ISSUE #3703: Allow customizing the executor used by `axum::serve`

### Motivation

`axum::serve` hardcodes `TokioExecutor` for spawning connection tasks and hyper's internal HTTP/2 tasks. This is the one thing that can't be customized from the outside, you have to reimplement the entire serve loop (~150 lines) just to swap the executor.

We hit this in [dial9-tokio-telemetry](https://github.com/dial9-rs/dial9-tokio-telemetry) where all spawned tasks need to be wrapped to capture scheduling delays and wake events. Our workaround was a full [reimplementation of `axum::serve`](https://github.com/dial9-rs/dial9-tokio-telemetry/blob/main/examples/metrics-service/src/axum_traced.rs).

### Proposal

Add a `Serve::with_executor()` builder method. Default executor stays `TokioExecutor`, so it's fully backward compatible.

```rust
axum::serve(listener, router)
    .with_executor(my_executor)
    .await;
```

### Alternatives

- Use hyper/hyper-util directly: works, but you lose all of axum's serve conveniences (graceful shutdown, accept loop, error handling). This is what the docs currently suggest.
- Reimplement `axum::serve`: ~150 lines of boilerplate that tracks axum internals and breaks across versions.

yulnr:
Closes #3703
 
 ## Motivation

`axum::serve` hardcodes `TokioExecutor` for spawning connection tasks and hyper's internal HTTP/2 tasks. This is the one thing that cannot be customized by wrapping axum's API from the outside, users are forced to reimplement the entire serve loop (~150 lines) just to swap the executor.

This is needed for use cases like runtime telemetry (e.g. [dial9-tokio-telemetry](https://github.com/dial9-rs/dial9-tokio-telemetry)) where we want to wrap task spawning, so that we can capture things like wake events.

**Edit:** the goal of this feature is to allow hooking into the current tokio spawns (by being able to set an executor that provides them, rather than hardcoding them) to attach instrumentation. The goal is not the make `serve` fully runtime agnostic.

## Solution

This PR adds `Serve::with_executor()` and `WithGracefulShutdown::with_executor()` builder methods that take an Executor, and an `axum::serve::Executor` trait for defining the Executor interface.

The default remains a new `TokioExecutor` (that spawns tokio spawns just like before), so we maintain backward compatibility.

Edit: initially I had this use `hyper_util::rt::tokio::TokioExecutor` as the default executor, but it meant leaking that in the public API so we opted for a new TokioExecutor that we own (it does practically same internally)

Small design note: this defines a new `axum::serve::Executor` trait rather than reuse of `hyper::rt::Executor<Fut>` directly, because the latter is generic at the trait level and I couldn't find a way to bound `E` to cover all the internal future types hyper needs. So this new trait uses a generic method instead, and then a `HyperExecutor<E>` adapter bridges to hyper.

### DIFF
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -26,7 +26,6 @@ allowed_external_types = [
     "futures_core::*",
     "futures_sink::*",
     "futures_util::*",
-    "hyper_util::rt::tokio::TokioExecutor",
     "tower_layer::*",
     "tower_service::*",
     # >=1.0
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -26,6 +26,7 @@ allowed_external_types = [
     "futures_core::*",
     "futures_sink::*",
     "futures_util::*",
+    "hyper_util::rt::tokio::TokioExecutor",
     "tower_layer::*",
     "tower_service::*",
     # >=1.0
--- a/axum/src/serve/mod.rs
+++ b/axum/src/serve/mod.rs
@@ -1059,7 +1059,11 @@ mod tests {
 
         // Verify the custom executor was used for spawning the connection task.
         // HTTP/1 should spawn exactly one task per connection.
-        assert_eq!(executor.count(), 1, "expected exactly one spawned task for the connection");
+        assert_eq!(
+            executor.count(),
+            1,
+            "expected exactly one spawned task for the connection"
+        );
     }
 
     #[crate::test]
--- a/axum/src/serve/mod.rs
+++ b/axum/src/serve/mod.rs
@@ -8,6 +8,7 @@ use std::{
     io,
     marker::PhantomData,
     pin::pin,
+    sync::Arc,
 };
 
 use axum_core::{body::Body, extract::Request, response::Response};
@@ -27,9 +28,10 @@ pub use self::listener::{ConnLimiter, ConnLimiterIo, Listener, ListenerExt, TapI
 
 /// Serve the service with the supplied listener.
 ///
-/// This method of running a service is intentionally simple and doesn't support any configuration.
+/// This method of running a service is intentionally simple and doesn't support much configuration.
 /// hyper's default configuration applies (including [timeouts]); use hyper or hyper-util if you
-/// need configuration.
+/// need more control. You can supply a custom [`Executor`] via [`Serve::with_executor`] to
+/// control how connection tasks are spawned.
 ///
 /// It supports both HTTP/1 as well as HTTP/2.
 ///
@@ -98,7 +100,7 @@ pub use self::listener::{ConnLimiter, ConnLimiterIo, Listener, ListenerExt, TapI
 /// [`HandlerWithoutStateExt::into_make_service_with_connect_info`]: crate::handler::HandlerWithoutStateExt::into_make_service_with_connect_info
 /// [`HandlerService::into_make_service_with_connect_info`]: crate::handler::HandlerService::into_make_service_with_connect_info
 #[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
-pub fn serve<L, M, S, B>(listener: L, make_service: M) -> Serve<L, M, S, B>
+pub fn serve<L, M, S, B>(listener: L, make_service: M) -> Serve<L, M, S, B, TokioExecutor>
 where
     L: Listener,
     M: for<'a> Service<IncomingStream<'a, L>, Error = Infallible, Response = S>,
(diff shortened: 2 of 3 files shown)


axum/Cargo.toml
```diff
@@ -26,6 +26,7 @@ allowed_external_types = [
     "futures_core::*",
     "futures_sink::*",
     "futures_util::*",
+    "hyper_util::rt::tokio::TokioExecutor",
```
rcoh: interesting side effect. I guess is a result of making it the default generic. It might be worth having our own TokioExecutor type to avoid having the default generic by the hyper_util type

jv1i: That's a good idea, I defined a new TokioExecutor type now so that we don't leak hyper's into the public API

rcoh (approved):
LGTM

axum/src/serve/mod.rs
```diff
@@ -111,21 +113,94 @@ where
...
+/// ```
+///
+/// If your executor is expensive to clone, wrap it in an `Arc`.
+/// A blanket implementation is provided for `Arc<T>` where `T: Executor`.
+#[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
+pub trait Executor: Clone + Send + Sync + 'static {
+    /// Execute a task.
+    fn execute<Fut>(&self, fut: Fut)
+    where
+        Fut: Future<Output = ()> + Send + 'static;
```
yanns: Do you think that this could be use to spawn other futures with a different output?
Would it make sense to copy the signature from tokio::spawn here?
Or is it better to keep it constraint to `()` for now, to make it easier to provide custom `Executor`

axum/src/serve/mod.rs
```diff
@@ -680,10 +867,50 @@ mod tests {
...
+        }
+    }
+
+    impl super::Executor for TestExecutor {
+        fn execute<Fut>(&self, fut: Fut)
+        where
+            Fut: std::future::Future<Output = ()> + Send + 'static,
+        {
+            self.0.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+            tokio::spawn(fut);
```
mladedav: The test should be independent of tokio if we're doing this. Something like `futures::excutor::LocalPool`.

My worry here is that this looks like it's supposed to be general, but from your motivation it seems you just want to be able to run some hooks before spawning tokio tasks and that's the scenario you've actually tested in your app. But this module still uses tokio's watch channel and Hyper uses `TokioIo` and I'm not entirely convinced that this would work out of the box with any other runtime (though I'm not sure, maybe it does).

yulnr: Good point, I see maybe this does read more generic than it is. I'll make that clear in the PR and also API/docs.
 
These changes does not make `serve` runtime-agnostic, I wrote this on the trait doc that we could give more emphasis:
```rust
/// Spawned futures rely on Tokio primitives internally, so the executor
/// must run them within a Tokio runtime context (e.g. via [`tokio::spawn`]).
```
So to confirm, there are multiple things in axum that need to run within a tokio runtime, and will otherwise panic with errors like: `there is no reactor running, must be called from the context of a Tokio 1.x runtime`. (the first offender you can see in the execute call is `TokioTimer`)

That was my initial read, seeing your comment now I ran a quick test with `futures::executor::LocalPool` just in case and I run exactly into those issues. 

We could make this runtime agnostic, but we'd be talking about a bigger refactor. I think it's possible if we abstract and decouple things like `TokioTimer`, `time::sleep`, `AsyncRead/AsyncWrite`: specifically the `Listener` trait defines `type Io: AsyncRead + AsyncWrite + Unpin ...` I need to see exactly how that cascades but I think changing that would be a semver break? (also I think actually hyper doesn't hardcode `TokioIo`, it's more of a utility we're using on top of hyper's own io trait).

With that said, perhaps some better naming would make this feature more clear? If `Executor` is too loaded perhaps: `Executor` -> `TokioSpawner`, and `TokioExecutor` -> `TokioSpawner`/`DefaultTokioSpawner`?

Also, picking up on @yanns point above, I missed that we could enforce this at the type level by having execute return tokio::task::JoinHandle<()>. (so is not just documented)

Would that make more sense?

Otherwise I can look into fully decoupling from tokio, but to get an idea of changes needed you can see how much is gated behind `#[cfg(feature = "tokio"])` around this.

axum/src/serve/mod.rs
yulnr: Thanks for bringing that up, based on the conversation here: https://github.com/tokio-rs/axum/pull/3704#discussion_r3079757826 I think is better to mirror tokio's signature, since we're either way not going all the way with supporting any custom executor/runtime. Updated to match that.

axum/src/serve/mod.rs
yulnr: Updated the trait to make `execute` mirror tokio's spawn, and removed the non-tokio tests (were checking for panics, now that code correctly wouldn't compile). What do we think of the current Executor naming?

axum/src/serve/mod.rs
```diff
@@ -365,12 +542,27 @@ where
     }
 }
 
-async fn handle_connection<L, M, S, B>(
+/// Adapts axum's [`Executor`] to hyper's `Executor<Fut>`.
```
mladedav:
```suggestion
/// Adapts axum's [`Executor`] to hyper's [`Executor<Fut>`](hyper::rt::Executor).
```

axum/src/serve/mod.rs
mladedav: I think the gating is mostly just convenience and nobody checked closely in a long time so just saying it's probably not the best indicator.

I lived under the impression that someone was running `axum` with other runtimes than tokio but that's definitely currently done manually and not through `serve` so maybe it's much harder than I originally thought.

I also misunderstood originally and though this really was supposed to be more general. This way it makes sense though, full disclosure, I hope one day we get rid of this in favor of the more general solution. But I wouldn't expect that to happen too soon so I'm fine having this here in the meantime.

mladedav:
Just to make sure, you believe this is not a breaking change, right? I originally thought it was but looking again, it really might not be?

yulnr:
> full disclosure, I hope one day we get rid of this in favor of the more general solution.

Agree!

> Just to make sure, you believe this is not a breaking change, right? I originally thought it was but looking again, it really might not be?

Correct, this shouldn't be breaking. for `serve` the signature:
```rust
pub fn serve<L, M, S, B>(listener: L, make_service: M) -> Serve<L, M, S, B, TokioExecutor>
```
keeps the same args, and has TokioExecutor as default for E, just to be explicit really because `Serve` is now: 
```rust
pub struct Serve<L, M, S, B, E = TokioExecutor> {
```
So `<L, M, S, B>` also works. If you had explicitly written something like `let server: Serve<L, M, S, B> = axum::serve(listener, app);`, the 5th param just defaults.

mladedav (approved):

PR #3714, opened 2026-04-06
Add tests to examples/global-404-handler

szabgab:
The CI already runs the tests of the examples, so IMHO it is a good idea to verify that every example works as expected.

See also #3263


davidpdrsn:
Like I wrote [here](https://github.com/tokio-rs/axum/issues/3263#issuecomment-4458454377) I don't think the examples necessarily need tests especially not very simple ones like this. So I'll close this for now.

PR #3635, opened 2026-01-19
feat(Redirect): allow owned locations

joeydewaal:
## Motivation
Before #3377 landed, `Redirect` stored the location as a `HeaderValue`. Now it stores it as a `String` internally, which means that it always has to clone the `&str` until the `HeaderValue` is created.

## Solution
Give people the option to create a `Redirect` with an owned value.

Alternatively, we could store it as a `Cow<'static, str>`. (and change the params to an `impl Into<Cow<'static, str>>`)

PR #3720, opened 2026-04-10
axum: Add missing PR link to changelog (#3635)

AMDmi3:
## Motivation

A changelog entry referes to a PR, but the link is missing

## Solution

Add the missing link


tottoto (approved):
Thanks!

PR #3724, opened 2026-04-12
test: add serve module integration tests

dihannahdi:
## Motivation

The `axum::serve` module is critical infrastructure, but its test coverage focuses mainly on compile-time checks (`if_it_compiles_it_works`), `local_addr()` verification, and graceful shutdown timeout behavior. There are no tests covering the actual HTTP request/response flow through the serve pipeline for common scenarios.

## Solution

Add four integration tests using the existing in-memory `ReadyListener` + `io::duplex` pattern (no real TCP connections needed):

- **`serve_post_with_body`** — verifies POST requests with a request body flow correctly through serve to a handler and back
- **`serve_not_found`** — verifies unmatched routes return HTTP 404 through the serve pipeline
- **`serve_multiple_routes`** — verifies multiple sequential requests (HTTP/1.1 keep-alive) on a single connection are correctly dispatched to different routes
- **`serve_with_state`** — verifies `Router::with_state()` propagates state correctly through the serve pipeline to `State<T>` extractors

All tests pass alongside the existing 7 tests (11 total).

### DIFF
--- a/axum/src/serve/mod.rs
+++ b/axum/src/serve/mod.rs
@@ -1043,6 +1043,126 @@ mod tests {
         assert_eq!(body, "Hello, World!");
     }
 
+    // Asserts the documented `with_graceful_shutdown` drain semantics: after the
+    // signal fires, an already-in-flight request is allowed to run to completion
+    // and only then does the `serve` future resolve. The existing
+    // `test_with_graceful_shutdown_request_header_timeout` only covers stalled
+    // requests being killed by hyper's header read timeout.
+    #[crate::test]
+    async fn graceful_shutdown_completes_inflight_request() {
+        use std::sync::Arc;
+
+        let started = Arc::new(tokio::sync::Notify::new());
+        let release = Arc::new(tokio::sync::Notify::new());
+        let (shutdown_tx, shutdown_rx) = tokio::sync::oneshot::channel::<()>();
+
+        let app = Router::new().route("/", {
+            let started = started.clone();
+            let release = release.clone();
+            get(move || {
+                let started = started.clone();
+                let release = release.clone();
+                async move {
+                    started.notify_one();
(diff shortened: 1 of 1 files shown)


davidpdrsn:
The vast majority of our tests uses our internal `TestClient` which does use `serve` https://github.com/tokio-rs/axum/blob/main/axum/src/test_helpers/test_client.rs#L24. So we do have some coverage.

Is there anything specific which we should be testing but aren't?

dihannahdi:
Fair point — `TestClient` does cover the basic request/response flow through `serve`, so the four tests I originally added were largely redundant with what the rest of the suite already exercises transitively. I've reworked the PR (force-pushed and rebased onto `main`) to target two specific gaps that aren't covered today:

- **`graceful_shutdown_completes_inflight_request`** — asserts `with_graceful_shutdown` drains in-flight requests rather than aborting them. The existing `test_with_graceful_shutdown_request_header_timeout` only proves that *stalled* requests get killed by hyper's header read timeout; the happy-path drain semantic (in-flight handler runs to completion, response is delivered, then the serve future resolves) had no regression test.
- **`tap_io_runs_on_each_accepted_connection`** — asserts `ListenerExt::tap_io` invokes its closure on every accepted connection. `limit_connections` has a direct unit test in `serve::listener::tests`; `tap_io` was only covered at the type level by `if_it_compiles_it_works`.

I also dropped the redundant POST/404/multi-route/`with_state` tests since those paths are well covered through `TestClient` and the existing `connect_info::socket_addr` end-to-end test. Let me know if you'd rather see this scoped down further or split.

davidpdrsn (approved):
LGTM! Thanks.

dihannahdi:
> LGTM! Thanks.

Me too, Thank you Mr. David! working on other PR you've asked about

PR #3725, opened 2026-04-12
fix(macros): support prefix/suffix in TypedPath segments

### LINKED ISSUE #3681: The TypedPath derive macro breaks for segments with a prefix or suffix

### Version

axum v0.8.8
axum-extra v0.12.5

### Crates

axum-extra

### Description

dihannahdi:
## Motivation

The `TypedPath` derive macro currently requires each path segment to be **entirely** a capture (`/{id}`) or **entirely** static (`/users`). This means routes with mixed segments—where a capture is embedded within static text—are rejected by the macro even though the underlying `matchit` router fully supports them.

Common real-world patterns that fail today:

```rust
#[typed_path("/@{username}")]       // prefix capture
struct AtUsername { username: String }

#[typed_path("/avatars/{id}.png")]   // suffix capture
struct AvatarPath { id: u32 }

#[typed_path("/files/{name}.tar.gz")] // prefix + suffix
struct FilePath { name: String }
```

These all produce a compile error because `parse_path()` only recognises segments that start with `{` and end with `}`.

Closes #3681

## Solution

Add a `CaptureWithAffix` variant to the internal `Segment` enum:

```rust
CaptureWithAffix {
    prefix: String,   // text before `{`
    capture: String,   // the capture name
    suffix: String,    // text after `}`
    span: Span,
}
```

`parse_path()` now scans each segment for `{…}` anywhere within it. When a capture is found with surrounding static text, the new variant is emitted. All downstream functions (`format_str_from_path`, `captures_from_path`, field expansion, unit-struct validation) are updated to handle it.

Escaped braces (`{{`, `}}`) are left alone as static segments—matching the existing behaviour.

## Testing

- New **pass** test (`prefix_suffix_capture.rs`) covering named structs, tuple structs, and paths with prefix-only, suffix-only, and prefix+suffix captures.
- All existing tests continue to pass (`cargo test -p axum-macros` — 28/28 passing).

### DIFF
--- a/axum-macros/src/typed_path.rs
+++ b/axum-macros/src/typed_path.rs
@@ -176,7 +176,7 @@ fn expand_unnamed_fields(
     let num_captures = segments
         .iter()
         .filter(|segment| match segment {
-            Segment::Capture(_, _) => true,
+            Segment::Capture(_, _) | Segment::CaptureWithAffix { .. } => true,
             Segment::Static(_) => false,
         })
         .count();
@@ -195,7 +195,9 @@ fn expand_unnamed_fields(
     let destructure_self = segments
         .iter()
         .filter_map(|segment| match segment {
-            Segment::Capture(capture, _) => Some(capture),
+            Segment::Capture(capture, _) | Segment::CaptureWithAffix { capture, .. } => {
+                Some(capture)
+            }
             Segment::Static(_) => None,
         })
         .enumerate()
@@ -286,7 +288,7 @@ fn expand_unit_fields(
 ) -> syn::Result<TokenStream> {
     for segment in parse_path(path)? {
         match segment {
-            Segment::Capture(_, span) => {
+            Segment::Capture(_, span) | Segment::CaptureWithAffix { span, .. } => {
                 return Err(syn::Error::new(
                     span,
                     "Typed paths for unit structs cannot contain captures",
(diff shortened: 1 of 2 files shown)


Samielakkad (commented):
This fixes the single-capture-with-affix case nicely. One edge case I think needs an explicit decision/test is a segment with more than one capture, for example:

```rust
#[typed_path(/files/{name}.{ext})]
```

`parse_path` currently finds only the first `{...}` pair and stores the rest as `suffix`. That means `format_str_from_path` would still emit `/files/{name}.{ext}`, so the router path contains both captures, but `captures_from_path` only returns `name` and the generated `Display` impl only formats one field.

If multiple captures per segment are not supported, this should probably be rejected with a clear compile error instead of being partially accepted. If they are intended to be supported, the parser likely needs to keep scanning the suffix rather than making it opaque static text.

dihannahdi:
Thanks for catching this. It turns out #3797, merged last week, already added this exact validation to the `TypedPath` macro. A segment with more than one capture is now rejected at compile time with the message "Cannot have multiple path parameters in a single segment", and prefix/suffix captures like `/avatars/{id}.png` are supported there too, via a cleaner rewrite that treats the affixes as ordinary static segments.

That covers both the feature this PR was adding and the edge case you raised, so I am going to close this as superseded by #3797. Appreciate the review.

PR #3726, opened 2026-04-12
fix(macros): improve debug_handler errors for WithRejection

### LINKED ISSUE #1337: `#[debug_handler]` doesn't improve error message for `WithRejection`
`WithRejection<T, E>` requires `E` to implement `IntoResponse`. If you forget that you get the familiar "handler not implemented" error and `#[debug_handler]` doesn't help:

```rust
use axum::extract::rejection::JsonRejection;
use axum::response::{IntoResponse, Response};
use axum::routing::{post, MethodRouter};
use axum::Json;
use axum_extra::extract::WithRejection;

#[tokio::main]
async fn main() {
    let _: MethodRouter = post(send);
}

#[axum_macros::debug_handler]
pub async fn send(
    _: WithRejection<Json<MessageSend>, ApiError>,
) -> Json<MessageSendResponse> {
    Json(MessageSendResponse { id: "".to_string() })
}

#[derive(serde::Deserialize)]
pub struct MessageSend {
    pub contents: String,
}

#[derive(serde::Serialize)]
pub struct MessageSendResponse {
    pub id: String,
}

#[derive(Debug, thiserror::Error)]
pub enum ApiError {
    #[error(transparent)]
    JsonExtractorRejection(#[from] JsonRejection),
}

// no `impl IntoResponse for ApiError`
```

The error message is

```
error[E0277]: the trait bound `fn(WithRejection<Json<MessageSend>, ApiError>) -> impl Future<Output = Json<MessageSendResponse>> {send}: Handler<_, _, _>` is not satisfied
   --> src/main.rs:9:32
    |
9   |     let _: MethodRouter = post(send);
    |                           ---- ^^^^ the trait `Handler<_, _, _>` is not implemented for `fn(WithRejection<Json<MessageSend>, ApiError>) -> impl Future<Output = Json<MessageSendResponse>> {send}`
    |                           |
    |                           required by a bound introduced by this call
    |
    = help: the following other types implement trait `Handler<T, S, B>`:
              <IntoHandler<H, T, S, B> as Handler<T, S, B>>
              <Layered<L, H, T, S, B> as Handler<T, S, B>>
              <Or<L, R, Lt, Rt, S, B> as Handler<(M, Lt, Rt), S, B>>
note: required by a bound in `post`
   --> /Users/davidpdrsn/.cargo/registry/src/github.com-1ecc6299db9ec823/axum-0.6.0-rc.1/src/routing/method_routing.rs:404:1
    |
404 | top_level_handler_fn!(post, POST);
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `post`
    = note: this error originates in the macro `top_level_handler_fn` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `WithRejection<Json<MessageSend>, ApiError>: FromRequest<(), Body, _>` is not satisfied
  --> src/main.rs:14:8
   |
14 |     _: WithRejection<Json<MessageSend>, ApiError>,
   |        ^^^^^^^^^^^^^ the trait `FromRequest<(), Body, _>` is not implemented for `WithRejection<Json<MessageSend>, ApiError>`
   |
   = help: the trait `FromRequest<S, B>` is implemented for `WithRejection<E, R>`
note: required by a bound in `__axum_macros_check_send_0_from_request_check`
```

The generate code is

```rust
#[allow(warnings)]
fn __axum_macros_check_send_0_from_request_check<M>()
where
    WithRejection<Json<MessageSend>, ApiError>:
        ::axum::extract::FromRequest<(), axum::body::Body, M> + Send,
{
}

#[allow(warnings)]
fn __axum_macros_check_send_0_from_request_call_check() {
    __axum_macros_check_send_0_from_request_check();
}
```

For some reason the additional `M` type parameter changes rust's error message. If you remove that and instead generate

```rust
#[allow(warnings)]
fn __axum_macros_check_send_0_from_request_check()
where
    WithRejection<Json<MessageSend>, ApiError>:
        ::axum::extract::FromRequest<(), axum::body::Body> + Send,
{
}
```

The error message becomes

```
error[E0277]: the trait bound `ApiError: IntoResponse` is not satisfied
  --> src/main.rs:36:5
   |
36 | /     WithRejection<Json<MessageSend>, ApiError>:
37 | |         ::axum::extract::FromRequest<(), axum::body::Body> + Send,
   | |_________________________________________________________________^ the trait `IntoResponse` is not implemented for `ApiError`
   |
```

We could look into this and see if its possible to improve this.

dihannahdi:
## Motivation

When `#[debug_handler]` is used with a `WithRejection<T, E>` extractor as the last or only handler argument, the compiler error messages are confusing and unhelpful. This is because the generated trait-bound check uses:

```rust
fn check<M>()
where
    WithRejection<T, E>: FromRequest<S, M> + Send,
{}
```

Since `WithRejection<T, E>` can implement `FromRequest` for **multiple values of `M`** (via both the direct `FromRequest<S>` impl and the blanket `FromRequest<S, ViaParts>` through `FromRequestParts`), rustc cannot resolve the ambiguity when a bound fails. The resulting error message doesn't explain *which specific constraint* is missing—it just says the trait bound is not satisfied.

For example, if the rejection type doesn't implement `IntoResponse`, the user currently gets an opaque error about `WithRejection<..>: FromRequest<(), _>` instead of a clear "`MyError: IntoResponse` is not satisfied."

Closes #1337

## Solution

Detect `WithRejection<T, E>` at the type level in the check generation and decompose it into two focused assertions:

1. **`T: FromRequest<S, M>`** — checks the inner extractor directly. Since most concrete extractors (`Query<_>`, `Json<_>`, etc.) only satisfy one `FromRequest` path, rustc can resolve `M` unambiguously and give a specific, helpful diagnostic.

2. **`E: IntoResponse`** — directly surfaces a missing `IntoResponse` impl on the custom rejection type.

### Before (unhelpful)
```
error[E0277]: the trait bound `WithRejection<Json<Payload>, BadRejection>: FromRequest<(), _>` is not satisfied
```

### After (clear and actionable)
```
error[E0277]: the trait bound `BadRejection: IntoResponse` is not satisfied
  --> handler.rs:22:32
   |
22 |     WithRejection(payload, _): WithRejection<Json<Payload>, BadRejection>,
   |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsatisfied trait bound
   |
help: the trait `IntoResponse` is not implemented for `BadRejection`
```

## Changes

- **`axum-macros/src/debug_handler.rs`**: Added `extract_with_rejection_types()` helper that detects `WithRejection<T, E>` and extracts the inner types. Modified `check_inputs_impls_from_request()` to use the inner extractor type `T` for the `FromRequest` bound check and generate a separate `E: IntoResponse` assertion.
- **`axum-macros/Cargo.toml`**: Added `with-rejection` to axum-extra dev-dependency features (needed for the new test).
- **`axum-macros/tests/debug_handler/pass/with_rejection.rs`**: New pass test verifying `#[debug_handler]` works correctly with `WithRejection`.
- **`axum-macros/tests/debug_handler/fail/with_rejection_not_into_response.rs`**: New fail test showing the improved error when the rejection type doesn't implement `IntoResponse`.

## Testing

All 36 `debug_handler` trybuild tests pass on nightly (24 handler + 12 middleware), including the 2 new tests.


eliottreich:
There's a $30 bounty for Axum router/macro improvements on [Code.TaskBounty](https://code.task-bounty.com/task/axum-web-framework-router-enhancement). Better error messages in debug_handler would be exactly the kind of DX improvement that fits.

Bounty: https://code.task-bounty.com/task/axum-web-framework-router-enhancement

Samielakkad (commented):
This should improve the common body-extractor case a lot. One remaining hole: the generated checks no longer seem to prove that the custom rejection can actually be constructed from the inner extractor rejection.

For `WithRejection<T, E>`, the useful bounds are roughly:

- `T: FromRequest<...>` / `FromRequestParts<...>`
- `E: From<T::Rejection>`
- `E: IntoResponse`

This PR checks the first and third pieces, but I do not see a focused check for `E: From<T::Rejection>`. So a type like `WithRejection<Json<Payload>, MyError>` where `MyError: IntoResponse` but does not implement `From<JsonRejection>` may still produce a less-targeted error, or possibly avoid the intended debug-handler check until the real handler bound is evaluated.

It would be good to add a fail test for “rejection implements IntoResponse but not From<inner rejection>” and, if needed, generate a second targeted assertion for that conversion.

dihannahdi:
You are right that the `From<T::Rejection>` piece was not being checked. I have added that bound to the same generated check function as the inner `FromRequest` check, under the same inferred `M`, since it mirrors what `WithRejection`'s own `FromRequest`/`FromRequestParts` impls require (verified against axum-extra's `with_rejection.rs`, which bounds `R: From<E::Rejection> + IntoResponse` on both).

There is a new fail test, `with_rejection_not_from_inner_rejection.rs`, where the rejection implements `IntoResponse` but not `From<inner rejection>`. It now fails on the generated check with a targeted `the trait bound BadRejection: From<JsonRejection> is not satisfied` pointing at the check function, rather than leaking to the handler body.

I also merged `main` to bring the branch current and regenerated the trybuild snapshots for the newer pinned nightly.

dihannahdi:
Short status so this does not drift out of view.

Samielakkad's point about the missing `From<T::Rejection>` check is in, with a fail test alongside it. At the same time I merged main and regenerated the trybuild snapshot for the newer pinned nightly. CI is green on the merged tree with `test-nightly` included, which was the part I was least confident about, since the stderr snapshots are specific to the compiler version.

Nothing is outstanding from my side. It is small and self-contained, so whenever a macros review comes around it should not take long to look at.

PR #3732, opened 2026-04-17
chore: update sha1 to 0.11

joeydewaal:
## Motivation
The `sha1` dep released `0.11`.

## Solution
Upgrade to the new version.


joeydewaal:
Looks like v0.11 requires the 2024 edition but axum's msrv is 1.80.

PR #3734, opened 2026-04-22
adding VecBinary(Vec<Bytes>) to ws::Message

whilesoftware:
## Motivation

zero-copy stuff

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -92,7 +92,7 @@
 
 use self::rejection::*;
 use super::FromRequestParts;
-use crate::{body::Bytes, response::Response, Error};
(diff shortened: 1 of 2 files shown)


axum/src/extract/ws.rs
```diff
@@ -928,6 +940,15 @@ impl From<Message> for Vec<u8> {
...
 }
 
+fn concat_vec_bytes(vec: &[Bytes]) -> Bytes {
+    let total: usize = vec.iter().map(|b| b.len()).sum();
+    let mut buf = BytesMut::with_capacity(total);
+    for bytes in vec {
+        buf.extend_from_slice(bytes);
+    }
+    buf.freeze()
+}
```
davidpdrsn, line 950: Couldn't users do this themselves and then use `Message::Binary(Bytes)`?

PR #3737, opened 2026-04-27
remove ECOSYSTEM.md

davidpdrsn:
I don't think ECOSYSTEM.md makes much sense anymore:

- It creates a bunch of churn and drive by PRs.
- Its sometimes hard to decide which things should or shouldn't be included.
- Its likely full of stale projects and I don't think it should be anyone's job to prune it.

I do think it made sense when axum was new but given its current maturity I don't think so.

So yeah I propose that we just remove it. I posted this [in Discord](https://discord.com/channels/500028886025895936/1406236163819180185/1498224564935594034) and people seemed to agree.


Darksonn (approved):

szabgab:
A while ago I included it in the Axum book I am writing. I'll be happy to keep maintaining it there.
https://axum.code-maven.com/ecosystem
source: https://github.com/szabgab/axum.code-maven.com/edit/main/src/ecosystem.md

NishantCoder108:
Half an hour ago, I was exploring tutorials to build an application using Axum and also include CI/cd, PostgreSQL stuff. I just reloaded the page, and this page is gone.
I'm not sure if this is right or wrong to remove, but I was referring to the Ecosystem.md page from a week ago to understand how I can build an application.
But anyway, thanks @szabgab for putting all resources in one place, I bookmarked it.

ISensuiI:
Thank you very much. I don't think this file is useless.  @szabgab

szabgab:
I wonder if it would make sense to create a temporary replacement of the `ECOSYSTEM.md` file with a link to the page where I am hosting it. So people who used to look at the file or who linked to that file will be able to find the alternative. After some time it could be removed.

szabgab:
As a second thought, I moved the updated content to a new repository called [awesome-axum](https://github.com/szabgab/awesome-axum).

davidpdrsn:
> I wonder if it would make sense to create a temporary replacement of the `ECOSYSTEM.md` file with a link to the page where I am hosting it. So people who used to look at the file or who linked to that file will be able to find the alternative. After some time it could be removed.

I don't want to promote any specific list. That was one of the problems with ECOSYSTEM.md in the first place.

sekunho:
@szabgab Thanks for adding my projects! :bow:

PR #3739, opened 2026-04-28
integrate release-plz

yanns:
Before merging this, we need to:
- change GitHub permissions so that this action can create PR: https://release-plz.dev/docs/github/quickstart#1-change-github-actions-permissions
- Set the CARGO_REGISTRY_TOKEN secret: https://release-plz.dev/docs/github/quickstart#2-set-the-cargo_registry_token-secret

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
       - &checkout
         name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           fetch-depth: 0
           persist-credentials: false
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -28,7 +28,7 @@ jobs:
         with:
           toolchain: stable
       - name: Run release-plz
-        uses: release-plz/action@v0.5
+        uses: release-plz/action@1528104d2ca23787631a1c1f022abb64b34c1e11 # v0.5
         with:
           command: release
         env:
@@ -50,7 +50,7 @@ jobs:
       - *checkout
       - *install-rust
       - name: Run release-plz
-        uses: release-plz/action@v0.5
+        uses: release-plz/action@1528104d2ca23787631a1c1f022abb64b34c1e11 # v0.5
         with:
           command: release-pr
         env:
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -1,0 +1,56 @@
+name: Release-plz
+
+on:
+  push:
+    branches:
+      - main
+      - v0.8.x
+
+jobs:
+  # Release unpublished packages.
+  release-plz-release:
+    name: Release-plz release
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'tokio-rs' }}
+    permissions:
+      contents: write
+      pull-requests: read
+    steps:
+      - &checkout
+        name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+      - &install-rust
+        name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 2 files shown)


Darksonn:
How do I add the token in a way that is not easily stolen by anyone for whom the test suite automatically runs (i.e. anyone who has submitted a PR in the past)?

yanns:
> How do I add the token in a way that is not easily stolen by anyone for whom the test suite automatically runs (i.e. anyone who has submitted a PR in the past)?

We'd need to add a secret like this: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets
Once added, those secrets are not visible anymore.
They are redacted from logs by GitHub, so no one should be able to see them. 
There's still some risks with that approach:
- a bug in GitHub actions leading to a secret being logged and leaked
- someone would could SSH into an action run, and see the secrets.

An alternative would be to set https://crates.io/docs/trusted-publishing.  By thinking more about it, I think that this would be more secure. If we want to go that way, I'll remove the mention of `CARGO_REGISTRY_TOKEN`.

jlizen:
If it is helpful, I recently looked at a lot of repos' ci and synthesized a good generic config:
https://github.com/battery-pack-rs/battery-pack/blob/main/battery-packs/ci-battery-pack/templates/trusted-publishing/.github/workflows/release.yml

This one will get you automatic PRs staging the version bumps and changelogs, that you can merge to release (or commit on top of to tweak the changelogs). That workflow is nice for keeping track of what is staged.

The [crate readme](https://github.com/battery-pack-rs/battery-pack/tree/main/battery-packs/ci-battery-pack#release-plz) has more discussion of necessary permissions.

You can also use the `cargo bp` tool to materialize the template and fold it into your workspace, but you've already done a lot of the legwork :)

jlizen:
I do recommend hash pinning to avoid supply chain attacks on the action.

And yeah, use crates.io trusted publishing, my template had an example. For that you can use GitHub's natively managed actions token (which I can only assume is ephemeral and thus somewhat more secure than a traditional secret).

yanns:
I've removed `CARGO_REGISTRY_TOKEN` with the idea that we'll use trusted publishing.
I've also pinned the version. In each commit, the message contains a link to the pinned commit.

.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,56 @@
...
+
+  # Create a PR with the new versions and changelog, preparing the next release.
+  release-plz-pr:
+    name: Release-plz PR
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'tokio-rs' }}
+    permissions:
+      contents: write
+      pull-requests: write
+    concurrency:
```
jlizen: This seems a bit odd, was it causing issues?

jlizen (approved):
Looks good to me

Darksonn:
What is the list of crates that the crates.io token will be used to publish? I can restrict it to be usable with only those crates.

yanns:
@Darksonn we'd like to use https://crates.io/docs/trusted-publishing instead of using a token. It seems more secure as we won't need a long living token. Does it sound good to you?

By looking again at the documentation, I think I've missed that we need the following:
```
    - uses: rust-lang/crates-io-auth-action@v1
      id: auth
```
I'll add this.

yanns:
I just needed to add `id-token: write`.

From https://release-plz.dev/docs/github/quickstart#2-set-the-cargo_registry_token-secret:

> If you want to use trusted publishing, don't use the rust-lang/crates-io-auth-action action, and don't set the CARGO_REGISTRY_TOKEN secret (if you already use CARGO_REGISTRY_TOKEN, remove it from your workflow file entirely). release-plz implements the same crates-io API calls of the rust-lang/crates-io-auth-action action, and uses them to obtain a token when necessary.
> 
> Set id-token: write in the permissions of the job that runs release-plz release.

Darksonn:
Ah, cool, I did not know about this feature. Although I can publish axum via the tokio-rs groups, I'm not a full owner of the crate, so I don't have access to this setting. David or Carl should be able to set it up.

davidpdrsn (approved):
🚀

Issue #3741, opened 2026-05-01
Handling of HEAD requests violates RFC 9110


lorenzleutgeb:
## Bug Report

Axum responds to HEAD requests with `content-length: 0`, which (in almost all cases) violates HTTP semantics. Refer to [RFC 9110, Sec. 9.3.2, Paragraph 2](https://datatracker.ietf.org/doc/html/rfc9110#section-9.3.2-2).

If the value for the `content-length` header that would be produced in response to a GET request is not known at the time of responding to a HEAD request, then it should be omitted, not included with a value of 0.

Real-world example: A handler that archives data into a `*.zip` file and responds with the archive. Assume that there is quite a lot of data, so archiving just to know the length of the hypothetical response to GET would be considered wasteful. It would be incorrect to respond with `content-length: 0`, because the archive itself will likely not be the empty file.

The same argument is used in https://github.com/tokio-rs/axum/pull/755#issuecomment-1037931480, so I allow myself to mention the author: @davidpdrsn

Previous issues that seem to be related:
 - https://github.com/tokio-rs/axum/issues/730
 - https://github.com/tokio-rs/axum/issues/747

### Version

Checked out https://github.com/tokio-rs/axum.git at commit `9f4d52e310a6053e49f3db09ce02f723a7a3f5bb` which is `main` as of 2026-05-01.

### Platform

NixOS on Windows Subsystem for Linux, likely irrelevant.

```
Linux wsl 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 GNU/Linux
```

### Description

I tried this code: https://github.com/tokio-rs/axum/blob/9f4d52e310a6053e49f3db09ce02f723a7a3f5bb/examples/handle-head-request/src/main.rs

Actual:

```shell-session
$ git clone https://github.com/tokio-rs/axum.git
$ cd axum/example/handle-head-request
$ cargo run -p example-handle-head-request
[build output]
listening on 127.0.0.1:3000
$ curl -s -D - -o /dev/null http://127.0.0.1:3000/get-head
HTTP/1.1 200 OK
content-type: text/plain; charset=utf-8
x-some-header: header from GET
content-length: 13
date: [current date and time]
$ curl -s -D - -o /dev/null --head http://127.0.0.1:3000/get-head
HTTP/1.1 200 OK
x-some-header: header from HEAD
content-length: 0
date: [current date and time]
```

[...]

PR #3742, opened 2026-05-02
body: Introduce Bodies of Unknown Size

lorenzleutgeb:
As reported in <https://github.com/tokio-rs/axum/issues/3741>, in some responses to HEAD requests, the header `content-type: 0` would be included, even though the size of the body was not known and also not zero.

The reason for this was that `axum::body::Body::empty`, which represents a body that is *known to be empty* was used, even though *nothing was known about the body*.

To remedy, introduce a new utility type `axum::body::Unknown`, which represents an unknown body. This type could be considered for upstreaming to the `http-body-util` crate in the future.

### DIFF
--- a/axum-core/src/body.rs
+++ b/axum-core/src/body.rs
@@ -1,5 +1,8 @@
 //! HTTP body utilities.
 
+mod unknown;
+pub(crate) use unknown::Unknown;
+
 use crate::{BoxError, Error};
 use bytes::Bytes;
 use futures_core::{Stream, TryStream};
@@ -51,6 +54,30 @@ impl Body {
     /// Create an empty body.
     pub fn empty() -> Self {
         Self::new(http_body_util::Empty::new())
+    }
+
+    /// Create a body of unknown size.
+    ///
+    /// This is useful in cases where a body is required to construct a
+    /// response, but the size of the body is not known.
+    ///
+    /// For example, this can be used to respond to `HEAD` requests,
+    /// for which the body of the corresponding `GET` request would be expensive
+    /// to compute. Note that this particular case is also mentioned in
+    /// [RFC 9110 (Section 9.3.2, Paragraph 2)].
+    ///
+    /// The most notable difference compared to an empty body as returned by
+    /// [`Body::empty`] lies in the upper bound returned by [`Body::size_hint`]:
+    /// The upper bound of the size of an empty body is 0 bytes, while there
+    /// is no upper bound for the size of an unknown body.
+    ///
+    /// Other than the size hint, an unknown body behaves like an empty body,
+    /// i.e., [`Body::is_end_stream`] returns `true`, and when polled via
+    /// [`Body::poll_frame`], it immediately returns `Poll::Ready(None)`.
+    ///
+    /// [RFC 9110 (Section 9.3.2, Paragraph 2)]: https://datatracker.ietf.org/doc/html/rfc9110#section-9.3.2-2
+    pub fn unknown() -> Self {
+        Self::new(Unknown::new())
     }
 
     /// Create a new `Body` from a [`Stream`].
--- a/axum-core/src/response/into_response.rs
+++ b/axum-core/src/response/into_response.rs
@@ -128,7 +128,7 @@ impl IntoResponse for StatusCode {
 
 impl IntoResponse for () {
     fn into_response(self) -> Response {
-        Body::empty().into_response()
+        Body::unknown().into_response()
     }
 }
 
--- a/axum/src/routing/route.rs
+++ b/axum/src/routing/route.rs
@@ -239,10 +239,70 @@ impl Future for InfallibleRouteFuture {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::{routing::get, test_helpers::*, Router};
 
     #[test]
     fn traits() {
         use crate::test_helpers::*;
         assert_send::<Route<()>>();
+    }
+
+    #[crate::test]
+    async fn regression_3741() {
+        const BODY: &str = "Very expensive body.";
+        let content_length: HeaderValue = HeaderValue::from_str(&BODY.len().to_string()).unwrap();
+
+        async fn handler(method: http::Method) -> Response {
+            if method == http::Method::HEAD {
+                ().into_response()
+            } else {
+                BODY.into_response()
+            }
+        }
+
+        let client = TestClient::new(Router::new().route("/", get(handler)));
+
+        let get = client.get("/").await;
+        assert_eq!(get.status(), http::StatusCode::OK);
+        assert_eq!(get.headers().get(CONTENT_LENGTH), Some(&content_length));
+
+        let head = client.head("/").await;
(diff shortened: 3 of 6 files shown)


lorenzleutgeb:
While this works to resolve the issue I faced when reporting #3741, I think this is conceptually not perfectly clean.

Consider the definition of [`http_body::Body`](https://docs.rs/http-body/1.0.1/http_body/trait.Body.html). It states:

> The `size_hint` function provides insight into the total number of bytes that will be streamed.

While the value returned from [`http_body::SizeHint::new`](https://docs.rs/http-body/1.0.1/http_body/struct.SizeHint.html#method.new) is not *wrong* here, it is slightly misleading. If ever the size hint of the newly introduced `axum::body::Unknown` is used to allocate buffers, they will be unnecessarily large.

I lack knowledge about axum to judge how much of a problem this might be.

davidpdrsn:
> While the value returned from http_body::SizeHint::new is not wrong here, it is slightly misleading. If ever the size hint of the newly introduced axum::body::Unknown is used to allocate buffers, they will be unnecessarily large.
> 
> I lack knowledge about axum to judge how much of a problem this might be.

I think that's fine. It's like Iterator's size hint, which is allowed to be wrong and shouldn't be relied on for soundness.

lorenzleutgeb:
I am adding two tests.

axum-core/src/response/into_response.rs
```diff
@@ -128,7 +128,7 @@ impl IntoResponse for StatusCode {
 
 impl IntoResponse for () {
     fn into_response(self) -> Response {
-        Body::empty().into_response()
+        Body::unknown().into_response()
```
lorenzleutgeb, line 131: @davidpdrsn please note this change in particular. Tests are passing and it seems to achieve what I want, but for an outsider like me it is hard to judge whether this actually is a sane change.

What is difficult for me to assess is the intention behind the semantics of converting `()` into a response. Up to now, it meant an empty body, i.e. one of known size zero.

This seems to have been introduced in https://github.com/tokio-rs/axum/commit/18f613ff986cf8d5ec62468137f32d6c04be8b05#diff-6f9f8f882acd334c5bf0b598b309afef5f90f63d20b4cc6c26751d66bcb1dba6R19 with a slight change from `::default` to `::empty` in https://github.com/tokio-rs/axum/commit/a005427d40523a625a79a89a394f182077491910#diff-6f9f8f882acd334c5bf0b598b309afef5f90f63d20b4cc6c26751d66bcb1dba6L18-R15

If it is intentional that `()` should convert to the empty body, then my suggestion here is probably bad and should be rejected.

If however it is accidental that `()` converts to the empty body, and it actually fits better to convert to a body of unknown size, then maybe I am on the right track.

axum/src/routing/route.rs
```diff
@@ -239,10 +239,70 @@ impl Future for InfallibleRouteFuture {
...
+        async fn handler() -> Response {
+            ().into_response()
+        }
+
+        let client = TestClient::new(Router::new().route("/", get(handler)));
+
+        let get = client.get("/").await;
+        assert_eq!(get.status(), http::StatusCode::OK);
+        assert_eq!(get.headers().get(CONTENT_LENGTH), Some(&content_length));
+    }
```
davidpdrsn, line 307: Okay I think I see what's going on now and why this test passes:

- `Unknown::is_end_stream` is true, which hyper detects and treats as no body (`None`):
  https://github.com/hyperium/hyper/blob/v1.8.1/src/proto/h1/dispatch.rs#L339-L349
- Later, for `None | Some(BodyLength::Known(0))`, hyper calls
  `Server::can_have_implicit_zero_content_length`. For this GET response it returns true,
  so hyper writes `content-length: 0`:
  https://github.com/hyperium/hyper/blob/v1.8.1/src/proto/h1/role.rs#L907-L916

axum-core/src/body.rs
```diff
@@ -53,6 +56,11 @@ impl Body {
         Self::new(http_body_util::Empty::new())
     }
 
+    /// Create an unknown body.
+    pub fn unknown() -> Self {
+        Self::new(Unknown::new())
+    }
```
davidpdrsn: Might be good to expand a bit on what exactly an "unknown body" is. Maybe we could copy the docs from `pub(crate) struct Unknown<D>` to this method as well, since `Unknown` isn't public.

davidpdrsn (commented):
I've come around! Since it doesn't impact GET responses with `()` I think this solution is good.

One small comment about docs but otherwise LGTM!

Would you mind adding a line in axum and axum-core's changelogs as well?

axum-core/src/body.rs
lorenzleutgeb: I moved the documentation for `Unknown` to this function, and left a note there to refer back to this function, since it is `pub`.

That's done in https://github.com/tokio-rs/axum/pull/3742/commits/f65a1407.

axum/src/routing/route.rs
lorenzleutgeb: Makes sense. Thanks for the explanation!

lorenzleutgeb:
> Would you mind adding a line in axum and axum-core's changelogs as well?

I did that in https://github.com/tokio-rs/axum/pull/3742/commits/f65a1407 but am not sure I managed to match your style of changelogs well enough. Happy to work on it with some more feedback.

axum/src/routing/route.rs
teohhanhui: Do you think putting some of this explanation as a comment would help future travellers?

axum-core/src/response/into_response.rs
teohhanhui: As a user, I think this would be very confusing for the typical user to see, and thus probably deserves a comment?

lorenzleutgeb: Sorry, I am not following. What would the confusion be that should be clarified in that comment?

teohhanhui: To quote yourself: 

> What is difficult for me to assess is the intention behind the semantics of converting () into a response. Up to now, it meant an empty body, i.e. one of known size zero.

I think a comment explaining why it's unknown rather than empty would help? But I'm just a user, so let's see what the maintainers think...

lorenzleutgeb: In the part you were quoting, I was actually pointing this out for a maintainer who would review, so it still is not clear how to approach it from a user's perspective.

Also note that there's a comment at https://github.com/tokio-rs/axum/pull/3742/changes#diff-a2a1ba736818ea9c311768df47cde954acaa0089aa330267431e7b2de07e47b2R59-R78

teohhanhui: Actually what I'm trying to say is just that the semantics of `()` is non-obvious here. So as you've run into the same confusion while making this change, I think it'd help future readers of the code, users and maintainers alike, to have a comment explaining why it uses `unknown` here.

i.e. why `()` translates to `unknown`, not `empty`

davidpdrsn (approved):
LGTM!

PR #3751, opened 2026-05-07
feat(axum-extra): support deserializing Vec<Enum> from repeated query parameters

### LINKED ISSUE #3696: Supporting deserialize Vec<Enum> from query strings

## Feature Request

I have the following handler to return multiple accounts based on query strings:

```rust
async fn list(State(cx): State<Arc<Api>>, Query(params): Query<Vec<List>>) -> Result<Json<Vec<Account>>, Error> {
    let mut res = Vec::new();

    for q in params {
        let a = match q {
            List::Id(v) => cx.account_db.load_by_id(v).await?,
            List::Username(v) => cx.account_db.load_by_username(v).await?,
        };

        res.push(a);
    }

    Ok(Json(res))
}

#[derive(Deserialize)]
#[serde(rename_all = "camelCase")]
enum List {
    Id(u32),
    Username(String),
}
```

The query can specify multiple predicates like the following:

```
id=123&username=abc&id=345&username=def
```

Currently it is not possible to do this with both built-in `Query` and the one from `axum-extra`.

### Alternatives

Currently I built a custom extractor to achieve this. However I can't share the code since it is belong to my employer. I think this pattern is quite common so I created this request.

Zelys-DFKH:
Fixes #3696

## Summary

This PR adds two new extractors to handle a real pain point: query strings with mixed enum variants. You can now write handlers that accept `?id=123&username=alice&id=456` and get back a properly-typed `Vec<SearchFilter>` with all the variants in order.

## Problem

Axum's query parameter handling works great for flat structs. But if you need to accept different parameter names as different enum variants — like filtering by ID or username in the same request — you're stuck. `serde_html_form` doesn't support this pattern because it doesn't know that the parameter *name* should pick the enum variant.

Your options were: hand-parse the query string, build a custom extractor, or restructure the API. None of them are fun.

## Solution

Two new extractors:
- **`QueryList<T>`** - Deserializes all query parameters into `Vec<T>`, where `T` is an enum. The parameter name becomes the variant discriminator.
- **`OptionalQueryList<T>`** - Same as `QueryList`, but returns an empty vec if there are no parameters (no error on missing data).

### How it works

The trick is simple: for each query parameter pair, create a JSON object `{ variant_name: value }` and let serde's enum deserializer do the work. Serde already knows how to pick the right enum variant from the object key. We just had to bridge the gap between query strings and JSON objects.

The implementation also handles value types correctly: `id=123` deserializes as the number 123 (not the string "123"), and `active=true` becomes a boolean, just like you'd expect from JSON semantics.

## Example

```rust
#[derive(Deserialize)]
#[serde(rename_all = "lowercase")]
enum SearchFilter {
    Id(u32),
    Username(String),
}

async fn search(QueryList(filters): QueryList<SearchFilter>) {
    for filter in filters {
        match filter {
            SearchFilter::Id(id) => println!("Filter by ID: {}", id),
            SearchFilter::Username(name) => println!("Filter by user: {}", name),
        }
    }
}

// GET /search?id=123&username=alice&id=456
// Receives: [Id(123), Username("alice"), Id(456)]
```

## Testing

Nine tests cover the realistic cases:
- Alternating enum variants in any order
- The same variant repeated (e.g., `?id=1&id=2&id=3`)
- Empty query strings
- Unknown variant names (400 Bad Request)
- Invalid values (type errors)
- Both sync and async extraction paths
- `OptionalQueryList` behavior with and without parameters

All existing tests pass. No breakage.

## Changes

- `axum-extra/src/extract/query.rs` — Added `QueryList<T>`, `OptionalQueryList<T>`, the `deserialize_pair` helper, and rejection types
- `axum-extra/src/extract/mod.rs` — Exported the new types
- `axum-extra/Cargo.toml` — Added `serde_json` dependency for the "query" feature

## Notes

- Reuses the existing `FailedToDeserializeQueryString` rejection type for consistency
- Follows the same patterns as `Query`/`OptionalQuery` so the API surface is familiar
- Works with any `DeserializeOwned` enum — full type safety
- Values parse as JSON types: `id=123` is a number, `active=true` is a boolean, `name=alice` is a string

Natural fit for axum-extra's extraction toolkit. The approach is simple and delegates the hard work to serde, which already knows how to deserialize enums correctly.

### DIFF
--- a/axum-extra/src/extract/query.rs
+++ b/axum-extra/src/extract/query.rs
@@ -956,23 +956,23 @@ mod tests {
 
         let app = Router::new().route(
             "/",
-            get(|OptionalQueryList(params): OptionalQueryList<Param>| async move {
-                params
-                    .iter()
-                    .map(|p| match p {
-                        Param::Name(n) => format!("name:{n}"),
-                        Param::Tag(t) => format!("tag:{t}"),
-                    })
-                    .collect::<Vec<_>>()
-                    .join("|")
-            }),
+            get(
+                |OptionalQueryList(params): OptionalQueryList<Param>| async move {
+                    params
+                        .iter()
+                        .map(|p| match p {
+                            Param::Name(n) => format!("name:{n}"),
+                            Param::Tag(t) => format!("tag:{t}"),
+                        })
+                        .collect::<Vec<_>>()
+                        .join("|")
+                },
+            ),
         );
 
         let client = TestClient::new(app);
 
-        let res = client
-            .get("/?name=john%20doe&tag=hello%26world")
-            .await;
+        let res = client.get("/?name=john%20doe&tag=hello%26world").await;
         assert_eq!(res.status(), StatusCode::OK);
         assert_eq!(res.text().await, "name:john doe|tag:hello&world");
     }
--- a/axum-extra/src/extract/query.rs
+++ b/axum-extra/src/extract/query.rs
@@ -175,6 +175,196 @@ composite_rejection! {
     }
 }
 
+fn deserialize_pair<T: DeserializeOwned>(
+    key: String,
+    value: String,
+) -> Result<T, serde_json::Error> {
+    let mut map = serde_json::Map::new();
+    let parsed_value: serde_json::Value = match value.parse::<serde_json::Number>() {
+        Ok(num) => serde_json::Value::Number(num),
+        Err(_) => match value.as_str() {
+            "true" => serde_json::Value::Bool(true),
+            "false" => serde_json::Value::Bool(false),
+            "null" => serde_json::Value::Null,
+            _ => serde_json::Value::String(value),
+        },
+    };
+    map.insert(key, parsed_value);
+    serde_json::from_value(serde_json::Value::Object(map))
+}
+
+/// Extractor that deserializes query strings into `Vec<T>` where each query parameter
+/// becomes an enum variant.
+///
+/// This is useful for deserializing alternating query parameters of different types,
+/// like `?id=123&username=abc&id=456` into `Vec<List>` where `List` is an enum with
+/// `Id(u32)` and `Username(String)` variants.
+///
+/// `T` is expected to be an enum that implements the `Deserialize` trait with externally-tagged
+/// variants where each variant name matches a query parameter name.
+///
+/// # Example
+///
+/// ```rust,no_run
+/// use axum::{routing::get, Router};
+/// use axum_extra::extract::QueryList;
+/// use serde::Deserialize;
+///
+/// #[derive(Deserialize)]
+/// #[serde(rename_all = "lowercase")]
+/// enum Param {
+///     Id(u32),
+///     Username(String),
+/// }
+///
+/// async fn handler(QueryList(params): QueryList<Param>) {
+///     for param in params {
+///         match param {
+///             Param::Id(id) => println!("ID: {}", id),
+///             Param::Username(name) => println!("Username: {}", name),
+///         }
+///     }
+/// }
+///
+/// let app = Router::new().route("/", get(handler));
+/// # let _: Router = app;
+/// ```
+#[cfg_attr(docsrs, doc(cfg(feature = "query")))]
+#[derive(Debug, Clone, Default)]
+pub struct QueryList<T>(pub Vec<T>);
+
+impl<T, S> FromRequestParts<S> for QueryList<T>
+where
+    T: DeserializeOwned,
+    S: Send + Sync,
+{
+    type Rejection = QueryListRejection;
+
+    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
+        let query = parts.uri.query().unwrap_or_default();
+        let pairs: Vec<(String, String)> = form_urlencoded::parse(query.as_bytes())
+            .map(|(k, v)| (k.into_owned(), v.into_owned()))
+            .collect();
+
+        let mut result = Vec::new();
+        for (key, value) in pairs {
+            let item = deserialize_pair::<T>(key, value)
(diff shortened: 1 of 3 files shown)


yanns:
Could you add some tests where the parameter values have special characters (like `&` or whitespace) that needs to be encoded / decoded.

axum-extra/src/extract/query.rs
```diff
@@ -175,6 +175,196 @@ composite_rejection! {
...
+{
+    type Rejection = QueryListRejection;
+
+    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
+        let query = parts.uri.query().unwrap_or_default();
+        let pairs: Vec<(String, String)> = form_urlencoded::parse(query.as_bytes())
+            .map(|(k, v)| (k.into_owned(), v.into_owned()))
+            .collect();
+
+        let mut result = Vec::new();
```
yanns: In case we keep `pairs`, can we use Vec with capacity of the `pairs` to avoid allocations?

axum-extra/src/extract/query.rs
```diff
@@ -175,6 +175,196 @@ composite_rejection! {
...
+impl<T, S> FromRequestParts<S> for QueryList<T>
+where
+    T: DeserializeOwned,
+    S: Send + Sync,
+{
+    type Rejection = QueryListRejection;
+
+    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
+        let query = parts.uri.query().unwrap_or_default();
+        let pairs: Vec<(String, String)> = form_urlencoded::parse(query.as_bytes())
```
yanns: Do we need to create this temporary Vec or could we use iterator all way long?

axum-extra/src/extract/query.rs
```diff
@@ -212,8 +403,8 @@ mod tests {
 
     #[tokio::test]
     async fn correct_rejection_status_code() {
-        #[derive(Deserialize)]
         #[allow(dead_code)]
+        #[derive(Deserialize)]
```
yanns, line 395: Nitpick: this change is not necessary.

Zelys-DFKH:
Thanks for catching that special character requirement! Added tests with URL-encoded whitespace and ampersand. Also fixed a couple of clippy warnings that showed up in the new run.

patelshudhanshu1999-maker:
I implemented the same feature 
suggestion: the temporary Vec allocation can be avoided 
by iterating directly over form_urlencoded::parse() 
the iterator works fine without collecting first.

Zelys-DFKH:
Good eye. Applied it across all three sites: `QueryList::from_request_parts`, `try_from_uri`, and `OptionalQueryList::from_request_parts` now chain directly into `collect::<Result<Vec<_>, _>>()`. One fewer allocation per call, and cleaner to read.

axum-extra/src/extract/query.rs
```diff
@@ -175,6 +175,184 @@ composite_rejection! {
...
+    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
+        let query = parts.uri.query().unwrap_or_default();
+        let result = form_urlencoded::parse(query.as_bytes())
+            .map(|(k, v)| {
+                deserialize_pair::<T>(k.into_owned(), v.into_owned())
+                    .map_err(FailedToDeserializeQueryString::from_err)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        Ok(Self(result))
```
yanns, line 252: This code is repeated 3 times no? Can you extract it? Or one is using the others?

patelshudhanshu1999-maker: Happy to collaborate! My PR #3756 addresses 
the cap feature from the original issue.
Feel free to incorporate it or we can 
work on a combined solution.

patelshudhanshu1999-maker: In my PR #3756, this duplication is avoided — 
the parsing logic uses a shared deserialize_pair 
helper and QueryList lives in a separate file. 
No repetition.

davidpdrsn:
I think this makes more sense as a standalone crate rather then being part of axum-extra.

We already have [`axum_extra::extract::Query`](https://docs.rs/axum-extra/latest/axum_extra/extract/struct.Query.html) which is primarily intended to support data the way HTML forms work. So I'm reluctant to add yet another query param extractor.

Shudhanshu-nile:
Published a standalone crate for this: [axum-query-list](https://crates.io/crates/axum-query-list)
Supports:

QueryList<T> — deserializes repeated query params into Vec<T>
QueryList<T, CAP> — same but with compile-time cap on total items
OptionalQueryList<T> — returns empty Vec if no params present

As suggested by @davidpdrsn in #3756

davidpdrsn:
~~Great! In that case I'll close this PR.~~

Sorry I probably closed this a bit too quickly but given the overlap between this and https://github.com/tokio-rs/axum/pull/3756 I think a better way forward is for you to collaborate on https://crates.io/crates/axum-query-list

axum-extra/src/extract/query.rs
Zelys-DFKH: @yanns — you're right. Here's the extracted helper:

```rust
fn parse_query_params<T: DeserializeOwned>(
    query: &str,
) -> Result<Vec<T>, FailedToDeserializeQueryString> {
    form_urlencoded::parse(query.as_bytes())
        .map(|(k, v)| {
            deserialize_pair::<T>(k.into_owned(), v.into_owned())
                .map_err(FailedToDeserializeQueryString::from_err)
        })
        .collect()
}
```

All three sites collapse to `Ok(Self(parse_query_params(query)?))`. Should have caught that myself.

If axum-query-list is taking contributions, I can bring this fix along with the URL-encoding and special-character test suite.

PR #3754, opened 2026-05-12
Add test coverage for hello-world example

farrosfr:
#3263

This updates the `hello-world` example to expose an `app()` constructor and adds a focused test for
  `GET /`.

The test verifies that:
- the route responds with `200 OK`
- the response body matches the expected HTML

Tested with:
- cargo fmt --all --check
- cargo test -p example-hello-world

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -1343,7 +1343,9 @@ name = "example-hello-world"
 version = "0.1.0"
 dependencies = [
  "axum",
+ "http-body-util",
  "tokio",
+ "tower",
 ]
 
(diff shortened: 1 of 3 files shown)


davidpdrsn:
Sorry but I don't think we need to add tests to this example. The purpose of it is giving users something they can copy and immediately understand and experiment with. I don't think adding tests helps with that. There are other examples that show how to do testing.

farrosfr:
Thanks for the review. That makes sense.

I’ll close this PR and look for an example where tests would add value without making the example less approachable.

PR #3756, opened 2026-05-15
feat(axum-extra): add QueryList and OptionalQueryList with optional cap

### LINKED ISSUE #3696: Supporting deserialize Vec<Enum> from query strings

patelshudhanshu1999-maker:
## Motivation

Fixes #3696

axum's Query extractor doesn't support deserializing `Vec<Enum>` 
from query strings. The original issue also requested a cap on 
total items (`CappedVec<List, 10>`), which existing PR #3751 
doesn't implement.

## Solution

Two new extractors in `axum-extra`:

- `QueryList<T>` — deserializes repeated query params into `Vec<T>`
- `QueryList<T, 10>` — same but with a compile-time cap on total items
- `OptionalQueryList<T>` — returns empty `Vec` if no params present

Implementation avoids intermediate Vec allocation by chaining 
directly into `collect::<Result<Vec<_>, _>>()`.

7 tests added covering:
- Mixed enum variants in order
- Same variant repeated
- Empty query string
- Unknown variant (400 error)
- Invalid value (400 error)  
- Special characters
- Cap limit enforcement

### DIFF
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -80,6 +80,7 @@ query = [
     "dep:serde_core",
     "dep:serde_html_form",
     "dep:serde_path_to_error",
+    "dep:serde_json",
 ]
 tracing = ["axum-core/tracing", "axum/tracing", "dep:tracing"]
 typed-header = ["dep:headers"]
--- a/axum-extra/Cargo.toml
+++ b/axum-extra/Cargo.toml
@@ -77,6 +77,7 @@ protobuf = ["dep:prost"]
 routing = ["axum/original-uri", "dep:rustversion"]
 query = [
     "dep:form_urlencoded",
+    "dep:serde",
     "dep:serde_core",
     "dep:serde_html_form",
     "dep:serde_path_to_error",
--- a/axum-extra/src/extract/mod.rs
+++ b/axum-extra/src/extract/mod.rs
@@ -18,6 +18,9 @@ mod json_deserializer;
 #[cfg(feature = "query")]
 mod query;
 
+#[cfg(feature = "query")]
+mod query_list;
+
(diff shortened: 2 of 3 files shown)


davidpdrsn:
I feel the same way about this that I feel about https://github.com/tokio-rs/axum/pull/3751

> I think this makes more sense as a standalone crate rather then being part of axum-extra.
> 
> We already have [`axum_extra::extract::Query`](https://docs.rs/axum-extra/latest/axum_extra/extract/struct.Query.html) which is primarily intended to support data the way HTML forms work. So I'm reluctant to add yet another query param extractor.

[🔗](https://github.com/tokio-rs/axum/pull/3751#issuecomment-4458111596)

patelshudhanshu1999-maker:
Closing as per feedback from @davidpdrsn. 
Will publish as a standalone crate instead.

Issue #3017, opened 2024-11-06
Make UrlParams public


Diggsey:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

Make `UrlParams` public.

### Motivation

I want to be able to nest a router under a dynamic path (`/:foo`) and use a middleware to extract the value of `:foo` and do something with it. The middleware should then be able to remove `foo` from the `UrlParams` so that when it calls `Next`, the nested router only sees path parameters that it understands.

### Proposal

Make `UrlParams` public.

### Alternatives

Provide a hygenic way to nest a router under a dynamic path.

Diggsey:
This branch adds a method to remove parameters from the request: https://github.com/tokio-rs/axum/compare/main...Diggsey:axum:axum-v0.7.7-patched?expand=1

mladedav:
I'm not sure about allowing such changes to the url params. I feel like it could lead to more surprises.

But it's true that nested handlers shouldn't have to be concerned about parameters of routers under which they've been nested.

Would it be good enough for your use case if we added another `Path`-like extractor which would expose only captures from the innermost router?

I feel like that somewhat follows the design of `Uri` and `OriginalUri`.

Diggsey:
It would solve my use-case, but I think it could be confusing. Especially if `Path` continues to be the "default".

Axum already hides what constitutes a single router - for example, when you call `nest` or `merge` the routers kindof get combined into one. Whereas if you explicitly call `oneshot` on the inner router it's clearly doing two separate routing steps.

It's unclear how such a Path-like extractor would interact with these.

Another option would be to provide a `.isolate()` method on router, which would be equivalent to:

```rust
    fn isolate(self) -> axum::Router<S> {
        axum::Router::new().fallback(
            |axum::extract::State(state): axum::extract::State<S>,
             mut request: axum::extract::Request| async move {
                clear_url_params(request.extensions_mut());
                self.clone().with_state(state).oneshot(request).await
            },
        )
    }
```

That way you could truly separate two routers.

mladedav:
It would be of course up to discussion if we should keep the current `Path` as default for better backwards compatibility or change it for example to `OriginalPath` for better consistency with `OriginalUri`.

[...]

PR #3757, opened 2026-05-15
feat: Add `RawPathParams::from_request_extensions`

davidpdrsn:
Adds `RawPathParams::from_request_extensions` which can be called outside an `async` context akin to `Json::from_bytes`. Requested here https://github.com/tokio-rs/axum/issues/3017#issuecomment-2733227558


mladedav (approved):
👍

PR #3760, opened 2026-05-15
fix(axum-macros): use OptionalFromRequest for Option<T> fields

### LINKED ISSUE #3623: `FromRequest` derive with `Option<T>` field unexpected semantics

patelshudhanshu1999-maker:
This supersedes #3665 and closes #3623 by implementing proper rejection mapping using OptionalFromRequest/OptionalFromRequestParts and adding regression tests for JSON parsing errors.

## Motivation

Currently, Option<T> fields in #[derive(FromRequest)] and #[derive(FromRequestParts)] use .ok() to handle the extracted result. This approach silently swallows all rejections (such as a JsonRejection when a payload is malformed or invalid) and converts them into None. The expected behavior is that an invalid payload should properly propagate its rejection to the client, and None should only be returned when the payload or part is genuinely missing (relying on the extractor's specific optional semantics).

## Solution

Replaced the .ok() chaining for Option<T> fields with the OptionalFromRequest and OptionalFromRequestParts traits. This delegates the decision of when to return None versus when to propagate a rejection directly to the specific extractor's implementation.

Additionally, added two regression tests (option_via_json.rs and option_via_parts.rs) to ensure that missing data gracefully resolves to None, while malformed data (like invalid JSON) properly surfaces the correct rejection/HTTP error.

### DIFF
--- a/axum-macros/CHANGELOG.md
+++ b/axum-macros/CHANGELOG.md
@@ -9,8 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **breaking:** `#[from_request(via(Extractor))]` now uses the extractor's
   rejection type instead of `axum::response::Response` ([#3261])
+- **breaking:** `Option<T>` fields in `#[derive(FromRequest)]` and
+  `#[derive(FromRequestParts)]` now use `OptionalFromRequest` /
+  `OptionalFromRequestParts` instead of calling `.ok()` on the result. This
+  means the extractor decides when to return `None` vs an error, instead of
+  silently converting all rejections to `None`. Extractors used with
+  `Option<T>` must implement `OptionalFromRequest` or
+  `OptionalFromRequestParts`. ([#3623])
 
 [#3261]: https://github.com/tokio-rs/axum/pull/3261
+[#3623]: https://github.com/tokio-rs/axum/issues/3623
 
 # 0.5.1
 
--- a/axum-macros/src/from_request/mod.rs
+++ b/axum-macros/src/from_request/mod.rs
@@ -476,19 +476,25 @@ fn extract_fields(
 
             if peel_option(&field.ty).is_some() {
                 let field_ty = into_outer(via.as_ref(), ty_span, peel_option(&field.ty).unwrap());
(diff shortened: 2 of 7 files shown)


davidpdrsn (commented):
Awesome! LGTM!

Would you mind adding one more test case. Something like:

```rust
#[tokio::main]
async fn main() {
    let _: Router = Router::new().route("/", get(handler))
}

struct Payload;

#[derive(FromRequest)]
struct Args {
    payload: Option<Payload>,
}

async fn handler(_: Args) {}

// No `impl<S> OptionalFromRequest<S> for Payload`
```

which should fail with a message that tells the user that `Payload` doesn't implement `OptionalFromRequest`. I just tested it manually and the error does say this but would still be nice with a test.

patelshudhanshu1999-maker:
Done! ✅ Added the test case as requested.

The new test file `option_without_optional_from_request.rs` verifies that:
- When a type doesn't implement `OptionalFromRequest`, 
- The compiler gives a clear error message: "the trait `OptionalFromRequest<_, _>` is not implemented for `Payload`"
- It also suggests implementing types like `Json<T>` that do implement the trait

All tests pass (51 passed in axum-macros). The error message properly guides users when they encounter this common mistake.

davidpdrsn (approved):
Thanks!

PR #3761, opened 2026-05-16
Drop `sync_wrapper` dependency

zerosnacks:
Note: _this PR was created with assistance of Amp (AI) and manually reviewed. The PR description is largely done with assistance of Amp. Any comments posted are by me alone._ 

## Motivation

`sync_wrapper` is a tiny external dependency that `axum` and `axum-core` pull in for a single purpose: to make stream-wrapping bodies `Sync` even when the inner stream is only `Send`. Only two call sites use it:

- `StreamBody` in `axum-core::body`
- `SseBody` in `axum::response::sse`

Both wrap the entire surface of `sync_wrapper` we actually exercise in ~40 lines: a `#[repr(transparent)]` newtype with `new`, `get_pin_mut`, and `unsafe impl<T> Sync`.

## Changes

- Remove the `sync_wrapper` dependency from both `axum` and `axum-core`.
- Add a minimal in-tree `SyncWrapper<T>` in `axum-core` (re-exported via a `#[doc(hidden)]` path so `axum` can use the same type without duplicating the file).
- Keep the `pin_project!` field as `#[pin] SyncWrapper<S>` in both bodies. **This is important:** `pin_project_lite` generates both `project()` (`Pin<&mut Field>`) and `project_ref()` (`Pin<&Field>`). Wrapping in `SyncWrapper<S>` ensures the shared projection only ever yields `Pin<&SyncWrapper<S>>`, never `Pin<&S>`, so the soundness story is identical to the upstream crate.
- Relax the workspace `unsafe_code` lint from `forbid` to `deny` so the two audited `unsafe` sites in `SyncWrapper` can opt in via `#[allow(unsafe_code)]` with `SAFETY` comments.
- `axum-extra` and `axum-macros` (which contain no `unsafe`) re-assert `#![forbid(unsafe_code)]` at the crate root so their existing safety guarantees are unchanged.
- Add `clippy::undocumented_unsafe_blocks = "deny"` at the workspace level so every `unsafe { ... }` must carry a SAFETY comment.
- Update the `Safety` sections of the `axum` and `axum-core` READMEs to describe the new policy.

## Behavioral equivalence

The polling path is byte-for-byte equivalent:

```rust
// before (external crate)
ready!(this.event_stream.get_pin_mut().poll_next(cx))
// after (in-tree wrapper)
ready!(this.event_stream.get_pin_mut().poll_next(cx))
```

`SyncWrapper::get_pin_mut` is the same `Pin::map_unchecked_mut` identity projection. The in-tree wrapper is API- and ABI-compatible with the subset of `sync_wrapper::SyncWrapper` axum actually used. Both `StreamBody<S>` and `SseBody<S>` remain unconditionally `Sync` for any `S`, same as before.

## Why not just `unsafe impl Sync` on the outer types?

An earlier iteration tried that — inline the field as `#[pin] S` and put `unsafe impl<S> Sync` on `StreamBody`/`SseBody` directly. It compiles and tests pass, but it weakens the soundness invariant: `pin_project_lite`'s generated `project_ref` would then hand out `Pin<&S>` to safe code in the defining module, whereas the wrapper approach guarantees it can only ever yield `Pin<&SyncWrapper<S>>`. Keeping the wrapper preserves the same encapsulation property `sync_wrapper` provides.


mladedav:
I don't think we want to allow unsafe code and vendor this code just to remove one small dependency.

Is there a reason to get rid of that dependency other than "it exists and is small"? Is it unmaintained or anything like that?

zerosnacks:
Hi @mladedav, thanks for your reply!

> I don't think we want to allow unsafe code and vendor this code just to remove one small dependency.

In my opinion there is no difference between using a crate that is unsafe internally or having a vendored version that is unsafe. In my opinion this should be addressed with additional lints and rejected on the PR level.

> Is there a reason to get rid of that dependency other than "it exists and is small"? Is it unmaintained or anything like that?

There is no immediate reason that prompted this other than I think it is becoming obvious due to [supply chain attacks](https://kerkour.com/rust-supply-chain-nightmare) that it is important for large successful projects to shift towards preferring tailored vendored extracts for any crates not managed by large reputable organizations that are trivial to replace.

mladedav:
I'll have to ask you to create an issue first. For this to have impact, we would probably have to vendor much more code and this would be just the first of many similar PRs.

So I think we need a broader consensus first, because what you're proposing would most likely result in increased maintenance burden by basically moving code from other projects here.

If you have any concrete idea what the end goal is (which I assume isn't "no dependencies except hyper and tokio") it would be great to discuss that there too.

zerosnacks:
Thanks @mladedav, I'll do that. I agree that broader consensus is required before moving forward. I'll close the PR for the time being.

PR #3762, opened 2026-05-16
examples: fix `using_serve_dir_with_assets_fallback` to not expose assets at root

### LINKED ISSUE #2809: Unexpected behaviour in example `static-file-server`

### Version

0.7.5

### Description

Unexpected behaviour in example `static-file-server`, function [`using_serve_dir_with_assets_fallback()`](https://github.com/tokio-rs/axum/blob/035c8a36b591bb81b8d107c701ac4b14c0230da3/examples/static-file-server/src/main.rs#L44):

It serves assets (in this case `script.js`) from the url `/assets/script.js`, but unexpectedly also from `/script.js`, because the whole `assets` folder is used as fallback instead of just the `index.html` file.

I expected to see this happen:  
url `/script.js` should fallback to `index.html`

Instead, this happened:  
url `/script.js` responded with `script.js`

Zelys-DFKH:
Closes #2809.
Credit to @tobx, who identified the root cause and expected behavior in the issue thread. This is the implementation.

## Motivation

The SPA example in `using_serve_dir_with_assets_fallback` is supposed to serve assets at `/assets/*` and return `index.html` for everything else. In practice, the router fallback pointed at the full `ServeDir`, so `/script.js` served the actual script rather than the index page.

## Solution

Change the fallback from `ServeDir` to a `ServeFile` for `index.html`. The `/assets` route is unchanged. Files still load from there, and missing ones fall back to `index.html`. Only the root-level fallback changes.

`ServeFile: Clone`, so both uses share one binding instead of two identical constructor calls.

Adds a regression test: `GET /script.js` returns `index.html` content, not the script.

### DIFF
--- a/examples/static-file-server/src/tests.rs
+++ b/examples/static-file-server/src/tests.rs
@@ -1,0 +1,29 @@
+use super::using_serve_dir_with_assets_fallback;
+use axum::{body::Body, http::Request, http::StatusCode};
+use http_body_util::BodyExt;
+use tower::ServiceExt;
+
+const INDEX_HTML_CONTENT: &str = include_str!("../assets/index.html");
+
+// Assets nested under `/assets` must not be reachable at the root.
+// Previously, `fallback_service(serve_dir)` served the whole assets directory
+// as a fallback, so `/script.js` would return the actual script.js instead of
+// the SPA fallback (index.html).
+#[tokio::test]
+async fn assets_not_served_at_root() {
+    let app = using_serve_dir_with_assets_fallback();
+    let response = app
+        .oneshot(
+            Request::builder()
+                .uri("/script.js")
+                .body(Body::empty())
+                .unwrap(),
+        )
+        .await
+        .unwrap();
+
+    assert_eq!(response.status(), StatusCode::OK);
+    let body = response.into_body().collect().await.unwrap().to_bytes();
+    let body_str = String::from_utf8(body.to_vec()).unwrap();
+    assert_eq!(body_str, INDEX_HTML_CONTENT);
+}
--- a/examples/static-file-server/src/tests.rs
+++ b/examples/static-file-server/src/tests.rs
@@ -6,9 +6,6 @@ use tower::ServiceExt;
 const INDEX_HTML_CONTENT: &str = include_str!("../assets/index.html");
 
 // Assets nested under `/assets` must not be reachable at the root.
-// Previously, `fallback_service(serve_dir)` served the whole assets directory
-// as a fallback, so `/script.js` would return the actual script.js instead of
-// the SPA fallback (index.html).
 #[tokio::test]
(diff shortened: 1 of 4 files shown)


examples/static-file-server/src/tests.rs
```diff
@@ -0,0 +1,29 @@
...
+use axum::{body::Body, http::Request, http::StatusCode};
+use http_body_util::BodyExt;
+use tower::ServiceExt;
+
+const INDEX_HTML_CONTENT: &str = include_str!("../assets/index.html");
+
+// Assets nested under `/assets` must not be reachable at the root.
+// Previously, `fallback_service(serve_dir)` served the whole assets directory
+// as a fallback, so `/script.js` would return the actual script.js instead of
+// the SPA fallback (index.html).
```
davidpdrsn: Lets remove this. It'll just confuse people reading it later.

```suggestion
```

examples/static-file-server/src/tests.rs
```diff
@@ -0,0 +1,29 @@
...
+        .oneshot(
+            Request::builder()
+                .uri("/script.js")
+                .body(Body::empty())
+                .unwrap(),
+        )
+        .await
+        .unwrap();
+
+    assert_eq!(response.status(), StatusCode::OK);
```
davidpdrsn: Shouldn't this return 404?

```suggestion
    assert_eq!(response.status(), StatusCode::NOT_FOUND);
```

davidpdrsn (changes requested):

examples/static-file-server/src/tests.rs
Zelys-DFKH: Fair question — I had to look twice at this myself. `using_serve_dir_with_assets_fallback` ends with `.fallback_service(index_html)`, so a request to `/script.js` misses the `/assets` nest and lands on that fallback, which returns 200 with the index.html body. A 404 would only show up without a fallback, like in `using_serve_dir_with_handler_as_service` just below.

The body check is what catches the regression: before the fix, `/script.js` returns the script source; after, it returns index.html. Status is 200 either way.

Happy to rework the test if you'd prefer. You know this codebase far better than I do.

davidpdrsn: I think returning `index.html` is fine, it just seems weird to me to respond with 200 when we don't send `script.js`. The client says "gimme `script.js`" and we say "yeah sure here `index.html`, no errors" 😅

You can change it to return 400 with

```rust
let index_html = SetStatus::new(ServeFile::new("assets/index.html"), StatusCode::NOT_FOUND);
```

The comment above would also need to be changed.

Zelys-DFKH:
Oops. Yeah, 200 there is kind of a lie. Wrapped it in `SetStatus` and updated the test and comments.

davidpdrsn (approved):
Thanks!

PR #3766, opened 2026-05-17
docs: update route_service example for axum 0.7+ Body types

wyf027:
## Summary

Fixes #2536.

Updates `Router::route_service` doc example comments: remove references to `axum::body::BoxBody` (removed in 0.7) and describe `Body` / `IntoResponse` instead.

Same intent as #3410 (rebased on current `main`).

## Test plan

- [x] Documentation only
- [ ] CI

PR #3765, opened 2026-05-17
docs(ws): document WebSocket recv cancellation safety

wyf027:
## Motivation

Fixes #3282.

Users are unsure whether `WebSocket::recv` (and read paths after `split`) are safe to use in `tokio::select!` without losing messages when another branch completes first.

## Solution

Document that:

- `WebSocket::recv` forwards to `StreamExt::next` and inherits its **cancel-safe** semantics.
- After `split`, the read half should use `StreamExt::next` (same guarantees).
- `send` points readers to `SinkExt::send` for write-side cancellation semantics.

This follows the discussion on #3657 (closed after confirming `recv` is cancel-safe, not cancel-unsafe).

## Test plan

- [x] Documentation-only change (`//!` / `///` on `ws.rs`)
- [ ] `cargo test -p axum` (if CI runs on fork)

PR #3767, opened 2026-05-17
docs: WebSocket cancel safety and route_service example updates

### LINKED ISSUE #2536: `Router::route_service()` example code is out of date

### LINKED ISSUE #3282: Cancellation safety for WebSocket's async methods is not documented

wyf027:
Combines two small documentation fixes for easier review:

1. **WebSocket cancellation safety** (`axum/src/extract/ws.rs`) — documents that `WebSocket::recv` is cancel-safe via `StreamExt::next`, with guidance for split streams and `tokio::select!`. Fixes #3282.

2. **route_service example** (`axum/src/docs/routing/route_service.md`) — removes obsolete `BoxBody` references; uses current `Body` / `IntoResponse` types for axum 0.7+. Fixes #2536.

Supersedes #3765 and #3766 (will close those if this is preferred).

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -89,6 +89,24 @@
 //! ```
 //!
 //! [`StreamExt::split`]: https://docs.rs/futures/0.3.17/futures/stream/trait.StreamExt.html#method.split
+//!
+//! # Cancellation safety
+//!
+//! [`WebSocket::recv`] is cancel-safe: it forwards to [`StreamExt::next`], which only
+//! holds a reference to the stream, so dropping the future before it completes does not
+//! discard a received message. This makes it suitable for use in
+//! [`tokio::select!`](https://docs.rs/tokio/latest/tokio/macro.select.html) alongside
+//! other cancel-safe futures (for example [`broadcast::Receiver::recv`]).
+//!
+//! When using [`StreamExt::split`], the read half is a [`SplitStream`] that implements
+//! [`Stream`]; use [`StreamExt::next`] on it for the same cancel-safety guarantees.
+//!
+//! [`WebSocket::send`] uses the [`Sink`] implementation. See [`SinkExt::send`] for its
+//! cancellation semantics when used in `select!`.
+//!
+//! [`broadcast::Receiver::recv`]: https://docs.rs/tokio/latest/tokio/sync/broadcast/struct.Receiver.html#method.recv
+//! [`SplitStream`]: https://docs.rs/futures-util/latest/futures_util/stream/struct.SplitStream.html
+//! [`SinkExt::send`]: https://docs.rs/futures-util/latest/futures_util/sink/trait.SinkExt.html#method.send
 
 use self::rejection::*;
 use super::FromRequestParts;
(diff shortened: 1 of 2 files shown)


axum/src/extract/ws.rs
```diff
@@ -89,6 +89,24 @@
...
+//! holds a reference to the stream, so dropping the future before it completes does not
+//! discard a received message. This makes it suitable for use in
+//! [`tokio::select!`](https://docs.rs/tokio/latest/tokio/macro.select.html) alongside
+//! other cancel-safe futures (for example [`broadcast::Receiver::recv`]).
+//!
+//! When using [`StreamExt::split`], the read half is a [`SplitStream`] that implements
+//! [`Stream`]; use [`StreamExt::next`] on it for the same cancel-safety guarantees.
+//!
+//! [`WebSocket::send`] uses the [`Sink`] implementation. See [`SinkExt::send`] for its
+//! cancellation semantics when used in `select!`.
```
davidpdrsn, line 105: I'm not an expert in cancel safety but is this actually true? Wouldn't it depend on the cancel safety of [`tokio_tungstenite::WebSocketStream`](https://docs.rs/tokio-tungstenite/0.29.0/tokio_tungstenite/struct.WebSocketStream.html) given that we're just wrapping that?

mladedav: I agree it depends on tungstenite, not just on whether we hold references. Not to mention `StreamExt::next` holds a mutable reference so I don't really get the reasoning at all.

As an aside, the account that opened the PR has authored about 800 commits in that one day so I wouldn't expect a reply here and I'd argue for closing this one as too low effort as the reasoning is completely wrong.

PR #3769, opened 2026-05-17
docs(cookie): clarify cookie-key-expansion for Key::derive_from (fixes #2646)

### LINKED ISSUE #2646: Documentation about Key::derive_from lists wrong feature name

### Version

Latest

### Platform

N/A but I'm on Linux

### Crates

axum-extra

### Description

The documentation for `Key::derive_from` (
https://docs.rs/axum-extra/latest/axum_extra/extract/cookie/struct.Key.html#method.derive_from) says its gated behind the feature key-expansion. It should be [cookie-key-expansion](https://github.com/tokio-rs/axum/blob/b6b203b3065e4005bda01efac8429176da055ae2/axum-extra/Cargo.toml#L21) instead. fwict this code is reexported from the cookie crate. However in the process of reexporting the code axum_extra modifies the name of the feature flag, adding the cookie- prefix. This is not reflected in the documentation, which lists key-expansion as the feature to enable. I'm not sure if there is something we can do to make sure the docs are updating when re-exporting the method or is this a limitation of docs.rs?

wyf027:
## Summary
- Document in `extract::cookie` that `Key::derive_from` requires `cookie-key-expansion` on `axum-extra`, not `key-expansion` from the `cookie` crate.
- Add module-level note linking to the crate feature flags table.

## Test plan
- [x] Docs-only change

Fixes #2646

### DIFF
--- a/axum-extra/src/extract/cookie/mod.rs
+++ b/axum-extra/src/extract/cookie/mod.rs
@@ -1,6 +1,12 @@
 //! Cookie parsing and cookie jar management.
 //!
 //! See [`CookieJar`], [`SignedCookieJar`], and [`PrivateCookieJar`] for more details.
+//!
+//! ## `Key::derive_from`
+//!
+//! The [`Key::derive_from`] method is provided by the underlying [`cookie`] crate, whose docs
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
Thanks!

PR #3770, opened 2026-05-17
fix(examples): restore query-params-with-empty-strings example

### LINKED ISSUE #3691: Docs mention query-params-with-empty-strings which doesn't exist

wyf027:
## Summary
- Restores the `query-params-with-empty-strings` example referenced from `Query` docs (present in 0.8.x releases but missing on `main`)
- Re-adds doc links for empty vs missing query values and multi-value params

Fixes #3691

## Test plan
- [ ] `cargo test -p example-query-params-with-empty-strings`
- [ ] Example runs with `cargo run -p example-query-params-with-empty-strings`

### DIFF
--- a/axum/src/extract/query.rs
+++ b/axum/src/extract/query.rs
@@ -36,6 +36,16 @@ use serde_core::de::DeserializeOwned;
 ///
 /// If the query string cannot be parsed it will reject the request with a `400
 /// Bad Request` response.
+///
+/// For handling values being empty vs missing see the [query-params-with-empty-strings][example]
+/// example.
+///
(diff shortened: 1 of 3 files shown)


davidpdrsn:
The example was intentionally removed in https://github.com/tokio-rs/axum/pull/3594 because `serde_html_form` handles this automatically. The user doesn't need at add `#[serde(default, deserialize_with = "empty_string_as_none")]`. The tests all pass even without that attribute.

PR #3771, opened 2026-05-17
docs(routing): document path must start with  in route()

### LINKED ISSUE #3752: Router path compile-time check

### Motivation

I just had an application compile fine, only to immediately crash on boot. It gave me the error message that "Paths must start with a `/`". A reasonable limitation - I was careless. That said, the path was a string literal in this case and should in a perfect world be checked at compile time.

<details><summary>Context</summary>

The goal was to include dependency attribution on a nested router. I was unsure if the nested router needed a leading slash or if that would cause axum to consider this an absolute path instead of a nested one. Without reading the docs - again the crash is on me and this is early dev - I just tried it out, expecting the compile to fail if this was incorrect. Unfortunately, it failed at runtime instead.

<details><summary><code>attribution.rs</code></summary>

```Rust
use axum::{Router, http::{HeaderValue, StatusCode, header}, body::Body, response::Response, routing::get};

pub fn router() -> Router {
    Router::new()
        .route("Cargo.lock", get(dependencies)) // Note the missing slash
}

async fn dependencies() -> Response {
    const CARGO_LOCK: &str = include_str!("../../../Cargo.lock");

    let mut res = Response::new(Body::from(CARGO_LOCK));
    *res.status_mut() = StatusCode::OK;
    res.headers_mut().insert(header::CONTENT_TYPE, HeaderValue::from_static("application/toml"));

    res
}
```

</details>

</details>

### Proposal

At first I just accepted this as a design limitation but then came across http's `HeaderValue::from_static` while trying to figure out the failure modes of `Response::builder` and if it was okay to `.expect` it in my scenario.

I made the flawed assumption that the `const fn from_static` would be executed and panic at compile time but testing it revealed that that's a runtime error (panic in this case) as well. Maybe I misunderstood <https://oneuptime.com/blog/post/2026-01-30-rust-compile-time-constants/view#string-processing-in-const-fn> [^1]

My last idea would be to use a macro that does the check. The error appears to be produced by the `validate_path` function in `/axum/src/routing/path_router.rs`

https://github.com/tokio-rs/axum/blob/c853e44ffce77b771069b6df6bd5adc1b78f1b2f/axum/src/routing/path_router.rs#L22-L34

I'm guessing this is not as easy as just switching this out for a macro due to the need to support dynamic paths but perhaps something can be done to catch errors earlier.

### Alternatives

Be explicit about this in the docstring of `Router::route`, which currently only lists duplicate routes and empty routes under `Panics`. While the examples don't show any routes that don't start with a `/`, none of them are nested so the original confusion about absolute or relative path remains.

Cheers!

Edit: Another consideration may be to have a type that serializes to a correct path string and has a compile-time constructor as well as a runtime one. Or skip the type and provide a `static_route` method that is compile-time checked.

[^1]: Anyways, at this point I spent too much time looking around for possible solutions to just delete this.

wyf027:
## Summary
Documents that `Router::route` panics when `path` does not start with `/`, including the nested-router case from the issue report.

Fixes #3752

## Test plan
- [ ] Docs render correctly on docs.rs

### DIFF
--- a/axum/src/docs/routing/route.md
+++ b/axum/src/docs/routing/route.md
@@ -176,4 +176,6 @@ let app = Router::new()
 The static route `/foo` and the dynamic route `/{key}` are not considered to
 overlap and `/foo` will take precedence.
 
-Also panics if `path` is empty.
+Also panics if `path` is empty or does not start with `/` (for example `"Cargo.lock"`).
+When nesting routers with [`Router::nest`](crate::Router::nest), nested paths are still
+relative segments and must start with `/` (use `"/Cargo.lock"` rather than `"Cargo.lock"`).


davidpdrsn (approved):
Thanks!

PR #3772, opened 2026-05-17
Support RawForm extraction from HEAD query strings

oab24413gmai:
RawForm already extracts query strings for GET requests. This extends the same query extraction behavior to HEAD requests and adds coverage for both empty and populated query strings.

Verification:
- cargo test -p axum raw_form

### DIFF
--- a/axum/src/extract/raw_form.rs
+++ b/axum/src/extract/raw_form.rs
@@ -9,8 +9,8 @@ use super::{
 
 /// Extractor that extracts raw form requests.
 ///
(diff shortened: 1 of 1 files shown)


davidpdrsn (commented):
LGTM!

Do you wanna update the changelog as well?

mladedav:
We've since decided on Discord to fill changelog based on commit messages so I think this is ready for merge.

I've just taken the liberty of updating the docs to also mention `HEAD` requests.

mladedav (approved):

PR #3776, opened 2026-05-22
fix(axum-extra): escape multipart Content-Disposition params

### LINKED ISSUE #3775: axum-extra: Escape multipart Content-Disposition parameters in MultipartForm

### Version

- `axum-extra v0.12.6`
- current `main` also appears affected (`axum-extra/src/response/multiple.rs`)

### Platform

`Darwin MacBookPro.local 25.2.0 Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:40 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T6000 arm64`

### Crates

- `axum-extra` (`response::multiple::MultipartForm`)

### Description

`MultipartForm` currently serializes part `Content-Disposition` headers by interpolating `Part::name` and `Part::filename` directly into quoted-string parameters.

The current implementation in `axum-extra/src/response/multiple.rs` looks like this:

```rust
let mut serialized_part = format!("Content-Disposition: form-data; name=\"{}\"", self.name);

if let Some(filename) = &self.filename {
    serialized_part += &format!("; filename=\"{filename}\"");
}
```

I tried this code:

```rust
use axum_extra::response::multiple::Part;

let part = Part::file(
    "field\"; injected=\"1",
    "evil\"; filename*=UTF-8''pwned.txt; x=\"",
    b"hi".to_vec(),
);

let body = String::from_utf8(part.serialize()).unwrap();
println!("{body}");
```

I expected to see this happen:

- `name` and `filename` remain within a single quoted-string parameter (or invalid values are rejected)
- behavior consistent with the escaping that was added for `Attachment` / `FileStream`

Instead, this happened:

- quotes and backslashes are not escaped before being inserted into the part header
- the serialized multipart body contains a `Content-Disposition` line whose parameter structure can be changed by the input values
- because this path writes raw multipart bytes directly, it also does not get the `HeaderValue` validation that protects the response-header code paths

This looks like the same class of issue that was previously reported in #3646 and fixed in #3664 for `Attachment` and `FileStream`, but `MultipartForm` appears to still have a direct interpolation call site.

A minimal regression test could assert that `name` / `filename` are escaped the same way as the existing `Content-Disposition` helper used elsewhere in `axum-extra`.

fallintoplace:
Closes #3775

## Summary

- reuse the internal `Content-Disposition` quoted-string escaping helper for `MultipartForm`
- reject CR/LF in multipart field names and filenames instead of serializing malformed part headers
- add regression tests for quoted-parameter escaping and invalid multipart metadata

## Root cause

`MultipartForm` serialized `Part::name` and `Part::filename` directly into the part `Content-Disposition` header without escaping or newline validation. That left multipart responses out of line with the earlier `Attachment` / `FileStream` fix and allowed malformed parameter structure to be emitted into the multipart body.

## Testing

- `cargo test -p axum-extra --features "multipart attachment file-stream"`
- `cargo check -p axum-extra --no-default-features --features multipart`
- `cargo check -p axum-extra --no-default-features --features "attachment file-stream"`

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -22,9 +22,12 @@ and this project adheres to [Semantic Versioning].
       .route_with_tsr("/path", get(/* handler */))
       .route_with_tsr("/path", post(/* handler */))
   ```
+- **fixed:** Escape multipart `Content-Disposition` parameters and reject
+  newlines in field names and filenames ([#3776])
 
 [#3599]: https://github.com/tokio-rs/axum/pull/3599
 [#3586]: https://github.com/tokio-rs/axum/pull/3586
+[#3776]: https://github.com/tokio-rs/axum/pull/3776
 
 # 0.12.6
 
--- a/axum-extra/src/response/attachment.rs
+++ b/axum-extra/src/response/attachment.rs
@@ -1,4 +1,4 @@
-use super::content_disposition::EscapedFilename;
+use super::content_disposition::EscapedQuotedString;
 use axum_core::response::IntoResponse;
 use http::{header, HeaderMap, HeaderValue};
 use tracing::error;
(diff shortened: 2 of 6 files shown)


yanns (approved):

yanns:
Would you add an entry in the CHANGELOG file?

fallintoplace:
@yanns I can have a look later in the day.

yanns:
GH actions are not working now: https://www.githubstatus.com/
Let's wait for the incident to be over.

fallintoplace:
@yanns I will catch up on this later.

fallintoplace:
@yanns The CI is finally green.

yanns (approved):
👍

Issue #3753, opened 2026-05-12
Connection rotation / max lifetime


adriangb:
We run a lot of bursty applications on Kubernetes. They tend to look like 1–3 pods of a low-replica coordinator and 4→128 worker pods (depending on load). A problem I've consistently encountered across systems, made worse by this setup, is that connection pools don't respond to scale-up events: the connection pools (e.g. `reqwest`) on the controller pods keep sending work to the same initial 4 worker pods even when those are saturated and another 124 pods are idle waiting for work. To get work to the new pods we'd need to establish *new* connections so that the k8s Service round-robins them across the full pod set.

Unfortunately most HTTP clients don't have any sort of connection rotation feature. But servers *can* pressure clients to rotate connections — or force them to — by closing connections after some bounded lifetime. This works for both HTTP/1 (`Connection: close` on the next response, then close) and HTTP/2 / gRPC (`GOAWAY` with a grace period for in-flight streams).

I'll open a sibling issue against `reqwest`, but I think it's valuable to have this option on the server as well:

1. **Belt and suspenders / defense in depth.** Even if the client supports rotation, having the server enforce it bounds the worst case.
2. **The author of the server is not always the author of the client.** The client's connection pool may not have these options, or the client's author may not be aware of how the server / infra works.

Database connection pools have long had similar options ([`max_lifetime` in sqlx](https://docs.rs/sqlx/latest/sqlx/pool/struct.PoolOptions.html#method.max_lifetime)). On the gRPC side, **tonic already exposes exactly this** via [`Server::max_connection_age`](https://docs.rs/tonic/latest/tonic/transport/struct.Server.html#method.max_connection_age) (plus `max_connection_age_grace` and `max_connection_idle`) — so there's clear precedent for this feature living in an axum-/hyper-adjacent server crate. Envoy has `max_connection_duration` for the same reason.

## Proposed API

```rust
use std::time::Duration;
use axum_server::ConnectionLimits;

let limits = ConnectionLimits::default()
    // Soft cap on total connection lifetime. When reached:
    //   - HTTP/1: set `Connection: close` on the next response, then close
    //     once the current request finishes.
    //   - HTTP/2: send GOAWAY; new streams are refused, existing streams
    //     get up to `max_connection_age_grace` to finish.
    .max_connection_age(Duration::from_secs(10 * 60))

    // Per-connection random jitter added to `max_connection_age`. Critical
    // for avoiding synchronized reconnect storms when many connections
[...]

PR #3779, opened 2026-05-26
feat(axum): add `ConnectionLimits` for bounding connection lifetime

adriangb:
Implements #3753.

Adds `axum::serve::ConnectionLimits`, applied via `Serve::connection_limits` (and the `WithGracefulShutdown` equivalent), to bound the lifetime of individual connections and force clients to rotate connections.

## Motivation

Behind a load balancer that round-robins *new* connections across backends (e.g. a Kubernetes `Service`), a client's connection pool keeps sending work to the backends it first connected to, even after the pool scales up — most HTTP clients don't rotate connections on their own. Servers can pressure clients to rotate by closing connections after a bounded lifetime. This mirrors [`tonic::transport::Server::max_connection_age`](https://docs.rs/tonic/latest/tonic/transport/struct.Server.html#method.max_connection_age) and Envoy's `max_connection_duration`.

## API

```rust
use std::time::Duration;
use axum::serve::ConnectionLimits;

let limits = ConnectionLimits::new()
    .max_connection_age(Duration::from_secs(10 * 60))
    .max_connection_age_jitter(Duration::from_secs(60))
    .max_connection_age_grace(Duration::from_secs(30));

axum::serve(listener, app)
    .connection_limits(limits)
    .await;
```

- **`max_connection_age`** — soft cap on total connection lifetime. When it elapses the connection is gracefully shut down: HTTP/1 sends `Connection: close` after the in-flight request finishes; HTTP/2 (incl. gRPC) sends `GOAWAY`, refusing new streams while in-flight ones finish.
- **`max_connection_age_jitter`** — random per-connection jitter in `[0, jitter]` added to the age limit, to avoid synchronized reconnect storms when many connections were established at once.
- **`max_connection_age_grace`** — hard cap on how long to wait for in-flight work after the age limit fires before forcibly closing.

## Implementation notes

- The mechanism reuses hyper's per-connection `graceful_shutdown` — the same primitive `with_graceful_shutdown` already calls per connection. The per-connection `select!` loop gains an age timer (which triggers `graceful_shutdown`) and a grace timer (which force-closes).
- Jitter uses `std::collections::hash_map::RandomState` for cheap per-connection randomness, avoiding a new RNG dependency.
- `connection_limits` composes with `with_graceful_shutdown` and `with_executor` in any order.

## Scope

Per discussion in #3753, this is age-based rotation only. Deliberately left out (possible follow-ups):

- `max_connection_idle` (needs per-connection IO activity tracking).
- On-demand / "lame-duck" rotation that rotates live connections while still accepting new ones. Note that terminal `with_graceful_shutdown` *already* signals idle connections (`GOAWAY` / `Connection: close`) today.

## Tests

- `random_duration_is_bounded_and_varies`
- `max_connection_age_closes_idle_connection` (HTTP/1) and `_http2` (`GOAWAY`)
- `max_connection_age_grace_force_closes_stuck_connection`
- Extended `if_it_compiles_it_works` type-level coverage

Changelog entry will follow once this PR has a number.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/axum/src/serve/mod.rs
+++ b/axum/src/serve/mod.rs
@@ -137,51 +137,40 @@ where
 /// It also bounds the worst case when a client's connection pool has no
 /// rotation of its own.
 ///
-/// The mechanism differs by protocol but the knobs are the same:
-///
-/// - **HTTP/1**: the next response gets a `Connection: close` header and the
-///   connection is closed once the in-flight request finishes.
-/// - **HTTP/2**: a `GOAWAY` is sent, so new streams are refused
-///   while in-flight streams are allowed to finish.
-///
-/// In both cases in-flight work is waited on for as long as it takes, unless
-/// [`max_connection_age_grace`] is set: once the grace period elapses the
-/// connection is closed even if a request is still in flight. See
-/// [`max_connection_age_grace`] for the trade-off.
+/// Each limit is configured with its own value type, so a limit's modifiers
+/// cannot be set without the limit itself. Currently the only limit is
+/// [`MaxConnectionAge`].
 ///
 /// # Example
 ///
 /// ```
 /// use std::time::Duration;
-/// use axum::{Router, routing::get, serve::ConnectionLifetimeLimits};
+/// use axum::{Router, routing::get, serve::{ConnectionLifetimeLimits, MaxConnectionAge}};
 ///
 /// # async {
 /// let router = Router::new().route("/", get(|| async { "Hello, World!" }));
 /// let listener = tokio::net::TcpListener::bind("0.0.0.0:3000").await.unwrap();
 ///
-/// let limits = ConnectionLifetimeLimits::new()
-///     // Stop accepting new requests on a connection after this long.
-///     .max_connection_age(Duration::from_secs(10 * 60))
-///     // Random per-connection jitter added to the age, to avoid synchronized
-///     // reconnect storms when many connections were established at once.
-///     .max_connection_age_jitter(Duration::from_secs(60))
-///     // Hard cap on how long to wait for in-flight work after the age limit
-///     // fires before forcibly closing.
-///     .max_connection_age_grace(Duration::from_secs(30));
+/// let limits = ConnectionLifetimeLimits::new().max_connection_age(
+///     MaxConnectionAge::new(Duration::from_secs(10 * 60))
+///         // Random per-connection jitter added to the age, to avoid
+///         // synchronized reconnect storms when many connections were
+///         // established at once.
+///         .jitter(Duration::from_secs(60))
+///         // Hard cap on how long to wait for in-flight work after the age
+///         // limit fires before forcibly closing.
+///         .grace(Duration::from_secs(30)),
+/// );
 ///
 /// axum::serve(listener, router)
 ///     .connection_lifetime_limits(limits)
 ///     .await;
 /// # };
 /// ```
-///
-/// [`max_connection_age_grace`]: ConnectionLifetimeLimits::max_connection_age_grace
 #[derive(Clone, Debug, Default)]
 #[must_use]
 pub struct ConnectionLifetimeLimits {
-    max_connection_age: Option<Duration>,
-    max_connection_age_jitter: Duration,
-    max_connection_age_grace: Option<Duration>,
+    max_connection_age: Option<MaxConnectionAge>,
 }
 
 impl ConnectionLifetimeLimits {
@@ -190,29 +179,66 @@ impl ConnectionLifetimeLimits {
         Self::default()
     }
 
-    /// Set a cap on how long a connection keeps accepting new requests.
+    /// Cap how long a connection keeps accepting new requests.
     ///
-    /// Once a connection has been open for this long, a graceful shutdown of
-    /// that connection is started: HTTP/1 connections close after the in-flight
-    /// request completes (sending `Connection: close`), and HTTP/2 connections
-    /// send a `GOAWAY`, refusing new streams while letting in-flight ones finish
-    /// (bounded by [`max_connection_age_grace`] if set).
+    /// See [`MaxConnectionAge`] for what the limit does and how to configure
+    /// its jitter and grace period.
     ///
-    /// Consider also setting [`max_connection_age_jitter`] to avoid all
-    /// connections opened around the same time tearing down simultaneously.
-    ///
-    /// The clock starts when the connection task is spawned, which happens
-    /// after the connection has been accepted and the service for it has been
-    /// created, rather than at accept time.
-    ///
-    /// [`max_connection_age_grace`]: ConnectionLifetimeLimits::max_connection_age_grace
-    /// [`max_connection_age_jitter`]: ConnectionLifetimeLimits::max_connection_age_jitter
-    pub fn max_connection_age(mut self, age: Duration) -> Self {
-        self.max_connection_age = Some(age);
+    /// Accepts a [`MaxConnectionAge`] or an `Option<MaxConnectionAge>`, so an
+    /// optional value derived from configuration can be passed straight
+    /// through. `None` leaves connection age unbounded, which is the default.
+    pub fn max_connection_age(mut self, age: impl Into<Option<MaxConnectionAge>>) -> Self {
+        self.max_connection_age = age.into();
         self
     }
+}
 
-    /// Set the maximum random jitter added to [`max_connection_age`].
+/// A cap on how long a connection keeps accepting new requests, applied by
+/// [`serve`] via [`ConnectionLifetimeLimits::max_connection_age`].
+///
+/// Once a connection has been open for [`new`]'s duration (plus [`jitter`]), a
+/// graceful shutdown of that connection is started. The clock starts when the
+/// connection task is spawned, which happens after the connection has been
+/// accepted and the service for it has been created, rather than at accept
+/// time. The mechanism differs by protocol:
+///
+/// - **HTTP/1**: the next response gets a `Connection: close` header and the
+///   connection is closed once the in-flight request finishes.
+/// - **HTTP/2**: a `GOAWAY` is sent, so new streams are refused while in-flight
+///   streams are allowed to finish.
+///
+/// In both cases in-flight work is waited on for as long as it takes, unless
+/// [`grace`] is set: once the grace period elapses the connection is closed even
+/// if a request is still in flight. See [`grace`] for the trade-off.
+///
+/// [`new`]: MaxConnectionAge::new
+/// [`jitter`]: MaxConnectionAge::jitter
+/// [`grace`]: MaxConnectionAge::grace
+#[derive(Clone, Debug)]
+#[must_use]
+pub struct MaxConnectionAge {
+    age: Duration,
+    jitter: Duration,
+    grace: Option<Duration>,
+}
+
+impl MaxConnectionAge {
+    /// Create a new `MaxConnectionAge` that stops a connection from accepting
+    /// new requests once it has been open for `age`.
+    ///
+    /// Consider also setting [`jitter`] to avoid all connections opened around
+    /// the same time tearing down simultaneously.
+    ///
+    /// [`jitter`]: MaxConnectionAge::jitter
+    pub fn new(age: Duration) -> Self {
+        Self {
+            age,
+            jitter: Duration::ZERO,
+            grace: None,
+        }
+    }
+
+    /// Set the maximum random jitter added to the age.
     ///
     /// Each connection adds a random duration in `[0, jitter]` to its age limit.
     /// This is important for avoiding synchronized reconnect storms when many
@@ -220,34 +246,29 @@ impl ConnectionLifetimeLimits {
     /// deploy): without it, every connection opened in the same instant tears
     /// down in the same instant once the age limit elapses.
     ///
-    /// Defaults to `Duration::ZERO`, i.e. no jitter. Has no effect unless
-    /// [`max_connection_age`] is also set.
-    ///
-    /// [`max_connection_age`]: ConnectionLifetimeLimits::max_connection_age
-    pub fn max_connection_age_jitter(mut self, jitter: Duration) -> Self {
-        self.max_connection_age_jitter = jitter;
+    /// Defaults to `Duration::ZERO`, i.e. no jitter.
+    pub fn jitter(mut self, jitter: Duration) -> Self {
+        self.jitter = jitter;
         self
     }
 
-    /// Set a hard cap on how long to wait for in-flight work after
-    /// [`max_connection_age`] fires before forcibly closing the connection.
+    /// Set a hard cap on how long to wait for in-flight work after the age
+    /// limit fires before forcibly closing the connection.
     ///
-    /// Without a grace period, [`max_connection_age`] only stops new requests:
-    /// the server waits however long it takes for in-flight work to finish
-    /// before closing the connection. Setting a grace period turns
-    /// `max_connection_age` (+ jitter) + grace into a hard deadline: when it
-    /// elapses the connection is closed *even if a request is still in flight*,
-    /// and the client never receives a response for it. This applies to HTTP/1
-    /// requests as well as HTTP/2 streams, so a handler that runs longer than
-    /// the age limit plus the grace period will never complete successfully.
-    /// Only set a grace period if bounding connection lifetime matters more
-    /// than letting slow requests finish.
+    /// Without a grace period the age limit only stops new requests: the server
+    /// waits however long it takes for in-flight work to finish before closing
+    /// the connection. Setting a grace period turns age (+ jitter) + grace into
+    /// a hard deadline: when it elapses the connection is closed *even if a
+    /// request is still in flight*, and the client never receives a response for
+    /// it. This applies to HTTP/1 requests as well as HTTP/2 streams, so a
+    /// handler that runs longer than the age limit plus the grace period will
+    /// never complete successfully. Only set a grace period if bounding
+    /// connection lifetime matters more than letting slow requests finish.
     ///
-    /// This has no effect unless [`max_connection_age`] is also set.
-    ///
-    /// [`max_connection_age`]: ConnectionLifetimeLimits::max_connection_age
-    pub fn max_connection_age_grace(mut self, grace: Duration) -> Self {
-        self.max_connection_age_grace = Some(grace);
+    /// Accepts a `Duration` or an `Option<Duration>`. `None` waits for in-flight
+    /// work for as long as it takes, which is the default.
+    pub fn grace(mut self, grace: impl Into<Option<Duration>>) -> Self {
+        self.grace = grace.into();
         self
     }
 }
@@ -822,10 +843,11 @@ async fn handle_connection<L, M, S, B, E>(
         // elapses we start a graceful shutdown of this connection and re-arm the
         // timer with the grace period (if any), which then bounds how long we
         // wait before forcibly closing.
-        let max_age = connection_lifetime_limits.max_connection_age.map(|age| {
-            let jitter = random_duration(connection_lifetime_limits.max_connection_age_jitter);
-            age.saturating_add(jitter)
-        });
+        let max_connection_age = connection_lifetime_limits.max_connection_age;
+        let max_age = max_connection_age
+            .as_ref()
+            .map(|limit| limit.age.saturating_add(random_duration(limit.jitter)));
+        let grace = max_connection_age.and_then(|limit| limit.grace);
         let mut timer = pin!(sleep_or_pending(max_age));
         let mut age_fired = false;
 
@@ -853,9 +875,7 @@ async fn handle_connection<L, M, S, B, E>(
                     age_fired = true;
                     trace!("max connection age reached, starting graceful shutdown");
                     conn.as_mut().graceful_shutdown();
-                    timer.set(sleep_or_pending(
-                        connection_lifetime_limits.max_connection_age_grace,
-                    ));
+                    timer.set(sleep_or_pending(grace));
                 }
                 Either::Right((Either::Right(_), _)) => {
                     trace!("max connection age grace period elapsed, closing connection");
@@ -947,7 +967,7 @@ mod tests {
 
     #[cfg(unix)]
     use super::IncomingStream;
-    use super::{serve, ConnectionLifetimeLimits, Listener};
+    use super::{serve, ConnectionLifetimeLimits, Listener, MaxConnectionAge};
     #[cfg(unix)]
     use crate::extract::connect_info::Connected;
     use crate::{
@@ -1132,10 +1152,14 @@ mod tests {
         .with_executor(exec);
 
         // connection_lifetime_limits, composable with the other builder methods in any order
+        let optional_age: Option<MaxConnectionAge> = None;
         let limits = ConnectionLifetimeLimits::new()
-            .max_connection_age(Duration::from_secs(60))
-            .max_connection_age_jitter(Duration::from_secs(10))
-            .max_connection_age_grace(Duration::from_secs(5));
+            .max_connection_age(optional_age)
+            .max_connection_age(
+                MaxConnectionAge::new(Duration::from_secs(60))
+                    .jitter(Duration::from_secs(10))
+                    .grace(Duration::from_secs(5)),
+            );
         serve(TcpListener::bind(addr).await.unwrap(), router.clone())
             .connection_lifetime_limits(limits.clone());
         serve(TcpListener::bind(addr).await.unwrap(), router.clone())
@@ -1546,7 +1570,8 @@ mod tests {
         tokio::spawn(
             serve(listener, app)
                 .connection_lifetime_limits(
-                    ConnectionLifetimeLimits::new().max_connection_age(Duration::from_secs(10)),
+                    ConnectionLifetimeLimits::new()
+                        .max_connection_age(MaxConnectionAge::new(Duration::from_secs(10))),
                 )
                 .into_future(),
         );
@@ -1599,11 +1624,9 @@ mod tests {
 
         tokio::spawn(
             serve(listener, app)
-                .connection_lifetime_limits(
-                    ConnectionLifetimeLimits::new()
-                        .max_connection_age(Duration::from_secs(10))
-                        .max_connection_age_grace(Duration::from_secs(5)),
-                )
+                .connection_lifetime_limits(ConnectionLifetimeLimits::new().max_connection_age(
+                    MaxConnectionAge::new(Duration::from_secs(10)).grace(Duration::from_secs(5)),
+                ))
                 .into_future(),
         );
 
@@ -1662,7 +1685,8 @@ mod tests {
         tokio::spawn(
             serve(listener, app)
                 .connection_lifetime_limits(
-                    ConnectionLifetimeLimits::new().max_connection_age(Duration::from_secs(10)),
+                    ConnectionLifetimeLimits::new()
+                        .max_connection_age(MaxConnectionAge::new(Duration::from_secs(10))),
                 )
                 .into_future(),
         );
@@ -1705,7 +1729,8 @@ mod tests {
         tokio::spawn(
             serve(listener, app)
                 .connection_lifetime_limits(
-                    ConnectionLifetimeLimits::new().max_connection_age(Duration::from_secs(10)),
+                    ConnectionLifetimeLimits::new()
+                        .max_connection_age(MaxConnectionAge::new(Duration::from_secs(10))),
                 )
                 .into_future(),
         );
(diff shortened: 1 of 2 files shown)


adriangb:
@davidpdrsn @yanns this is the first piece of the work discussed in #3753. This seemed like a good stopping point for one PR. The code is LLM generated based on the discussion in the issue + my guidance. I reviewed the code by hand and it looks good to me, although I am not that knowledgeable about Axum internals.

axum/src/serve/mod.rs
```diff
@@ -625,6 +824,18 @@ async fn handle_connection<L, M, S, B, E>(
...
+                    age_fired = true;
+                    trace!("max connection age reached, starting graceful shutdown");
+                    conn.as_mut().graceful_shutdown();
+                    if let Some(grace) = connection_limits.max_connection_age_grace {
+                        grace_timer.set(OptionFuture::from(Some(tokio::time::sleep(grace))));
+                    }
+                }
+                Some(()) = grace_timer.as_mut() => {
+                    trace!("max connection age grace period elapsed, closing connection");
+                    break;
```
yanns, line 878: With this, a task that runs longer than max_connection_age + max_connection_age_grace will never complete.

Example of a test reproducing that:
```

    async fn max_connection_age_grace_does_not_cut_inflight_request() {
        use std::sync::Arc;
        use tokio::sync::Notify;

        let started = Arc::new(Notify::new());
        let released = Arc::new(Notify::new());

        let app = Router::new().route("/", {
            let started = started.clone();
            let released = released.clone();
            get(move || {
                let started = started.clone();
                let released = released.clone();
                async move {
                    started.notify_one();
                    released.notified().await;
                    "done"
                }
            })
        });

        let (client, server) = io::duplex(1024);
        let listener = ReadyListener(Some(server));

        // age=60s, grace=5s: the request starts before the age limit, runs
        // past age+grace (65s), and should still complete when released.
        tokio::spawn(
            serve(listener, app)
                .connection_limits(
                    ConnectionLimits::new()
                        .max_connection_age(Duration::from_secs(60))
                        .max_connection_age_grace(Duration::from_secs(5)),
                )
                .into_future(),
        );

        let stream = TokioIo::new(client);
        let (mut sender, conn) = hyper::client::conn::http1::handshake(stream).await.unwrap();
        tokio::spawn(conn);

        let request = Request::builder().uri("/").body(Body::empty()).unwrap();
        let send = tokio::spawn(async move { sender.send_request(request).await });

        // Wait until the handler is actually running.
        started.notified().await;

        // Advance paused time past age(60s) + grace(5s) = 65s. The age timer
        // fires at t=60s and calls graceful_shutdown() — for HTTP/1 this
        // should let the in-flight request finish. The grace timer must not
        // override that and force-close the connection at t=65s.
        tokio::time::sleep(Duration::from_secs(66)).await;

        // Release the handler. The response should arrive successfully because
        // the connection is still alive (graceful_shutdown() for HTTP/1 waits
        // for the current response before closing).
        released.notify_one();

        let result = tokio::time::timeout(Duration::from_secs(5), send)
            .await
            .expect("send task did not complete")
            .unwrap();

        assert!(
            result.is_ok(),
            "expected in-flight request to complete after max_connection_age fired, \
             but the grace timer force-closed the connection: {result:?}",
        );
    }
```

Is it what we want?
Is it up to the users to decide if they want to use the grace and kill existing connections?
If this is the case, maybe we should mention this effect in the documentation.

axum/src/serve/mod.rs
```diff
@@ -613,6 +799,19 @@ async fn handle_connection<L, M, S, B, E>(
...
         let mut signal_closed = pin!(signal_tx.closed().fuse());
 
+        // Soft cap on the connection's lifetime (with optional jitter). When it
+        // elapses we start a graceful shutdown of this connection, and the grace
+        // timer (if any) bounds how long we then wait before forcibly closing.
+        let max_age = connection_limits.max_connection_age.map(|age| {
+            let jitter = connection_limits
+                .max_connection_age_jitter
+                .map_or(Duration::ZERO, random_duration);
+            tokio::time::sleep(age + jitter)
```
yanns: This panics on `Duration` overflow.

`std::ops::Add` for `Duration` is implemented as:

```rust
fn add(self, rhs: Duration) -> Duration {
    self.checked_add(rhs).expect("overflow when adding durations")
}
```

In real-world, the users should pick values small enough to that it never happens. If we want to avoid the panic, we can use:
```
tokio::time::sleep(age.saturating_add(jitter))
```
and then the max duration would be used in case of overflow.

yanns:
@adriangb have you seen my comments on this PR?

adriangb:
> @adriangb have you seen my comments on this PR?

sorry i missed them! i will address today or tomorrow

axum/src/serve/mod.rs
adriangb: Good catch, straightforward fix. I went with `tokio::time::sleep(age.saturating_add(jitter))` as you suggested.

axum/src/serve/mod.rs
adriangb: I do think this is what we want. This is the same behavior as Tonic\*, which I am largely modeling this feature after.

> With this, a task that runs longer than max_connection_age + max_connection_age_grace will never complete.

I think it is worth clarifying what happens because "never completes" could mean it hangs, etc:
- On the client side: the client will receive an abrupt connection error.
- On the server side: the task will be dropped, all work inside of it will be immediately and forcefully aborted.

>  Is it up to the users to decide if they want to use the grace and kill existing connections?

Yes, they can not set `max_connection_age_grace()` (or set it to some very large number) and then connections will not be killed.

>  If this is the case, maybe we should mention this effect in the documentation.

Added a reference.

\*: as Tonic's _documented_ behavior. Tonic's current implementation seems to actually deviate from its own docs (it never sends the GOAWAY when a grace is set).

adriangb:
Thanks for the review @yanns! I think I've addressed your points in a way that seems satisfactory. Looking forward to the next round of review.

adriangb:
@yanns could I bug you for another review? Thanks!

yanns:
I'm available again in August

adriangb:
fyi @yanns I've rebased and resolved conflicts if you have time to review

yanns (approved):
Thanks, this is looking good to me!

axum/src/serve/mod.rs
```diff
@@ -117,10 +121,176 @@ where
...
+/// individual connections.
+///
+/// Closing connections after a bounded lifetime pressures clients to establish
+/// *new* connections, which is useful behind a load balancer (e.g. a Kubernetes
+/// `Service`) that round-robins new connections across the current set of
+/// backends: without rotation, a client's connection pool keeps sending work to
+/// whichever backends it first connected to, even after the pool has scaled up.
+/// It also bounds the worst case when a client's connection pool has no
+/// rotation of its own. This mirrors `tonic`'s `max_connection_age` and Envoy's
+/// `max_connection_duration`.
```
mladedav: I don't think we need this rationale of similar features in our docs. The rest is fine.

axum/src/serve/mod.rs
```diff
@@ -117,10 +121,176 @@ where
...
+///
+/// The mechanism differs by protocol but the knobs are the same:
+///
+/// - **HTTP/1**: the next response gets a `Connection: close` header and the
+///   connection is closed once the in-flight request finishes.
+/// - **HTTP/2** (including gRPC): a `GOAWAY` is sent, so new streams are refused
```
mladedav: I'd rather delete gRPC.

- gRPC can be used over all three HTTP versions if someone tries hard enough
- axum doesn't care about gRPC specifically, it's just another protocol inside http

axum/src/serve/mod.rs
```diff
@@ -117,10 +121,176 @@ where
...
+///   while in-flight streams are allowed to finish.
+///
+/// In both cases in-flight work is waited on for as long as it takes, unless
+/// [`max_connection_age_grace`] is set: once the grace period elapses the
+/// connection is closed even if a request is still in flight. See
+/// [`max_connection_age_grace`] for the trade-off.
+///
+/// Note that this is distinct from [`ListenerExt::limit_connections`], which
+/// bounds the *number* of concurrent connections rather than their lifetime.
```
mladedav: Maybe `ConnectionLimits` isn't a good name then. If it's specifically just for age limits the name should reflect that.

Ideally the names are distinct enough that this note isn't necessary.

axum/src/serve/mod.rs
```diff
@@ -117,10 +121,176 @@ where
...
+/// ```
+/// use std::time::Duration;
+/// use axum::{Router, routing::get, serve::ConnectionLimits};
+///
+/// # async {
+/// let router = Router::new().route("/", get(|| async { "Hello, World!" }));
+/// let listener = tokio::net::TcpListener::bind("0.0.0.0:3000").await.unwrap();
+///
+/// let limits = ConnectionLimits::new()
+///     // Soft cap on total connection lifetime.
```
mladedav: This is a cap on how long the connection is receiving new requests. I don't think "soft cap" conveys what this does.

axum/src/serve/mod.rs
```diff
@@ -117,10 +121,176 @@ where
...
+/// [`max_connection_age_grace`]: ConnectionLimits::max_connection_age_grace
+/// [`ListenerExt::limit_connections`]: crate::serve::ListenerExt::limit_connections
+#[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
+#[derive(Clone, Copy, Debug, Default)]
+#[must_use]
+pub struct ConnectionLimits {
+    max_connection_age: Option<Duration>,
+    max_connection_age_jitter: Option<Duration>,
+    max_connection_age_grace: Option<Duration>,
+}
```
mladedav: I'd argue here for making public fields and using the builder pattern. I don't think it gives us anything here.

Separately, I also think if we want the builder pattern, I'd argue to let people pass `Option` inside the methods so that they can disable the options if need be.

axum/src/serve/mod.rs
```diff
@@ -117,10 +121,176 @@ where
...
+    }
+}
+
+/// Returns a pseudo-random [`Duration`] in `[Duration::ZERO, max]`.
+///
+/// Uses [`RandomState`], whose keys are seeded by the OS and bumped on each
+/// construction, to get cheap per-connection randomness without pulling in a
+/// dedicated RNG dependency.
+///
+/// [`RandomState`]: std::collections::hash_map::RandomState
```
mladedav: This implementation detail that shouldn't be in docs.

axum/src/serve/mod.rs
```diff
@@ -117,10 +121,176 @@ where
...
+/// # };
+/// ```
+///
+/// [`max_connection_age_grace`]: ConnectionLimits::max_connection_age_grace
+/// [`ListenerExt::limit_connections`]: crate::serve::ListenerExt::limit_connections
+#[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
+#[derive(Clone, Copy, Debug, Default)]
+#[must_use]
+pub struct ConnectionLimits {
+    max_connection_age: Option<Duration>,
```
mladedav: `max_connection_age` shouldn't be optional if it's required for this to make any sense.

axum/src/serve/mod.rs
```diff
@@ -117,10 +121,176 @@ where
...
+    if max.is_zero() {
+        return Duration::ZERO;
+    }
+
+    let rand = std::collections::hash_map::RandomState::new()
+        .build_hasher()
+        .finish();
+
+    let max_nanos = max.as_nanos();
+    let nanos = u128::from(rand) % (max_nanos + 1);
```
mladedav: `rand` is `u64`, therefore `max` larger than `u64::MAX` nanoseconds doesn't make sense. Which is over 584 years if I'm not mistaken.

That said any operations with `u128` don't make much sense either. I think it might be faster to cast `max_nanos` to `u64` instaed, or `u64::MAX` if it's outside the range. The result should be the same.

axum/src/serve/mod.rs
```diff
@@ -117,10 +121,176 @@ where
...
+}
+
+/// Returns a pseudo-random [`Duration`] in `[Duration::ZERO, max]`.
+///
+/// Uses [`RandomState`], whose keys are seeded by the OS and bumped on each
+/// construction, to get cheap per-connection randomness without pulling in a
+/// dedicated RNG dependency.
+///
+/// [`RandomState`]: std::collections::hash_map::RandomState
+#[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
```
mladedav: This gate is already on the whole module so this is redundant.

axum/src/serve/mod.rs
```diff
@@ -117,10 +121,176 @@ where
...
+    }
+
+    let rand = std::collections::hash_map::RandomState::new()
+        .build_hasher()
+        .finish();
+
+    let max_nanos = max.as_nanos();
+    let nanos = u128::from(rand) % (max_nanos + 1);
+    // `nanos <= max_nanos` and realistic jitter fits comfortably in `u64`;
+    // saturate in the absurd case rather than truncating.
```
mladedav: Don't mention jitter here, this function doesn't need to be specific for that. I don't think this comment is needed at all.

And as mentioned earlier, this isn't really reachable either since `rand` is `u64` which then binds `nanos` so the `try_from` would succeed everytime anyway.

axum/src/serve/mod.rs
```diff
@@ -117,10 +121,176 @@ where
...
+
+/// Sleeps for `duration`, or never completes if it's `None`.
+///
+/// Used to model an unset timer without having to poll a future that would
+/// otherwise complete immediately.
+#[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
+fn sleep_or_pending(duration: Option<Duration>) -> Either<tokio::time::Sleep, Pending<()>> {
+    match duration {
+        Some(duration) => Either::Left(tokio::time::sleep(duration)),
+        None => Either::Right(std::future::pending()),
```
mladedav, line 295: I think this might resurrect #3112, can you check please?

axum/src/serve/mod.rs
```diff
@@ -117,10 +121,176 @@ where
...
+/// ```
+///
+/// [`max_connection_age_grace`]: ConnectionLimits::max_connection_age_grace
+/// [`ListenerExt::limit_connections`]: crate::serve::ListenerExt::limit_connections
+#[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
+#[derive(Clone, Copy, Debug, Default)]
+#[must_use]
+pub struct ConnectionLimits {
+    max_connection_age: Option<Duration>,
+    max_connection_age_jitter: Option<Duration>,
```
mladedav: `None` is the same thing as `Some(Duration::ZERO)` so I don't think this should be optional.

mladedav (commented):
Some of these might be just personal opinions. @yanns would you mind weighing in on some, at least at the API regarding builder pattern vs normal public struct?

axum/src/serve/mod.rs
adriangb: Removed in 2fa591de (also dropped the matching `Mirrors tonic's max_connection_age_grace` line from the grace docs for consistency).

axum/src/serve/mod.rs
adriangb: Removed the mention in b381ba36.

axum/src/serve/mod.rs
adriangb: Makes sense. Reworded in 81cdd328: the setter is now documented as "a cap on how long a connection keeps accepting new requests", and I removed "soft cap" from some other spots that used that terminology.

axum/src/serve/mod.rs
adriangb: 👍🏻 jitter is now a non-optional `Duration` defaulting to `Duration::ZERO` (38ffa3da)

axum/src/serve/mod.rs
adriangb: Good point, done in cfc29228, using clamped u64s as you suggested.

axum/src/serve/mod.rs
adriangb: 👍🏻 removed

axum/src/serve/mod.rs
adriangb: Both resolved by cfc29228

axum/src/serve/mod.rs
adriangb: I dropped all four gates this PR had added in 829d61e0. The remaining ones in the file predate this PR (main has 15 with the same redundancy). I suggest those be tackled in a separate PR. I'm happy to open a tracking issue if you think it's worth it.

axum/src/serve/mod.rs
adriangb: Renamed to `ConnectionLifetimeLimits` in 511536ac, let me know if you think that resolves the ambiguity.

adriangb:
Thanks for the review @mladedav.

The main remaining issue I see is use of a builder pattern vs. public fields.

The reasons why I think a builder is a better approach here are:

1. We may add more fields in the future. I elected to not add `max_connection_idle` (described in #3753) because of implementation complexity, but it could be added in the future. If we did want to add more fields having an exhaustive struct with public fields would make doing so a breaking change. We'd have to mark it as `#[non_exhaustive]`, at which point users can't use struct literal syntax anyway and are forced into `let mut limits = ConnectionLifetimeLimits::default(); limits.max_connection_age = Some(...)` which seems unergonomic.

2. It lets us express relationships or invariants that should be upheld, as well as evolve with less breakage. For example, your flag that "max_connection_age shouldn't be optional if it's required for this to make any sense" is a good callout. With a builder we have the options of restructuring how these knobs relate without breaking anyone: we could later group the age modifiers structurally (e.g. `.max_connection_age_with(age, jitter, grace)` or a nested `MaxConnectionAge` value `.max_connection_age(MaxConnectionAge::new(age_limit).with_jitter(...).with_grace(...))` or add sibling limits like `max_connection_idle`, all additively. 

On letting the setters take `Option` to allow un-setting: I think it'd make the common case worse (`.max_connection_age(Some(dur))`) and programmatically un-setting isn't a common workflow: the struct is `Copy` and built once at startup. We can always add `Option` taking setter variants later.

So the one choice I will flag: if we want to make the current config builder as ergonomic and correct for _today_ as we can I would suggest changing it to:

```rust
ConnectionLifetimeLimits::new(max_age)
  .jitter(Duration::from_secs(60))
  .grace(Duration::from_secs(30));
```

This means you can't set `jitter` without setting `max_age`.

On the other hand, if we want to group these options right now so that it's nicer once/if we introduced an option like `max_connection_idle` I would suggest:

```rust
ConnectionLifetimeLimits::default()
  .with_max_age(
    MaxConnectionAge::new(max_age)
        .jitter(...)
        .grace(...)
  )
  // hypothetical new option to be added
  .with_max_idle(...)
```

I'm curious what @yanns thinks.

yanns:
@adriangb do you think that this hypothetical new option (max idle) is something that might come up? Do you have an use-case for this?

In general, to decide for/against the builder pattern, we need to check its complexity versus its value. I tend to be for it so that we can evolve it and keeping it backwards compatible.

yanns:
@mladedav ok for you going with this change?

adriangb:
> In general, to decide for/against the builder pattern, we need to check its complexity versus its value. I tend to be for it so that we can evolve it and keeping it backwards compatible.

I made https://github.com/adriangb/axum/pull/2 to show the tradeoff. I also lean towards the builder (i.e. keeping the PR as is).

> do you think that this hypothetical new option (max idle) is something that might come up? Do you have an use-case for this?

I do think it will come up, it's a common option. The main reason to defer it here is complexity. I would suggest we build so that it's possible to add later but still defer it to be added later.

adriangb:
I made a minor change, removing `Copy` in favor of `Clone`.

Sorry for the noise with the closing and opening (hit the wrong button) and force pushes. I took a detailed look at the API as it stands now and the paths we could go down.

On public fields vs. builders: I'm quite convinced we should keep the builder as it is in this PR now, and I think that's your leaning too @yanns. Just so we can see the difference, https://github.com/adriangb/axum/pull/2 is this same PR with public fields instead. It works out to about 17 lines of `impl` block once you discount the doc comments, since those just move onto the fields.

On the optionality of `max_connection_age`: this is @mladedav's point in https://github.com/tokio-rs/axum/pull/3779#discussion_r3757040143. Setting `max_connection_age_jitter` without `max_connection_age` compiles, runs and does nothing, and we're handwaving over it with a doc line on each setter.

Either way, the shape currently in this PR isn't the right answer:

1. https://github.com/adriangb/axum/pull/3 keeps `ConnectionLifetimeLimits` as a container and moves `jitter` and `grace` onto a `MaxConnectionAge` value. Adding a sibling limit later is one more field behind the method we already have.

2. https://github.com/adriangb/axum/pull/4 drops the container entirely: `ConnectionLifetimeLimits` becomes `MaxConnectionAge` and is applied via `Serve::max_connection_age`. Adding a sibling limit later is a new method on `Serve` and on `WithGracefulShutdown`.

Both of them make jitter-without-an-age unrepresentble, and either can add `max_connection_idle`. The real cost of (2) is that every future limit has to be mirrored as top level methods on both builders (`Serve` and `WithGracefulShutdown`).

What we have today is the one option neither answer justifies: it does not enforce that `max_connection_age_jitter` can't be set without setting `max_connection_age` and we can't migrate to either shape later without breaking people. 

I lean towards (1), since I do think `max_connection_idle` will come up. 

@mladedav since this is your thread, does either of those look right to you? Happy to squash whichever we pick into this PR.

axum/src/serve/mod.rs
mladedav: That's a good name, thanks

axum/src/serve/mod.rs
mladedav: Yeah, you definitely don't need to fix it in this PR. We want to pull the whole serve out possibly so then we'll hopefully clean this up.

mladedav:
Yeah, after all I agree the builder is probably better here.

That said, I still think the constructor could take a value for the lifetime as no lifetime is invalid configuration when we want to build this. Then only valid states should be representable I think.

Did you have some time to check the mentioned issue from #3112? It's fine if not and if you believe it's out of scope of what you're doing here, but _someone_ should.

axum/src/serve/mod.rs
jlizen: Strongly agree, `new()` should force max age, and I would remove the `Default` implementation unless we want to set a reasonable default (my knee-jerk is to punt on that since it varies so widely based on use case)

axum/src/serve/mod.rs
jlizen: ++ will need an answer on this @adriangb before we can move forward with this

axum/src/serve/mod.rs
jlizen: nit: "intentionally simple" is getting stretched a bit thin now that we have two knobs, might need rephrasing

axum/src/serve/mod.rs
```diff
@@ -620,21 +814,49 @@ async fn handle_connection<L, M, S, B, E>(
         let mut conn = pin!(builder.serve_connection_with_upgrades(io, hyper_service));
         let mut signal_closed = pin!(signal_rx.changed().fuse());
 
+        // Age limit for the connection (with optional jitter). When it
```
jlizen, line 838: Might be worth noting that the clock starts on task task, rather than accept?

axum/src/serve/mod.rs
```diff
@@ -1271,4 +1510,221 @@ mod tests {
             app.into_make_service(),
         );
     }
+
+    #[test]
```
jlizen, line 1535: Some additional tests I'd like to see:
- http/2 grace period test
- a request *finishing* within the grace period succeeds
- a more integration-y test that jitter is actually added to the deadline
- something testing the interaction of a graceful shutdown signal and age timer, eg:

```
age=10s + grace=5s + with_graceful_shutdown(rx), start a stuck handler, fire the signal at t=1s, advance past t=15s, assert the in-flight request errors and the serve future resolves rather than hanging.
```

axum/src/serve/mod.rs
adriangb: How do you folks feel about the proposal in https://github.com/adriangb/axum/pull/3?

jlizen: Seems good to me

jlizen: I tend to prefer typestate pattern + flat per field builder methods, but the single one holding a struct works too. Not a strong feeling.

axum/src/serve/mod.rs
adriangb: added these tests in 6aed36e

axum/src/serve/mod.rs
adriangb: tweaked this comment in 6aed36e

axum/src/serve/mod.rs
adriangb: Reworded in 6aed36e

axum/src/serve/mod.rs
adriangb: okay thank you, i will merge that in

axum/src/serve/mod.rs
adriangb: I did take a look at this. Disclosure: I'm not familiar with the issue and these are evidently complex things to reason about so I used AI assistance to reason through it, but I checked the conclusions to the best of my understanding.

My understanding of #3112: the leaked task was a `tokio::spawn`ed task whose entire
body was `pending()`. `pending()` returns `Pending` without ever registering a
waker, so nothing could ever wake that task, and the runtime parked it and never
dropped it. That's what tokio-console reports as a lost waker.

I don't think either half of that applies here:

1. The `pending()` is one branch of the per-connection `select`, alongside the
   hyper connection and `signal_rx.changed()`, and both of those do register
   wakers. So the task always has a live waker.
2. It's owned by the connection task and dropped with it, rather than outliving
   the server.

We measured this with `RuntimeMetrics::num_alive_tasks()` on a current thread
runtime, counting alive tasks after the client connection closes and then again
after the serve future is dropped:

| | after conn closes | after serve dropped |
|---|---|---|
| no limits set | 1 (the serve task) | 0 |
| age fired, no grace (timer re-armed to `pending()`) | 1 (the serve task) | 0 |
| control: `with_graceful_shutdown(pending())` | 1 | 1 |

That last row does look like the original bug still present on main if a user supplies `pending()` (we fixed the case where _we_ supply it via `Serve`). We can track in a followup issue if you think that makes sense.

axum/src/serve/mod.rs
```diff
@@ -117,10 +121,185 @@ where
...
+    }
+
+    /// Cap how long a connection keeps accepting new requests.
+    ///
+    /// See [`MaxConnectionAge`] for what the limit does and how to configure
+    /// its jitter and grace period.
+    ///
+    /// Accepts a [`MaxConnectionAge`] or an `Option<MaxConnectionAge>`, so an
+    /// optional value derived from configuration can be passed straight
+    /// through. `None` leaves connection age unbounded, which is the default.
```
adriangb: For the setters accepting an `Option<Duration>` this makes sense. In this case I don't think it does. This should accept `MaxConnectionAge` or `Option<MaxConnectionAge>` (I'd lean towards the former).

adriangb: Also I don't think these comments should narrate reasoning for why we accept the parameter types we do.

axum/src/serve/mod.rs
adriangb: I think this has been addressed and superseded by your later review (https://github.com/tokio-rs/axum/pull/3779#issuecomment-5439451344).

We've gone with the builders, enforced that `jitter` and `grace` cannot be set if `max_age` isn't set and we've allowed passing in `Option<Duration>` for `grace` (`jitter` is disabled by passing in 0).

axum/src/serve/mod.rs
mladedav: Nah, I'm happy with this, thanks.

axum/src/serve/mod.rs
```diff
@@ -18,7 +20,9 @@ use futures_util::{
 };
 use http_body::Body as HttpBody;
 use hyper::body::Incoming;
-use hyper_util::rt::{TokioIo, TokioTimer};
+use hyper_util::rt::TokioIo;
+#[cfg(feature = "http1")]
+use hyper_util::rt::TokioTimer;
```
mladedav, line 25: Why this change? I don't see change in use of `TokioTimer` in the diff so I'm not sure why this would be needed.

mladedav (approved):
One last small thing.

Is everyone happy to merge or anyone still has some outstanding concerns/requests?

axum/src/serve/mod.rs
jlizen: I can take a look later tonight

jlizen (approved):
Changes look good, thanks.

One other thing I just noticed - this wouldn't properly apply to websocket connections since hyper passes off the socket upon upgrade. I would just add a note in the setter docs that it doesn't apply to upgraded websocket connections. (Or CONNECT upgrades as well)

adriangb:
> One other thing I just noticed - this wouldn't properly apply to websocket connections since hyper passes off the socket upon upgrade. I would just add a note in the setter docs that it doesn't apply to upgraded websocket connections. (Or CONNECT upgrades as well)

Thanks @jlizen, applied. Also rebased. This is g2g from my end.

adriangb:
Hi folks, anything I can do here to avoid having this bitrot with 3 approvals? Thanks!

jlizen:
Sorry for the delay! I've been traveling and just catching up on the backlog. I should be able to review by EOD.

axum/src/serve/mod.rs
jlizen: TokioTimer is only used under http1 (line 838), so main  emits an unused-import warning in http2-only builds; this gate fixes it.

jlizen:
New changes LGTM, thanks!

PR #3780, opened 2026-05-27
docs(axum): document cancellation safety for WebSocket::recv

### LINKED ISSUE #3282: Cancellation safety for WebSocket's async methods is not documented

JustinCMR:
Closes #3282.

Adds a `# Cancel safety` note to `WebSocket::recv`. The citation routes through the inner [`WebSocketStream`](https://docs.rs/tokio-tungstenite/latest/tokio_tungstenite/struct.WebSocketStream.html)'s `# Cancel safety` section, which landed via [snapview/tokio-tungstenite#378](https://github.com/snapview/tokio-tungstenite/pull/378) (merged 2026-05-27).

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -549,6 +549,15 @@ impl WebSocket {
     /// Receive another message.
     ///
     /// Returns `None` if the stream has closed.
+    ///
+    /// # Cancel safety
+    ///
+    /// This method is cancel-safe. `recv` is `self.next().await` over the inner
+    /// [`WebSocketStream`], whose `Stream` impl gained a `# Cancel safety`
+    /// section in [snapview/tokio-tungstenite#378].
+    ///
+    /// [`WebSocketStream`]: https://docs.rs/tokio-tungstenite/latest/tokio_tungstenite/struct.WebSocketStream.html
+    /// [snapview/tokio-tungstenite#378]: https://github.com/snapview/tokio-tungstenite/pull/378
     pub async fn recv(&mut self) -> Option<Result<Message, Error>> {
         self.next().await
     }


JustinCMR:
Heads up: the failing `check (examples)` job is unrelated to this PR. It's `clippy::useless_borrows_in_formatting` firing on `examples/form/src/main.rs:73` under the current clippy with `-D warnings`, and it fails on every open PR right now, not just this one. This PR only touches `axum/src/extract/ws.rs` and modifies no example. I opened #3781 to fix it.

mladedav:
Does that mean that axum should provide a guarantee based on observation of undocumented behavior of 3rd party crate?

What happens if `tokio-tungestine` decides to make the method cancel-unsafe?

I think the most valuable path forward would be contacting upstream and adding the same guarantee to the methods we depend on in this way. Otherwise, I don't think we should "own" the guarantee and at best state that right now it might be cancel safe and it probably will be later, but we don't know and don't promise anything.

JustinCMR:
That makes sense, agreed on both counts. Owning a guarantee that rests on undocumented upstream behavior is the wrong call, and a future tokio-tungstenite change could break it silently. I'll open an issue on snapview/tokio-tungstenite to get the Stream/poll_next cancel-safety documented there first, then come back to this with a citation to a real contract. Marking this draft in the meantime. Thanks for the read.

axum/src/extract/ws.rs
```diff
@@ -88,7 +88,60 @@
...
+//! }
+//! ```
+//!
+//! When using [`StreamExt::split`], the read half is a [`SplitStream<WebSocket>`] that
+//! implements [`Stream`]. Calling [`StreamExt::next`] on it has the same cancel-safety
+//! guarantees as [`WebSocket::recv`].
+//!
+//! Writes through [`WebSocket::send`] (or the [`Sink`] implementation) are not cancel-safe:
+//! dropping a send future mid-flush may leave a partially written message in the inner
+//! stream. Drive sends to completion before selecting again.
```
vilgotf: Drive-by review, take it or leave it.

This seems overly verbose (I could not highlight everything as it overflows my screen).

Traits (like `Stream`) and non-async methods (like `Stream::poll_next` are _not_ cancel safe because cancellation-safety is a property of _async_ methods. The `StreamExt::next` method is widely known and documented as being cancel safe, so this PR should, in my opinion, merely document that `WebSocket::recv` is cancel safe (as it just forwards to another cancel safe method).

JustinCMR:
Thanks @vilgotf, that is a much cleaner framing and you are right that I was conflating poll-level behavior with the cancel-safety contract. Pushed a revision that shortens this to a single `# Cancel safety` note on `recv` pointing at `StreamExt::next`'s documented guarantee. Diff is now seven lines on `ws.rs`.

@mladedav, does this framing address your concern? `WebSocket::recv` is `self.next().await`, so the guarantee chains through `StreamExt::next`'s documented cancel-safety rather than through any tokio-tungstenite-specific behavior. If you agree, happy to undraft. The upstream tokio-tungstenite docs PR (snapview/tokio-tungstenite#378) can ship independently as a docs improvement either way.

axum/src/extract/ws.rs
```diff
@@ -549,6 +549,13 @@ impl WebSocket {
     /// Receive another message.
     ///
     /// Returns `None` if the stream has closed.
+    ///
+    /// # Cancel safety
+    ///
+    /// This method is cancel-safe. It forwards to
+    /// [`StreamExt::next`](https://docs.rs/futures/0.3.17/futures/stream/trait.StreamExt.html#method.next),
```
davidpdrsn: This link doesn't say anything about cancel safety.

JustinCMR:
Fixed in d5e88d39. The previous link did not document cancel safety; citation now routes through the inner `WebSocketStream`'s `# Cancel safety` section, which landed via snapview/tokio-tungstenite#378 (merged 2026-05-27).

axum/src/extract/ws.rs
```diff
@@ -549,6 +549,15 @@ impl WebSocket {
     /// Receive another message.
     ///
     /// Returns `None` if the stream has closed.
+    ///
+    /// # Cancel safety
+    ///
+    /// This method is cancel-safe. `recv` is `self.next().await` over the inner
+    /// [`WebSocketStream`], whose `Stream` impl gained a `# Cancel safety`
+    /// section in [snapview/tokio-tungstenite#378].
```
vilgotf, line 557: I think David meant for something similar to tokio's description (which highlights its interaction with `select`). Tokio documents cancel safety like:

> This method is cancel safe. If `recv` is used as the event in a [`tokio::select!`](https://docs.rs/tokio/latest/tokio/macro.select.html) statement and some other branch completes first, it is guaranteed that no messages were received on this channel.

edit: The above seems gramatically incorrect. _s/the event/an event/_ or even _s/the event/a branch/_

> `recv` is `self.next().await` [...]

Why the method is cancel safe does not matter to consumers. The equivalence of `Stream`/`recv` and `Sink`/`send` should be documented, or in my opinion, one set of APIs should be removed.

davidpdrsn: >     /// This method is cancel-safe. `recv` is `self.next().await` over the inner
>     /// [`WebSocketStream`], whose `Stream` impl gained a `# Cancel safety`
>     /// section in [snapview/tokio-tungstenite#378].

and

> Fixed in https://github.com/tokio-rs/axum/commit/d5e88d395c8fff80fe507f994a6b1c6eda7ce560. The previous link did not document cancel safety; citation now routes through the inner WebSocketStream's # Cancel safety section, which landed via https://github.com/snapview/tokio-tungstenite/pull/378 (merged 2026-05-27).

feels off. I'm starting to sense that @JustinCMR is just feeding our comments to an LLM and pasting the response back...

@vilgotf would you be up for submitting a PR?

JustinCMR:
That's fair, this is all already covered. I leaned heavily on an LLM for the rewrites here and the citations suffered for it. Closing in favor of @vilgotf taking this forward if they are up for it. Thanks for the time here and the merge on #3781.

PR #3782, opened 2026-05-29
Support affixed typed path captures

### LINKED ISSUE #3681: The TypedPath derive macro breaks for segments with a prefix or suffix

farrosfr:
Fixes #3681

## Motivation

`#[derive(TypedPath)]` currently only recognizes captures when the whole path segment is a capture, such as `{id}`.

That means paths with captures plus a prefix or suffix, such as `/@{username}` or `/files/{name}.json`, cannot be derived even though they are useful route patterns.

## Solution

Update the typed path parser to collect captures inside path segments while preserving the original `TypedPath::PATH`.

This also keeps display formatting percent-encoding capture values and adds trybuild coverage for prefixed, suffixed, mixed, and tuple captures.

Validation:

- `cargo fmt --check`
- `cargo test -p axum-macros --features __private`

### DIFF
--- a/axum-macros/src/typed_path.rs
+++ b/axum-macros/src/typed_path.rs
@@ -362,8 +362,7 @@ fn format_str_from_path(segments: &[Segment]) -> String {
             Segment::Capture(capture, _) => format!("{{{capture}}}"),
             Segment::Static(segment) => segment.to_owned(),
         })
-        .collect::<Vec<_>>()
-        .join("/")
+        .collect()
 }
 
 fn captures_from_path(segments: &[Segment]) -> Vec<syn::Ident> {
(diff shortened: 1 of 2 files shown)


axum-macros/src/typed_path.rs
```diff
@@ -387,23 +386,43 @@ fn parse_path(path: &LitStr) -> syn::Result<Vec<Segment>> {
...
+            segments.push(Segment::Static(rest[..start].to_owned()));
+        }
+
+        rest = &rest[start + 1..];
+
+        if let Some(end) = rest.find('}') {
+            let capture = &rest[..end];
+            if capture.is_empty()
+                || capture.starts_with('{')
+                || capture.ends_with('}')
```
mladedav: This doesn't really make sense, we've found the first `}` so the string cannot contain it.

axum-macros/tests/typed_path/pass/affixed_captures.rs
```diff
@@ -0,0 +1,69 @@
...
+#[typed_path("/users/{user_id}/teams/team-{team_id}")]
+struct MixedCaptures {
+    user_id: u32,
+    team_id: u32,
+}
+
+#[derive(TypedPath, Deserialize)]
+#[typed_path("/@{username}")]
+struct UnnamedPrefixedCapture(String);
+
```
mladedav, line 32: Also add tests for paths with e.g. `/{{escpaed_capture}}/{id}` which should match stuff like `/{escaped_capture}/7`. I think the parsing code does not handle this and says that it's invalid capture.

axum-macros/src/typed_path.rs
```diff
@@ -387,23 +386,43 @@ fn parse_path(path: &LitStr) -> syn::Result<Vec<Segment>> {
...
+                || capture.contains('{')
```
mladedav: This makes the `starts_with('{')` redundant. But see the later comment on test.

mladedav (changes requested):

farrosfr:
Thanks for the review Sir @mladedav . I updated the parser to skip doubled `{` when finding capture starts, removed the redundant capture checks, and added a regression test for `/{{escaped_capture}}/{id}`.

mladedav (approved):

mladedav:
Thanks!

PR #3784, opened 2026-05-31
Exclude tokio/net for WASM builds

itpetey:
## Motivation

I want to implement a custom `axum::serve::Listener` for my WASM runtime, but `axum::serve` needlessly relies on tokio/net, which doesn't compile on WASM.

## Solution

Made tokio/net an optional feature.

### DIFF
--- a/axum/src/serve/listener.rs
+++ b/axum/src/serve/listener.rs
@@ -4,7 +4,6 @@ use std::{
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
-    time::Duration,
 };
 
 use pin_project_lite::pin_project;
@@ -245,6 +244,7 @@ where
     }
 }
 
+#[cfg(feature = "tokio-net")]
 async fn handle_accept_error(e: io::Error) {
     if is_connection_error(&e) {
         return;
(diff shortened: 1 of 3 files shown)


mladedav:
Just to understand, the rest of `tokio` compiles on wasm correctly? So you want to use the `tokio` feature just without the `tokio/net` part?

itpetey:
Yep that's right. More specifically I want to use the `axum::serve` module, which depends on `tokio`. You can test compilation by running: `cargo build --target wasm32-unknown-unknown --no-default-features --features http1,tokio`.

mladedav:
One more thing is the way it's written now, it's a breaking change (if someone used the `impl`s and explicitly listed features, they need to add the new one to the list).

Non-breaking alternative would be adding a `tokio-wasm` feature and leave the `tokio` as is.

itpetey:
I thought of a simpler solution that avoids features altogether.

axum/src/serve/listener.rs
```diff
@@ -32,8 +30,9 @@ pub trait Listener: Send + 'static {
     fn local_addr(&self) -> io::Result<Self::Addr>;
 }
 
-impl Listener for TcpListener {
-    type Io = TcpStream;
+#[cfg(all(feature = "tokio", not(target_arch = "wasm32")))]
```
mladedav: This whole module is already gated behind `tokio`. Let's use just the target arch here. The same applies on lines 244 and 265. On line 53 we don't need to change anything.

mladedav (commented):
Yeah, this is better than changing features, nice idea.

mladedav (approved):

mladedav:
Thanks!

PR #3785, opened 2026-06-01
keep github actions updated

yanns:
Documentation: https://release-plz.dev/docs/github/update
Given this comment, dependabot should continue using pin dependencies: https://github.com/dependabot/dependabot-core/issues/7913#issuecomment-2748641653

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    # Check for updates every Monday
+    schedule:
+      interval: "weekly"


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    # Check for updates every Monday
```
mladedav, line 5: Nit, from my understanding this will run when we merge this and then every 7 days. So this would be true if we happen to merge this on a Monday, otherwise it's wrong.

mladedav (approved):
We don't use dependabot elsewhere, right?

There are [opinions](https://words.filippo.io/dependabot/) to not use dependabot at all since it's often just a churn machine and we'd rather test for security than for new releases. But I think for github actions it should be ok since there won't be that many releases anyway.

.github/dependabot.yml
yanns: It's possible. I understood that "weekly" runs between Sunday and Monday, but. might be wrong.

PR #3786, opened 2026-06-02
document ws cancellation & delegation

vilgotf:
Documents `recv` as cancel safe as well as the difference between `StreamExt::next`/`recv` and `SinkExt::send`/`send` (there is no difference).

Ref: #3780, #3282

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -549,6 +549,13 @@ impl WebSocket {
     /// Receive another message.
     ///
     /// Returns `None` if the stream has closed.
+    ///
+    /// # Cancel safety
+    ///
+    /// This method is cancel safe. If used as a branch in [`tokio::select!`] and
+    /// another branch completes first, then no message is received.
+    ///
+    /// [`tokio::select!`]: https://docs.rs/tokio/latest/tokio/macro.select.html
     pub async fn recv(&mut self) -> Option<Result<Message, Error>> {
         self.next().await
     }
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -550,6 +550,8 @@ impl WebSocket {
     ///
     /// Returns `None` if the stream has closed.
     ///
+    /// This method delegates to [`Stream`].
+    ///
     /// # Cancel safety
     ///
     /// This method is cancel safe. If used as a branch in [`tokio::select!`] and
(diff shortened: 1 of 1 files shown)


axum/src/extract/ws.rs
```diff
@@ -549,11 +549,22 @@ impl WebSocket {
...
+    ///
+    /// [`tokio::select!`]: https://docs.rs/tokio/latest/tokio/macro.select.html
     pub async fn recv(&mut self) -> Option<Result<Message, Error>> {
         self.next().await
     }
 
     /// Send a message.
+    ///
+    /// This method delegates to [`Sink`].
     pub async fn send(&mut self, msg: Message) -> Result<(), Error> {
```
mladedav, line 568: Can we also add here then that not polling to completion means the message is lost?

This is from tokio's mpsc bounded sender' send, I think we can use almost the same thing if I'm not mistaken? Or is this actually the place where we should refer to the implementation's guarantees?

Cancel safety
If send is used as the event in a tokio::select! statement and some other branch completes first, then it is guaranteed that the message was not sent. However, in that case, the message is dropped and will be lost.

mladedav (approved):
This is improvement either way. If you don't want to go back to the send methods cancel safety, we can merge this and revisit this later. (Though with how long it took to add the docs for the simple case someone either has to do it now or never. Feel free to tell me if you don't want to do it though)

axum/src/extract/ws.rs
vilgotf: `SinkExt::send` consists of 3 parts, which makes it not cancel safe:

1. `poll_ready`: make space for the incoming message (flush the internal buffer)
2. `start_send`: buffer the message
3. `poll_flush`: flush the internal buffer

The message may therefore be buffered and/or partially flushed, if another branch completes first.

mladedav: Right, so it's even worse and if the user drops the future, it can cause even protocol errors? Or at least non-deterministically send parts of messages.

Do you want to add that to the documentation here or would you rather we merge this as is?

vilgotf: It should not cause protocol errors as the sink retains the buffered message (and will eventually flush the rest). But whether or not the message is sent at the moment of another branch completes cannot be determined.

> Do you want to add that to the documentation here or would you rather we merge this as is?

I don't feel comfortable extending its guarantees here with Axum. I have not seen anyone document what I mentioned above, so it seems strange to start here (rather than upstream in `futures-rs`).

mladedav: Makes sense, thanks!

PR #3792, opened 2026-06-10
 fix(macros): support prefix and suffix captures in TypedPath segments

### LINKED ISSUE #3681: The TypedPath derive macro breaks for segments with a prefix or suffix

patelshudhanshu1999-maker:
## Motivation
Resolves #3681. Currently, the `TypedPath` derive macro fails to handle path segments with prefixes or suffixes (e.g. `/@{username}`, `/files/{name}.json`), silently treating them as static segments. This forces users to implement `TypedPath` manually for such routes.

## Solution
Instead of splitting path segments by `/` and parsing them, this PR updates the parser (`parse_path`) to scan the path linearly using `.find('{')` and `.find('}')`. This yields alternating `Segment::Static` and `Segment::Capture` segments. 

This approach is highly elegant because:
1. It does not introduce any new variants to the `Segment` enum.
2. It requires zero modifications to the existing code generator logic (`expand_named_fields`, `expand_unnamed_fields`, `expand_unit_fields`, etc.).
3. It naturally supports multiple captures in a single segment (e.g. `/{first}-{last}`).

Added new trybuild pass tests in `axum-macros/tests/typed_path/pass/prefixed_segments.rs` covering all the prefix/suffix scenarios.

### DIFF
--- a/axum-macros/src/typed_path.rs
+++ b/axum-macros/src/typed_path.rs
@@ -362,8 +362,7 @@ fn format_str_from_path(segments: &[Segment]) -> String {
             Segment::Capture(capture, _) => format!("{{{capture}}}"),
             Segment::Static(segment) => segment.to_owned(),
         })
-        .collect::<Vec<_>>()
-        .join("/")
+        .collect()
 }
 
 fn captures_from_path(segments: &[Segment]) -> Vec<syn::Ident> {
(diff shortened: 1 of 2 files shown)


mladedav:
Is this different to #3782?

patelshudhanshu1999-maker:
I noticed #3782 implements a similar scan loop, but this PR goes a step further by introducing compile-time safety checks to prevent runtime matchit router panics (such as multiple parameters in a single segment and misplaced catch-alls) and properly supports escaped braces ({{ and }}). I've updated the PR with these features and test cases.

mladedav:
It looks incredibly similar to the other one to the point that it makes me wonder if you maybe forgot to credit the original author. I also don't see support or tests for the additional features. Do you mind elaborating?

mladedav:
It does look incredibly similar to the other one, did you maybe forget to credit the original author?

patelshudhanshu1999-maker:
Apologies for the confusion. I realized #3782 has been merged and it resolved the initial prefix/suffix capture along with escaping.

However, while testing, I noticed that the current implementation on main still has two critical safety issues that will cause runtime panics in matchit (the router):

Multiple parameters in a single segment (e.g. #[typed_path("/user-{first}-{last}")]) compiles successfully but panics the router at runtime.
Misplaced catch-alls (e.g. #[typed_path("/{*rest}/foo")]) also compiles fine but panics at runtime.
I have opened a new, clean PR focusing purely on adding compile-time safety validations for these cases along with trybuild fail tests: https://github.com/tokio-rs/axum/pull/3797

PR #3795, opened 2026-06-13
Add cargo-fuzz targets for OSS-Fuzz integration

canolgun:
## cargo-fuzz targets for OSS-Fuzz

Adds 3 fuzz targets covering pre-auth HTTP boundaries:
- `fuzz_request` — full HTTP request routing
- `fuzz_uri` — URI parsing (pre-auth)
- `fuzz_headers` — HTTP header parsing (pre-auth)

Part of OSS-Fuzz integration for continuous security testing.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,0 +1,34 @@
+[package]
+name = "axum-fuzz"
+version = "0.0.0"
+edition = "2021"
+publish = false
+
+[package.metadata]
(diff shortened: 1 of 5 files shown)


mladedav:
Closing as the upstream project rejected this as the author created several similar PRs and they've been rejected as ai slop.

If someone actually did this properly, including communication with Google which manages the service, I think this would be great and we'd at least consider it.

PR #3796, opened 2026-06-14
Make SSE `KeepAliveStream::new` public

michaelvanstraten:
## Motivation

I have a route that returns updates for a job that is being executed by the service. If the job has already finished I want to return a `futures::stream::once` with the final state of the job. To make this possible I need to box the stream that is returned.

The current issue is that I can not box the `KeepAliveStream` because the only way to construct it is to call `keep_alive` on `Sse` which returns a `Sse<KeepAliveStream<S>>`.

## Solution

Just make the `KeepAliveStream::new` public.

Wrapping the `futures::stream::once` in a `KeepAliveStream` would probably also work, but just by chance.

### DIFF
--- a/axum/src/response/sse.rs
+++ b/axum/src/response/sse.rs
@@ -587,7 +587,25 @@ pin_project! {
 
 #[cfg(feature = "tokio")]
 impl<S> KeepAliveStream<S> {
-    fn new(keep_alive: KeepAlive, inner: S) -> Self {
+    /// Create a new `KeepAliveStream` from a `KeepAlive` configuration and stream to wrap.
+    ///
+    /// The wrapped stream should yield SSE events as `Result<Event, E>` items. Prefer
+    /// [`Sse::keep_alive`] unless you need to add keep-alive frames without going through [`Sse`].
+    ///
+    /// # Example
+    ///
+    /// ```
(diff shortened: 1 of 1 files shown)


Samielakkad (commented):
Retracted.

michaelvanstraten:
@Samielakkad, I’m sorry to bother you, but do you have any objections to merging this?

michaelvanstraten:
@jlizen, any chance you could look at this?

mladedav:
> If the job has already finished I want to return a futures::stream::once with the final state of the job. To make this possible I need to box the stream that is returned.

So basically you want to either return a full response or an SSE response? This sounds to me more like you'd want to have something like `Either<impl IntoResponse, Sse<S>>` or am I misunderstanding? The full response should also go over SSE and just not send any more updates after the initial answer?

michaelvanstraten:
> The full response should also go over SSE and just not send any more updates after the initial answer?

It’s not a full response in the sense; it is just an `Sse` response which sends a single event. If I were to switch this to `Either<impl IntoResponse, Sse<S>>` and then return just the event json I could no longer use [`EventSource`](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) on the route. I could, of course, just try to give a response that looks like a `text/event-stream`, but what’s the point of it if this could just be public?

PR #3797, opened 2026-06-14
feat(macros): add compile-time validations to TypedPath macro

patelshudhanshu1999-maker:
Currently, the `TypedPath` derive macro compiles successfully but causes runtime panics during router setup in two cases:
1. Multiple captures in a single segment - (e.g. `/user-{first}-{last}`): Panics with `Only one parameter is allowed per path segment`.
2. Misplaced catch-all parameters - (e.g. `/{*rest}/foo`): Panics with `catch-all parameters are only allowed at the end of a path`.

This PR introduces compile-time validation checks for both of these cases directly inside `parse_path` using the macro's native brace-escaped scanner (`find_first_not_double`), raising compile errors instead of runtime crashes.

Includes:
- Compile-time segment capture limit check.
- Compile-time catch-all placement check.
- UI trybuild tests for both failure modes.

### DIFF
--- a/axum-macros/src/typed_path.rs
+++ b/axum-macros/src/typed_path.rs
@@ -404,14 +404,6 @@ fn parse_path(path: &LitStr) -> syn::Result<Vec<Segment>> {
                 return Err(syn::Error::new_spanned(path, "invalid capture in path"));
             }
 
-            let is_catch_all = capture.starts_with('*');
-            if is_catch_all && !rest[end + 1..].is_empty() {
-                return Err(syn::Error::new_spanned(
-                    path,
-                    "Wildcards/catch-alls must be at the end of the path",
-                ));
-            }
-
             segments.push(Segment::Capture(
                 capture.strip_prefix('*').unwrap_or(capture).to_owned(),
                 path.span(),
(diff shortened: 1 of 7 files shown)


axum-macros/src/typed_path.rs
```diff
@@ -402,6 +404,14 @@ fn parse_path(path: &LitStr) -> syn::Result<Vec<Segment>> {
                 return Err(syn::Error::new_spanned(path, "invalid capture in path"));
             }
 
+            let is_catch_all = capture.starts_with('*');
+            if is_catch_all && !rest[end + 1..].is_empty() {
```
mladedav:
```suggestion
            if is_catch_all && rest.len() != end {
```

axum-macros/src/typed_path.rs
```diff
@@ -402,6 +404,14 @@ fn parse_path(path: &LitStr) -> syn::Result<Vec<Segment>> {
...
+                return Err(syn::Error::new_spanned(
+                    path,
+                    "Wildcards/catch-alls must be at the end of the path",
```
mladedav:
```suggestion
                    "Wildcards must be at the end of the path",
```

axum-macros/src/typed_path.rs
```diff
@@ -402,6 +404,14 @@ fn parse_path(path: &LitStr) -> syn::Result<Vec<Segment>> {
...
+                ));
+            }
```
mladedav: Why are the checks here and not in the new dedicated function?

axum-macros/src/typed_path.rs
```diff
@@ -421,6 +431,24 @@ fn parse_path(path: &LitStr) -> syn::Result<Vec<Segment>> {
     Ok(segments)
 }
 
+fn validate_path_segments(value: &str, path: &LitStr) -> syn::Result<()> {
+    for segment in value.split('/') {
+        let mut count = 0;
```
mladedav: Better name please. Count of what?

mladedav (changes requested):
Thist looks good, let's just clean it up a little bit.

axum-macros/src/typed_path.rs
patelshudhanshu1999-maker: Condition thread: Applied the simplification, but with end + 1 — since end is the index of }, rest.len() != end is always true and would reject valid paths like /{*rest}. Used rest.len() != end + 1, equivalent to the original !rest[end + 1..].is_empty().
Message thread: Done.
Dedicated-function thread: moved it into validate_path_segments.
Naming thread: Renamed to capture_count.

axum-macros/src/typed_path.rs
```diff
@@ -421,6 +423,40 @@ fn parse_path(path: &LitStr) -> syn::Result<Vec<Segment>> {
...
+            break;
+        };
+        if rest[..end].starts_with('*') && rest.len() != end + 1 {
+            return Err(syn::Error::new_spanned(
+                path,
+                "Wildcards must be at the end of the path",
+            ));
+        }
+        rest = &rest[end + 1..];
+    }
```
mladedav: I think it would be clearer if we iterated just once and checked the segment in the previous loop if it's a wildcard capture. Then if we have another iteration, that's an error. Or if the wildcard doesn't cover the whole segment.

patelshudhanshu1999-maker: Restructured to a single pass as suggested: the segment loop now tracks wildcard captures — another segment after one is an error, and so is a wildcard that doesn't cover its whole segment (e.g. /files/{*rest}.txt, which previously panicked at router setup). Added a trybuild test for that case.

mladedav (approved):
Thanks

PR #3801, opened 2026-06-17
feat: add QUERY method routing

### LINKED ISSUE #3799: Support for the HTTP QUERY Method

### Motivation

Recently the [RFC 10008](https://www.rfc-editor.org/info/rfc10008/) standardized the HTTP QUERY method, which is also registered at the [IANA HTTP Method Registry](https://www.iana.org/assignments/http-methods/http-methods.xml#methods). In short, this new method is a mix of GET and POST method: it's safe and idempotent like GET but also has a body like POST.

### Proposal

The QUERY HTTP method should be added to `axum::routing::MethodFilter` as well as the `axum::routing::method_routing` module and any other place where HTTP methods might be handled.

### Alternatives

There is no known real alternative other than using either GET or POST method, which is not ideal.

yunz-dev:
Add `MethodFilter::QUERY`, `routing::query[_service]` and `MethodRouter::query[_service]` and `RouterExt::typed_query`

## Motivation
- resolves #3799
> Recently the [RFC 10008](https://www.rfc-editor.org/info/rfc10008/) standardized the HTTP QUERY method, which is also registered at the [IANA HTTP Method Registry](https://www.iana.org/assignments/http-methods/http-methods.xml#methods). In short, this new method is a mix of GET and POST method: it's safe and idempotent like GET but also has a body like POST.

## Solution

> It's already supported in http (https://github.com/hyperium/http/pull/798) but I don't think that it's been released yet, so you can develop against the git version or wait for a release. https://github.com/tokio-rs/axum/issues/3799#issuecomment-4728231376

While developing I put this in the root `Config.toml`
```rust
[patch.crates-io]
http = { git = "https://github.com/hyperium/http.git" }
```

### DIFF
--- a/axum/src/routing/method_filter.rs
+++ b/axum/src/routing/method_filter.rs
@@ -26,23 +26,25 @@ impl MethodFilter {
     /// [extended CONNECT]: https://www.rfc-editor.org/rfc/rfc8441.html#section-4
     /// [HTTP Upgrade Token Registry]: https://www.iana.org/assignments/http-upgrade-tokens/http-upgrade-tokens.xhtml
     /// [`WebSocketUpgrade`]: crate::extract::WebSocketUpgrade
-    pub const CONNECT: Self = Self::from_bits(0b0_0000_0001);
+    pub const CONNECT: Self = Self::from_bits(0b00_0000_0001);
     /// Match `DELETE` requests.
-    pub const DELETE: Self = Self::from_bits(0b0_0000_0010);
+    pub const DELETE: Self = Self::from_bits(0b00_0000_0010);
     /// Match `GET` requests.
-    pub const GET: Self = Self::from_bits(0b0_0000_0100);
+    pub const GET: Self = Self::from_bits(0b00_0000_0100);
     /// Match `HEAD` requests.
-    pub const HEAD: Self = Self::from_bits(0b0_0000_1000);
+    pub const HEAD: Self = Self::from_bits(0b00_0000_1000);
     /// Match `OPTIONS` requests.
-    pub const OPTIONS: Self = Self::from_bits(0b0_0001_0000);
+    pub const OPTIONS: Self = Self::from_bits(0b00_0001_0000);
     /// Match `PATCH` requests.
-    pub const PATCH: Self = Self::from_bits(0b0_0010_0000);
+    pub const PATCH: Self = Self::from_bits(0b00_0010_0000);
     /// Match `POST` requests.
-    pub const POST: Self = Self::from_bits(0b0_0100_0000);
+    pub const POST: Self = Self::from_bits(0b00_0100_0000);
     /// Match `PUT` requests.
-    pub const PUT: Self = Self::from_bits(0b0_1000_0000);
+    pub const PUT: Self = Self::from_bits(0b00_1000_0000);
     /// Match `TRACE` requests.
-    pub const TRACE: Self = Self::from_bits(0b1_0000_0000);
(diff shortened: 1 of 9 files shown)


yunz-dev:
CI will fail until https://github.com/hyperium/http/pull/798 in release

axum/src/routing/method_filter.rs
```diff
@@ -26,23 +26,25 @@ impl MethodFilter {
...
-    pub const POST: Self = Self::from_bits(0b0_0100_0000);
+    pub const POST: Self = Self::from_bits(0b00_0100_0000);
     /// Match `PUT` requests.
-    pub const PUT: Self = Self::from_bits(0b0_1000_0000);
+    pub const PUT: Self = Self::from_bits(0b00_1000_0000);
     /// Match `TRACE` requests.
-    pub const TRACE: Self = Self::from_bits(0b1_0000_0000);
+    pub const TRACE: Self = Self::from_bits(0b01_0000_0000);
+    /// Match `QUERY` requests.
+    pub const QUERY: Self = Self::from_bits(0b10_0000_0000);
```
yunz-dev, line 47: Unsure if the bit mask is meant to follow a certain protocol or it doesn’t matter.

aabolfazl:
I'like to add more sections of RFC if not implemented yet.

Samielakkad (commented):
Retracted.

mladedav:
Considering we have no such tests for options or trace I don't think that bar is fair here. We definitely don't need a test per http verb for "merge conflict detection" and so on. I'd even argue against the allow header checks, that's niche and since query doesn't participate in complex behaviors like the head/get combo does, I don't think that needs explicit tests. Having a well crafted test that tests all the methods would of course be great, but I don't think this PR is place for that.

It would be nice to have a test trying at least a basic request here though.

---

@Samielakkad correct me if I'm wrong, but these reviews you were posting lately are generated by an LLM, right? I don't think they're always very productive and it actually seems like the prompt makes the machine try to find _anything_ to comment on or improve to seem productive instead of being helpful. I don't use AI for reviews regularly myself, but from what I've heard, there is a big noise-to-signal ratio which I feel is happening here.

We're happy for contributions and engagement from the community, including (and maybe especially) in reviews, but I think your workflow would benefit from a bit more hands-on approach to better steer the tools you're using. I understand if that's not something you want to invest your time into and that's fine too.

It would be also helpful to improve the tests. But that's also something where LLMs from my experience really don't shine and instead crank out a bunch of tests testing pretty much the same thing. But if you want to help by making a few tests that you believe would be beneficial, that would be also appreciated.

aatifsyed (approved):

yunz-dev:
Seems like http v1.5.0 is released, I'll fix this up when I get home.

mladedav:
It's not enough to update cargo.lock, you have to bump the version in cargo.toml too since this is a new .minimum version we require to compile.

That's why the msrv check fails, which actually also checks minimum dependency versions.

yunz-dev:
> It's not enough to update cargo.lock, you have to bump the version in cargo.toml too since this is a new .minimum version we require to compile.

ahh.. right. Silly me haha. Should be fixed now

mladedav (approved):
Thanks!

mohsin:
Has this not landed yet on a release? Seem to be unable to consume it yet.

mladedav:
Yes, this is merged but not yet released. We are currently working toward 0.9 release which I believe we should be able to make within a month. Until then you'll probably want to keep using the `any` workaround mentioned in the issue if you need this right now.

PR #3802, opened 2026-06-19
Add `Redirect::found` and `Redirect::moved_permanently`

neeko-cat:
This PR adds the `Redirect::found` and `Redirect::moved_permanently` constructors.
<sub><sup>Note: No AI has been used in the making of this PR. :)</sup></sub>

## Motivation

I am implementing a web server following a strict spec that among other things specifies the HTTP Status Code I must use on the replies. To comply I must use `302 Found` in some cases, and swapping it to the more modern `307 Temporary Redirect` is sadly not acceptable in this case.

Currently `Redirect` only has constructors for `303 See Other`, `307 Temporary Redirect`, and `307 Permanent Redirect`. However, inspecting the code I found a *private* constructor allowing arbitrary status codes and a cute comment that reads:

> We're open to adding more constructors upon request, if they make sense :)

Although unfortunate, I'd say my use-case *makes sense*.

## Solution

I added a function for `302 Found`, and another for `301 Moved Permanently` (following the same reasoning). They're basically clones of the existing constructors with a different Status Code.

## Unresolved questions

- It might be wise to expand the docs of the two functions I added pointing confused users to 307 and 308 instead, unless they know what they are doing.
- Maybe name it `Redirect::moved` instead of `Redirect::moved_permanently`.

### DIFF
--- a/axum/src/response/redirect.rs
+++ b/axum/src/response/redirect.rs
@@ -25,6 +25,20 @@ pub struct Redirect {
 }
 
 impl Redirect {
+    /// Create a new [`Redirect`] that uses a [`301 Moved Permanently`][mdn] status code.
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301
+    pub fn moved_permanently(uri: impl Into<String>) -> Self {
+        Self::with_status_code(StatusCode::MOVED_PERMANENTLY, uri.into())
+    }
+
+    /// Create a new [`Redirect`] that uses a [`302 Found`][mdn] status code.
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302
+    pub fn found(uri: impl Into<String>) -> Self {
+        Self::with_status_code(StatusCode::FOUND, uri.into())
+    }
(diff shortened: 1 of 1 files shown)


mladedav:
I think this is fine to add.

You're right it might be best adding the notes about 307 and 308 being preferred. The full `moved_permanently` name is better than just `moved` in my opinion.

Samielakkad (commented):
This looks like a sensible addition. One small release-note/docs thing before merge: since this adds two public constructors, it should probably get an `axum/CHANGELOG.md` entry.

Also, the docs for `moved_permanently`/`found` may be worth cross-linking to the existing 307/308 constructors or briefly saying these preserve the historical 301/302 semantics. That would help users pick between `found` vs `temporary_redirect` and `moved_permanently` vs `permanent_redirect` without having to know the redirect-method semantics by memory.

mladedav:
Feel free to add changelog, but we plan to backfill it by commit messages, so it's not necessary anymore.

I'd argue against linking 301 from 307 and similar, we want people gently to push towards the newer methods so linking away from them is contra productive. If someone cares about the semantics, they can always look them up in the already provided links to mdn.

PR #3804, opened 2026-06-20
axum-extra: avoid multipart boundary collisions

### LINKED ISSUE #3794: Ensure generated multipart boundaries do not occur in part data
## Summary

`MultipartForm` documents that a multipart boundary should not occur in the data, but the implementation generates one boundary and serializes all parts without checking for collisions.

## Affected code

- `axum-extra/src/response/multiple.rs:36` generates the boundary.
- `axum-extra/src/response/multiple.rs:52` serializes the parts.
- `axum-extra/src/response/multiple.rs:212` documents the boundary invariant.

## Why this matters

The chance of a random collision is tiny, but this module already has all parts in memory. Enforcing the invariant would make the implementation match the documented behavior at low cost.

## Proposed fix

Add a private `generate_boundary_for_parts(&[Part])` helper that retries if the candidate boundary appears in any part body or relevant serialized metadata.

Use a bounded retry loop and return an internal server error only if repeated attempts fail.

## Acceptance criteria

- Boundary generation checks candidate boundaries against all parts.
- A test forces a collision and verifies a different boundary is chosen.
- Normal multipart responses are unchanged.

Sabnock01:
## Motivation

Fixes: #3794

`MultipartForm` generated a random boundary once and used it without checking whether that boundary appeared in any serialized part data. Multipart boundaries are expected to be unique delimiters that do not occur in the enclosed part contents. If the generated boundary appears inside a part body or serialized metadata, the response can become ambiguous for multipart parsers because part data may be interpreted as a delimiter.

The probability of a random collision is extremely small and not addressed in many libs, but `MultipartForm` already owns all parts in memory before serialization, so it can cheaply enforce the documented boundary invariant rather than relying only on randomness.

## Solution

Add bounded retry logic for multipart boundary generation. A candidate boundary is rejected if it appears in any serialized part metadata or body content. If all retry attempts collide, return an internal server error instead of producing an invalid multipart response.

This also adds regression coverage by seeding boundary generation to force a collision, then asserting that the response chooses a different boundary and that the colliding boundary only appears as part content, not as a multipart delimiter.

### DIFF
--- a/axum-extra/src/response/multiple.rs
+++ b/axum-extra/src/response/multiple.rs
@@ -32,8 +32,14 @@ impl MultipartForm {
 
 impl IntoResponse for MultipartForm {
     fn into_response(self) -> Response {
-        // see RFC5758 for details
-        let boundary = generate_boundary();
+        // see RFC 7578 for details
+        let Some(boundary) = generate_boundary_for_parts(&self.parts) else {
+            return (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                "Unable to generate multipart boundary",
+            )
+                .into_response();
+        };
         let mut headers = HeaderMap::new();
         let mime_type: Mime = match format!("multipart/form-data; boundary={boundary}").parse() {
             Ok(m) => m,
@@ -63,7 +69,7 @@ impl IntoResponse for MultipartForm {
 // Valid settings for that header are: "base64", "quoted-printable", "8bit", "7bit", and "binary".
 /// A single part of a multipart form as defined by
 /// <https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4>
-/// and RFC5758.
+/// and RFC 7578.
 #[derive(Debug)]
 pub struct Part {
     // Every part is expected to contain:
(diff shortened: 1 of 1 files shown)


Samielakkad (commented):
The retry loop is a good fix for the ambiguity issue. One test concern: `generated_boundary_does_not_appear_in_part_contents` relies on `fastrand::seed(0)`, which mutates the process-global RNG state.

Because Rust tests can run concurrently, a global seed like this can make the test order-dependent if another test also calls `generate_boundary()` around the same time. It also leaves the global RNG in a deterministic state for later tests in the same process.

It may be safer to make the collision test deterministic without touching global RNG state, for example by factoring `generate_boundary_for_parts` so the candidate generator can be injected in tests, or by testing `Part::contains_boundary` plus a small local retry helper. That would keep the regression coverage without introducing shared-RNG coupling between tests.

mladedav (approved):
Thanks and sorry it took so long.

mladedav:
Also about the previous comment: `fastrand::seed` is thread-local. So this test should be safe. It's true that it might be better to do `get_seed` athe start and then restore it at the end, but the tests don't require nondeterministic random numbers, so this is also fine.

PR #3805, opened 2026-06-20
Prevent "?&" in TypedPath query params

michielp1807:
## Motivation

When using `TypedPath`, serializing query parameters used to produce an ampersand after the question mark in the path:

```rs
#[derive(TypedPath)]
#[typed_path("/users")]
struct Users;

#[derive(Serialize)]
struct Pagination {
    page: u32,
    per_page: u32,
}

let path = Users.with_query_params(Pagination {
    page: 1,
    per_page: 10,
});

assert_eq!(path.to_uri(), "/users?&page=1&per_page=10"); // notice the "?&"
```

While this is technically not a problem (because it is allowed according to the spec as mentioned by the now removed comment in one of the test cases), I think this is not ideal behavior, especially if you end up exposing these paths to the end-users of your website.

I just think it is nicer (and more conventional) to get `/users?page=1&per_page=10` rather than `/users?&page=1&per_page=10`.

## Solution

The query parameter `Serializer` from the `form_urlencoded` crate provides a `for_suffix` constructor that can be used to indicate the starting position of the query parameters. Using this constructor makes it not insert the ampersand for the first query parameter after the question mark, as shown in the updated test cases.

I've also added a test case to make sure paths that end with a question mark are handled correctly.

### DIFF
--- a/axum-extra/src/routing/typed.rs
+++ b/axum-extra/src/routing/typed.rs
@@ -256,7 +256,7 @@ pub trait TypedPath: std::fmt::Display {
     ///     per_page: 10,
     /// });
     ///
-    /// assert_eq!(path.to_uri(), "/users?&page=1&per_page=10");
+    /// assert_eq!(path.to_uri(), "/users?page=1&per_page=10");
     /// ```
     ///
     /// # Panics
@@ -291,10 +291,12 @@ where
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let mut out = self.path.to_string();
-        if !out.contains('?') {
+
+        let params_start = out.find('?').map(|i| i + 1).unwrap_or_else(|| {
             out.push('?');
-        }
-        let mut urlencoder = form_urlencoded::Serializer::new(&mut out);
+            out.len()
+        });
+        let mut urlencoder = form_urlencoded::Serializer::for_suffix(&mut out, params_start);
         self.params
             .serialize(serde_html_form::ser::Serializer::new(&mut urlencoder))
             .unwrap_or_else(|err| {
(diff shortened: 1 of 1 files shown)


axum-extra/src/routing/typed.rs
```diff
@@ -430,7 +427,26 @@ mod tests {
...
 
-        assert_eq!(uri, "/users/1?&foo=foo&bar=123&baz=true&qux=1337");
+        assert_eq!(uri, "/users/1?foo=foo&bar=123&baz=true&qux=1337");
+    }
+
+    #[test]
+    fn with_params_question_mark_no_params() {
+        #[derive(TypedPath)]
+        #[typed_path("/test?")]
+        struct EndsWithQuestionMark;
```
mladedav, line 437: From what I gather this worked before this PR (and there was event the `out.contains('?')` check so I guess it might have been intentional) but I can't tell if this is something we _want_ to support. I'd have to take a closer look but I think this would probably not work for routing so it would exist just for the `to_uri` implementations?

Out of curiosity, why did you add this test? Because of the implementation or did you see this documented somewhere (I can't find it anywhere right now).

michielp1807: I added this test to verify that the edge case got handled correctly. Without the `.map(|i| i + 1)` this test would fail but the others still succeed. I'm fine with removing this test case if paths ending in a question mark are undesired and this test sets a bad example.

mladedav: I'm a bit torn here, because I think it is undesirable, but the code was seemingly built to work with this even before this change.

I'm fine with keeping the test around. If we ever forbid this, it would happen on the macro anyway and at that point this test will fail and we'll know to remove it.

mladedav (approved):

axum-extra/src/routing/typed.rs
michielp1807: Sounds good to me!

PR #3810, opened 2026-06-28
fix(axum-extra): support explicit 0-0 file stream ranges

fallintoplace:
## Summary

- make `FileStream::try_range_response` treat `end` as an explicit inclusive byte offset
- add `FileStream::try_range_response_open_ended` for `bytes=start-` style ranges
- add regression coverage for `bytes=0-0` and open-ended range parsing

## Root cause

`try_range_response` used `end == 0` as an internal sentinel for open-ended ranges. That made `bytes=0-0` indistinguishable from `bytes=0-`, so callers could not request only the first byte even though that is a valid HTTP byte range.

## Notes

Open-ended ranges still have a dedicated API via `try_range_response_open_ended`, but explicit ranges now keep their literal end offset.

## Testing

- `cargo test -p axum-extra --features file-stream`
- `cargo test -p axum-extra --doc --features file-stream`
- `cargo check -p axum-extra --no-default-features --features "attachment file-stream"`

### DIFF
--- a/axum-extra/CHANGELOG.md
+++ b/axum-extra/CHANGELOG.md
@@ -22,9 +22,12 @@ and this project adheres to [Semantic Versioning].
       .route_with_tsr("/path", get(/* handler */))
       .route_with_tsr("/path", post(/* handler */))
   ```
+- **fixed:** Let `FileStream::try_range_response(..., 0, 0)` return the first
+  byte and add `try_range_response_open_ended` for open-ended ranges ([#3810])
 
 [#3599]: https://github.com/tokio-rs/axum/pull/3599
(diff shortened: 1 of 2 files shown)


mladedav:
Looking at the implementation, ranges are much more broken by not supporting suffix ranges and multiple ranges. Doing this much work just support a single byte requests doesn't make sense when the code needs much bigger overhaul.

fallintoplace:
I focused on the end == 0 sentinel making bytes=0-0 unrepresentable, but I agree the current range model is incomplete overall.

mladedav:
I'm going to close this now as it's too much work for almost no benefit.

PR #3811, opened 2026-06-28
fix(axum-core): skip non-data frames in BytesMut

fallintoplace:
## Summary

- make the `BytesMut` extractor skip non-data body frames instead of stopping at the first one
- add a regression test covering an interleaved `data / non-data / data` body
- verify `BytesMut` now matches `Bytes` for that frame sequence

## Root cause

`body_to_bytes_mut` returned early when `frame.into_data()` failed. That meant the `BytesMut` extractor could truncate request bodies after a trailer or any other non-data frame, even though `Body::into_data_stream` already discards non-data frames and keeps polling.

## Testing

- `cargo test -p axum-core`
- `cargo check -p axum-core`

### DIFF
--- a/axum-core/CHANGELOG.md
+++ b/axum-core/CHANGELOG.md
@@ -14,9 +14,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **changed:** `impl IntoResponse for ()` (which gets called by
   `impl IntoResponse for HeaderMap`, `impl IntoResponse for Extensions` and
   others) now returns a body of unknown size.
+- **fixed:** Make the `BytesMut` extractor skip non-data body frames instead
+  of truncating the remaining buffered data ([#3811])
 
 [#3721]: https://github.com/tokio-rs/axum/pull/3721
 [#3742]: https://github.com/tokio-rs/axum/pull/3742
+[#3811]: https://github.com/tokio-rs/axum/pull/3811
 
 # 0.5.6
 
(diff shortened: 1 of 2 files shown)


mladedav:
Aren't trailers sent after the body? Then those can't cause truncation. Do you have an example of a valid request that would have been malformed?

fallintoplace:
The issue I’m trying to fix is more about the http_body::Body abstraction and extractor consistency: BytesMut currently treats the first non-data frame as end-of-body, while Bytes continues consuming the body and only returns the collected data. So for the same Body, BytesMut can produce a different/truncated result compared to Bytes.

mladedav:
Do you have an example request that would show that?

It is an inconsistency and we might want to unify it, but as far as I can tell, both are correct. But first, I want to know whether the stated issue is real or not.

fallintoplace:
I looked into this more, and I don’t think I can provide a valid wire-level HTTP request where trailers are followed by more data. For normal HTTP semantics, trailers are terminal.

The concrete reproducer is an in-process/custom `http_body::Body` passed via `Body::new`, where the same `Request<Body>` extracts differently as `Bytes` vs `BytesMut`: `Bytes` skips non-data frames via `into_data_stream()`, while `BytesMut` currently stops at the first one.

So I’m happy to narrow the PR framing to extractor consistency rather than claiming a valid HTTP request is truncated.

@mladedav

mladedav:
Ok, then please drop the test since it's testing a non-compliant request.

And since you're looking into this, do you see any reason (performance or otherwise) not to use the `BytesMut` and then `freeze`? That way the two implementations can't drift again.

fallintoplace:
I dropped the test since it was relying on a non-compliant frame sequence, and I narrowed the changelog wording to custom-body extractor consistency rather than malformed HTTP.
On the BytesMut + freeze() idea: I looked into that too. I think there is a small performance reason to keep Bytes as-is for now, since the current collect().to_bytes() path can take optimized/less-copying paths in http-body-util, whereas routing everything through BytesMut would always buffer into a new mutable allocation first.
So I kept this PR scoped to making BytesMut match Bytes when custom/in-process bodies yield non-data frames, without changing the existing Bytes fast path. If you’d still prefer the shared implementation for maintenance reasons, I’m happy to switch it.

mladedav (approved):
Thanks

PR #3812, opened 2026-06-29
fix!: resolve inconsistent trailing-slash behavior when nesting routers

### LINKED ISSUE #2659: Consider changing the behavior of / route when nesting

ficolas2:
Fixes #2659

## Motivation

  Fixes inconsistent behavior of `Router::nest()`, making trailing slashes distinctions explicit. These are breaking changes.

  - `nest("/foo", Router::new().route("", handler))` was not possible before, and only matches `/foo` (no trailing slash)
  - `nest("/foo", Router::new().route("/", handler))` matches `/foo/` (with trailing slash)
  - `nest("/foo/", ...)` now panics, trailing slashes on the nest path are disallowed

  ## Solution
  The PR consists of 4 smaller behavioral changes:
  - Allow route("", _)
      - Needed to be able to match the nest prefix exactly without a trailing slash when using nest()
      - This commit is self contained, all tests working.
  - Panic on nesting paths ending with `/`
      - Disallows the ambiguous nest paths like `/foo/`, reducing special cases.
      - The tests are left broken by this commit, and fixed by the two test commits that follow
  - Simplify nested path route to a plain concatenation
      - Now that trailing slashes are disallowed, path_for_nested_route can be simplified to a plain concatenation
      - Tests are broken by this commit, needing the following fix commit, and the last test commit to be fixed
  - Stop adding trailing slashes when stripping prefixes from an exact match
      - Required for the `route("", _)` vs `route("/", _)` distinction to work correctly. Without this change, the path `/foo` being stripped of the prefix `/foo` turns into `/`, matching the handler on route("/", _) when it should match the one on route("", _)

The route("", _) edge case without nesting, that was recommended in the original issue to do nothing, doesn't panic, it is simply unreachable, since HTTP request paths always start with "/"

No new tests have been added, some redundant ones have been deleted, and many have been modified covering the new behaviors.

### DIFF
--- a/axum/src/routing/strip_prefix.rs
+++ b/axum/src/routing/strip_prefix.rs
@@ -86,6 +86,18 @@ fn strip_prefix(uri: &Uri, prefix: &str) -> Option<Uri> {
                     break;
                 }
             }
+            // the path has a trailing slash, but the prefix doesn't, so we don't
+            // remove it
+            //
+            // For example
+            // prefix = /foo
+            // path = /foo/
+            //
+            // The prefix matches, and the new path should be `/`
+            Item::First("") => {
+                *matching_prefix_length.as_mut().unwrap() -= 1;
+                break;
+            }
             // the path had more segments than the prefix but we got a match.
             //
             // For example:
@@ -108,11 +120,13 @@ fn strip_prefix(uri: &Uri, prefix: &str) -> Option<Uri> {
     // panic
     let after_prefix = uri.path().split_at(matching_prefix_length?).1;
 
-    let new_path_and_query = match (after_prefix.starts_with('/'), path_and_query.query()) {
-        (true, None) => after_prefix.parse().unwrap(),
-        (true, Some(query)) => format!("{after_prefix}?{query}").parse().unwrap(),
-        (false, None) => format!("/{after_prefix}").parse().unwrap(),
-        (false, Some(query)) => format!("/{after_prefix}?{query}").parse().unwrap(),
+    let new_path_and_query = match (after_prefix.chars().next(), path_and_query.query()) {
+        (None, None) => "".parse().unwrap(),
+        (None, Some(query)) => format!("?{query}").parse().unwrap(),
+        (Some('/'), None) => after_prefix.parse().unwrap(),
+        (Some('/'), Some(query)) => format!("{after_prefix}?{query}").parse().unwrap(),
+        (Some(_), None) => format!("/{after_prefix}").parse().unwrap(),
+        (Some(_), Some(query)) => format!("/{after_prefix}?{query}").parse().unwrap(),
     };
 
     let mut parts = uri.clone().into_parts();
@@ -241,13 +255,13 @@ mod tests {
         };
     }
 
-    test!(empty, uri = "/", prefix = "/", expected = Some("/"),);
+    test!(empty, uri = "/", prefix = "/", expected = Some(""),);
 
     test!(
         single_segment,
         uri = "/a",
         prefix = "/a",
-        expected = Some("/"),
+        expected = Some(""),
     );
 
     test!(
@@ -276,7 +290,7 @@ mod tests {
         single_segment_trailing_slash,
         uri = "/a/",
         prefix = "/a/",
-        expected = Some("/"),
+        expected = Some(""),
     );
 
     test!(
@@ -325,7 +339,7 @@ mod tests {
         multi_segment_trailing_slash,
         uri = "/a/b/",
         prefix = "/a/b/",
-        expected = Some("/"),
+        expected = Some(""),
     );
 
     test!(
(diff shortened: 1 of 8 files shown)


axum/src/routing/strip_prefix.rs
mladedav, line 81: This might need changing too (maybe just the comment, maybe the code itself, didn't check too closely)

axum/src/routing/strip_prefix.rs
```diff
@@ -86,6 +86,18 @@ fn strip_prefix(uri: &Uri, prefix: &str) -> Option<Uri> {
...
+            //
+            // For example
+            // prefix = /foo
+            // path = /foo/
+            //
+            // The prefix matches, and the new path should be `/`
+            Item::First("") => {
+                *matching_prefix_length.as_mut().unwrap() -= 1;
+                break;
+            }
```
mladedav: How is this different from the path being longer?

If the prefix is `/foo` then `/foo` would return `/foo`, as would `/foo/`, but `/foo/bar` would return `/foo/`?

In either case the match a bit lower checking whether the rest of the path starts with a slash or not would take care of this anyway, wouldn't it?

mladedav (commented):
Thanks, I think this is moving in the right direction but I'm now unsure about the case of people being able to route empty paths on the top-level router which would be unreachable.

I want to think on possibilities of having something like a split between nestable routers nad top level routers so that we could prevent this kinfld of misuse but that might be too high cognitive load for that edge case.

axum/src/routing/strip_prefix.rs
ficolas2: It is not redundant, for example:

path: /foo/
prefix: /foo
matching_prefix_length = 5
after_prefix = /foo/
(None, None)  => ""

So /foo/ - /foo gets turned into "", but the correct would be /

I think now that there arent special cases with the slashes, the strip_prefix code can be simplified by quite a bit, if instead of doing all this logic, we instead of split('/') we tokenized, but that was a larger refactor not related to this PR.

With a tokenized version, we just remove things from the path 1:1, without doing anything weird with the slashes.

I now notice that this can actually be simplified slightly, I created a new commit to do it. 

I think a follow-up of doing it with tokenization is a nice cleanup from this PR.

ficolas2:
I agree, the route("", _) case is a bit awkward, on the other hand, I doubt anyone would trip on it, or trip on it for long, if anyone ever does route("").

On the other hand, I cant think of a way to do it. Specially changing or creating new interfaces I think would remove the niceness of having them be lean.

Not my call though, happy to modify however you guys think its best!

mladedav:
I added a comment to [the issue](https://github.com/tokio-rs/axum/issues/2659#issuecomment-4923043501). I'd like to look at some other options and compare how they look so please keep this PR as is.

Just so you know what the state is. And thanks for pushing this along!

mladedav:
@ficolas2 We've discussed it in the parent issue and came to https://github.com/tokio-rs/axum/issues/2659#issuecomment-5464349516. Feel free to voice any opinions there if you see something differently.

Do you want to work on this issue more and implement that here or should we write the alternative?

In either case, thanks for opening this PR and driving it forward when it was in a bit if limbo.

ficolas2:
Yes, I will implement what was decided in the issue

mladedav:
Hey @ficolas2, do you know when you will be available to continue with this? We want to move towards 0.9 release and this is one of the breaking changes we'd like to include there.

ficolas2:
> Hey @ficolas2, do you know when you will be available to continue with this? We want to move towards 0.9 release and this is one of the breaking changes we'd like to include there.

Working on it today, unsure how much i'll be able to work on it, but want to have it by friday at most

ficolas2:
> Hey @ficolas2, do you know when you will be available to continue with this? We want to move towards 0.9 release and this is one of the breaking changes we'd like to include there.

What time do you need? I'm unsure if i'll be able to get it by the weekend, and I cant work on it this weekend

mladedav:
If that means that you think you could have it done either this or next week, then that's fine. There's no hard deadline for the release or anything, it's more of a trying to do it reasonably soon at this point.

ficolas2:
> If that means that you think you could have it done either this or next week, then that's fine. There's no hard deadline for the release or anything, it's more of a trying to do it reasonably soon at this point.

Started working on it today, sorry a bit late with it, I've had two busy weekends in a row

PR #3815, opened 2026-07-03
add Box::leak &'static approach to dependency-injection example

FalkWoldmann:
## Motivation
The dependency-injection example only shows dependencies shared behind an `Arc` (as `Arc<dyn UserRepo>` or a cloned concrete type). That leaves out a pattern that's common in long-running services: leaking a process-lifetime dependency with `Box::leak` to get a `&'static` reference.

## Solution

Enhance the DI example showing how to leak a dependency to a `&'static` reference as an alterantive to `Arc`, covering the Copy upside and the memory-leak downside

### DIFF
--- a/examples/dependency-injection/src/main.rs
+++ b/examples/dependency-injection/src/main.rs
@@ -32,24 +32,59 @@ async fn main() {
 
     let user_repo = InMemoryUserRepo::default();
 
-    // We generally have two ways to inject dependencies:
+    // Two separate decisions go into injecting a dependency.
     //
-    // 1. Using trait objects (`dyn SomeTrait`)
-    //     - Pros
-    //         - Likely leads to simpler code due to fewer type parameters.
-    //     - Cons
-    //         - Less flexible because we can only use object safe traits
-    //         - Small amount of additional runtime overhead due to dynamic dispatch.
-    //           This is likely to be negligible.
-    // 2. Using generics (`T where T: SomeTrait`)
-    //     - Pros
-    //         - More flexible since all traits can be used.
-    //         - No runtime overhead.
-    //     - Cons:
-    //         - Additional type parameters and trait bounds can lead to more complex code and
-    //           boilerplate.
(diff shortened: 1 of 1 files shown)


mladedav:
This has nothing to do with dependency-injection. Even your example completely ignores the whole point of the other exmaple and hardcodes a static in memory repository.

I think we have an example (or maybe just docs) somewhere about using either `State` or thread locals/captures for state injection into handlers and this could theoretically live there.

FalkWoldmann:
Thanks for the feedback but I disagree. Right now, the example includes the different concepts which I only tried to make more explicit. Also, having a static memory repository (or some other structs) does not exclude the code to adhere to DI principles, since the dependencies are still effectively injected. In other popular frameworks like Spring Boot it is totally viable to do it like that and not to introduce some `interface` just for the sake of it. 

In addition, could you please link me to the part of the docs where you think this content would fit better?

mladedav:
The docs mention different ways to get state your handlers [here](https://docs.rs/axum/latest/axum/#sharing-state-with-handlers). There is a mention of using an `Arc` and arguably it could also mention `Box::leak` or a static variable.

Now I get where you're coming from, but the the whole `State` concept is then dependency injection and with a tiny bit of squinting, so are `Extensions` (as a scoped dependency) and with a lot more squinting so are all other extractors (as transient dependencies). And arguably all of that lacks any dependency resolution so maybe some people wouldn't call dependency injection anything in axum.

The point of this example is to show how to get replaceable or mockable structs in handlers and maybe we could rename it to something else. What that should be, I don't know though.

I'm sure there's been a GitHub discussion mentioning the leaked state but I can't find it now and I thought there was an example showing the different ways to propagate state but maybe I'm mistaken. But it's two orthogonal concepts, this example is showing generics/dynamic dispatch in state, another example could show `State` with `impl Clone`/`Arc`/`&'static` vs task locals vs closure captures. But I wouldn't want to mix those two concepts in one example.

PR #3818, opened 2026-07-06
feat: add RFC-aware FileStream range handling

### LINKED ISSUE #3814: Byte range requests in File stream do not support all features defined by RFC.

### Version

0.8.9

### Crates

`axum-extra`

### Description

In `FileStream`

We currently support only ranges like `100-199` but not suffix ranges like `-100` or multiple ranges like `100-199, 300-399`

ByteBaker:
## Motivation

**Make byte range request support RFC compliant.**

Existing support for the [HTTP `Range` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Range) in `axum-extra` does not support several valid byte-range forms described by [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-range-requests), as highlighted by #3814.

Consequently, valid open-ended, suffix, and multiple-range requests could return `416 Range Not Satisfiable`.

### API change

Since this change will ship in a breaking release, `try_range_response` retains its existing name but changes its signature to accept the range header directly:

```rust
FileStream::<ReaderStream<File>>::try_range_response(
    file_path: impl AsRef<Path>,
    range: Option<&HeaderValue>,
)
```

This removes the need for callers to parse the header into a single `(start, end)` pair.

- A missing header or an unknown range unit returns the full file response.
- Malformed byte ranges return an empty `416 Range Not Satisfiable` response with `Content-Range: bytes */{file_size}`.
- Unsatisfiable ranges are omitted. If none are satisfiable, the response is `416`.
- A single satisfiable range returns a normal `206 Partial Content` response.
- Multiple satisfiable ranges return a streamed `multipart/byteranges` response.
- The parameter accepts a header value directly, allowing callers to pass `headers.get(header::RANGE)` while still supporting custom header names.

### Implementation

- Add a private `RangeSpec` type representing one byte-range specification before the file size is known.
- `RangeSpec` supports bounded (`start-end`), open-ended (`start-`), and suffix (`-length`) forms using `Option<i64>` values.
- Add a private `RangeSpecs` collection backed by a fixed-size array and an explicit length, avoiding heap allocation.
- Parse and retain up to eight ranges. Additional ranges are ignored, as HTTP allows a server to return only a portion of the requested ranges.
- Resolve parsed ranges after obtaining the file size, clamping end offsets and omitting unsatisfiable ranges.
- Keep parsing and normalization in a separate private module.
- Stream bounded file sections without buffering the complete file.

Fixes #3814

### DIFF
--- a/axum-extra/src/response/file_stream.rs
+++ b/axum-extra/src/response/file_stream.rs
@@ -5,13 +5,21 @@ use axum_core::{
 };
 use bytes::Bytes;
 use futures_core::TryStream;
-use http::{header, StatusCode};
-use std::{io, path::Path};
+use futures_util::{stream, TryStreamExt};
+use http::{header, HeaderValue, StatusCode};
+use std::{
+    io,
+    path::{Path, PathBuf},
+};
 use tokio::{
     fs::File,
     io::{AsyncReadExt, AsyncSeekExt},
 };
 use tokio_util::io::ReaderStream;
+
+mod range;
+
+use self::range::{normalize_range_specs, RangeSpecs, MAX_RANGES};
 
 /// Encapsulate the file stream.
 ///
@@ -126,6 +134,8 @@ where
     pub fn into_range_response(self, start: u64, end: u64, total_size: u64) -> Response {
         let mut resp = Response::builder().header(header::CONTENT_TYPE, "application/octet-stream");
         resp = resp.status(StatusCode::PARTIAL_CONTENT);
+        resp = resp.header(header::ACCEPT_RANGES, "bytes");
+        resp = resp.header(header::CONTENT_LENGTH, end - start + 1);
 
         resp = resp.header(
             header::CONTENT_RANGE,
@@ -142,39 +152,31 @@ where
             })
     }
 
-    /// Attempts to return RANGE requests directly from the file path.
+    /// Attempts to return a response for an HTTP `Range` header directly from the file path.
     ///
     /// # Arguments
     ///
     /// * `file_path` - The path of the file to be streamed
-    /// * `start` - The start position of the range
-    /// * `end` - The end position of the range
-    ///
-    /// # Note
-    ///
-    /// * If `end` is 0, then it is used as `file_size - 1`
-    /// * If `start` > `file_size` or `start` > `end`, then `Range Not Satisfiable` is returned
+    /// * `range` - The value returned by `headers.get(header::RANGE)`
     ///
     /// # Examples
     ///
     /// ```
     /// use axum::{
-    ///     http::StatusCode,
+    ///     http::{header, HeaderMap, StatusCode},
     ///     response::IntoResponse,
     ///     Router,
     ///     routing::get
     /// };
-    /// use std::path::Path;
     /// use axum_extra::response::file_stream::FileStream;
     /// use tokio::fs::File;
     /// use tokio_util::io::ReaderStream;
-    /// use tokio::io::AsyncSeekExt;
     ///
-    /// async fn range_stream() -> impl IntoResponse {
-    ///     let range_start = 0;
-    ///     let range_end = 1024;
-    ///
-    ///     FileStream::<ReaderStream<File>>::try_range_response("CHANGELOG.md", range_start, range_end).await
+    /// async fn range_stream(headers: HeaderMap) -> impl IntoResponse {
+    ///     FileStream::<ReaderStream<File>>::try_range_response(
+    ///         "CHANGELOG.md",
+    ///         headers.get(header::RANGE),
+    ///     ).await
     ///         .map_err(|e| (StatusCode::NOT_FOUND, format!("File not found: {e}")))
     /// }
     ///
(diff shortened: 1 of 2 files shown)


axum-extra/src/response/file_stream.rs
```diff
@@ -181,40 +250,267 @@ where
...
+
+fn response_416(total_size: u64) -> Response {
+    Response::builder()
+        .status(StatusCode::RANGE_NOT_SATISFIABLE)
+        .header(header::CONTENT_RANGE, format!("bytes */{total_size}"))
+        .body(body::Body::from("Range Not Satisfiable"))
+        .unwrap_or_else(|e| {
+            (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                format!("build FileStream response error: {e}"),
```
mladedav: This is not a response error

axum-extra/src/response/file_stream.rs
```diff
@@ -181,40 +250,267 @@ where
...
     }
+
+    (ranges, len)
+}
+
+fn response_416(total_size: u64) -> Response {
+    Response::builder()
+        .status(StatusCode::RANGE_NOT_SATISFIABLE)
+        .header(header::CONTENT_RANGE, format!("bytes */{total_size}"))
+        .body(body::Body::from("Range Not Satisfiable"))
```
mladedav: No need for a body here

axum-extra/src/response/file_stream.rs
```diff
@@ -181,40 +250,267 @@ where
...
+        .unwrap_or_else(|e| {
+            (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                format!("build FileStream response error: {e}"),
+            )
+                .into_response()
```
mladedav: If you used this directly,.you wouldn't have to unwrap. In any case, keep the 416 even if the header cannot be built (which should never happen)

axum-extra/src/response/file_stream.rs
```diff
@@ -181,40 +250,267 @@ where
...
+        })
+}
+
+fn invalid_range_response() -> Response {
```
mladedav: This is inconsistent with the other method name. But shouldn't we include the header here too anyway? And if we also drop the body, we can live with just.the other method.

axum-extra/src/response/file_stream.rs
```diff
@@ -5,14 +5,79 @@ use axum_core::{
...
 };
 use tokio_util::io::ReaderStream;
 
+const MAX_RANGES: usize = 8;
+
+#[derive(Clone, Copy, Default)]
+struct RangeSpec {
+    start: Option<i64>,
+    end: Option<i64>,
+}
```
mladedav: Please extract the range header logic to its own file, this is already long enough.

axum-extra/src/response/file_stream.rs
```diff
@@ -144,6 +211,8 @@ where
 
     /// Attempts to return RANGE requests directly from the file path.
     ///
+    /// Deprecated: use [`Self::try_range_response_from_header`] for HTTP `Range` headers.
```
mladedav: We'll do this in a breaking release so just delete whatever is not necessary anymore.

mladedav (changes requested):

axum-extra/src/response/file_stream.rs
ByteBaker: How about I keep the existing name then? Change only the signature?

axum-extra/src/response/file_stream.rs
ByteBaker: Hmm 🤔 makes sense

axum-extra/src/response/file_stream.rs
mladedav: Sure, since the parameters changed we can keep the name. I don't mind either way to be honest.

ByteBaker:
@mladedav I've updated the PR based on your suggestions. Please review.

Samielakkad (commented):
Retracted.

axum-extra/src/response/file_stream/range.rs
```diff
@@ -0,0 +1,128 @@
...
+/// - `start-end`: both values are inclusive byte offsets.
+/// - `start-`: `end` is `None`.
+/// - `-length`: `start` is `None` and `end` stores the suffix length.
+///
+/// A successfully parsed specification never has both fields set to `None`.
+#[derive(Clone, Copy, Default)]
+struct RangeSpec {
+    start: Option<i64>,
+    end: Option<i64>,
+}
```
mladedav, line 22: Let's do an enum then. Three cases, the invalid `None, None` state disappears, no ambiguity about `end` becoming `length` when `start` is `None`.

Also, the numbers should be unsigned in all cases.

axum-extra/src/response/file_stream/range.rs
```diff
@@ -0,0 +1,128 @@
...
+
+impl TryFrom<&str> for RangeSpec {
```
mladedav, line 24: Nit, there is `FromStr` which is almost the same thing so you might use that directly.

axum-extra/src/response/file_stream/range.rs
```diff
@@ -0,0 +1,128 @@
...
+    type Error = ();
```
mladedav, line 25: Please create an enum for this with real failure cases. The users should be able to find out what went wrong.

We're currently the only consumer, but we might decide to make this public and having a clean errors can help anyone who maintains this code later.

axum-extra/src/response/file_stream/range.rs
```diff
@@ -0,0 +1,128 @@
...
+/// One parsed byte-range specification before the file size is known.
+///
+/// The fields represent the three supported forms as follows:
+///
+/// - `start-end`: both values are inclusive byte offsets.
+/// - `start-`: `end` is `None`.
+/// - `-length`: `start` is `None` and `end` stores the suffix length.
+///
+/// A successfully parsed specification never has both fields set to `None`.
+#[derive(Clone, Copy, Default)]
```
mladedav, line 18: I don't think this should derive `Default`. Especially since the default is an invalid state, it's better to explicitly use `Option<RangeSpec>` instead in that case.

axum-extra/src/response/file_stream/range.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/// A fixed-capacity collection of parsed byte-range specifications.
+///
+/// Only `ranges[..len]` contains parsed values. Parsing fails if any specification is malformed
+/// within the first [`MAX_RANGES`] ranges. Any remaining ranges are ignored.
+#[derive(Default)]
+pub(super) struct RangeSpecs {
+    ranges: [RangeSpec; MAX_RANGES],
+    len: usize,
```
mladedav, line 47: We can have `[Option<RangeSpec>; MAX_RANGES]`.

Why did you choose a fixed size array over a vector? To save the allocation, to prevent DoS with too many overlapping ranges, or something else?

axum-extra/src/response/file_stream.rs
```diff
@@ -183,38 +185,171 @@ where
     /// ```
     pub async fn try_range_response(
         file_path: impl AsRef<Path>,
-        start: u64,
-        mut end: u64,
+        range: Option<&HeaderValue>,
```
mladedav, line 188: I'd rather we take the range unconditionally. If it's missing, the caller should handle that.

axum-extra/src/response/file_stream.rs
```diff
@@ -183,38 +185,171 @@ where
...
+            }
+            Some((_, range_values)) => Some(range_values),
+            None => None,
+        };
+
+        let file = File::open(path).await?;
         let metadata = file.metadata().await?;
         let total_size = metadata.len();
+        let Some(range_values) = range_values else {
+            return Ok(response_416(total_size));
```
mladedav: This seems inconsistent, if the header contains `=` we ignore the header and respond with full representation and if it doesn't contain `=` we return 416.

The spec says we can ignore or reject the header, but I'd rather we were consistent. So here, also ignore it please.

axum-extra/src/response/file_stream.rs
```diff
@@ -183,38 +185,171 @@ where
...
+                    (path, ranges, len, total_size, boundary, index, 2, None),
+                )));
+            }
 
-        Ok(FileStream::new(stream).into_range_response(start, end, total_size))
+            let (start, end) = ranges[index];
+            if phase == 0 {
+                return Ok::<_, io::Error>(Some((
+                    Bytes::from(format!(
+                        "--{boundary}\r\nContent-Type: application/octet-stream\r\nContent-Range: bytes {start}-{end}/{total_size}\r\n\r\n"
```
mladedav, line 288: There is already `axum_extra::response::MultipartForm` and I don't think we should reimplement that here.

It might need some tweaking to support the headers we need here (and discard some that that implementation requires) but at least the core serialization should be shared.

If you'd rather not do refactoring like that, feel free to only implement the single range cases, ignoring headers with multiple ranges (which we currently fail to even parse I believe) and we'll finish this later. It's your call.

axum-extra/src/response/file_stream/range.rs
ByteBaker: Well, I used `TryFrom` because so far an error in parsing was possible. Once we switch to the enum, this would change. I'll check it out.

axum-extra/src/response/file_stream/range.rs
ByteBaker: The motivation, as mentioned in the PR description, was to save allocation. I didn't think anyone would send beyond a certain number of max ranges (if they're sane).
> Add a private RangeSpecs collection backed by a fixed-size array and an explicit length, avoiding heap allocation.

I guess I should've also added a small comment in the code.

axum-extra/src/response/file_stream.rs
ByteBaker: I'll take care of it. No worries.

ByteBaker:
@mladedav thanks for the comprehensive review again. Otherwise I was worried this thread had gone cold. 😅

mladedav:
Sorry for the slow reviews, I'll try to be faster next time.

ByteBaker:
That's alright. We all have bunch of things to take care of. Do it at your own pace please. I'm trying to push the updates ASAP though while you're still able to review it.

ByteBaker:
Hey @mladedav, I've finished the changes. Do take a look. During my own review, I found a few things that could've been problematic, so I took care of them too. Nothing fancy, just integer overflow, skipping malformed range, and overall RFC compliance.

axum-extra/src/response/file_stream.rs
```diff
@@ -142,79 +154,165 @@ where
...
+        if index >= len {
+            return None;
         }
 
-        file.seek(std::io::SeekFrom::Start(start)).await?;
+        let (start, end) = ranges[index];
+        let content_range = format!("bytes {start}-{end}/{total_size}");
+        let path = path.to_path_buf();
+        let reader = stream::once(async move {
+            let mut file = File::open(path).await?;
```
mladedav: Open the file just once and seek into it in each iteration.

axum-extra/src/response/multiple.rs
```diff
@@ -201,27 +211,26 @@ impl Part {
...
+        let part = multipart::Part::new(
+            boundary,
+            [
+                ("Content-Disposition", disposition.as_str()),
+                ("Content-Type", content_type.as_str()),
+            ],
+            body::Body::from(self.contents),
+        );
+
+        Ok((part, body_len))
```
mladedav, line 233: Let's not return the body length here and instead make it retrievable from the `Part` struct as the body's size hint.

Better yet, that method should be able to tell us the length of the whole part so that all the logic concerning how many additional bytes are dashes and crlf shouldn't be needed here, it just takes the length of the body and the resolved prefix.

And in the best case scenario, I think we should be able to `impl Body for Vec<Part>` (or a newtype thereof) and let that implementation handle size hints of all parts provide exact hints. Then the implementation here would be much simpler I think.

My worry is that any change could easily break the content length calculations which could create malformed responses.

axum-extra/src/response/multipart.rs
```diff
@@ -0,0 +1,119 @@
...
+use bytes::Bytes;
+use futures_core::TryStream;
+use futures_util::{stream, TryStreamExt};
+
+/// A prepared multipart section with buffered framing and a streamed body.
+pub(super) struct Part {
+    /// Opening boundary, headers, and blank line; excludes the body and its trailing CRLF.
+    pub(super) prefix: Bytes,
+    /// Content emitted after the prefix, without buffering it in the encoder.
+    body: Body,
```
mladedav: Why is this `Body`? From what I've seen so far, we only build these parts to the encode them into an actual body and send that away. This could just be a `Vec<u8>`, right?

The problems I see here are that because you have a body here then where you try to consume it you need to have an async context, that makes you use a stream for the resulting body and you try to manually track content length there.

If we just didn't touch async here, I think everything will end up simpler.

mladedav (commented):
The file stream side looks good to me now, there's just a bit of complexity regarding the multipart content length I'd rather avoid if we can help it.

PR #3827, opened 2026-07-09
Return 501 Not Implemented for unrecognized HTTP methods

### LINKED ISSUE #3826: Unrecognised HTTP methods return 405 Method Not Allowed instead of 501 Not Implemented
Currently, requests using an HTTP method that the server does not recognise (e.g. FOOBAR) are answered with 405 Method Not Allowed by default. Per [RFC 9110](https://www.rfc-editor.org/info/rfc9110/), they should be answered with 501 Not Implemented.

Requests using an HTTP method that the server does recognise, but which it is unable to respond to for this resource, should continue to be answered with 405 Method Not Allowed.

From [RFC9110 § 15.5.6](https://www.rfc-editor.org/info/rfc9110/#section-15.5.6):

> The 405 (Method Not Allowed) status code indicates that the method received in the request-line is known by the origin server but not supported by the [target resource](https://www.rfc-editor.org/info/rfc9110/#target.resource). The origin server MUST generate an [Allow](https://www.rfc-editor.org/info/rfc9110/#field.allow) header field in a 405 response containing a list of the target resource's currently supported methods.

From [RFC9110 § 15.6.2](https://www.rfc-editor.org/info/rfc9110/#section-15.6.2):

> The 501 (Not Implemented) status code indicates that the server does not support the functionality required to fulfill the request. This is the appropriate response when the server does not recognize the request method and is not capable of supporting it for any resource.

From [RFC9110 § 9.1](https://httpwg.org/specs/rfc9110.html#method.overview):

> An origin server that receives a request method that is unrecognized or not implemented SHOULD respond with the [501 (Not Implemented)](https://httpwg.org/specs/rfc9110.html#status.501) status code. An origin server that receives a request method that is recognized and implemented, but not allowed for the target resource, SHOULD respond with the [405 (Method Not Allowed)](https://httpwg.org/specs/rfc9110.html#status.405) status code.

---

This issue was identified using the conformance test suite developed in this paper: Rautenstrauch, J., & Stock, B. (2024). Who’s Breaking the Rules? Studying Conformance to the HTTP Specifications and its Security Impact. In Proceedings of the 19th ACM Asia Conference on Computer and Communications Security (pp. 843–855). ACM. https://doi.org/10.1145/3634737.3637678. I am however not affiliated with these authors.

bardiharborow:
Per [RFC 9110](https://www.rfc-editor.org/info/rfc9110/), 405 Method Not Allowed is for methods the server recognizes but the target resource does not allow, while 501 Not Implemented is for methods the server does not recognize at all. MethodRouter's default fallback previously returned 405 for both cases.

The default fallback now returns 501 for methods that don't map to a MethodFilter, keeping 405 (with the Allow header) for known methods. Custom methods routed via `any` or a user-set fallback are unaffected, since those replace the default fallback.

Fixes #3826

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+- **breaking:** Requests using an unrecognized HTTP method now receive `501 Not
+  Implemented` instead of `405 Method Not Allowed` from `MethodRouter`'s default
+  fallback, per RFC 9110. Routing custom methods via `any` or a custom fallback
+  is unaffected
 - **breaking:** Router fallbacks are now properly merged for nested routers ([#3158])
 - **breaking:** `#[from_request(via(Extractor))]` now uses the extractor's
   rejection type instead of `axum::response::Response` ([#3261])
--- a/axum/src/routing/method_routing.rs
+++ b/axum/src/routing/method_routing.rs
@@ -794,11 +794,17 @@ impl<S, E> MethodRouter<S, E>
 where
     S: Clone,
 {
-    /// Create a default `MethodRouter` that will respond with `405 Method Not Allowed` to all
-    /// requests.
+    /// Create a default `MethodRouter` that will respond with `405 Method Not Allowed` to
+    /// requests using a known method, and `501 Not Implemented` to requests using an
+    /// unknown method.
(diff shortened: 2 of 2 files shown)


bardiharborow:
I am withdrawing this because 501 Not Implemented can only be returned if no resource on the server supports it, and MethodRouter cannot determine that. This will need to be fixed elsewhere, but where, I am not sure yet.

PR #3828, opened 2026-07-10
feat(path): deserialize wildcard captures into Vec by splitting on /

### LINKED ISSUE #3621: Path extractor Vec<String> deserialization support for wildcard paths

## Feature Request

The path extractor currently only supports greedy extraction of the entire wildcard path value into a `String` value. It should also support `Vec<T>` and handle the value extraction and parsing to `T`.

Though I personally only need support for `Vec<String>` :)

### Motivation

Currently wildcard values are generally deserialized into a `String` and the user has to handle value extraction.

For a project I would like to implement nested groups/namespaces, which requires using wildcards. This would require me to handle extracting the path values into a `Vec<String>` before I can run logic for the strings.

I don't see why axum's `Path` extractor shouldn't support `Vec<String>`, so here I am.

### Proposal

In `axum\src\extract\path\de.rs` have `ValueDeserializer` support serde deserialization with `deserialize_seq`.

The `PathDeserializer` would supply the wildcard path as string, which could then be split by `/` and pushed to the sequence.

### Alternatives

Manually implement wildcard `String` value splitting by `/` chars for each wildcard route or write a custom path extractor and maybe submit that for axum-extra.

patelshudhanshu1999-maker:
Fixes #3621

## Motivation

When a route uses a wildcard capture like `/{*path}`, there's no ergonomic way to extract the individual path segments. A `Path<Vec<String>>` deserializes a single capture into a one-element vector containing the entire matched string, and using a `Vec<T>` inside a struct field (e.g., `struct Params { key: Vec<String> }`) fails with an unsupported-type rejection.

Users working with file-system-like routes (`/files/{*path}`) or S3-style APIs (`/{bucket}/{*key}`) need to manually split the captured string, which is boilerplate-heavy and error-prone.

## Solution

When a route has a **single path parameter** (typically a wildcard), the `Path` extractor now splits the captured value on `/` and deserializes each non-empty segment individually into the target sequence type. This works for:

- **Top-level**: `Path<Vec<String>>`, `Path<Vec<i32>>`, etc.
- **Struct fields**: `struct Params { bucket: String, key: Vec<String> }`
- **Tuple elements**: `(String, Vec<String>)`

### Key design decisions

- **Empty segments are skipped** — trailing slashes and double slashes don't produce empty strings in the output.
- **`Vec<(String, String)>` is unaffected** — it continues to deserialize into `(name, value)` pairs of all captures, preserving backward compatibility.
- **Percent-decoding happens before splitting** — since `matchit` delivers percent-decoded values, an encoded `/` (`%2F`) also acts as a separator.
- **All-empty segments edge case** — when the value has only empty segments (e.g., `/`), the behavior depends on context: in a struct field the result is an empty `Vec`, while at the top level the original single-element behavior is preserved.

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -21,6 +21,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **added:** `IntoResponseParts` impl for `Redirect`, allowing it to be combined
   with a body in a response tuple ([#3721])
 - **added:** Add `RawPathParams::from_request_extensions` ([#3757])
+- **breaking:** `Path` now deserializes a sequence such as `Vec<String>` from a single
+  capture by splitting it on `/` (skipping empty segments), so wildcard captures can
+  be extracted as segments, both as the top-level type and as a struct field or tuple
+  element. Previously a top-level `Vec` deserialized a single capture into one element
+  and struct fields failed with an unsupported-type rejection ([#3621])
 - **changed:** `serve` has an additional generic argument and can now work with any response body
   type, not just `axum::body::Body` ([#3205])
 - **changed:** `Redirect` constructors now accept any `impl Into<String>` ([#3635])
@@ -39,6 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#3721]: https://github.com/tokio-rs/axum/pull/3721
 [#3742]: https://github.com/tokio-rs/axum/pull/3742
 [#3757]: https://github.com/tokio-rs/axum/pull/3757
+[#3621]: https://github.com/tokio-rs/axum/issues/3621
 
 # 0.8.9
 
--- a/axum/src/docs/routing/route.md
+++ b/axum/src/docs/routing/route.md
@@ -77,6 +77,34 @@ Note that `/{*key}` doesn't match empty segments. Thus:
 - `/{*key}` doesn't match `/` but does match `/a`, `/a/`, etc.
 - `/x/{*key}` doesn't match `/x` or `/x/` but does match `/x/a`, `/x/a/`, etc.
 
+Wildcard captures can be extracted as one string or deserialized into a
+sequence of segments:
+
+```rust
+use axum::{
+    Router,
+    routing::get,
+    extract::Path,
+};
+
+let app: Router = Router::new()
+    .route("/assets/{*path}", get(|Path(path): Path<String>| async move {
+        // for `/assets/css/main.css`, `path` is "css/main.css"
+    }))
+    .route("/files/{*segments}", get(|Path(segments): Path<Vec<String>>| async move {
+        // for `/files/foo/bar/baz`, `segments` is `["foo", "bar", "baz"]`
+    }));
+```
+
+When deserializing into a sequence, the capture is split on `/` and empty
+segments are skipped. Since captures are percent-decoded before they are
+split, a percent-encoded slash (`%2F`) also separates segments; this applies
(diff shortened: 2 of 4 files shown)


Samielakkad (commented):
There is an inconsistency in the all-empty-segment case for a single capture.

The docs say sequence deserialization splits the decoded capture on `/` and skips empty segments, and the struct-field test follows that behavior:

```rust
struct Params { key: Vec<String> }
// key = "/" => key: vec![]
```

But top-level `Path<Vec<String>>` currently falls back to the old ?one capture becomes one element? path when the split has no non-empty segment:

```rust
// value = "/" or "//" => vec!["/".to_owned()] / vec!["//".to_owned()]
```

That comes from `PathDeserializer::deserialize_seq`: it only constructs `SingleParamSeqAccess` after finding a non-empty first segment, otherwise it falls through to `SeqDeserializer`.

This can show up with a decoded `%2F` / `%2F%2F` capture, and it means `Path<Vec<T>>` and `Path<StructWithVec>` disagree for the same captured value. I think the top-level sequence path should commit to split-mode whenever there is exactly one capture, even if the resulting segment list is empty, or the docs should explicitly call out the fallback behavior.

mladedav:
Why was previous work in #3680 not pointed out and what complexity is this solving on top of that?

jlizen:
> Why was previous work in https://github.com/tokio-rs/axum/pull/3680 not pointed out and what complexity is this solving on top of that?

In my read, the main thing is, this doesn't break `Vec<(String, String)>` and `deserialize_any`-based element types since it does the first-element sniffing (which is also why it is more complicated). It also allows keyed error messages.

Though, in exchange it makes the all-empty capture case more ambiguous and forces that fallback.

I could probably go either way on this. How much do we care about breaking the above cases, if we are anyway prepping `0.9` @mladedav ? Erring on the side of simpler code and no fallbacks is awfully tempting...

mladedav:
I think we should support the struct variant. I'm not convinced we should split the wildcard in `Vec<T>` and make it behave so much differently to `Vec<(String, T)>`.

Reading the description also makes me think that having paths `/{foo}` and `/{foo}/{bar}` would behave differently if `foo` contains escaped slashes. At least that's my reading of the "When a route has a single path parameter (typically a wildcard)" part. I'd prefer not to be inconsistent like that.

Also the motivation mentions `/{bucket}/{*path}` where if someone tries `Vec<String>` they get two strings and they probably want to use `(String, Vec<String>)` instead. But `Vec<String>` would have worked for the other case of `/files/{*path}`.

So if it's simple to do, I think I'd prefer to split this to support the struct case (the original feature request I believe) and decide about the rest in that or another issue because I don't think it's entirely clear what would be the obvious thing to do from a user's perspective.

---

Breakage is fine as long as it fails to compile or similarly loudly compiles (e.g. the v07 checks when matchit was upgraded).

jlizen:
@patelshudhanshu1999-maker : Could you narrow this to supporting vectors in struct fields and tuple elements, leaving top-level `Path<Vec<_>>` behavior unchanged?

That should let us remove the first-element dispatch and move this part forward while we discuss top-level behavior separately. Feel free to open an issue for the latter if you personally have a use case for the top level behavior (and be sure to include a motivating example).

Could you also add a docs example for `/{bucket}/{*key}`, showing both a struct with `key: Vec<String>` and the equivalent `Path<(String, Vec<String>)>`? Please also show that a single wildcard extracted as `Path<Vec<String>>` remains one unsplit element. A solo-wildcard example using `Path<(Vec<String>,)>` would be good too.


And then, we need regression tests preserving existing top-level vector behavior, including name/value pairs.

PR #3829, opened 2026-07-11
feat(sse): add Event::raw for fully custom payloads

### LINKED ISSUE #2415: Support custom SSE event

## Feature Request
Enable advanced customization of the SSE (Server-Sent Events) Event struct by introducing the EventExt trait.

### Motivation
In certain use cases, users may require additional customization options for the SSE Event struct in the Axum library. For instance, [Vercel AI SDK](https://github.com/vercel/ai) doesn't follow the standard and only accepts raw text. So it is impossible to use Axum which always includes the `key` such as `data`, `id`, etc. 

### Proposal
The proposed solution involves introducing a trait, EventExt, with methods for customizing the SSE Event. Users can implement this trait to add their own logic and fields to the event. The trait includes methods for setting the data field and finalizing the event.

```rust
pub trait EventExt {
    fn data<T>(mut self, data: T) -> Self
    where
        T: AsRef<str>;

    fn finalize(self) -> Bytes;
}
```
Users can then implement this trait for the SSE Event struct at the application level to provide their own customizations.

```rust
impl EventExt for Event {
    fn data<T>(mut self, data: T) -> Self
    where
        T: AsRef<str>,
    {
        // Custom logic to handle data customization
        self.data = data.as_ref().to_string();
        self
    }

    fn finalize(self) -> Bytes {
        // Custom logic for finalizing the event
        // Convert the event to Bytes and return
        // ...
    }
}

fn handle_event() -> Sse<impl Stream<Item = Result<TextEvent, Infallible>>, TextEvent> {
  ...
   yield Ok(TextEvent::default().data(content));
  ...
}
```

This approach allows users to have fine-grained control over the SSE Event without modifying the core Axum library.

I also tried to implement in in this branch https://github.com/linrium/axum/tree/feat/support-custom-sse-event

dihannahdi:
## Motivation

`sse::Event` can only be built through the field-oriented builder (`data`, `event`, `id`, `retry`, `comment`), so the exact bytes of every event are decided by axum. #2415 asks for control over that serialization, the concrete cases being proxying pre-serialized events from an upstream SSE source, and emitting payloads for clients that consume the stream but don't accept the standard field framing.

The objection raised in the issue thread, that users can hand-roll a `Response` with a stream body, is fair for the framing itself. What it loses is everything `Sse` provides beyond two headers. `Sse::keep_alive` only operates on streams of `Event`, and `KeepAliveStream::new` is private, so anyone leaving the type also reimplements the keep-alive timer. That is exactly the boilerplate axum exists to remove.

Internally the escape hatch already exists. `Event::DEFAULT_KEEP_ALIVE` is built from the private `Event::finalized(Bytes)`, which bypasses framing entirely. This PR exposes that path with one constructor.

## Solution

Add `Event::raw`:

```rust
pub fn raw<T>(bytes: T) -> Self
where
    T: Into<Bytes>,
{
    Self::finalized(bytes.into())
}
```

The payload is emitted verbatim as a single body frame, with no `data:` framing, no field validation, and no added terminator. The docs make the contract explicit, so callers targeting standard `EventSource` clients must supply the complete event themselves, for example `"data: hi\n\n"`. They also cover the two interactions worth knowing, which are that raw events reset the keep-alive timer like any other event, and that calling builder methods after `raw` appends fields after the raw bytes, which is allowed but flagged as usually unintended. A "Custom event payloads" section was added to the module docs.

Deliberately **not** included is the `EventExt` trait and `Sse<S, T>` generic sketched in the issue. A trait-based design adds a public trait plus a type parameter that downstream signatures mentioning `Sse` may need to spell, forces `KeepAliveStream` to become generic the same way, and buys nothing over the constructor, since a custom event type can always be mapped into the stream via `Event::raw` with zero extra allocation. The payload has to become `Bytes` either way.

The tests cover a byte-exact round-trip including an empty payload and a payload with no trailing newline, builder-after-raw semantics, an integration test asserting byte-exact chunks alongside intact `content-type` and `cache-control`, and a paused-time keep-alive interplay test. There are doctests on the method and on the new module section.

No `unsafe`, no new dependencies, MSRV untouched, and the change is semver-additive.

Fixes #2415

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **added:** `IntoResponseParts` impl for `Redirect`, allowing it to be combined
   with a body in a response tuple ([#3721])
 - **added:** Add `RawPathParams::from_request_extensions` ([#3757])
-- **added:** `sse::Event::raw` for events with fully custom payloads ([#2415])
+- **added:** `sse::Event::raw` for events with fully custom payloads ([#3829])
 - **changed:** `serve` has an additional generic argument and can now work with any response body
   type, not just `axum::body::Body` ([#3205])
 - **changed:** `Redirect` constructors now accept any `impl Into<String>` ([#3635])
@@ -40,7 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#3721]: https://github.com/tokio-rs/axum/pull/3721
 [#3742]: https://github.com/tokio-rs/axum/pull/3742
 [#3757]: https://github.com/tokio-rs/axum/pull/3757
-[#2415]: https://github.com/tokio-rs/axum/issues/2415
+[#3829]: https://github.com/tokio-rs/axum/pull/3829
 
 # 0.8.9
 
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **added:** `IntoResponseParts` impl for `Redirect`, allowing it to be combined
   with a body in a response tuple ([#3721])
 - **added:** Add `RawPathParams::from_request_extensions` ([#3757])
+- **added:** `sse::Event::raw` for events with fully custom payloads ([#2415])
 - **changed:** `serve` has an additional generic argument and can now work with any response body
   type, not just `axum::body::Body` ([#3205])
 - **changed:** `Redirect` constructors now accept any `impl Into<String>` ([#3635])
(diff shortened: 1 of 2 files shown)


dihannahdi:
This one has been quiet since it opened, so let me add the piece of evidence that was not available when I wrote it.

The case in the description rests on keep-alive. You can hand-roll an SSE response easily enough, and the objection on #2415 was right about that much. But `Sse::keep_alive` only works over streams of `Event`, and `KeepAliveStream::new` is private, so leaving the type means reimplementing the idle timer as well. The real choice today is between using `Sse` and giving up custom payloads, or writing your own payloads and giving up keep-alive.

The new part is #3796, which asks for `KeepAliveStream::new` to be made public. Different author, different shape, same wall. They want to reach the keep-alive machinery from outside and cannot. That is the same gap this PR closes from the other side, and the two are probably worth weighing together rather than separately, since whichever way you lean likely answers both.

I have also tidied the description and swapped a couple of em dashes in the docs for plain prose. Nothing about the implementation changed.

Mostly I would like a steer on direction. If the view is that keep-alive should be reachable directly rather than through a raw event, that is genuinely useful to know, because it would make #3796 the better fix and this one unnecessary.

jlizen (approved):
Approach seems reasonable to me, thanks.

PR #3830, opened 2026-07-11
feat(ws): add MockUpgrade for testing WebSocket routes

### LINKED ISSUE #3688: Support testing of websocket routes

### Motivation

For regular routes you can test them without listening on any tcp port using `router.oneshot(Request::builder()...)`. For websockets however the `WebSocketUpgrade` extractor needs the `hyper::upgrade::OnUpgrade` extension which does not have a public constructor. Adding one has been rejected by hyper in https://github.com/hyperium/hyper/issues/4034.

### Proposal

Create an alternative to `OnUpgrade` that has a public constructor accepting an `AsyncRead + AsyncWrite` object that can be added as extension to a `Request` which is checked by `WebSocketUpgrade` if there is no `hyper::upgrade::OnUpgrade` extension.

### Alternatives

Locally patch hyper to make the `OnUpgrade` constructor public. This would effectively require vendoring hyper to be able to run in CI and make it impossible to publish the project in question to crates.io.

dihannahdi:
## Motivation

Every axum route can be exercised in-process with `tower::ServiceExt::oneshot`, except WebSocket routes. `WebSocketUpgrade` requires the `hyper::upgrade::OnUpgrade` request extension, which cannot be constructed outside hyper, and hyper declined to open that constructor (hyperium/hyper#4034). That leaves binding a real TCP listener as the only way to test a WebSocket route. As the author of `axum-test` noted on the issue, websockets are the one transport that can't be mocked today.

## Solution

This implements the mechanism proposed in #3688, an alternative to `OnUpgrade` with a public constructor accepting an `AsyncRead + AsyncWrite` object, checked by `WebSocketUpgrade` when no hyper `OnUpgrade` is present.

- New `extract::ws::MockUpgrade` with `MockUpgrade::new(io)` where `io: AsyncRead + AsyncWrite + Unpin + Send + 'static`, typically one end of `tokio::io::duplex`. It is a request extension. `Clone` is required by `http::Extensions` and is provided by an `Arc<Mutex<Option<Box<dyn ...>>>>` whose lock is only ever held for a synchronous `take()`, never across an await.
- `from_request_parts` tries `hyper::upgrade::OnUpgrade` first, so an injected `MockUpgrade` can never shadow a live server connection. Handshake validation is byte-for-byte identical on both paths, so tests must send a real handshake request and exercise the real extraction logic. Response construction (101 or HTTP/2 200, `sec-websocket-accept`, subprotocol echo) is shared.
- `WebSocket`'s inner io becomes a private two-variant enum, either `TokioIo<Upgraded>` or `Box<dyn AsyncReadWrite>`. Both variants are `Unpin`, so delegation is `Pin::new` on `get_mut()`, with no pin-projection and no `unsafe`. `poll_write_vectored` and `is_write_vectored` are forwarded so the hyper path keeps its vectored-write behavior, and the only added cost there is one enum-discriminant branch per poll. No public type leaks this change.

A `# Testing` section in the module docs contains a complete runnable example, with a duplex pair, a handshake request carrying the extension, `oneshot`, then driving the client end with `tokio_tungstenite::WebSocketStream::from_raw_socket(.., Role::Client, ..)`.

Because the I/O is single-use while the extension itself is `Clone`, a `# One-shot I/O` section on the type spells out that clones share one stream and that each connection needs a fresh `MockUpgrade`, and `MockUpgrade::new` points at it.

The tests cover an echo round-trip over a duplex via `oneshot`, asserting the RFC 6455 `sec-websocket-accept` for the spec's example key, handshake validation still being enforced on the mock path, `on_failed_upgrade` firing when the io was already consumed, subprotocol negotiation over the mock path, and two requests carrying clones of one `MockUpgrade`, where the second reaches `on_failed_upgrade`. The existing TCP-based `integration_test` and `http2` tests cover the unchanged hyper path.

On naming, `MockUpgrade` follows the issue's framing, and I am happy to rename it to `TestUpgrade` or `ManualUpgrade` if either reads better.

No new dependencies, everything is under the existing `ws` feature, and the change is semver-additive.

Fixes #3688

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **added:** `IntoResponseParts` impl for `Redirect`, allowing it to be combined
   with a body in a response tuple ([#3721])
 - **added:** Add `RawPathParams::from_request_extensions` ([#3757])
+- **added:** `extract::ws::MockUpgrade` for testing WebSocket routes without a TCP connection ([#3688])
 - **changed:** `serve` has an additional generic argument and can now work with any response body
   type, not just `axum::body::Body` ([#3205])
 - **changed:** `Redirect` constructors now accept any `impl Into<String>` ([#3635])
@@ -39,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#3721]: https://github.com/tokio-rs/axum/pull/3721
 [#3742]: https://github.com/tokio-rs/axum/pull/3742
 [#3757]: https://github.com/tokio-rs/axum/pull/3757
+[#3688]: https://github.com/tokio-rs/axum/issues/3688
 
 # 0.8.9
 
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **added:** `IntoResponseParts` impl for `Redirect`, allowing it to be combined
   with a body in a response tuple ([#3721])
 - **added:** Add `RawPathParams::from_request_extensions` ([#3757])
-- **added:** `extract::ws::MockUpgrade` for testing WebSocket routes without a TCP connection ([#3688])
+- **added:** `extract::ws::MockUpgrade` for testing WebSocket routes without a TCP connection ([#3830])
 - **changed:** `serve` has an additional generic argument and can now work with any response body
   type, not just `axum::body::Body` ([#3205])
 - **changed:** `Redirect` constructors now accept any `impl Into<String>` ([#3635])
@@ -40,7 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#3721]: https://github.com/tokio-rs/axum/pull/3721
 [#3742]: https://github.com/tokio-rs/axum/pull/3742
 [#3757]: https://github.com/tokio-rs/axum/pull/3757
-[#3688]: https://github.com/tokio-rs/axum/issues/3688
+[#3830]: https://github.com/tokio-rs/axum/pull/3830
 
 # 0.8.9
 
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -89,6 +89,79 @@
 //! ```
 //!
 //! [`StreamExt::split`]: https://docs.rs/futures/0.3.17/futures/stream/trait.StreamExt.html#method.split
+//!
+//! # Testing
+//!
+//! WebSocket routes can be tested without a real TCP server by inserting a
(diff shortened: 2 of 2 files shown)


Samielakkad (commented):
One public-API edge case worth documenting/testing: `MockUpgrade` is `Clone`, but the underlying I/O is still one-shot.

That follows from the `Arc<Mutex<Option<_>>>` design, and it is probably the right implementation tradeoff because request extensions need to be cheaply movable/clonable. But from a user?s perspective it is easy to build a reusable request template or clone the same `MockUpgrade` into two requests and have the second request still return `101 Switching Protocols`; the actual failure only surfaces later through `on_failed_upgrade`.

There is already a good internal test for the ?I/O was consumed? branch by manually taking the slot. I?d add either:

- a public-facing test that sends two requests with clones of the same `MockUpgrade` and asserts the second one triggers `on_failed_upgrade`, or
- a doc note on `MockUpgrade::new` / the type docs that clones share the same single-use I/O and each WebSocket connection needs a fresh `MockUpgrade`.

That would make the one-shot behavior clear before people start copying request builders around in tests.

dihannahdi:
Thanks for the review. I went with both suggestions.

There is now a "One-shot I/O" section on the `MockUpgrade` docs, plus a pointer to it from `MockUpgrade::new`, spelling out that clones share one single-use I/O and that each WebSocket connection needs a fresh `MockUpgrade`.

I also added a test, `mock_upgrade_clone_shares_single_use_io`, that sends two requests using clones of the same `MockUpgrade`. The first upgrade consumes the shared duplex stream, and the second still returns `101 Switching Protocols` but then fires `on_failed_upgrade`, so the one-shot behavior is pinned down end to end rather than only documented.

Pushed as a follow-up commit.

dihannahdi:
A short note so this does not just sit quietly.

Everything Samielakkad raised is in, and the branch is green and merges cleanly. I have also pushed a small commit replacing the em dashes in the new comments with plain prose, since nothing else under `axum/src` uses them, and I refreshed the description so it covers the one-shot section and the clone test instead of stopping where the PR originally did.

What I still do not have is a read from a maintainer on whether the shape is right, and that is the part I cannot settle on my own, since it adds public API. The constraint underneath it is hyperium/hyper#4034, where seanmonstar said he did not want to open up the constructor of the upgrade extension publicly. If WebSocket routes are going to be testable without a real socket, something on the axum side has to carry it, and `MockUpgrade` is my attempt at the smallest version of that.

If the name is wrong, or it belongs somewhere else, or you would rather axum not grow this at all, I would much rather hear it than keep guessing.

jlizen:
I haven't given this a close read yet, but the overall approach seems reasonable, and it does seem useful.

I tend to prefer putting stuff like this behind a `test-util` feature. But, we don't have one yet. Any opinion @mladedav ?

mladedav:
Yeah, sounds good. We don't have such feature yet but we can add that.

I'd just like for it to have at most performance overhead and not cause behavior change in the normal case. Specifically, if any crate enables this feature, anyone who depends on that crate cannot disable the feature but their code should still work. The PR so far adheres to that I believe, but I'd like to be specific about that.

Issue #2673, opened 2024-03-24
Open SSE connections with pending streams block graceful shutdown


sulami:
- [X] I have looked for existing issues (including closed) about this

### Version

- axum 0.7.4
- axum-core 0.4.3

### Platform

`Darwin Continuitiy.local 23.3.0 Darwin Kernel Version 23.3.0: Wed Dec 20 21:30:44 PST 2023; root:xnu-10002.81.5~7/RELEASE_ARM64_T6000 arm64`

(M1 Pro, macOS 14.3.1)

### Description

In one of my projects, I have an SSE endpoint that listens on a broadcast channel and forwards items from that through the SSE endpoint. If I want to gracefully shutdown the application, the upstream channel is still pending which in turn prevents the graceful shutdown, until an event is actually sent. If no event is sent, the application hangs forever.

Naively I would've assumed the connection would be closed when a shutdown event occurs, probably after sending out a `204 No Content` to avoid the client trying to reconnect. The code in question used to be a websocket, which does close the connection on shutdown.

I haven't looked at the axum code in question yet, but I'd be ready to submit a change to that effect, if so desired.

**Reproduction**

I've setup a minimal example for this:

```toml
[package]
name = "axum-sse-shutdown"
version = "0.1.0"
edition = "2021"

[dependencies]
axum = "0.7"
futures = "0.3"
tokio = { version = "1", features = ["full"] }
```

```rust
use std::convert::Infallible;

use axum::{
    response::{sse::Event, IntoResponse, Sse},
    routing::get,
    Router,
};
use tokio::{net::TcpListener, select, signal};

#[tokio::main]
async fn main() {
    let app = Router::new().route("/sse", get(sse_handler));
    let listener = TcpListener::bind("0.0.0.0:8080").await.unwrap();
    axum::serve(listener, app.into_make_service())
        .with_graceful_shutdown(shutdown_handler())
        .await
        .unwrap();
}

async fn sse_handler() -> impl IntoResponse {
    let stream = futures::stream::pending::<Result<Event, Infallible>>();
    Sse::new(stream)
}

async fn shutdown_handler() {
    let ctrl_c = async {
        signal::ctrl_c()
            .await
            .expect("failed to install Ctrl+C handler");
    };

    let terminate = async {
        signal::unix::signal(signal::unix::SignalKind::terminate())
            .unwrap()
            .recv()
            .await;
    };

    select! {
        _ = ctrl_c => {},
        _ = terminate => {},
    }
}
```

[...]

PR #3819, opened 2026-07-06
Extract `axum::serve` and friends into new `axum-serve` crate

FalkWoldmann:
## Motivation
See #3433.

## Solution
This is my best shot at extracting the shared logic 😆. Lmk if I missed anything important

PR #3831, opened 2026-07-11
fix(serve): include WebSocket tasks in graceful shutdown

### LINKED ISSUE #3003: Graceful Shutdown ignores Websocket handler task
### Version

axum 0.7.4

### Platform

```
Linux [...] 6.11.3-200.fc40.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Oct 10 22:31:19 UTC 2024 x86_64 GNU/Linux
```

### Crates

axum

### Description

I have the following scenario:

* I need websocket connections which contain shared references (`Arc`) to the application state
* Sometimes I need to stop the service and be sure all shared state is dropped
* I am using graceful shutdown

The concrete reason the state needs to be dropped is that it may contain a file handle that is exclusive and that I want to access after shutting down.

The problem:  
The task spawned in `ws.on_upgrade` is not connected to the graceful shutdown future at all. I can listen to a global `CancellationToken` and quit the connection myself, but that still does not guarantee me that the connection is closed when the graceful shutdown future completes.  
If I don't immediately quit the application after the server quits, the websocket connection happily continues running. Try placing a `sleep` below the `axum::serve`. 

I would expect websockets to behave the same as SSE, i.e. blocking graceful shutdown until completion  (#2673).

dihannahdi:
## Motivation

`axum::serve(..).with_graceful_shutdown(signal)` stops accepting, drains connections, and waits for every connection task via a `close_rx` guard. WebSocket handlers escape all of it. `WebSocketUpgrade::on_upgrade` spawns a bare `tokio::spawn`, and once hyper hands off the upgraded io the connection task exits. The shutdown future then resolves while sockets are still live, so handler state (`Arc`s, file handles) outlives "shutdown", and nothing even tells the peer to close (#3003).

This matches the direction already sketched in the issue threads: "Propagating some kind of a channel there so that we keep track of it for the purpose of graceful shutdown should be easy enough" (#3003), the `ShutdownToken`-via-extension idea from #2673, and the extension-based proposal later in the #3003 thread.

## Solution

Two halves.

**Tracking.** `handle_connection` already holds the `signal_tx` and `close_rx` for the connection, so it builds a `pub(crate) GracefulPeer { shutdown, guard }` from those two and inserts it into every request's extensions. `WebSocketUpgrade` picks it up during extraction, and the spawned handler task holds the `close_rx` clone for its lifetime, so `close_tx.closed()` waits for WebSocket handlers exactly as it does for in-flight requests.

There is no second channel. The socket observes the same `signal` channel the accept loop uses, watching it from the sender side via `closed()`, since a receiver clone would keep the accept loop's own `signal_tx.closed()` from ever firing.

Under plain `serve` the extension is inserted just the same, but `Serve::run` holds its signal receiver for the lifetime of a future that never returns, so the signal can never fire and both halves stay inert.

**Closing.** On upgrade the socket receives a boxed observer for that signal, one allocation per WebSocket upgrade and nothing per message. `WebSocket`'s `Stream` impl polls it before reading. When shutdown begins it enqueues a `Close` frame through the Sink half, best-effort, with a small `Idle`/`Queued`/`Flushing` state machine that tolerates already-closing sockets and `Pending` sinks by falling through to reads so the connection keeps progressing. Reads keep flowing, the peer's Close reply ends the stream, and the typical `while let Some(msg) = socket.recv().await` loop returns on its own with no application changes. This is tungstenite's sanctioned close-initiation path, where `start_send(Close)` sets `ClosedByUs` and `read()` auto-flushes pending close frames.

**Behavior note, intentional and documented.** With graceful shutdown configured, a handler that ignores socket closure and never returns now blocks shutdown, where before it was silently orphaned. That is the same trust model as any in-flight response body, since an unfinished SSE stream blocks shutdown the same way (#2673). It is documented in the ws module docs, which `on_upgrade` links to.

The regression test reproduces #3003 end to end, with a recv-loop handler holding an outside `Arc`, a real client over TCP, and the signal fired, asserting that the client sees a Close frame, that the serve future completes, and that `Arc::strong_count` returns to 1. With the tracking neutered to simulate the old behavior, both graceful-shutdown tests fail with the client hung and the socket never closed. A second test covers a handler that also sends, where send-after-Close errors and the handler exits.

Anticipated questions:

- *Why not a public `ShutdownToken` extractor?* More API surface, and it pushes work onto every application. The `pub(crate)` handle is the natural seed for that as a follow-up if wanted.
- *Can the injected Close corrupt a concurrent user send via `split()`?* No. `SplitSink`/`SplitStream` serialize through a `BiLock` and frames are discrete. Post-Close user sends get tungstenite's documented `SendAfterClosing` error, and only during shutdown.
- *Middleware semantics?* Layers still see the request complete at response time, which is pre-existing and noted in #3003. This PR narrows the tracking gap only.
- *Does a peer that never replies to Close hang shutdown?* The handler's recv loop ends on peer Close, or on error or EOF. A silent peer keeps the socket open, matching the SSE trust model. A hard deadline can still be enforced by racing the serve future against a timeout, unchanged from today.

Everything is feature-gated exactly like the `serve` module (`tokio` plus `http1`/`http2`), since `ws` alone doesn't imply a server. Per-feature combos (`tokio` alone, `http1` alone, `ws` alone, no-default) compile clean under `-D warnings`.

No `unsafe`, no new dependencies beyond plain `tokio::sync::watch`, MSRV untouched. The CHANGELOG carries a single `fixed` entry.

Fixes #3003

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -25,7 +25,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   type, not just `axum::body::Body` ([#3205])
 - **changed:** `Redirect` constructors now accept any `impl Into<String>` ([#3635])
 - **changed:** Updated `matchit` allowing for routes with captures and static prefixes and suffixes ([#3702])
+- **changed:** `Serve::with_graceful_shutdown` now includes WebSocket connections in the drain:
+  open sockets are sent a close frame and the server waits for their handler tasks. A handler
+  that never returns will now hold shutdown open (as with any in-flight response body) ([#3003])
 - **fixed:** Responses to `HEAD` will not accidentally reply with `content-length: 0` anymore ([#3742])
+- **fixed:** graceful shutdown no longer ignores WebSocket handler tasks; the server now waits
+  for them and initiates a close handshake on open sockets ([#3003])
 
 [#3158]: https://github.com/tokio-rs/axum/pull/3158
 [#3261]: https://github.com/tokio-rs/axum/pull/3261
@@ -39,6 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#3721]: https://github.com/tokio-rs/axum/pull/3721
 [#3742]: https://github.com/tokio-rs/axum/pull/3742
 [#3757]: https://github.com/tokio-rs/axum/pull/3757
+[#3003]: https://github.com/tokio-rs/axum/issues/3003
 
 # 0.8.9
 
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -25,12 +25,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   type, not just `axum::body::Body` ([#3205])
 - **changed:** `Redirect` constructors now accept any `impl Into<String>` ([#3635])
 - **changed:** Updated `matchit` allowing for routes with captures and static prefixes and suffixes ([#3702])
-- **changed:** `Serve::with_graceful_shutdown` now includes WebSocket connections in the drain:
-  open sockets are sent a close frame and the server waits for their handler tasks. A handler
-  that never returns will now hold shutdown open (as with any in-flight response body) ([#3831])
 - **fixed:** Responses to `HEAD` will not accidentally reply with `content-length: 0` anymore ([#3742])
-- **fixed:** graceful shutdown no longer ignores WebSocket handler tasks; the server now waits
-  for them and initiates a close handshake on open sockets ([#3831])
+- **fixed:** `Serve::with_graceful_shutdown` now waits for WebSocket handler tasks and sends open
+  sockets a close frame, rather than leaving them running once the server has shut down. A handler
+  that never returns will now hold shutdown open, as with any in-flight response body ([#3831])
 
 [#3158]: https://github.com/tokio-rs/axum/pull/3158
 [#3261]: https://github.com/tokio-rs/axum/pull/3261
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -27,10 +27,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **changed:** Updated `matchit` allowing for routes with captures and static prefixes and suffixes ([#3702])
 - **changed:** `Serve::with_graceful_shutdown` now includes WebSocket connections in the drain:
   open sockets are sent a close frame and the server waits for their handler tasks. A handler
-  that never returns will now hold shutdown open (as with any in-flight response body) ([#3003])
+  that never returns will now hold shutdown open (as with any in-flight response body) ([#3831])
 - **fixed:** Responses to `HEAD` will not accidentally reply with `content-length: 0` anymore ([#3742])
 - **fixed:** graceful shutdown no longer ignores WebSocket handler tasks; the server now waits
-  for them and initiates a close handshake on open sockets ([#3003])
+  for them and initiates a close handshake on open sockets ([#3831])
 
 [#3158]: https://github.com/tokio-rs/axum/pull/3158
 [#3261]: https://github.com/tokio-rs/axum/pull/3261
@@ -44,7 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#3721]: https://github.com/tokio-rs/axum/pull/3721
 [#3742]: https://github.com/tokio-rs/axum/pull/3742
 [#3757]: https://github.com/tokio-rs/axum/pull/3757
-[#3003]: https://github.com/tokio-rs/axum/issues/3003
+[#3831]: https://github.com/tokio-rs/axum/pull/3831
 
 # 0.8.9
 
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -88,11 +88,32 @@
 //! }
 //! ```
 //!
+//! # Graceful shutdown
+//!
+//! When the server is run with [`axum::serve`] and
+//! [`Serve::with_graceful_shutdown`], WebSocket connections take part in the
+//! shutdown. Once the shutdown signal fires, each open [`WebSocket`] sends a
+//! Close frame to its peer, and the server waits for the handler tasks to
+//! finish before resolving.
+//!
+//! In practice this means a handler with the usual read loop needs no changes:
+//! after the Close handshake completes, [`WebSocket::recv`] returns `None` and
+//! the loop ends. Handlers should treat `None` (or an `Err`) from `recv` as a
+//! signal to return. A handler that never polls the socket and never returns
+//! will still hold shutdown open, just like any other in-flight response body
+//! (Server-Sent Events behave the same way).
+//!
 //! [`StreamExt::split`]: https://docs.rs/futures/0.3.17/futures/stream/trait.StreamExt.html#method.split
+//! [`axum::serve`]: crate::serve()
+//! [`Serve::with_graceful_shutdown`]: crate::serve::Serve::with_graceful_shutdown
 
 use self::rejection::*;
 use super::FromRequestParts;
 use crate::{body::Bytes, response::Response, Error};
+// Only available when `axum::serve` is compiled in; the `ws` feature does not
+// imply `http1`/`http2`, so this must be gated exactly like the `serve` module.
+#[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
+use crate::serve::GracefulPeer;
 use axum_core::body::Body;
 use futures_core::{FusedStream, Stream};
 use futures_sink::Sink;
@@ -140,6 +161,10 @@ pub struct WebSocketUpgrade<F = DefaultOnFailedUpgrade> {
     on_upgrade: hyper::upgrade::OnUpgrade,
     on_failed_upgrade: F,
     sec_websocket_protocol: BTreeSet<HeaderValue>,
+    /// Set when `axum::serve` is running with graceful shutdown; lets the
+    /// resulting socket participate in it. See [`GracefulPeer`].
+    #[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
+    graceful: Option<GracefulPeer>,
 }
 
 impl<F> std::fmt::Debug for WebSocketUpgrade<F> {
@@ -337,11 +362,34 @@ impl<F> WebSocketUpgrade<F> {
             on_upgrade: self.on_upgrade,
             on_failed_upgrade: callback,
             sec_websocket_protocol: self.sec_websocket_protocol,
+            #[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
+            graceful: self.graceful,
         }
     }
 
     /// Finalize upgrading the connection and call the provided callback with
     /// the stream.
+    ///
+    /// # Graceful shutdown
+    ///
+    /// When the server is run with [`axum::serve`] and
+    /// [`with_graceful_shutdown`], the spawned handler task participates in the
+    /// shutdown:
+    ///
+    /// * the server waits for this task to finish before resolving, and
+    /// * once shutdown begins the [`WebSocket`] automatically sends a Close
+    ///   frame to the peer.
+    ///
+    /// A handler that reads from the socket in the usual way — e.g.
+    /// `while let Some(msg) = socket.recv().await { .. }` — will therefore
+    /// observe the stream end and return on its own. A handler that never polls
+    /// the socket and never returns will still block shutdown, the same as any
+    /// other in-flight response body.
+    ///
+    /// See the [module docs](self) and [`axum::serve`] for details.
+    ///
+    /// [`axum::serve`]: crate::serve()
+    /// [`with_graceful_shutdown`]: crate::serve::Serve::with_graceful_shutdown
     #[must_use = "to set up the WebSocket connection, this response must be returned"]
     pub fn on_upgrade<C, Fut>(self, callback: C) -> Response
     where
@@ -355,7 +403,31 @@ impl<F> WebSocketUpgrade<F> {
 
         let protocol = self.protocol.clone();
 
+        // Derive the graceful-shutdown observer and the guard that keeps the
+        // server draining until this task ends. Both are `None` (a no-op) when
+        // graceful shutdown is not configured.
+        #[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
+        let (shutdown, guard) = match self.graceful {
+            Some(peer) => {
+                let mut rx = peer.shutdown;
+                let shutdown: Pin<Box<dyn Future<Output = ()> + Send>> = Box::pin(async move {
+                    // Either a value is sent or all senders drop; both mean
+                    // "shutdown has begun".
+                    let _ = rx.changed().await;
+                });
+                (Some(shutdown), Some(peer.guard))
+            }
+            None => (None, None),
+        };
+        #[cfg(not(all(feature = "tokio", any(feature = "http1", feature = "http2"))))]
+        let shutdown: Option<Pin<Box<dyn Future<Output = ()> + Send>>> = None;
+
         tokio::spawn(async move {
+            // Hold the guard (if any) for the whole task so that
+            // `close_tx.closed()` in `axum::serve` waits for this handler.
+            #[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
+            let _guard = guard;
+
             let upgraded = match on_upgrade.await {
                 Ok(upgraded) => upgraded,
                 Err(err) => {
@@ -371,6 +443,8 @@ impl<F> WebSocketUpgrade<F> {
             let socket = WebSocket {
                 inner: socket,
                 protocol,
+                shutdown,
+                close_state: CloseState::Idle,
             };
             callback(socket).await;
         });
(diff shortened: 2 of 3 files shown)


axum/src/serve/mod.rs
```diff
@@ -454,9 +470,19 @@ where
         } = self;
 
         let (signal_tx, signal_rx) = watch::channel(());
+        // A second channel dedicated to telling upgraded connections (e.g.
+        // WebSockets) that shutdown has begun. `signal_rx` being dropped wakes
+        // the per-connection tasks so hyper can start its graceful shutdown,
+        // but an upgraded connection's io has been handed off and is no longer
+        // tracked by hyper, so it needs its own notification.
+        let (shutdown_tx, shutdown_rx) = watch::channel(());
```
mladedav: Why does it need its own notification and we can't just reuse the other channel?

axum/src/serve/mod.rs
```diff
@@ -475,6 +501,10 @@ where
                 &mut make_service,
                 &signal_tx,
                 &close_rx,
+                Some(GracefulPeer {
+                    shutdown: shutdown_rx.clone(),
+                    guard: close_rx.clone(),
+                }),
```
mladedav: Is this the only `handle_connection` callsite? Whi is this optional?

mladedav (commented):
Please walk through the generated code and remove superfluous comments/documentation. It's good to have those things but keep it to a level where it provides value.

axum/src/serve/mod.rs
dihannahdi: Right, it doesn't. The comment was rationalizing. hyper drops the upgraded io, but the shutdown signal isn't hyper's, it's our own `signal` watch channel, which the socket can watch directly.

b26a5bc drops the second channel. `GracefulPeer` holds a `signal_tx` clone now and the socket awaits `signal_tx.closed()`, the same edge the accept loop and per-connection tasks already wait on.

Only subtlety: the channel fires by dropping its receiver, so it can only be watched from the sender side. A `signal_rx` clone in the socket would count as a receiver and keep the accept loop's `signal_tx.closed()` from firing.

axum/src/serve/mod.rs
dihannahdi: Two callsites — `Serve::run` passes `None`, this one passes `Some`.

Optional because the plain serve path has no shutdown signal at all: `run` there is `-> !` and its signal receiver is never dropped, so there's nothing to observe and no drain to hold open. `None` also means non-graceful upgrades skip the extension, so that path is unchanged.

dihannahdi:
Done. Cut the editorializing and the comments that just restated the code. The graceful-shutdown explanation was duplicated three times (on `with_graceful_shutdown`, `on_upgrade`, and the module docs), so it now lives in the module docs and the other two link to it.

axum/CHANGELOG.md
```diff
@@ -25,7 +25,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   type, not just `axum::body::Body` ([#3205])
 - **changed:** `Redirect` constructors now accept any `impl Into<String>` ([#3635])
 - **changed:** Updated `matchit` allowing for routes with captures and static prefixes and suffixes ([#3702])
+- **changed:** `Serve::with_graceful_shutdown` now includes WebSocket connections in the drain:
+  open sockets are sent a close frame and the server waits for their handler tasks. A handler
+  that never returns will now hold shutdown open (as with any in-flight response body) ([#3831])
 - **fixed:** Responses to `HEAD` will not accidentally reply with `content-length: 0` anymore ([#3742])
+- **fixed:** graceful shutdown no longer ignores WebSocket handler tasks; the server now waits
+  for them and initiates a close handshake on open sockets ([#3831])
```
mladedav: Why are there two entries about this?

axum/src/extract/ws.rs
```diff
@@ -539,10 +604,38 @@ fn header_contains(headers: &HeaderMap, key: &HeaderName, value: &'static str) -
 /// A stream of WebSocket messages.
 ///
 /// See [the module level documentation](self) for more details.
-#[derive(Debug)]
 pub struct WebSocket {
     inner: WebSocketStream<TokioIo<hyper::upgrade::Upgraded>>,
     protocol: Option<HeaderValue>,
+    /// Resolves once a graceful shutdown has begun, prompting the socket to
+    /// send a Close frame. `None` when graceful shutdown is not configured, in
+    /// which case the socket behaves exactly as before.
```
mladedav: This is a comment that will be here possibly for years to come, "before" doesn't really make sense.

axum/src/extract/ws.rs
```diff
@@ -588,9 +681,51 @@ impl FusedStream for WebSocket {
 impl Stream for WebSocket {
     type Item = Result<Message, Error>;
 
-    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // `WebSocket` is `Unpin`, so we can operate on the fields directly.
```
mladedav: Go through the comments again.

axum/src/serve/mod.rs
```diff
@@ -475,6 +488,10 @@ where
                 &mut make_service,
                 &signal_tx,
                 &close_rx,
+                Some(GracefulPeer {
+                    shutdown: signal_tx.clone(),
+                    guard: close_rx.clone(),
+                }),
```
mladedav: Now we're sending clones of stuff we're already sending inside.

Would it be possible for you to run benchmarks to see if performance is degraded if we simply added the shutdown channels to all requests regardless of whether graceful shutdown was configured and possibly even whether upgrade header was sent?

axum/src/serve/mod.rs
```diff
@@ -230,10 +230,22 @@ where
     /// }
     /// ```
     ///
+    /// # WebSocket connections
+    ///
+    /// WebSocket connections (via [`WebSocketUpgrade`]) take part in the
+    /// shutdown: once the signal fires each open socket is sent a Close frame,
+    /// and this future waits for the handler tasks to finish before resolving.
+    /// A handler that never returns still blocks shutdown, like any other
+    /// in-flight response body. See the [`ws` module docs] for details.
```
mladedav: This is too much detail, just don't mention was at all in this method's docs.

mladedav (changes requested):

axum/CHANGELOG.md
dihannahdi: Collapsed to one. It's a bug fix at heart, so it's a single `fixed` entry now, with the behaviour change folded into the last clause.

axum/src/extract/ws.rs
dihannahdi: Good point, dropped that clause. The doc now just says what the field is. It resolves once shutdown begins and prompts the Close frame, and it's `None` when graceful shutdown isn't configured.

axum/src/extract/ws.rs
dihannahdi: Went through them again. Removed the ones that only narrate the code (the `Unpin` note, the "don't poll it again", the per-arm asides) and kept one comment up top for why we send Close but keep reading.

axum/src/serve/mod.rs
dihannahdi: Right, that was redundant. `handle_connection` already has `signal_tx` and `close_rx`, so it builds the peer from those now and nothing extra is passed in.

On perf, I ran the `benches/` harness (rewrk, four 10s runs) comparing `main` against a build that inserts the peer into every request, including the plain `serve` path where nothing ever reads it. The most overhead-sensitive scenarios, four runs each.

| scenario | main (req/s) | always-insert (req/s) | delta |
| --- | --- | --- | --- |
| minimal | 48049 | 47642 | -0.8% |
| basic | 41189 | 40278 | -2.2% |
| routing | 42721 | 41502 | -2.9% |

Those deltas sit inside the run-to-run noise. `main` on its own swung 6 to 10 percent between runs, and the always-insert runs were actually tighter at about 2 percent. Past a trivial handler the single insert disappears.

So the gating is gone. The peer goes into every request, the `Upgrade`/`CONNECT` check is removed, and both `handle_connection` callsites are identical now, no more `Option`/`None`. If you'd rather the plain `serve` path not carry an extension it never uses, I can gate the insert on graceful shutdown being configured and still keep the header check gone. Happy either way.

axum/src/serve/mod.rs
dihannahdi: Removed. `with_graceful_shutdown` no longer mentions websockets, that detail lives only in the ws module docs now.

dihannahdi:
Coming back to this after a couple of quiet weeks. Two things from my side, and one question about ordering.

First, I left you with a choice at the end of the last round rather than a decision, which was not useful. I said I could gate the insert on graceful shutdown being configured if you preferred it that way. I am not going to. Your question was whether the plain path could simply carry the channels regardless, the benchmarks said it could, and putting a configuration check back would only reintroduce a branch to save something the numbers could not detect. If you read the code and still want it gated, I will do it, but you should not have been left holding that decision.

Second, I reread the PR description and found it was still describing the design from before your review. It talked about a second watch channel, and about inserting only on requests carrying an upgrade header, both of which you asked me to remove and both of which have been gone from the code for two rounds now. Anyone arriving fresh would have been reading about an implementation that no longer exists. I have rewritten it to match what is actually on the branch. No code moved for that, only the text.

There is also one small commit that swaps the em dashes in the new comments for plain prose, since nothing else under `axum/src` uses them.

The ordering question is #3819, which moves `serve` into its own crate and touches the same file this one grows. Whichever of the two lands second will need rework, and it is much cheaper for that to be this one. If you would rather park this until #3819 is settled, say so and I will wait, rather than hand you a conflict later.

axum/src/extract/ws.rs
```diff
@@ -637,9 +669,39 @@ impl FusedStream for WebSocket {
...
+        };
+        if shutdown_fired {
+            this.shutdown = None;
+            this.close_state = CloseState::Queued;
+        }
+
+        if this.close_state == CloseState::Queued {
+            match this.inner.poll_ready_unpin(cx) {
+                Poll::Ready(Ok(())) => {
+                    this.close_state = match this.inner.start_send_unpin(ts::Message::Close(None)) {
```
jlizen: Probably `Close(Some(CloseFrame { code: 1001, .. })` is better?

axum/src/serve/mod.rs
```diff
@@ -563,9 +556,24 @@ where
...
 
+/// Lets an upgraded connection (e.g. a WebSocket) take part in graceful
+/// shutdown. Inserted into every request's extensions and read by
+/// [`WebSocketUpgrade`].
+///
+/// [`WebSocketUpgrade`]: crate::extract::ws::WebSocketUpgrade
+// Only read by the `ws` extractor.
+#[cfg_attr(not(feature = "ws"), allow(dead_code))]
+#[derive(Clone, Debug)]
+pub(crate) struct GracefulPeer {
```
jlizen: nit: `ConnectionShutdownHandle` is clearer at usage site?

axum/src/extract/ws.rs
```diff
@@ -1313,4 +1375,198 @@ mod tests {
             tungstenite::Message::Pong(Bytes::from_static(b"ping"))
         );
     }
+
+    // Graceful-shutdown interaction. These need `axum::serve`, which is only
+    // available with `http1`/`http2` (the `ws` feature alone does not imply it).
```
jlizen: We should also have:
- a test that websockets under "plain" serve (no graceful shutdown) do NOT get close frames
- tests using `split()`:
   - graceful shutdown, both sides see the proper frames
   - write half dropped early, read half sees proper frame
   - read half dropped early, write half sees proper frames

axum/CHANGELOG.md
```diff
@@ -21,15 +21,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-- **added:** `sse::Event::raw` for events with fully custom payloads ([#3829])
 - **changed:** `serve` has an additional generic argument and can now work with any response body
   type, not just `axum::body::Body` ([#3205])
-- **changed:** Reduced contention in `axum::serve` shutdown notification with many
-  active connections ([#3867])
 - **changed:** `Redirect` constructors now accept any `impl Into<String>` ([#3635])
 - **changed:** Updated `matchit` allowing for routes with captures and static prefixes and suffixes ([#3702])
 - **fixed:** Responses to `HEAD` will not accidentally reply with `content-length: 0` anymore ([#3742])
-- **added:** Add `MethodFilter::QUERY`, `routing::query[_service]` and `MethodRouter::query[_service]` ([#3801])
+- **fixed:** `Serve::with_graceful_shutdown` now waits for WebSocket handler tasks and sends open
```
jlizen: This is a bit of a behavioral change, since an orphaned handler now will keep the connection open. Probably the right choice, but it'd be nice to qualify this as a breaking behavioral change rather than just a fix and loop it into the upcoming 0.9 release

axum/src/serve/mod.rs
```diff
@@ -592,16 +599,27 @@ async fn handle_connection<L, M, S, B, E>(
...
         .call(IncomingStream {
             io: &io,
             remote_addr,
         })
         .await
         .unwrap_or_else(|err| match err {})
-        .map_request(|req: Request<Incoming>| req.map(Body::new));
+        .map_request(move |req: Request<Incoming>| {
+            let mut req = req.map(Body::new);
+            req.extensions_mut().insert(graceful.clone());
```
jlizen: I'm not suggesting that we check the headers every request, but could we not pass through `Option<GracefulPeer>` based on whether graceful shutdown is enabled on the connection at all? Saves us an allocation and some extra polling per connection without any meaningful increase in code complexity.

axum/src/extract/ws.rs
```diff
@@ -134,14 +88,29 @@
...
+//!
+//! When the server is run with [`axum::serve`] and
+//! [`Serve::with_graceful_shutdown`], WebSocket connections take part in the
+//! shutdown: once the signal fires each open [`WebSocket`] sends a Close frame
+//! to its peer, and the server waits for the handler tasks to finish before
+//! resolving.
+//!
+//! A handler with the usual read loop needs no changes. After the Close
+//! handshake [`WebSocket::recv`] returns `None`, so the loop ends and the task
+//! returns. A handler that never polls the socket and never returns will still
```
jlizen: Should we include a sample showing how to race the serve future against a hard timeout, to handle this case?

jlizen (commented):
I started to merge against main before realizing main changed a fair bit underneath us, so I undid the merge. Please do so when you have time @dihannahdi .

Some other comments are in this review.

axum/src/extract/ws.rs
dihannahdi: Done. `1001` with reason `server shutting down`, built through `Message::Close` + `close_code::AWAY` so it goes through the same conversion as a user close. The tests assert the code, not just that some close frame arrived.

axum/src/serve/mod.rs
dihannahdi: Renamed to `ConnectionShutdownHandle`.

axum/CHANGELOG.md
dihannahdi: Moved to `breaking:` and reworded to say the serve future now waits and an orphaned handler holds shutdown open.

axum/src/extract/ws.rs
dihannahdi: Added to the ws module docs, right under the paragraph about handlers that never return.

axum/src/serve/mod.rs
dihannahdi: Done. `handle_connection` now takes `Option<&ConnectionShutdownHandle>`: `Serve::run` passes `None`, `WithGracefulShutdown::run` builds one handle before the accept loop and passes `Some(&graceful)`. Plain serve skips the extension insert, and with no handle in the request the socket has no future to poll, so `poll_shutdown_close` is one enum compare.

The merge with main made this cheaper than it was. Since #3867 the signal fires by dropping the sender, so the handle carries a `watch::Receiver` and the sender-side `closed()` workaround this branch had is gone.

axum/src/extract/ws.rs
dihannahdi: Added all four, plus the existing two reworked onto a shared harness.

The write-half-alone case needed a real fix, not just a test. The close frame was only injected in `Stream::poll_next`, so a handler that splits and drops the read half never saw shutdown and held the server open forever. `poll_shutdown_close` now runs from `Sink::poll_ready` and `poll_flush` as well, and `WebSocket::send` goes through the `Sink` impl instead of straight to `inner` so the plain (unsplit) send path gets it too.

Checked the tests actually bite:

- drop the `Sink` side call: `split_write_half_alone_sees_shutdown` fails, serve future times out
- drop the `Stream` side call: `waits_for_reading_socket` and `split_read_half_alone_sees_shutdown` fail

`plain_serve_never_closes_sockets` echoes twice under plain `serve` and asserts no frame arrives in between.

dihannahdi:
Merged main and worked through the review, thanks @jlizen.

#3867 changed the shape of this. Shutdown now fires by dropping the signal sender and connections watch `signal_rx.changed()`, so the handle carries a `watch::Receiver` and the sender-side `closed()` trick this branch was using is gone. Simpler than what was here before.

Beyond the merge:

- `GracefulPeer` is now `ConnectionShutdownHandle`, and `handle_connection` takes it as an `Option`. Plain `serve` passes `None`, so no extension insert and nothing to poll on the socket.
- Close frame carries `1001` and a reason.
- The close injection also runs from `Sink::poll_ready`/`poll_flush`, not just `Stream::poll_next`. Without that a `split()` handler that drops the read half never sees shutdown and hangs the server. `WebSocket::send` routes through the `Sink` impl now for the same reason.
- Six tests: the two originals plus split with both halves live, read half alone, write half alone, and plain serve never closing a socket. Confirmed each side of the injection is load bearing by removing it and watching the matching tests fail.
- CHANGELOG entry is `breaking:` now.

One defect of my own found while checking: `poll_shutdown_close` bailed out once the signal had been consumed, so a close frame left `Queued` behind a `Pending` sink was never retried. Fixed in ec44d1e, the early return keys off the state instead.

Green locally: `cargo test -p axum --all-features` (only pre-existing `routing::tests::body_limited_by_default` fails, also fails on main), doc tests, `cargo clippy --all-features --all-targets -D warnings`, `cargo doc -D warnings`, and `tokio` / `http1` / `ws` / `ws,tokio` / `http1,tokio` / no-default feature combos.

PR #3837, opened 2026-07-20
feat(routing): support the HTTP QUERY method

### LINKED ISSUE #3799: Support for the HTTP QUERY Method

farrosfr:
# feat(routing): support HTTP QUERY method (RFC 10008)

Resolves #3799.

## Overview

Adds routing filters, handlers, and typed routing support for the newly standardized HTTP `QUERY` method (RFC 10008).

### `axum`
* **`MethodFilter::QUERY`**: Defined a new bitflag (`0b10_0000_0000`).
* **`TryFrom<Method> for MethodFilter`**: Added fallback matching (`other if other.as_str() == "QUERY"`).
* **`MethodRouter`**:
  * Added `query: MethodEndpoint<S, E>` field.
  * Added top-level `query` and `query_service` functions.
  * Added chained `.query(handler)` and `.query_service(service)` methods.
  * Updated the internal `call!` macro to match `"QUERY"` string slices for routing.
* **Testing**: Added a generic `.request(method, url)` to the test client and an integration test verifying that `QUERY` requests route to the `query` handler.

### `axum-extra`
* **`RouterExt::typed_query`**: Implemented typed routing support for `QUERY` requests on `TypedPath`.

## Compatibility Notes

Because the `http` crate has not yet released a new version containing `Method::QUERY`, the implementation uses string comparisons (`req.method().as_str() == "QUERY"`). This compiles on current stable dependencies and remains forward-compatible.

### DIFF
--- a/axum-extra/src/routing/mod.rs
+++ b/axum-extra/src/routing/mod.rs
@@ -234,6 +234,19 @@ pub trait RouterExt<S>: sealed::Sealed {
         T: SecondElementIs<P> + 'static,
         P: TypedPath;
 
+    /// Add a typed `QUERY` route to the router.
+    ///
+    /// The path will be inferred from the first argument to the handler function which must
+    /// implement [`TypedPath`].
+    ///
+    /// See [`TypedPath`] for more details and examples.
+    #[cfg(feature = "typed-routing")]
+    fn typed_query<H, T, P>(self, handler: H) -> Self
+    where
+        H: axum::handler::Handler<T, S>,
+        T: SecondElementIs<P> + 'static,
+        P: TypedPath;
+
     /// Add another route to the router with an additional "trailing slash redirect" route.
     ///
     /// If you add a route _without_ a trailing slash, such as `/foo`, this method will also add a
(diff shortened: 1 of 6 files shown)


mladedav:
There's already #3801 which waits for upstream `http` release. I'd rather we already use the newer `http::Method` with the method than to compare strings here.

You can check that PR and discuss there if you think anything else is missing in the implementation.

farrosfr:
Thanks @mladedav  I am sorry because I didn't check for existing open PRs before submitting.

Okay i will check if there is any need for implementation in that PR.

PR #3838, opened 2026-07-20
Add docs describing how, why and when to use &'static state

FalkWoldmann:
## Motivation
Supersedes https://github.com/tokio-rs/axum/pull/3815

### DIFF
--- a/axum/src/lib.rs
+++ b/axum/src/lib.rs
@@ -186,7 +186,78 @@
 //! State is cloned for every request. Wrapping your state in `Arc` makes those
 //! clones cheap. If all fields are already cheap to clone (for example, each field
 //! is itself an `Arc` or a copy type), you can `#[derive(Clone)]` directly on the
-//! struct instead.
+//! struct instead. Many client types, such as [`reqwest::Client`], AWS SDK service
+//! clients, and [`mongodb::Client`], already use shared ownership internally and
+//! are cheap to clone. Such clients generally do not need another `Arc` solely to
+//! make cloning cheap; doing so adds another level of indirection.
+//!
+//! [`reqwest::Client`]: https://docs.rs/reqwest/latest/reqwest/struct.Client.html
+//! [`mongodb::Client`]: https://docs.rs/mongodb/latest/mongodb/struct.Client.html
+//!
+//! ### Using `&'static` state
+//!
+//! For state built once and intended to live until the process exits, [`Box::leak`]
+//! or a static [`LazyLock`] can provide a `&'static AppState` to the
+//! router.[^static-state] This is useful for state built from runtime configuration,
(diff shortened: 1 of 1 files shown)


axum/src/lib.rs
```diff
@@ -186,7 +186,78 @@
...
+//! make cloning cheap; doing so adds another level of indirection.
+//!
+//! [`reqwest::Client`]: https://docs.rs/reqwest/latest/reqwest/struct.Client.html
+//! [`mongodb::Client`]: https://docs.rs/mongodb/latest/mongodb/struct.Client.html
+//!
+//! ### Using `&'static` state
+//!
+//! For state built once and intended to live until the process exits, [`Box::leak`]
+//! or a static [`LazyLock`] can provide a `&'static AppState` to the
+//! router.[^static-state] This is useful for state built from runtime configuration,
```
mladedav: I'm not sure how I feel about the footnote. It does work on docs.rs (I didn't know that tbh) but it's at the bottome of the text before items of the crate and the last thing just before happnes to be the features table and that makes it look like continuation of that rather than footnotes.

Though it is a numbered list (with one element) and it has a backlink, so maybe it's completely fine. But I wonder if you can think of a way to embed the information into the text here directly, or drop it.

What do you think?

Image

axum/src/lib.rs
```diff
@@ -186,7 +186,78 @@
...
+//! }
+//!
+//! let app_state = AppState {
+//!     // Initialize fields at startup.
+//! };
+//! let app_state: &'static AppState = Box::leak(Box::new(app_state));
+//!
+//! let app = Router::new()
+//!     .route("/", get(handler))
+//!     .with_state(app_state);
```
mladedav: Wrap this into a `fn main`. I think that would make the example clearer as there would be better distinction between defining items and statements.

mladedav (commented):
Looks good, thanks. Just two minor things.

axum/src/lib.rs
FalkWoldmann: Fine with every possibilties tbh. I could try summarizing the difference more (ie. just mention the sync vs async construction difference), which could maybe be too much condensed info at once. Omitting it could confuse people new to Rust, which wouldn't be so nice since a lot of Rust newbies stumble across axum (which speaks to the quality of the crate!). Just putting the content inline would be a bit verbose. That's why I like elaborating via footnotes, but I also see your objections regarding the visual confusion, especially since the usage of footnotes in docs.rs is still quite sparse.

We could add an additional H1 headline "References" to further differentiate from crate features

FalkWoldmann: I inlined the description

mladedav (approved):
Thanks

Issue #3793, opened 2026-06-10
ws: prefer actor pattern to `StreamExt::split`


vilgotf:
Axum's WS documentation highlights the `StreamExt::split` method for independent send & recv tasks. This is, however, just an `Arc<Mutex>` wrapper (with all of its inherent performance overhead). The [actor pattern](https://ryhl.io/blog/actors-with-tokio/) for I/O resources is a higher performance alternative. Axum is a high performance web framework, so it's recommended patterns should similarly be high performance.

Refs: #2695

mubasharameen485-cloud:
i work on it

PR #3842, opened 2026-07-21
docs(axum): recommend actor pattern for concurrent WebSocket read/write

0xendale:
## Motivation

The WebSocket module docs suggest `StreamExt::split` as the way to read and write concurrently. Splitting wraps each half in a lock that must be acquired on every operation, which adds avoidable overhead. Upstream issue tokio-rs/axum#3793 proposes highlighting the actor pattern as the higher-performance alternative, and tokio-rs/axum#3786 already laid the groundwork by documenting that `WebSocket::recv` is cancel safe.

## Solution

Rework the "Read and write concurrently" section in `axum/src/extract/ws.rs`:

- Lead with the actor pattern: a single task owns the socket and uses `tokio::select!` to race `WebSocket::recv` (cancel safe) against an `mpsc` channel that other tasks use to send messages to the client. Links to [Actors with Tokio](https://ryhl.io/blog/actors-with-tokio/) for background.
- Keep `StreamExt::split` as a documented alternative, with a note about its per-operation locking overhead.

Both examples are compile-checked doc tests. Verified with `cargo test -p axum --all-features --doc ws` and `cargo doc` using CI's rustdoc flags (`-D rustdoc::all -A rustdoc::private-doc-tests`); the test-only hidden line does not appear in the rendered docs.

### DIFF
--- a/axum/src/extract/ws.rs
+++ b/axum/src/extract/ws.rs
@@ -65,8 +65,54 @@
 //!
 //! # Read and write concurrently
 //!
-//! If you need to read and write concurrently from a [`WebSocket`] you can use
-//! [`StreamExt::split`]:
+//! To read and write concurrently it's not necessary to split the socket into
+//! separate halves. Instead, have a single task own the [`WebSocket`] — the
+//! [actor pattern] — and use [`select!`] to race receiving a message from the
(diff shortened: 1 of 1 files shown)


mladedav (approved):
Thanks and sorry for the delay

PR #3850, opened 2026-07-28
fix(axum): don't enable `tokio/macros` feature

snejugal:
## Motivation

I'm trying to minimize the dependency tree of a project I'm working on. I noticed that `tokio` had its `macros` feature enabled only because of `axum`, and this feature pulls in the `tokio-macros` crate. It's one crate, but it's still nice to avoid compiling code that's unused anyway.

## Solution

I removed the `tokio/macros` features. I checked that `axum` still compiles without any changes to the code at all and that all tests pass. I also manually skimmed through uses of `tokio` with ripgrep and found no use of `tokio`'s macros outside of tests (and this feature is already once again enabled in `dev-dependencies` anyway).

I also found it strange that the `tokio` feature enables `tokio` _and_ lists some of its features, yet the table below also lists some of `tokio`'s features. That's why I moved the remaining `rt` feature to that table.

### DIFF
--- a/axum/Cargo.toml
+++ b/axum/Cargo.toml
@@ -72,9 +72,7 @@ query = [
 tokio = [
     "dep:hyper-util",
     "dep:tokio",
-    "tokio/rt",
     "tower/make",
-    "tokio/macros",
 ]
 tower-log = ["tower/log"]
 tracing = ["dep:tracing", "axum-core/tracing"]
@@ -102,10 +100,10 @@ __private_docs = [
 __private = ["tokio", "http1", "dep:reqwest"]
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-tokio = { package = "tokio", version = "1.44", features = ["time", "net"], optional = true }
+tokio = { package = "tokio", version = "1.44", features = ["time", "rt", "net"], optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-tokio = { package = "tokio", version = "1.44", features = ["time"], optional = true }
+tokio = { package = "tokio", version = "1.44", features = ["time", "rt"], optional = true }
 
 [dependencies]
 axum-core = { path = "../axum-core", version = "0.5.6" }


snejugal:
huh, CI failed and did find some use of `tokio::select`. I guess I forgot to run `cargo c` with `--all-features`, and `cargo test --all-features` just misses that. `select` from `futures_util` doesn't work easily, so I guess I'll live on with `tokio-macros` and close this PR.

mladedav:
If you want this and you're not using `axum::serve`, you can enable tokio macros if we have either http1 or http2 features.

Personally, I prefer the futures utils select function to the tokio select macro, so I wouldn't mind that either (assuming there are just two branches, didn't check) so if you want to do that, you also can.

If you decide that's too much work for small benefit, feel free to ignore this of course.

snejugal:
> Personally, I prefer the futures utils select function to the tokio select macro, so I wouldn't mind that either (assuming there are just two branches, didn't check) so if you want to do that, you also can.

The issue that I faced is that `select!` is used with `!Unpin` futures, but `futures`'s `select` function requires `Unpin`. The only way to fix this (that I'm aware of) is to box these futures, but `select!` avoids boxing. So it looks to me like a trade-off to be made, and I'm not really sure that removing `tokio-macros` from the dependency tree is worth it. I'm mostly looking for dependencies that end up unused or can be trivially replaced anyway.

mladedav:
The futures can also be pinned onto stack, see #3851

snejugal:
oh, I missed that `Pin<&mut T>` is `Unpin` on its own. That's nice, thanks!

Should I wait for your PR to get merged and reopen this one?

mladedav:
The select change is merged so you can look if your changes pass cargo hack now

snejugal:
got back to this PR just now. checks should pass now, but I rebased my branch and GitHub won't let me reopen this PR anymore, so I opened https://github.com/tokio-rs/axum/pull/3864 instead.

PR #3852, opened 2026-07-30
Configure axum http1 and http2 via closure

gmarti:
### Motivation

  Currently, axum::serve doesn't expose any way to configure the underlying hyper HTTP/1 or HTTP/2 connection settings. Users who need to set things like header_read_timeout, keep_alive, or max_concurrent_streams are directed to drop down to raw hyper/hyper-util — which is a significant ergonomics hit for what are often simple one-liner configurations.

###  Solution

Add a .with_hyper_configuration() method on Serve (and WithGracefulShutdown) that accepts a closure receiving the full hyper_util::server::conn::auto::Builder. This gives full access to hyper's configuration without axum needing to mirror every setting individually.

```
  axum::serve(listener, router)
      .with_hyper_configuration(|builder| {
          builder.http1().header_read_timeout(Duration::from_secs(5));
          builder.http2().max_concurrent_streams(200);
      })
      .await;
```

  When set, the closure replaces axum's default configuration (HTTP/1 timer, HTTP/2 CONNECT protocol) — the user has full control over the builder.

###  Design

  A sealed HyperConfigure<E> trait is introduced with two implementations:
  - () — applies axum's defaults (the zero-config path)
  - Fn(&mut Builder<HyperExecutor<E>>) — user-supplied full-control closure

  H is added as a type parameter on Serve (defaulting to ()), which gives compile-time enforcement of method ordering: with_executor() is only available when H = (), preventing a class of bugs where changing the executor type would invalidate a previously-set closure.

  Valid call orderings (all enforced at the type level):
```
  serve(listener, router).await;
  serve(listener, router).with_executor(exec).await;
  serve(listener, router).with_executor(exec).with_hyper_configuration(|b| {}).await;
  serve(listener, router).with_hyper_configuration(|b| {}).await;
  serve(listener, router).with_graceful_shutdown(signal).with_executor(exec).await;
  serve(listener, router).with_graceful_shutdown(signal).with_hyper_configuration(|b| {}).await;
```

  ---
  This PR was co-authored with Claude Code (Anthropic).

### DIFF
--- a/axum/src/serve/mod.rs
+++ b/axum/src/serve/mod.rs
@@ -100,7 +100,7 @@ pub use self::listener::{ConnLimiter, ConnLimiterIo, Listener, ListenerExt, TapI
 /// [`HandlerWithoutStateExt::into_make_service_with_connect_info`]: crate::handler::HandlerWithoutStateExt::into_make_service_with_connect_info
 /// [`HandlerService::into_make_service_with_connect_info`]: crate::handler::HandlerService::into_make_service_with_connect_info
 #[cfg(all(feature = "tokio", any(feature = "http1", feature = "http2")))]
-pub fn serve<L, M, S, B>(listener: L, make_service: M) -> Serve<L, M, S, B, TokioExecutor>
+pub fn serve<L, M, S, B>(listener: L, make_service: M) -> Serve<L, M, S, B>
 where
     L: Listener,
     M: for<'a> Service<IncomingStream<'a, L>, Error = Infallible, Response = S>,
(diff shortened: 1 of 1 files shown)


gmarti:
linked to #2939 and #2938

mladedav:
We don't want to expose hyper utils as public dependency of axum.

There's #3484 going in a different direction that could enable this.

gmarti:
@mladedav yeah thx for the link, but looks like its going to take a while !

I think we will use the fork in the meantime... the current solution with switching to hyper is not maintenable

I will close the MR

PR #3856, opened 2026-08-02
feat: add InnerPath extractor

### LINKED ISSUE #3017: Make UrlParams public

DanBondarenko:
Closes #3017

Add new extractor `InnerPath` capturing path params from the innermost router.

## Motivation

Nested handlers shouldn't have to be concerned about params of routers under which they've been nested. There's no way to exclude capture params from enclosing routers. `Path` returns every capture in the path. Routers using it can't be relocated or mounted at different prefixes.

## Solution

`InnerPath` deserializes captures only from the innermost router's route. Behavioral relationship of `InnerPath` and `Path` is comparable to `Uri` and `OriginalUri`.

Implementation:
- adds a `CountEnclosingCaptures` layer when router is nested, updating `EnclosingCapturesCount` in extensions
- `InnerPath` uses `UrlParams` just like `Path`, but skips this nr of params and deserializes the rest
- ~behind `inner-path` feature, off by default~
- extracted shared logic with `Path` to new helper functions

Happy to make changes. Idea for naming change: `Path` -> `OriginalPath`, `InnerPath` -> `Path` (like `OriginalUri` and `Uri`). Alternative: `InnerPath` -> `RouterPath`.

Also: should `InnerPathRejection` and `PathRejection` be unified? Kept them separate, but it introduced some complexity.

### DIFF
--- a/axum/src/extract/path/inner_path.rs
+++ b/axum/src/extract/path/inner_path.rs
@@ -230,10 +230,15 @@ where
     }
 
     fn call(&mut self, mut req: Request<B>) -> Self::Future {
-        let previous = req
-            .extensions_mut()
-            .get_or_insert_default::<EnclosingCapturesCount>();
-        previous.0 += self.additional_captures_count;
+        let extensions = req.extensions_mut();
+        let EnclosingCapturesCount(previous) = extensions
+            .get::<EnclosingCapturesCount>()
+            .copied()
+            .unwrap_or_default();
+
+        extensions.insert(EnclosingCapturesCount(
+            previous + self.additional_captures_count,
+        ));
 
         self.inner.call(req)
     }
--- a/axum/src/extract/path/inner_path.rs
+++ b/axum/src/extract/path/inner_path.rs
@@ -1,10 +1,7 @@
-use crate::extract::path::{
-    de, ErrorKind, FailedToDeserializePathParams, PathDeserializationError,
-};
-use crate::extract::rejection::{InnerPathRejection, MissingPathParams};
+use crate::extract::path::{get_params, serialize_path_params};
+use crate::extract::rejection::InnerPathRejection;
 use crate::extract::FromRequestParts;
 use crate::routing::strip_prefix::count_captures;
-use crate::routing::url_params::UrlParams;
 use crate::util::PercentDecodedStr;
 use axum_core::extract::{OptionalFromRequestParts, Request};
 use axum_core::RequestPartsExt;
@@ -151,44 +148,26 @@ where
     type Rejection = InnerPathRejection;
 
     async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
-        // Extracted into separate fn so it's only compiled once for all T.
-        fn get_params(
+        // Non-T computation is in standalone function preventing monomorphization over T
+        fn get_inner_params(
             parts: &Parts,
         ) -> Result<&[(Arc<str>, PercentDecodedStr)], InnerPathRejection> {
-            match parts.extensions.get::<UrlParams>() {
-                Some(UrlParams::Params(params)) => {
-                    let EnclosingCapturesCount(enclosing_captures_count) = parts
-                        .extensions
-                        .get::<EnclosingCapturesCount>()
-                        .copied()
-                        .unwrap_or_default();
+            let EnclosingCapturesCount(enclosing_captures_count) = parts
+                .extensions
+                .get::<EnclosingCapturesCount>()
+                .copied()
+                .unwrap_or_default();
 
-                    let inner_params = params
-                        .get(enclosing_captures_count..)
-                        .expect("Mismatch between url params and count of captures. This is bug in axum. Please file an issue.");
+            let inner_params = get_params::<InnerPathRejection>(parts)?
+                .get(enclosing_captures_count..)
+                .expect("Mismatch between url params and count of captures. This is bug in axum. Please file an issue.");
 
-                    Ok(inner_params)
-                }
-                Some(UrlParams::InvalidUtf8InPathParam { key }) => {
-                    let err = PathDeserializationError {
-                        kind: ErrorKind::InvalidUtf8InPathParam {
-                            key: key.to_string(),
-                        },
-                    };
-                    Err(FailedToDeserializePathParams(err).into())
-                }
-                None => Err(MissingPathParams.into()),
-            }
+            Ok(inner_params)
         }
 
-        fn failed_to_deserialize_path_params(err: PathDeserializationError) -> InnerPathRejection {
-            InnerPathRejection::FailedToDeserializePathParams(FailedToDeserializePathParams(err))
-        }
-
-        match T::deserialize(de::PathDeserializer::new(get_params(parts)?)) {
-            Ok(val) => Ok(Self(val)),
-            Err(e) => Err(failed_to_deserialize_path_params(e)),
-        }
+        get_inner_params(parts)
+            .and_then(serialize_path_params)
+            .map(Self)
     }
 }
 
@@ -206,7 +185,7 @@ where
         match parts.extract::<Self>().await {
             Ok(Self(params)) => Ok(Some(Self(params))),
             Err(InnerPathRejection::FailedToDeserializePathParams(e))
-                if matches!(e.kind(), ErrorKind::WrongNumberOfParameters { got: 0, .. }) =>
+                if e.kind().is_zero_params() =>
             {
                 Ok(None)
             }
--- a/axum/src/extract/path/inner_path.rs
+++ b/axum/src/extract/path/inner_path.rs
@@ -16,7 +16,7 @@ use tower_service::Service;
 /// and parse them using [`serde`].
 ///
 /// Any percent encoded parameters will be automatically decoded. The decoded
-/// parameters must be valid UTF-8, otherwise `Path` will fail and return a `400
+/// parameters must be valid UTF-8, otherwise `InnerPath` will fail and return a `400
 /// Bad Request` response.
 ///
 /// # `InnerPath` vs [`Path`]
@@ -27,7 +27,7 @@ use tower_service::Service;
 ///
 /// # `Option<InnerPath<T>>` behavior
 ///
-/// You can use `InnerPath<Path<T>>` as an extractor to allow the same handler to
+/// You can use `Option<InnerPath<Path<T>>>` as an extractor to allow the same handler to
 /// be used in a route with parameters that deserialize to `T`, and another
 /// route with no parameters at all.
 ///
@@ -52,7 +52,6 @@ use tower_service::Service;
 ///     // `user_id` is from the `{user_id}` capture,
 ///     // while `{tenant_id}` from the outer router isn't included,
 ///     // so this handler works no matter what it's nested under.
-///     todo!()
 /// }
 ///
 /// let api = Router::new().route("/users/{user_id}", get(user_info));
@@ -112,13 +111,12 @@ use tower_service::Service;
 ///     InnerPath(params): InnerPath<HashMap<String, String>>,
 /// ) {
 ///     // `params` contains keys "team_id" and "user_id"
-///     todo!()
 /// }
 ///
 /// async fn params_vec(
 ///     InnerPath(params): InnerPath<Vec<(String, String)>>,
 /// ) {
-///     todo!()
+///     // `params` contains a vector containing a single tuple
 /// }
 ///
 /// let api = Router::new()
--- a/axum/src/extract/path/inner_path.rs
+++ b/axum/src/extract/path/inner_path.rs
@@ -135,6 +135,8 @@ use tower_service::Service;
 /// [`serde`]: https://crates.io/crates/serde
 /// [`serde::Deserialize`]: https://docs.rs/serde/1.0.127/serde/trait.Deserialize.html
 /// [`customize-path-rejection`]: https://github.com/tokio-rs/axum/blob/main/examples/customize-path-rejection/src/main.rs
+/// [`Path`]: crate::extract::Path
+/// [`nest`]: crate::routing::Router::nest
 #[derive(Debug)]
 pub struct InnerPath<T>(pub T);
 
--- a/axum/src/extract/path/inner_path.rs
+++ b/axum/src/extract/path/inner_path.rs
@@ -1,16 +1,15 @@
-use crate::extract::path::{get_params, serialize_path_params};
-use crate::extract::rejection::InnerPathRejection;
+use crate::extract::path::{
+    serialize_path_params, ErrorKind, FailedToDeserializePathParams, PathDeserializationError,
+};
+use crate::extract::rejection::{InnerPathRejection, MissingPathParams};
 use crate::extract::FromRequestParts;
-use crate::routing::strip_prefix::count_captures;
+use crate::routing::url_params::UrlParams;
 use crate::util::PercentDecodedStr;
-use axum_core::extract::{OptionalFromRequestParts, Request};
+use axum_core::extract::OptionalFromRequestParts;
 use axum_core::RequestPartsExt;
 use http::request::Parts;
 use serde_core::de::DeserializeOwned;
 use std::sync::Arc;
-use std::task::{Context, Poll};
-use tower_layer::{layer_fn, Layer};
-use tower_service::Service;
 
 /// Extractor that will get captures from the path of the innermost router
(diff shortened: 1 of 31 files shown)


axum/src/extract/path/mod.rs
```diff
@@ -601,6 +601,46 @@ impl IntoResponse for InvalidUtf8InPathParam {
...
+}
+
+fn serialize_path_params<T, E>(params: &[(Arc<str>, PercentDecodedStr)]) -> Result<T, E>
+where
+    T: DeserializeOwned + Send,
+    E: From<FailedToDeserializePathParams>,
+{
+    // Non-T computation is in standalone function preventing monomorphization over T
+    let deserializer = de::PathDeserializer::new(params);
+    T::deserialize(deserializer).map_err(into_path_rejection)
```
mladedav: I'm not a fan of the `.into().into()` helper function.

Here you can do

```rust
    T::deserialize(deserializer)
        .map_err(FailedToDeserializePathParams::from)
        .map_err(E::from)
```

and in the other place

```rust
Err(E::from(FailedToDeserializePathParams::from(error)))
```

or some other variation on the same themes.

axum/src/extract/path/mod.rs
```diff
@@ -601,6 +601,46 @@ impl IntoResponse for InvalidUtf8InPathParam {
...
+        }
+    }
+}
+
+fn serialize_path_params<T, E>(params: &[(Arc<str>, PercentDecodedStr)]) -> Result<T, E>
+where
+    T: DeserializeOwned + Send,
+    E: From<FailedToDeserializePathParams>,
+{
+    // Non-T computation is in standalone function preventing monomorphization over T
```
mladedav: I don't get what you're trying to say here, could you elaborate please?

axum/src/extract/path/inner_path.rs
```diff
@@ -0,0 +1,760 @@
...
+use std::sync::Arc;
+use std::task::{Context, Poll};
+use tower_layer::{layer_fn, Layer};
+use tower_service::Service;
+
+/// Extractor that will get captures from the path of the innermost router
+/// and parse them using [`serde`].
+///
+/// Any percent encoded parameters will be automatically decoded. The decoded
+/// parameters must be valid UTF-8, otherwise `Path` will fail and return a `400
```
mladedav: `Path`

axum/src/extract/path/inner_path.rs
```diff
@@ -0,0 +1,760 @@
...
+///
+/// # `InnerPath` vs [`Path`]
+/// [`Path`] gives you every capture in the request path, including those matched by enclosing
+/// [`nest`] calls. `InnerPath` gives you only the ones from the route the innermost router
+/// matched. If the router isn't nested, or is nested only at static paths, `InnerPath` and [`Path`] see
+/// exactly the same captures.
+///
+/// # `Option<InnerPath<T>>` behavior
+///
+/// You can use `InnerPath<Path<T>>` as an extractor to allow the same handler to
```
mladedav: Did you mean `Option<InnerPath<T>>`?

axum/src/extract/path/inner_path.rs
```diff
@@ -0,0 +1,760 @@
...
+///     routing::get,
+///     Router,
+/// };
+/// use std::collections::HashMap;
+///
+/// async fn params_map(
+///     InnerPath(params): InnerPath<HashMap<String, String>>,
+/// ) {
+///     // `params` contains keys "team_id" and "user_id"
+///     todo!()
```
mladedav: Drop the todos (both in this example), the previous examples don't use them either.

axum/src/routing/path_router.rs
```diff
@@ -192,10 +194,18 @@ where
...
             let layer = (
                 StripPrefix::layer(prefix),
                 SetNestedPath::layer(path_to_nest_at),
+                CountEnclosingCaptures::layer(path_to_nest_at),
             );
+            #[cfg(not(feature = "inner-path"))]
+            let layer = (
+                StripPrefix::layer(prefix),
+                SetNestedPath::layer(path_to_nest_at),
+            );
```
mladedav: I like this (repeated) cfg if else less than just having the captures counting on even when not needed. Can we move the layer somewhere so that it's not behind a feature and have it on by default?

I'm not entirely sure about this one though, if you could run some benchmarks if it produces measurable difference, that would be best. Otherwise I can try and take a look when I have some time or just go with my guts.

mladedav (commented):
Thanks for the PR!

Looking at it, we might also have to add something similar for `MatchedPath`. Do you see something else that might need the same treatment?

---

I wrote a few comments on the current implementation but have you looked if we could just put the latest matches inside a new extension for `LatestAndGreatestUrlParams` (or extend `UrlParams` to contain a list and either an index or copy of the latest params) so that there is no need for any of the complexity here? Then `InnerPath` would just do the exact same thing as `Path` does, it would just use a different extension as the base data.

axum/src/routing/path_router.rs
DanBondarenko: Disclaimer: I'm not experienced with benchmarks. I used `cargo bench` to compare feature on/off. Results show near-negligible difference: <1%, close to noise. I didn't measure binary size differences. 

Based on this, I presume it's safe to remove the feature flag? I did this in the latest revision.

<details><summary>Detailed results of benchmarks</summary>
<p>

I used three existing benchmarks and two new ones for this:
```rust
    benchmark("nest-deep")
        .path("/level-0/level-1/level-2/level-3/level-4/leaf")
        .run(|| {
            let mut app = Router::new().route("/leaf", get(|| async {}));
            for level in (0..5).rev() {
                app = Router::new().nest(&format!("/level-{level}"), app);
            }
            app
        });

    benchmark("nest-deep-captures")
        .path("/level-0/0/level-1/1/level-2/2/level-3/3/level-4/4/leaf")
        .run(|| {
            let mut app = Router::new().route("/leaf", get(|| async {}));
            for level in (0..5).rev() {
                app = Router::new().nest(&format!("/level-{level}/{{id_{level}}}"), app);
            }
            app
        });
```

Results:

| Benchmark | Req/s with feature on | Req/s with feature off | Change |
| -- | -- | -- | -- |
| basic (control) | 116,437 | 117,153 | +0.62% |
| routing (control) | 120,677 | 120,288 | −0.32% |
| basic-nest | 113,866 | 112,619 | −1.10% |
| nest-deep | 110,497 | 109,616 | −0.80% |
| nest-deep-captures |  109,258 | 108,133 | −1.03% |


Output of `cargo bench -p axum -- basic routing basic-nest nest-deep nest-deep-captures`:
```
     Running benches/benches.rs (target/release/deps/benches-61545e76700a81b3)
Running "basic" benchmark
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:53296/a/b/c for 30 second(s)
    Latencies:
      0.09ms   0.03ms   0.01ms   0.75ms
    Requests:
      Total: 3493183 Req/Sec: 116436.66
    Transfer:
      Total: 433.08 MB Transfer Rate: 14.44 MB/Sec
Running "basic-nest" benchmark
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:53308/a/b/c for 30 second(s)
    Latencies:
      0.09ms   0.03ms   0.02ms   2.11ms
    Requests:
      Total: 3416066 Req/Sec: 113866.08
    Transfer:
      Total: 423.52 MB Transfer Rate: 14.12 MB/Sec
Running "nest-deep" benchmark
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:53320/level-0/level-1/level-2/level-3/level-4/leaf for 30 second(s)
    Latencies:
      0.09ms   0.03ms   0.02ms   0.82ms
    Requests:
      Total: 3315000 Req/Sec: 110497.30
    Transfer:
      Total: 237.11 MB Transfer Rate: 7.90 MB/Sec
Running "nest-deep-captures" benchmark
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:53333/level-0/0/level-1/1/level-2/2/level-3/3/level-4/4/leaf for 30 second(s)
    Latencies:
      0.09ms   0.03ms   0.02ms   1.99ms
    Requests:
      Total: 3277831 Req/Sec: 109258.23
    Transfer:
      Total: 234.45 MB Transfer Rate: 7.81 MB/Sec
Running "routing" benchmark
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:53344/foo/bar/baz for 30 second(s)
    Latencies:
      0.08ms   0.03ms   0.01ms   3.77ms
    Requests:
      Total: 3620405 Req/Sec: 120677.17
    Transfer:
      Total: 258.95 MB Transfer Rate: 8.63 MB/Sec
```

Output of `cargo bench -p axum --features inner-path -- basic routing basic-nest nest-deep nest-deep-captures`:
```
     Running benches/benches.rs (target/release/deps/benches-e05c5464431905c3)
Running "basic" benchmark
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:53362/a/b/c for 30 second(s)
    Latencies:
      0.09ms   0.03ms   0.02ms   1.14ms
    Requests:
      Total: 3514677 Req/Sec: 117152.81
    Transfer:
      Total: 435.74 MB Transfer Rate: 14.52 MB/Sec
Running "basic-nest" benchmark
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:53373/a/b/c for 30 second(s)
    Latencies:
      0.09ms   0.03ms   0.02ms   2.95ms
    Requests:
      Total: 3378639 Req/Sec: 112618.55
    Transfer:
      Total: 418.88 MB Transfer Rate: 13.96 MB/Sec
Running "nest-deep" benchmark
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:53385/level-0/level-1/level-2/level-3/level-4/leaf for 30 second(s)
    Latencies:
      0.09ms   0.03ms   0.02ms   0.94ms
    Requests:
      Total: 3288548 Req/Sec: 109615.59
    Transfer:
      Total: 235.22 MB Transfer Rate: 7.84 MB/Sec
Running "nest-deep-captures" benchmark
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:53396/level-0/0/level-1/1/level-2/2/level-3/3/level-4/4/leaf for 30 second(s)
    Latencies:
      0.09ms   0.03ms   0.02ms   1.88ms
    Requests:
      Total: 3244089 Req/Sec: 108133.45
    Transfer:
      Total: 232.04 MB Transfer Rate: 7.73 MB/Sec
Running "routing" benchmark
  Beginning round 1...
  Benchmarking 10 connections @ http://0.0.0.0:53407/foo/bar/baz for 30 second(s)
    Latencies:
      0.08ms   0.03ms   0.02ms   5.93ms
    Requests:
      Total: 3608749 Req/Sec: 120288.45
    Transfer:
      Total: 258.12 MB Transfer Rate: 8.60 MB/Sec
```

</p>
</details>

axum/src/extract/path/inner_path.rs
DanBondarenko: Sure thing! Done.

axum/src/extract/path/inner_path.rs
DanBondarenko: Ah yes. Fixed.

axum/src/extract/path/mod.rs
DanBondarenko: The original `from_request_parts` of `Path` used inner fns with the comment "Extracted into separate fn so it's only compiled once for all T.". I felt the obligation to somehow keep communicating that it's important to prevent binary size ballooning for each T. But seems this causes more confusion than benefit, so I removed it now.

axum/src/extract/path/mod.rs
DanBondarenko: Yeah, on second glance I agree. Inlined it.

DanBondarenko:
@mladedav Thank you for the review! I'll answer each of the topics separately.

---

> Looking at it, we might also have to add something similar for MatchedPath. Do you see something else that might need the same treatment?

AFAIK, two are missing:
* no inner variant for `RawPathParams`
* no inner variant for `MatchedPath`

---

> have you looked if we could just put the latest matches inside a new extension for LatestAndGreatestUrlParams (or extend UrlParams to contain a list and either an index or copy of the latest params) so that there is no need for any of the complexity here?

That was my first thought and impl as well. Problem? `Router::nest` flattens routers, producing a single combined router. UrlParams is inserted into extensions only once for this combined router, NOT per nesting level. 

Established pattern for maintaining "borders" between nesting levels: add service for each nesting level (e.g. `StripPrefix`/`SetNestedPath`). This is the pattern I followed. I didn't see other ways to track the nesting levels when a request arrives.

Why not read `UrlParams` and `EnclosingCapturesCount` in the service and insert `LatestAndGreatestUrlParams` into extensions immediately? IMO, this introduces even more complexity:
* The machinery of `UrlParams` would have to be duplicated for the "latest and greatest" variant. 
* No reduction of services/structs: all current ones would still be required.
* Moves logic from `InnerPath`'s `from_request_parts` to the service at the cost of an extra struct.

If complexity of `get_params` and `serialize_path_params` is a concern, it could be reduced by merging `PathRejection` and `InnerPathRejection`.

mladedav:
We could set a new latest field of `UrlParams` e.g. in the`StripPrefix` Middleware, couldn't we?

DanBondarenko:
@mladedav 
> We could set a new latest field of `UrlParams` e.g. in the`StripPrefix` Middleware, couldn't we?

Indeed, that makes sense. Less code this way too. I reimplemented it using `StripPrefix`. 

I was initially hesitant about relying on `StripPrefix` since I thought of it as "belonging to" the `Uri` extractor, so I didn't want to touch it. But on second glance, it's probably more maintainable to have one "fat" service rather than many small ones. If that's the direction, you may also consider merging `SetNestedPath` the same way.

As `StripPrefix` now has new responsibilities, it may be desireable to rename it. I'd consider something like `EnterNestedRouter`. Let me know if this is something you'd like me do.

axum/src/routing/mod.rs
```diff
@@ -34,7 +34,7 @@ mod method_filter;
 mod not_found;
 pub(crate) mod path_router;
 mod route;
-mod strip_prefix;
+pub(crate) mod strip_prefix;
```
mladedav, line 37: I don't think this change is needed anymore.

axum/src/extract/path/inner_path.rs
```diff
@@ -0,0 +1,722 @@
...
+///
+/// # `InnerPath` vs [`Path`]
+/// [`Path`] gives you every capture in the request path, including those matched by enclosing
+/// [`nest`] calls. `InnerPath` gives you only the ones from the route the innermost router
+/// matched. If the router isn't nested, or is nested only at static paths, `InnerPath` and [`Path`] see
+/// exactly the same captures.
+///
+/// # `Option<InnerPath<T>>` behavior
+///
+/// You can use `Option<InnerPath<Path<T>>>` as an extractor to allow the same handler to
```
mladedav, line 29: `Path` should not be here.

axum/src/extract/path/inner_path.rs
```diff
@@ -0,0 +1,722 @@
...
+/// async fn params_map(
+///     InnerPath(params): InnerPath<HashMap<String, String>>,
+/// ) {
+///     // `params` contains keys "team_id" and "user_id"
+/// }
+///
+/// async fn params_vec(
+///     InnerPath(params): InnerPath<Vec<(String, String)>>,
+/// ) {
+///     // `params` contains a vector containing a single tuple
```
mladedav, line 118: I think this is wrong. This should contain two tuples, one for `user_id` and one for `team_id` with only `tenant_id` missing.

mladedav (commented):
Thanks a lot for this, especially for the very thorough tests.

Just a few very minor things.

Also, on [Discord](https://discord.com/channels/500028886025895936/1406236163819180185/1546216248583397406) we've decided to keep the naming as is currently done with `InnerPath` which was I think the only outstanding decision.

DanBondarenko:
@mladedav Thanks for reviewing! I fixed the three issues you mentioned. 🙂

PR #3858, opened 2026-08-04
Fix nested error downcast

spellsaif:
## Motivation                                                                                                                                                                           
                                                                                                                                                                                          
  Currently, when a request body limit is exceeded, FailedToBufferBody::from_err checks if the underlying error is a LengthLimitError by stripping exactly two hardcoded layers of axum_core::Error wrappers.                                                                                                                                                              
                                                                                                                                                                                          
  This is fragile. In applications using custom nested extractors, logging middleware, or body-mapping layers, the request body can be wrapped multiple times. This results in three or more layers of axum_core::Error wrapping around the root error (e.g., Error → Error → Error → LengthLimitError).
  
  Because the code only downcasts two layers, the third layer is never unwrapped. The code falls back to UnknownBodyError and incorrectly returns 400 Bad Request instead of the expected 413 Payload Too Large.

## Solution
  
  This PR replaces the hardcoded double-downcast in FailedToBufferBody::from_err with a loop that recursively downcasts and strips axum_core::Error wrappers dynamically:
  
    let mut box_error = err.into();
    loop {
        match box_error.downcast::<Error>() {
            Ok(unwrapped) => {
                box_error = unwrapped.into_inner();
            }
            Err(err) => {
                box_error = err;
                break;
            }
        }
    }
  
  Since the downcast consumes ownership of the boxed error, we match on the Err variant to safely reclaim ownership of the original BoxError when no more axum_core::Error wrappers are found.
  
### Verification & Testing
  
  1. Added Unit Test: Added a new test nested_body_limit_rejection in axum/src/routing/tests/mod.rs that applies a custom body-limiting middleware multiple times, successfully producing  ≥3 layers of wraps. The test verifies that the server correctly returns 413 Payload Too Large.
  2. Workspace Verification: Verified that all 377 tests in the workspace compile and pass successfully.
  
### Performance Rationale
  
  • Happy Path: This function is only executed on the error path (inside .map_err(FailedToBufferBody::from_err)), meaning there is zero impact on successful request handling performance.
  • Error Path: Downcasting in Rust is a cheap, zero-allocation operation (simply comparing two TypeId integer values). The loop runs at most N times (typically 1 to 3), which takes a fraction of a microsecond and has no measurable overhead compared to the request handling lifecycle.

### DIFF
--- a/axum-core/src/extract/rejection.rs
+++ b/axum-core/src/extract/rejection.rs
@@ -19,17 +19,19 @@ impl FailedToBufferBody {
     where
         E: Into<BoxError>,
     {
-        // two layers of boxes here because `with_limited_body`
-        // wraps the `http_body_util::Limited` in an `axum_core::Body`
-        // which also wraps the error type
-        let box_error = match err.into().downcast::<Error>() {
-            Ok(err) => err.into_inner(),
-            Err(err) => err,
-        };
-        let box_error = match box_error.downcast::<Error>() {
-            Ok(err) => err.into_inner(),
-            Err(err) => err,
-        };
+        // peel any number of `axum_core::Error` wrappers to reach the underlying source error
+        let mut box_error = err.into();
+        loop {
+            match box_error.downcast::<Error>() {
+                Ok(unwrapped) => {
+                    box_error = unwrapped.into_inner();
+                }
+                Err(err) => {
+                    box_error = err;
+                    break;
+                }
+            }
+        }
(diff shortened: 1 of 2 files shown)


mladedav (commented):
Did you verify that the test is failing on main?

Also fix ci please.

spellsaif:
Let me verify again and fix it

spellsaif:
@mladedav all tests are passing and also fixed CI :)

mladedav (approved):

PR #3860, opened 2026-08-10
Expose MethodRouter call_with_state

Manishakumari21:

### DIFF
--- a/axum/src/routing/method_routing.rs
+++ b/axum/src/routing/method_routing.rs
@@ -1462,7 +1462,6 @@ mod tests {
         assert_eq!(status, StatusCode::OK);
         assert_eq!(body, "ok");
     }
-    
 
     #[crate::test]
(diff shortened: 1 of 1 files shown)


Manishakumari21:
Hi, I’ve implemented the requested change by exposing `MethodRouter::call_with_state` and added a regression test covering shared-reference usage.

I also verified the change with the relevant test, which passes successfully. Thank you for reviewing!

mladedav:
Could you link the issue where the change was requested please?

Also, I don't think the test actually tests what you say. It's a submodule so the visibility is not tested at all and the test itself is named as if it is supposed to test that this works with shared reference, but if the method required an owned object, I think it would still pass.

Also next time, you should write any information you want to share about the PR in the description (including motivation and the issue link) not into comments. It's better visible for anyone seeing the PR.

PR #3867, opened 2026-08-17
Avoid shared shutdown waiter contention

jthomson04:
## Motivation

`axum::serve` currently clones the shutdown `watch::Sender` for every accepted connection. Each connection polls `Sender::closed()`, so all connection shutdown waiters share synchronization state. This becomes expensive when many connections are active across many CPU cores.

Initial end-to-end benchmarks used [Dynamo](https://github.com/ai-dynamo/dynamo), a distributed inference serving framework that uses Axum for HTTP. Its frontend accepts OpenAI-compatible requests, performs tokenization and routing, and streams worker output to clients. These results were measured in a frontend-bound scenario. The only software difference between the two arms was the Axum build: unmodified Axum 0.8.9 versus Axum 0.8.9 with this patch.

| Dynamo end-to-end arm | Throughput | First response latency p50 | First response latency p95 | First response latency p99 |
|---|---:|---:|---:|---:|
| Unmodified Axum 0.8.9 | 1,000.3 req/s | 63.08 ms | 113.96 ms | 262.87 ms |
| Axum 0.8.9 with receiver fix | 1,897.9 req/s | 42.47 ms | 107.84 ms | 263.93 ms |

I also isolated this behavior in a standalone Axum benchmark. It used 4,096 persistent connections and finite responses with 32 256-byte SSE frames, one scheduler yield between frames, a stock `rewrk` client on another node, a 15-second warmup, and a 60-second measurement.

| Arm | Throughput | p99 latency | CPU/request |
|---|---:|---:|---:|
| Control | 151,990 req/s | 71.44 ms | 0.312 ms |
| Receiver fix | 180,091 req/s | 65.79 ms | 0.147 ms |

In an on-CPU profile of this benchmark, `Sender::closed` used 16.57% inclusive CPU in the control. Its shared mutex used 15.32% inclusive CPU, including 7.12% in the contended lock path. Those stacks disappeared with this change, while `Receiver::changed` used 4.14% inclusive CPU.

## Solution

Keep the shutdown sender in the server task and give each accepted connection its own cloned receiver. The accept loop and connection tasks wait on `Receiver::changed()`, and shutdown drops the sender to notify all receivers.

This keeps the public API and graceful shutdown behavior unchanged. Existing tests continue to cover graceful draining and custom executors.

## Validation

- `cargo test -p axum --all-features serve::tests`
- `cargo test --locked --workspace --all-features --all-targets`
- `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/axum/CHANGELOG.md
+++ b/axum/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **added:** Add `RawPathParams::from_request_extensions` ([#3757])
 - **changed:** `serve` has an additional generic argument and can now work with any response body
   type, not just `axum::body::Body` ([#3205])
+- **changed:** Reduced contention in `axum::serve` shutdown notification with many
+  active connections ([#3867])
 - **changed:** `Redirect` constructors now accept any `impl Into<String>` ([#3635])
 - **changed:** Updated `matchit` allowing for routes with captures and static prefixes and suffixes ([#3702])
 - **fixed:** Responses to `HEAD` will not accidentally reply with `content-length: 0` anymore ([#3742])
@@ -44,6 +46,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#3836]: https://github.com/tokio-rs/axum/pull/3836
 [#3757]: https://github.com/tokio-rs/axum/pull/3757
 [#3801]: https://github.com/tokio-rs/axum/pull/3801
+[#3867]: https://github.com/tokio-rs/axum/pull/3867
 
 # 0.8.9
 
--- a/axum/src/serve/mod.rs
+++ b/axum/src/serve/mod.rs
@@ -329,7 +329,6 @@ where
             _marker,
         } = self;
 
-        // Receiver-side waits avoid contention in `Sender::closed`'s shared notification state.
         let (_signal_tx, signal_rx) = watch::channel(());
         let (_close_tx, close_rx) = watch::channel(());
 
@@ -457,7 +456,6 @@ where
             _marker,
         } = self;
 
-        // Receiver-side waits avoid contention in `Sender::closed`'s shared notification state.
         let (signal_tx, mut signal_rx) = watch::channel(());
         executor.execute(async move {
             signal.await;
--- a/axum/src/serve/mod.rs
+++ b/axum/src/serve/mod.rs
@@ -329,14 +329,14 @@ where
             _marker,
         } = self;
 
-        let (signal_tx, _signal_rx) = watch::channel(());
+        let (_signal_tx, signal_rx) = watch::channel(());
         let (_close_tx, close_rx) = watch::channel(());
 
         loop {
             let (io, remote_addr) = listener.accept().await;
             handle_connection(
                 &mut make_service,
-                &signal_tx,
+                signal_rx.clone(),
                 &close_rx,
                 io,
                 remote_addr,
@@ -456,18 +456,18 @@ where
             _marker,
         } = self;
 
-        let (signal_tx, signal_rx) = watch::channel(());
+        let (signal_tx, mut signal_rx) = watch::channel(());
         executor.execute(async move {
             signal.await;
             trace!("received graceful shutdown signal. Telling tasks to shutdown");
-            drop(signal_rx);
+            drop(signal_tx);
         });
 
         let (close_tx, close_rx) = watch::channel(());
 
         loop {
             let (io, remote_addr) =
-                match select(pin!(listener.accept()), pin!(signal_tx.closed())).await {
+                match select(pin!(listener.accept()), pin!(signal_rx.changed())).await {
                     Either::Left((conn, _)) => conn,
                     Either::Right(_) => {
                         trace!("signal received, not accepting new connections");
@@ -477,7 +477,7 @@ where
 
             handle_connection(
                 &mut make_service,
-                &signal_tx,
+                signal_rx.clone(),
                 &close_rx,
                 io,
                 remote_addr,
@@ -562,7 +562,7 @@ where
 
 async fn handle_connection<L, M, S, B, E>(
     make_service: &mut M,
-    signal_tx: &watch::Sender<()>,
+    mut signal_rx: watch::Receiver<()>,
     close_rx: &watch::Receiver<()>,
     io: <L as Listener>::Io,
     remote_addr: <L as Listener>::Addr,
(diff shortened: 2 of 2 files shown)


axum/src/serve/mod.rs
```diff
@@ -329,14 +329,14 @@ where
             _marker,
         } = self;
 
-        let (signal_tx, _signal_rx) = watch::channel(());
+        let (_signal_tx, signal_rx) = watch::channel(());
         let (_close_tx, close_rx) = watch::channel(());
```
jthomson04: `_close_tx` is never awaited in this path — `close_tx.closed()` exists only in `WithGracefulShutdown::run` (line 496). So nothing observes this channel here, yet every accepted connection still pays `Arc::clone` + `ref_count.fetch_add` when `handle_connection` clones `close_rx` (line 601), and `fetch_sub` + a `notify_waiters()` branch on drop (line 634) — all on one shared cache line bounced across every core.

That's the same class of cross-core contention this PR removes for the signal channel, on the same accept hot path. Making the parameter `Option<watch::Receiver<()>>`, or hoisting drain-tracking out of `handle_connection`, would let the non-graceful path skip it entirely.

axum/src/serve/mod.rs
```diff
@@ -329,14 +329,14 @@ where
             _marker,
         } = self;
 
-        let (signal_tx, _signal_rx) = watch::channel(());
+        let (_signal_tx, signal_rx) = watch::channel(());
```
jthomson04: Worth a comment here on why the wait is receiver-side. `Sender::closed()` and `Receiver::changed()` are interchangeable spellings of the same wait as far as behaviour goes, so nothing in the test suite can catch someone swapping back — they'd silently give up the throughput win. A line at the channel construction (and ideally at the `changed()` call sites, 470 and 617) is the only guard.

The load-bearing `_signal_tx` is less fragile than it looks, for what it's worth: mutating it to `let (_, signal_rx)` makes `tap_io_runs_on_each_accepted_connection` fail, so CI does catch that one.

axum/src/serve/mod.rs
```diff
@@ -456,18 +456,18 @@ where
...
-            drop(signal_rx);
+            drop(signal_tx);
         });
 
         let (close_tx, close_rx) = watch::channel(());
 
         loop {
             let (io, remote_addr) =
-                match select(pin!(listener.accept()), pin!(signal_tx.closed())).await {
+                match select(pin!(listener.accept()), pin!(signal_rx.changed())).await {
```
jthomson04: `Either::Right(_)` discards the `changed()` result, so `Ok(())` (a value was sent) is treated the same as `Err(RecvError)` (last sender dropped) — only the latter is a real shutdown. Same at line 617, where it triggers `graceful_shutdown()`.

Latent today since nothing calls `send()` on this channel, but the old `Sender::closed()` had no second trigger to worry about. If the channel later carries anything — a shutdown reason, a second notification — the accept loop stops accepting and every live connection starts draining. Matching `Either::Right((Err(_), _))`, or a comment that `Ok` is unreachable, pins the invariant.

axum/src/serve/mod.rs
```diff
@@ -562,7 +562,7 @@ where
 
 async fn handle_connection<L, M, S, B, E>(
     make_service: &mut M,
-    signal_tx: &watch::Sender<()>,
+    mut signal_rx: watch::Receiver<()>,
```
jthomson04: Minor: `signal_rx` now arrives by value (caller clones, at 339 and 481) while `close_rx` is still by reference (callee clones, at 601). The by-value form is only needed because `changed()` takes `&mut self`, which the callee could just as well get from its own clone — `let mut signal_rx = signal_rx.clone();` — keeping both parameters `&watch::Receiver<()>` and both accept loops symmetric.

axum/src/serve/mod.rs
```diff
@@ -329,14 +329,14 @@ where
             _marker,
         } = self;
 
-        let (signal_tx, _signal_rx) = watch::channel(());
+        let (_signal_tx, signal_rx) = watch::channel(());
         let (_close_tx, close_rx) = watch::channel(());
 
         loop {
             let (io, remote_addr) = listener.accept().await;
```
jthomson04: Not for this PR, but worth noting: this loop and the one in `WithGracefulShutdown::run` (459) are near-identical — same channel setup, same six-argument `handle_connection` call, differing only in the signal `select` and the post-loop drain. It's why this single change had to be applied twice in lockstep, and why the now-dead close channel survives in only one of the two copies. A shared helper parameterized by an optional signal receiver would make the shutdown protocol a single point of truth.

axum/CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **added:** Add `RawPathParams::from_request_extensions` ([#3757])
 - **changed:** `serve` has an additional generic argument and can now work with any response body
   type, not just `axum::body::Body` ([#3205])
+- **changed:** Reduced contention in `axum::serve` shutdown notification with many
+  active connections ([#3867])
```
mladedav, line 27: Why did you put it in the middle of the list? Not that it really matters just wondering the reasoning here.

axum/src/serve/mod.rs
```diff
@@ -456,28 +457,32 @@ where
             _marker,
         } = self;
 
-        let (signal_tx, signal_rx) = watch::channel(());
+        // Receiver-side waits avoid contention in `Sender::closed`'s shared notification state.
```
mladedav: Drop the two comments

mladedav:
This makes sense, would you mind tying even with `tokio::sync::Notify` directly? That's what both of those end up using internally so we might as well do that too. Maybe it could still give a little more performance.

Anyway thanks, this looks good.

jthomson04:
Hey @mladedav. I tried out that `tokio::sync::Notify` approach. I found that it has significantly lower perf than both the baseline on main and the receiver patch. Under the hood, it looks like [`tokio::sync::watch::Receiver` uses a `BigNotify`](https://github.com/tokio-rs/tokio/blob/tokio-1.49.0/tokio/src/sync/watch.rs#L560), which is internal to tokio. We could reproduce that behavior in Axum, but unless tokio exposes `BigNotify`, we'd need to implement and maintain our own sharded notifier here.

mladedav (commented):
I am a bit surprised that raw `Notify` would be slower than the current `closed`-based variant since that one does use it in the background and we would prevent spurious wakes which I suspect were cause of the contention. But I missed the big notify so you're right we might want to just use what you had originally.

Thanks.

axum/CHANGELOG.md
jthomson04: I grouped it with the other axum serve change directly above.

mladedav (approved):

PR #3870, opened 2026-08-27
Treat CRLF as a single SSE line ending

aiqubits:
## Motivation

The server-sent events parsing algorithm treats CRLF (`\r\n`) as a single
line ending. `EventDataWriter` currently handles every CR and LF independently,
so serializing data such as `first\r\nsecond` inserts an extra `data: ` field
between the CR and LF. Clients consequently decode an unexpected blank line.

The same issue occurs when the CR and LF are written in separate
`fmt::Write::write_str` calls.

## Solution

Track when a write ends with a CR and defer the next `data: ` prefix until the
following byte is known. A following LF completes one CRLF line ending;
otherwise, the CR is handled as a lone line ending. Finalizing the writer also
handles a pending trailing CR.

The `EventDataWriter` documentation now describes CRLF, lone CR, and lone LF
explicitly. Regression tests cover CRLF in one write and every pair of
boundaries that divides mixed line-ending data into three writes.

## Testing

- `cargo test --locked --workspace --all-features --tests`
- `cargo clippy --locked -p axum --lib --all-features -- -D warnings -A clippy::double_must_use`
- `cargo doc --locked -p axum --all-features --no-deps`

### DIFF
--- a/axum/src/response/sse.rs
+++ b/axum/src/response/sse.rs
@@ -40,7 +40,6 @@ use http_body::Frame;
 use pin_project_lite::pin_project;
 use std::{
     fmt::{self, Write as _},
-    io::Write as _,
     mem,
     pin::Pin,
     task::{ready, Context, Poll},
@@ -179,8 +178,9 @@ pub struct Event {
 /// Expose [`Event`] as a [`std::fmt::Write`]
 /// such that any form of data can be written as data safely.
 ///
-/// This also ensures that newline characters `\r` and `\n`
-/// correctly trigger a split with a new `data: ` prefix.
+/// This also ensures that CRLF (`\r\n`), lone CR (`\r`), and lone LF (`\n`)
+/// line endings correctly trigger a split with a new `data: ` prefix. A CRLF
+/// pair is treated as a single line ending.
 ///
 /// # Panics
 ///
@@ -195,6 +195,10 @@ pub struct EventDataWriter {
     // this does not say anything about whether or not `event` contains
     // data or not.
     data_written: bool,
+
+    // A trailing CR might be followed by an LF in the next write. Defer the
+    // next `data: ` prefix until we know whether they form a CRLF pair.
+    pending_cr: bool,
 }
 
 impl Event {
@@ -218,6 +222,7 @@ impl Event {
         EventDataWriter {
             event: self,
             data_written: false,
+            pending_cr: false,
         }
     }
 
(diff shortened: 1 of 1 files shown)


axum/src/response/sse.rs
```diff
@@ -427,43 +432,67 @@ impl EventDataWriter {
...
 impl EventDataWriter {
-    // Assumption: underlying writer never returns an error:
-    // <https://docs.rs/bytes/latest/src/bytes/buf/writer.rs.html#79-82>
     fn write_buf(&mut self, buf: &[u8]) -> usize {
         if buf.is_empty() {
             return 0;
         }
 
-        let buffer = self.event.buffer.as_mut();
-
```
mladedav, line 448: Can you try not making random small changes like moving this line 15 lines lower?

This is just noise that makes the change more complicated than necessary and it does not improve the code.

axum/src/response/sse.rs
```diff
@@ -427,43 +432,67 @@ impl EventDataWriter {
...
+                buffer.put_u8(b'\n');
+                last_split = 1;
+            }
+            buffer.extend_from_slice(b"data: ");
+            self.pending_cr = false;
+        }
+
+        while let Some(delimiter) = memchr::memchr2(b'\n', b'\r', &buf[last_split..]) {
+            let delimiter = last_split + delimiter;
+            buffer.extend_from_slice(&buf[last_split..=delimiter]);
```
mladedav: Using `memchr2_iter` was perfectly reasonable here and did exactly what you're doing with the while let, and we couldn't break the code by forgetting to update `last_split`. So why the change?

mladedav (commented):
Thanks for the fix.

axum/src/response/sse.rs
aiqubits: You're right. I've restored the original buffer/writer structure and kept the changes limited to the CRLF state handling.

axum/src/response/sse.rs
aiqubits: I've restored memchr2_iter. CRLF pairs are now handled by skipping the CR and processing the pair at the LF position, while a trailing CR remains pending across writes.

mladedav (approved):

axum/src/response/sse.rs
mladedav: For the record, I didn't really mind the direct BytesMut usage instead of io::Write, but here specifically you just moved the variable assignment a few lines elsewhere without a reason.

But keeping the code as is is also fine.

PR #3871, opened 2026-08-27
Upgrade axum-macros to syn 3

jakobhellermann:
## Motivation

Can reduce dependencies if you have syn 3 already in the tree.
`syn` 3.0 has a MSRV of `1.71`, which is lower than this crate.

## Solution

Bump version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -224,7 +224,7 @@ dependencies = [
  "rustversion",
  "serde",
  "serde_json",
- "syn",
+ "syn 3.0.4",
  "tokio",
  "trybuild",
 ]
@@ -414,7 +414,7 @@ checksum = "97369cbbc041bc366949bc74d34658d6cda5621039731c6310521892a3a20ae0"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.114",
 ]
 
 [[package]]
@@ -518,7 +518,7 @@ checksum = "162ee34ebcb7c64a8abebc059ce0fee27c2262618d7b60ed8faf72fef13c3650"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.114",
 ]
 
 [[package]]
@@ -1182,7 +1182,7 @@ dependencies = [
  "itertools",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.114",
 ]
 
 [[package]]
@@ -1204,7 +1204,7 @@ checksum = "f71ee38b42f8459a88d3362be6f9b841ad2d5421844f61eb1c59c11bff3ac14a"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.114",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


mladedav:
- ci wants to change cargo.lock so maybe you didn't commit all changes?
- we transitively depend on `syn@2` so this actually cannot speed up anything by having less dependencies but can slow things down by having more. Until all our dependencies also upgrade.
- cargo deny might need a new setting to allow the syn duplication.

That said we still want to update it, it's just that the motivation misrepresents the situation, it's only benefitial if you don't have `syn@2` elsewhere in your tree and specifically if it also isn't in our tree.

jakobhellermann:
> ci wants to change cargo.lock so maybe you didn't commit all changes?

Interesting, I ran `cargo check` and committed the changes that it made. I can run `cargo update` and check that in, if it helps.

> we transitively depend on syn@2 so this actually cannot speed up anything by having less dependencies but can slow things down by having more. Until all our dependencies also upgrade.

You're right. I only checked `cargo tree -i syn --depth 1` on my crate and went through the dependencies that aren't on syn 3 already. Didn't mean to misrepresent the improvements.

```
cargo tree -i syn@2
syn v2.0.119
├── quickcheck_macros v1.2.0 (proc-macro)
│   [dev-dependencies]
├── synstructure v0.13.2
├── tracing-attributes v0.1.31 (proc-macro)
├── yoke-derive v0.8.2 (proc-macro)
└── zerofrom-derive v0.1.7 (proc-macro)
```

Seems like the only production dependency is quickcheck, I opened a PR for it https://github.com/BurntSushi/quickcheck/pull/353.
Until that is released I suppose there is no point in merging this

synstructure (merged): https://github.com/mystor/synstructure/pull/75
zerofrom and yoke: (merged): https://github.com/unicode-org/icu4x/pull/8292
tracing-attributes: https://github.com/tokio-rs/tracing/pull/3594

> cargo deny might need a new setting to allow the syn duplication.

Indeed it fails. Depending on whether you want to merge this before or after all (including dev) dependencies have released version targetting syn 3 I can add an exception, or wait until it isn't necessary anymore.

jakobhellermann:
> ci wants to change cargo.lock so maybe you didn't commit all changes?

Okay it was complaining about the `examples/Cargo.lock` file. I've updated it now (fully with `cargo update`, not just the syn change).

mladedav:
Maybe I'm misunderstanding the output (it does not make much sense to me anyway), but I think there are three more crates that pull it in even outside of dev dependencies:
```
> cargo tree -i syn@2 --depth 1 --all-features
syn v2.0.119
├── prost-derive v0.14.4 (proc-macro)
├── quickcheck_macros v1.2.0 (proc-macro)
│   [dev-dependencies]
├── synstructure v0.13.2
├── tracing-attributes v0.1.31 (proc-macro)
├── yoke-derive v0.8.2 (proc-macro)
└── zerofrom-derive v0.1.7 (proc-macro)
> cargo tree -i syn@2 --depth 1 --all-features --edges no-dev
syn v2.0.119
├── prost-derive v0.14.4 (proc-macro)
├── synstructure v0.13.2
├── yoke-derive v0.8.2 (proc-macro)
└── zerofrom-derive v0.1.7 (proc-macro)
```

I'd say we can merge it before all dependencies upgrade their versions. As long as they do it in a minor release we don't need to wait for them to benefit from it eventually.

PR #3872, opened 2026-08-28
Adding way to SSE Event (and Buffer) test for equality

AlexanderHD27:
## Motivation
I want to write some unit-Test for my API that use sse. I have a serpate function that creates the Event steam. It would be nice to just use `assert_eq()` . This requires the `PartialEq` Trait.

## Solution
Just adding it to derive

### DIFF
--- a/axum/src/response/sse.rs
+++ b/axum/src/response/sse.rs
@@ -143,7 +143,7 @@ where
 /// Once finalized, it's immutable and cheap to clone.
 /// The buffer is active during the event building, but eventually
 /// becomes finalized to send http body frames as [`Bytes`].
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone)]
 enum Buffer {
     Active(BytesMut),
     Finalized(Bytes),
@@ -169,7 +169,7 @@ impl Buffer {
 }
 
 /// Server-sent event
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone)]
 #[must_use]
 pub struct Event {
     buffer: Buffer,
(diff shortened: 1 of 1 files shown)


mladedav:
You can call `.finalize()` on your event and then directly compare the underlying `Bytes`. The derive also checks some internal state that should not be relevant and `finalize` has a side effect if the event is not tracked as closed (and it is called when we send the event) so it might be better to directly check that as that's the real output that will be sent on the wire.

Is that good enough for your case?

AlexanderHD27:
Yes, theoreticly that would work for me. But maybe I miss something but `.finalize()` and `buffer` field are private in `Event`, so some mechanism to get to the data is needed then. Maybe a `.data()` or `.buffer()` method?
(I am happy to implement that)

Thanks for your response

mladedav:
I don't think you actually want that at all. For your tests it shouldn't matter if you have "active" buffer that is one new line short which will be appended when you finalize it or if you have a finalized event already with said trailing new line there.

More pragmatically, I don't think axum should let you observe that difference either. Adding the partial eq (or the methods you propose) would allow you to do exactly that and strictly speaking, we shouldn't change that later in a non-breaking release which would make it harder to evolve the implementation.

So I'd rather not do this and have users check the finalized bytes directly.

AlexanderHD27:
I'm compleatly fine using the `.finalize()`, i get that. But still my problem is i cannot access because its private to Event.
```
error[E0624]: method `finalize` is private
   --> src/admin_dashboard.rs:207:67
    |
207 |             let event = sse_stream.next().await.unwrap().unwrap().finalize();
    |                                                                   ^^^^^^^^ private method
```

So is making finalize() pub okay?

guemax:
+1. I'd appreciate making `finalize()` public. I have the same/similar problem and, currently, I cannot use the proposed solution because `finalize()` is private.

mladedav (approved):
Sorry, this slipped through. Yeah, making it public seems reasonable to me.

guemax:
Thank you very much!

Issue #3878, opened 2026-09-04
Add `axum::extract::Result`


mladedav:
- [x] I have looked for existing issues (including closed) about this

### Motivation

Writing handlers that take `Result` as an extractor is hard because one has to either know the rejection type and name it directly or they have to write stuff like `Result<T, <T as FromRequestParts>::Rejection>`.

#3877 was what brought me to the idea. It wouldn't help directly, but the person there said they've tried `Result` but weren't able to make it work so maybe if they happened to find this new `Result`, they could work this out by themselves. But even then, having just one generic would be easier on them even if we had to respond to the question of how.

### Proposal

We can provide `pub type Result<T: FromRequestParts> = Result<T, <T as FromRequestParts>::Rejection>` and then users can do simply `axum::extract::Result<T>`.

### Alternatives

Do nothing if we think there is already enough `Result` types in the world.

webdevsamran:
Hi team, I would like to implement `axum::extract::Result`.

Having a dedicated `Result<T, E = <T as FromRequestParts>::Rejection>` extractor wrapper in `axum::extract` will make fallible extraction ergonomic without requiring callers to manually specify complex rejection types in handler signatures. I can implement the extractor, derive `FromRequestParts`/`FromRequest`, and add documentation and test coverage.

Could you please assign this to me? Thanks!

mladedav:
There's already an open PR, if you want to, you can see if there's something to help with there but I don't think you can write something different than is there.

mladedav:
The proposed type alias is not possible because we also need to include `S`.

It should also use `FromRequest` rather than `FromRequestParts` which would force us to also include its `M` in the `Result` alias making it three generics.

With that, I'll close this because I think the alias might not be an improvement. We could also evaluate adding `FromRequestResult<T, S>` and `FromRequestPartsResult<T, S>` but I don't think that would be that good either since the discoverability might be worse. We can reopen if there is appetite or if someone comes up with a way to make this more ergonomic.

More context https://github.com/tokio-rs/axum/pull/3879#discussion_r3966098907

PR #3879, opened 2026-09-05
Clarify how to handle extractor rejections

pederbe:
## Summary

Explain how to choose the error type in `Result<T, E>` when handling extractor rejections. Add a `Path` example alongside the existing `Json` example, and link the guide and rejection module in both directions.

The path example demonstrates a custom error response and tests successful and rejected requests.

Refs #3878 and #3877.

## Testing

- `cargo test --locked -p axum --all-features --doc`: 177 passed, 1 ignored. The new path example checks both response status and body; the existing JSON example compiles.
- `cargo doc --locked -p axum --all-features --no-deps` with `RUSTDOCFLAGS="-D rustdoc::all -A rustdoc::private-doc-tests"`: passed. Verified both links in the generated HTML.
- `cargo fmt --all --check`: passed.

Checks ran on Windows with Rust 1.98.1.

### DIFF
--- a/axum/src/extract/mod.rs
+++ b/axum/src/extract/mod.rs
@@ -2,6 +2,47 @@
 
 use http::header::{self, HeaderMap};
 
+/// A result whose error type is the rejection of a [`FromRequestParts`] extractor.
+///
+/// Use this to handle an extractor's rejection in the handler instead of automatically
+/// returning it as a response.
+///
+/// ```
+/// use axum::{extract::{Path, Result}, routing::get, Router};
+///
+/// async fn handler(path: Result<Path<u64>>) -> String {
+///     match path {
+///         Ok(Path(id)) => format!("Item {id}"),
+///         Err(rejection) => format!("Invalid path: {rejection}"),
+///     }
+/// }
+///
+/// let app = Router::new().route("/{id}", get(handler));
+/// # let _: Router = app;
+/// ```
+///
+/// `S` is the state type used by the extractor and defaults to `()`. For extractors
+/// that require application state, specify it explicitly:
+///
+/// ```
+/// use axum::{extract::{Result, State}, routing::post, Router};
+///
+/// #[derive(Clone)]
+/// struct AppState;
+///
+/// async fn handler(state: Result<State<AppState>, AppState>, body: String) -> String {
+///     let Ok(State(state)) = state;
+///     body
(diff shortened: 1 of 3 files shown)


axum/src/extract/mod.rs
```diff
@@ -2,6 +2,47 @@
...
+///     body
+/// }
+///
+/// let app = Router::new().route("/", post(handler)).with_state(AppState);
+/// # let _: Router = app;
+/// ```
+///
+/// This alias is for request-parts extractors. For body-consuming extractors such
+/// as `Json`, use [`std::result::Result`] with their rejection type.
+pub type Result<T, S = ()> = std::result::Result<T, <T as FromRequestParts<S>>::Rejection>;
```
jlizen: The type alias seems more confusing than not given the asymmetry of only working with `FromRequestParts`. Also, leaking `S` seems bad.

My preference would be to just expand `extract.md` to have examples for `Path` and `Json` and cross-link the `rejection` module. Seems like that would address the main confusion from #3877 ?

jlizen (changes requested):

axum/src/extract/mod.rs
mladedav: > The type alias seems more confusing than not given the asymmetry of only working with `FromRequestParts`

Yeah, that probably should be `FromRequest`, that way it should work with all extractors because we have effectively `impl<T: FromRequestParts> FromRequest for T`. Which would make the `Result` type even worse because we'd have to include both generics from `FromRequest<S, M>` and not just the state.

> Also, leaking S seems bad.

That one should be fine. Each handler has a specific state anyway. Very few extractors require a specific state so I think it would be very rare that the unit state would not apply here and even then I would hope that `_` would get resolved to the correct type.

---

But yeah, I didn't originally realize that we would need another generic even though I'd hope inference would always get it right. Having `Result` with two parameters where it's not the ok and err variants would probably be even more confusing than not doing this.

For now, I agree just tweaking the docs would probably be better unless we can figure out some trick to make this work more elegantly.

pederbe: Agreed. I removed the alias and expanded the rejection-handling docs instead, adding a `Path` example alongside the existing JSON example and links to the rejection module. The path example tests both successful extraction and a custom rejection response.

PR #3881, opened 2026-09-07
fix(axum-extra): preserve protobuf body buffering rejections

cuishuang:
## Motivation

`Protobuf::from_request` currently maps errors returned while collecting the request body to `ProtobufDecodeError`. As a result, exceeding `DefaultBodyLimit` returns `422 Unprocessable Entity`, even though protobuf decoding has not started.

This also bypasses the documented `BytesRejection` variant of `ProtobufRejection` for request body buffering failures.

## Solution

Buffer the request through `Bytes::from_request` and only map errors returned by `T::decode` to `ProtobufDecodeError`.

This makes request body buffering failures use `BytesRejection`, so an oversized body returns `413 Payload Too Large`, while invalid protobuf data continues to return `422 Unprocessable Entity`.

This uses the shared `Bytes` extractor rather than duplicating `FailedToBufferBody` error classification in `axum-extra`. It replaces the aggregate-buffer path introduced in #3241 because that path does not preserve the body buffering rejection type.

### DIFF
--- a/axum-extra/src/protobuf.rs
+++ b/axum-extra/src/protobuf.rs
@@ -5,11 +5,9 @@ use axum_core::__define_rejection as define_rejection;
 use axum_core::{
     extract::{rejection::BytesRejection, FromRequest, Request},
     response::{IntoResponse, IntoResponseFailed, Response},
-    RequestExt,
 };
-use bytes::BytesMut;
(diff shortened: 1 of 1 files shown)


mladedav (approved):
Thanks for the fix

PR #3886, opened 2026-09-08
fix(axum-extra): handle missing optional typed headers

cuishuang:
## Motivation

`Option<TypedHeader<T>>` currently decodes the header values before checking whether the header exists.

Some `headers::Header` implementations successfully decode an empty iterator. For example, `headers::Cookie` decodes an empty iterator into an empty cookie collection. As a result, `Option<TypedHeader<headers::Cookie>>` returns `Some(empty Cookie)` when the `Cookie` header is completely absent.

This contradicts the documented behavior of optional extractors, where a missing header should return `None`.

## Solution

Check whether the header exists before decoding its values.

If the header is absent, return `None` immediately. If it is present, preserve the existing decoding behavior and return `TypedHeaderRejection` when parsing fails.

### DIFF
--- a/axum-extra/src/typed_header.rs
+++ b/axum-extra/src/typed_header.rs
@@ -90,10 +90,12 @@ where
         _state: &S,
     ) -> Result<Option<Self>, Self::Rejection> {
         let mut values = parts.headers.get_all(T::name()).iter();
-        let is_missing = values.size_hint() == (0, Some(0));
+        if values.size_hint() == (0, Some(0)) {
+            return Ok(None);
+        }
+
         match T::decode(&mut values) {
-            Ok(res) => Ok(Some(Self(res))),
-            Err(_) if is_missing => Ok(None),
+            Ok(value) => Ok(Some(Self(value))),
             Err(err) => Err(TypedHeaderRejection {
                 name: T::name(),
                 reason: TypedHeaderRejectionReason::Error(err),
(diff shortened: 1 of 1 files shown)


axum-extra/src/typed_header.rs
```diff
@@ -89,16 +89,17 @@ where
...
-            Ok(res) => Ok(Some(Self(res))),
-            Err(_) if is_missing => Ok(None),
-            Err(err) => Err(TypedHeaderRejection {
+        T::decode(&mut values)
+            .map(|value| Some(Self(value)))
+            .map_err(|err| TypedHeaderRejection {
                 name: T::name(),
                 reason: TypedHeaderRejectionReason::Error(err),
-            }),
-        }
```
mladedav, line 101: Instead of these changes, just do `if values.size_hint() == (0, Some(0)) { return Ok(None) }` before the match.

The `is_missing` variable was correct, the issue is onlz that the `Ok` case didn't take it into account at all.

This way we do just one lookup into `headers`, your rework does one to check presence and then another to get the values.

mladedav (commented):
Thanks for spotting this, you're right this does not align with our docs.

axum-extra/src/typed_header.rs
cuishuang: Thanks for the suggestion. I updated the implementation to check `values.size_hint()` before decoding, while keeping the existing `get_all(...).iter()` lookup.

Please review it again.

mladedav (approved):
Thanks again

PR #1789, opened 2023-02-25
Remove `B` type param: Follow ups

davidpdrsn:
These are some follow-ups to https://github.com/tokio-rs/axum/pull/1778 and https://github.com/tokio-rs/axum/pull/1751. Both those branches should be merged first so this is just a draft.

Basically I think we should use `axum_core::body::Body` everywhere, for consistency and simplicity:

- Add a `Request<B = Body>` type alias.
- Use that type alias everywhere.
- Change `Next<B>` to just `Next`.
- Use `Body` in `axum_core::response::Response` and thus also in `IntoResponse`.
- Remove `BoxBody` and its `boxed` constructor.
- Implement `FromRequest` for `Body`.
- Remove `RawBody` extractor.
- Remove re-exports of `Empty` and `Full`

## TODO

- [x] Should we remove the re-exports of `Full` and `Empty`? `Body` should be able to do all the same things.
- [x] Double check changelog

PR #3892, opened 2026-09-14
Accept generic request bodies in function middleware

### LINKED ISSUE #2492: `RequestBodyLimitLayer` when applied via `ServiceBuilder` won't compile in v0.7
- [×] I have looked for existing issues (including closed) about this

### Version

- axum v0.7.3
- axum-core v0.4.2

### Platform

- Darwin Kernel Version 23.0.0: Fri Sep 15 14:41:34 PDT 2023; root:xnu-10002.1.13~1/RELEASE_ARM64_T8103 arm64

### Crates

```toml
axum = "0.7.3"
tokio = { version = "1.35.1", features = ["full"] }
tower = "0.4.13"
tower-http = { version = "0.5.0", features = ["limit"] }
```

### Description

 With `ServiceBuilder` I was applying multiple middleware, the first of which was `RequestBodyLimitLayer` and everything compiled in v0.6 (when the `B` type parameter was still available). With Axum now having its own body type in v0.7, the following example no longer compiles:
```rs
use axum::{
    extract::{Request, State},
    middleware::{self, Next},
    response::{Response, Result},
    routing::get,
    Router,
};
use tower::ServiceBuilder;

#[tokio::main]
async fn main() {
    let app = Router::new()
        .route("/", get(|| async { "Hello, Axum!" }))
        .layer(
            ServiceBuilder::new()
                .layer(RequestBodyLimitLayer::new(1_048_576))
                .layer(middleware::from_fn_with_state("state", foo)),
        );

    let listener = tokio::net::TcpListener::bind("0.0.0.0:8080").await.unwrap();

}

async fn foo(State(_state): State<&'static str>, req: Request, next: Next) -> Result<Response> {
    Ok(next.run(req).await)
}
```
```bash
error[E0277]:
the trait bound `axum::middleware::FromFn<fn(axum::extract::State<&'static str>,
axum::http::Request<axum::body::Body>, axum::middleware::Next) ->
impl std::future::Future<Output = std::result::Result<axum::http::Response<axum::body::Body>,
axum::response::ErrorResponse>> {foo}, &str, axum::routing::Route, _>:
tower::Service<axum::http::Request<tower_http::body::Limited<axum::body::Body>>>` is not satisfied
```

I noticed in the ["Fewer generics"](https://tokio.rs/blog/2023-11-27-announcing-axum-0-7-0#fewer-generics) section from the v0.7 announcement blog post that `ServiceBuilder` wasn't being used and sure enough things will compile if `RequestBodyLimitLayer` is moved to its own layer (like in the blog post):
```diff
 // compiles
 let app = Router::new()
     .route("/", get(|| async { "Hello, Axum!" }))
     .layer(
         ServiceBuilder::new()
-            .layer(RequestBodyLimitLayer::new(1_048_576))
             .layer(middleware::from_fn_with_state("state", foo)),
     )
+    .layer(RequestBodyLimitLayer::new(1_048_576));
```

To keep `RequestBodyLimitLayer` a part of `ServiceBuilder`, things will also compile if it is moved out of the first position (but I will eventually need it to be first):
```diff
 // also compiles, but limits need to be applied before everything else, so ultimately won't work
 let app = Router::new()
     .route("/", get(|| async { "Hello, Axum!" }))
     .layer(
         ServiceBuilder::new()
-            .layer(RequestBodyLimitLayer::new(1_048_576))
             .layer(middleware::from_fn_with_state("state", foo))
+            .layer(RequestBodyLimitLayer::new(1_048_576)),
     );
```

In #1110 (and mentioned in [the blog post](https://tokio.rs/blog/2023-11-27-announcing-axum-0-7-0#fewer-generics)) the solution back in v0.6 was to extract `Request<http_body::Limited<Body>>` instead of `Request<Body>` in subsequent middleware. But with the body type parameter being removed from v0.7, I'm not sure how to satisfy the trait bound in this situation (other than avoid using `RequestBodyLimitLayer` with `ServiceBuilder`).

KayanoLiam:
## Motivation

Fixes #2492.

Placing `RequestBodyLimitLayer` before `from_fn_with_state` in a `ServiceBuilder` fails to compile: the limit layer passes `Request<Limited<Body>>`, while `FromFn` only implements `Service<Request<Body>>`. The same restriction prevents directly layering function middleware over a service receiving another HTTP body type.

## Solution

Accept compatible HTTP request bodies and convert them to axum's `Body` before running the extractors, following the conversion used by `map_request`. The middleware function and `Next` continue to use axum's request type.

Regression tests cover the original stateful layer composition, successful body extraction, HTTP 413 for oversized bodies, and a direct service receiving `Full<Bytes>`.

### Compatibility

This expands the `Service` implementation and affects type inference for direct, unqualified `.ready()` calls. They may need `ServiceExt::<Request>::ready(&mut middleware)`; `.oneshot(request)` infers the input body from its argument. The changelog marks this as a breaking change, and the API docs explain the explicit request type. A separate downstream probe verified the old unqualified form on the baseline and the qualified form on both versions.

### Validation

- [Upstream CI](https://github.com/tokio-rs/axum/actions/runs/34839259836): all 17 jobs pass, including stable/beta workspace and example tests, MSRV, pinned nightly, documentation, feature combinations, and target checks.
- Both new tests fail to compile on the original source because the request body types are rejected.
- `cargo test -p axum --all-features --locked`: 385 unit tests, 1 integration test, and 178 doc tests pass; 1 doc test is ignored.
- `GITHUB_ACTIONS=true cargo test --locked --workspace --all-features --all-targets`: passes across the root workspace using its CI benchmark smoke path.
- Macro tests on the CI-pinned `nightly-2026-05-04`: 120 compile fixtures and 23 doc tests pass. Five relevant example packages compile with all features and targets.
- `cargo fmt -p axum -- --check` and `cargo doc -p axum --all-features --no-deps --locked`: pass.
- `cargo check -p axum --no-default-features --lib --locked`: passes.
- The unmodified baseline and this branch both hit an existing `clippy::double_must_use` error in `axum/src/extract/path/mod.rs` with Rust 1.98. `cargo clippy -p axum --all-features --all-targets -- -D warnings -A clippy::double_must_use` passes; the unrelated path code is unchanged.

AI assistance: OpenAI Codex assisted with the implementation, tests, compatibility review, and local validation.


mladedav:
Did you verify that it doesn't work on main? I think we've handled something like this some time ago but it's a breaking change and it's waiting for 0.9 but I might be confusing two different issues. (EDIT I think I was mistaken, what I was thinking about was specific to `serve`)

Does this in any way touch on #1110 though?

KayanoLiam:
Yes. I reproduced both new tests against `0b9325bf6624a8ec8840b35641a3ab3d88539845`, which is still current `main`. With only the regression tests added and the implementation unchanged, `cargo test -p axum --lib middleware::from_fn::tests` fails: the `ServiceBuilder` case requires `FromFn: Service<Request<Limited<Body>>>`, and the direct `Full<Bytes>` case is also rejected.

It is related to #1110. #1751 introduced accepting generic request bodies and converting them to axum's `Body` at service boundaries, including `FromFn`. #1789 subsequently removed `FromFn`'s generic input and that conversion while changing `Next<B>` to `Next`. In the current `RequestBodyLimitLayer -> FromFn -> Route` stack, the wrapped body reaches `FromFn` before the route can normalize it.

This PR accepts generic input at that boundary while keeping middleware arguments, extractors, and `Next` on axum's `Body`. The regressions pass with the fix, including HTTP 413 for oversized bodies. I also verified the direct `.ready()` inference change and marked it as breaking with the explicit request-type migration in the docs/changelog. All 17 upstream CI jobs pass.

PR #3896, opened 2026-09-18
feat(axum): add `max_connection_idle` to `ConnectionLifetimeLimits`

adriangb:
## Problem

#3779 adds `max_connection_age`, which caps how long a connection lives whether or not it is being used. The original proposal in #3753 also listed `max_connection_idle`, matching [`tonic::transport::Server::max_connection_idle`](https://docs.rs/tonic/latest/tonic/transport/struct.Server.html#method.max_connection_idle). This adds it.

The two limits answer different questions. `max_connection_age` forces clients to rotate, so that a load balancer can spread work over backends that appeared after the pool was built. `max_connection_idle` releases the memory and the file descriptor of a connection that a client holds open but does not use. A server with many mostly-idle clients pays for every one of those connections.

## What this adds

```rust
use std::time::Duration;
use axum::serve::ConnectionLifetimeLimits;

let limits = ConnectionLifetimeLimits::new()
    .max_connection_idle(Duration::from_secs(5 * 60));

axum::serve(listener, app)
    .connection_lifetime_limits(limits)
    .await;
```

The clock starts when the connection goes idle and resets as soon as a request arrives. A connection that keeps serving requests is never closed by this limit, however long it lives. That is what sets it apart from `max_connection_age`.

Once the limit elapses the connection is shut down gracefully, the same way the age limit does it: HTTP/1 closes after the current request, HTTP/2 sends a `GOAWAY`.

Both limits can be set together.

## Why a plain `Duration` and not a value type

`MaxConnectionAge` has `jitter` and `grace`. Neither earns its place here:

- **Jitter** exists to desynchronise connections that were all opened in the same instant. Idle moments are already desynchronised by traffic.
- **Grace** exists to bound how long to wait for in-flight work. An idle connection has none.

If a knob turns out to be needed, `max_connection_idle(Duration)` can become `max_connection_idle(impl Into<MaxConnectionIdle>)` without breaking callers.

## Implementation — two traps

hyper's `Connection` does not expose whether it is idle, so in-flight requests are counted in a private service wrapper. Two things make a naive version silently wrong:

1. **A request does not end when the handler returns a `Response`. It ends when the response body has been fully sent.** The in-flight guard travels from the response future into a body wrapper and is released when `poll_frame` returns `None`, or when the body is dropped part way. Without this, a streaming response looks idle while it is still streaming.
2. **HTTP/2 multiplexes, so "idle" means zero open streams**, not "the last request finished". The count is a counter, and the idle timer is only armed on the N→0 transition.

Both are covered by tests that were verified against a mutated implementation.

### Upgraded connections

If the response is `101`, or the request used `CONNECT` and the response is 2xx, the request's count is never given back and the connection counts as busy for the rest of its life. An upgraded HTTP/2 stream stays on the tracked connection but `serve` cannot see its traffic, so without this a live WebSocket would look idle and receive a `GOAWAY`.

This is a heuristic, because hyper gives no signal. It fails safe: a handler that returns `101` without a real upgrade pins its connection as busy rather than closing a live one. Flagging it explicitly for review.

## Tests

Seven new tests, following the existing `max_connection_age*` style: idle connection closed on HTTP/1 and HTTP/2; slow handler not closed; streaming response body not closed; HTTP/2 open stream not treated as idle; upgraded HTTP/2 connection never closed; idle beating age when both are set; age plus grace still force-closing a connection that is never idle.

The streaming test is worth a note. Its first version did not catch its own mutation, because an HTTP/1 graceful shutdown mid-stream still delivers the body in full, so the bug was invisible. It now sends a second request on the same connection, which fails if the connection was marked close-after-response.

`cargo test -p axum --all-features` passes. fmt clean. clippy clean on `tokio,http1`, `tokio,http2` and `tokio,http1,http2`. `RUSTDOCFLAGS="-D warnings" cargo doc --all-features` clean.

## Open questions

- The upgrade heuristic above.
- **No grace period for idle**, which is a deliberate asymmetry with `max_connection_age`. If a request arrives between the timer firing and `graceful_shutdown` taking effect, HTTP/1 waits for it with no bound. Forcing a close there would kill a request that arrived legitimately, so waiting seems right.
- The in-flight wrapper wraps every connection's service, with the counter set to `None` when no idle limit is configured. That costs an `Option` check per body frame. It can be gated harder if that matters.

Part of #3753.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


adriangb:
@yanns @mladedav @jlizen do any of you have an interest in this feature? It's much less important for us than #3779 so I'm not sure how much time I'll be able to dedicate but it would be nice to have for completeness, I can help push it across the line if there's interest.

jlizen:
Yes, I am interested in it. I've had limited maintenance cycles because I've been traveling but I'd like to circle back to reviewing it, hopefully next week.

adriangb:
Okay, took it out of draft so it can be reviewed. No rush on my end.

PR #3898, opened 2026-09-21
match TraceLayer documentation link version

abd0-omar:
adjust code comment to match and point to the same tower-http crate version

## Motivation

the code comment be on par with the crate version

## Solution

update the code comment


yanns (approved):
Wondering if we should use https://docs.rs/tower-http/latest/tower_http/trace/ to avoid having to update this comment all the time

abd0-omar:
> Wondering if we should use https://docs.rs/tower-http/latest/tower_http/trace/ to avoid having to update this comment all the time

yeah, we risk doc drift if `Cargo.toml` updates without updating the version comment, but tower-http `latest` could show breaking changes that don't match the example code (which I doubt). I lean toward keeping it pinned for accuracy, but happy to switch to latest if you'd rather avoid the ongoing maintenance

